PR #3, opened 2018-07-14
Add /embed/:id

### LINKED ISSUE #2: Embedding support
It would be nice if it could support being embedded in web pages via /embed/videoID.

omarroth:
Would close #2.

This currently supports all options supported by player on watch page (`start`, `t`, `end`, `listen`) and can be embedded the same as the default YT player:
```
<iframe id="ytplayer" type="text/html" width="640" height="360"
  src="https://invidio.us/embed/M7lc1UVf-VE"
  frameborder="0"></iframe>
```
Large [amount of options](https://developers.google.com/youtube/player_parameters) aren't yet supported.
Leaving open for a bit for feedback.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -373,6 +373,79 @@ get "/watch" do |env|
   templated "watch"
 end
 
+get "/embed/:id" do |env|
+  if env.params.url["id"]?
+    id = env.params.url["id"]
+  else
+    next env.redirect "/"
+  end
+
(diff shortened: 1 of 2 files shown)


Zero3K:
Which option controls auto-play? If there isn't one, then it should be added.

omarroth:
@Zero3K Absolutely, as a subsequent PR it would be good to implement several other features.

PR #186, opened 2018-09-26
basic docker-compose cluster

flourgaz:
Run `docker-compose up` in the base directory of the project, it should build two images and start two containers, one for the app, other for postgres db.

Find the IP of the app container with `docker inspect invidious_invidious_1 | grep IPAddress` and access that IP address with port 3000 in your browser.

To rebuild the images (after changing anything in the code/dockerfile) run `docker-compose build` or `docker-compose build --no-cache`.

The postgres data is in stored in a volume which is set up on first run. If you want to delete that and setup again, remove the volume with `docker volume rm invidious_postgresdata` (the name might be slightly different if you have the project in different directory, find it with `docker volume ls`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,15 @@
+FROM archlinux/base
+
+RUN pacman -Sy --noconfirm shards crystal imagemagick librsvg \
+    which pkgconf gcc ttf-liberation
+# base-devel contains many other basic packages, that are normally assumed to already exist on a clean arch system
+
+ADD . /invidious
+
+WORKDIR /invidious
+
+RUN sed -i 's/host: localhost/host: postgres/' config/config.yml && \
+    shards && \
+    crystal build src/invidious.cr
+
+CMD [ "/invidious/invidious" ]
--- a/setup.sh
+++ b/setup.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 createdb invidious
-createuser kemal
+#createuser kemal
+psql -c "CREATE USER kemal WITH PASSWORD 'kemal';"
 psql invidious < config/sql/channels.sql
 psql invidious < config/sql/videos.sql
 psql invidious < config/sql/channel_videos.sql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: '3'
+services:
+  postgres:
+    build:
+      context: .
+      dockerfile: docker/Dockerfile.postgres
+    restart: unless-stopped
+    volumes:
+      - postgresdata:/var/lib/postgresql/data
+  invidious:
+    build:
+      context: .
+      dockerfile: docker/Dockerfile
+    restart: unless-stopped
+    ports:
+      - "3000:3000"
+    depends_on:
+      - postgres
+
+volumes:
+  postgresdata:
--- a/docker/Dockerfile.postgres
+++ b/docker/Dockerfile.postgres
@@ -1,0 +1,10 @@
+FROM postgres:10
+
+ENV POSTGRES_USER postgres
+
+ADD ./setup.sh /setup.sh
+ADD ./config/sql /config/sql
+ADD ./docker/entrypoint.postgres.sh /entrypoint.sh
+
+ENTRYPOINT [ "/entrypoint.sh" ]
+CMD [ "postgres" ]
--- a/docker/entrypoint.postgres.sh
+++ b/docker/entrypoint.postgres.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+CMD="$@"
+if [ ! -f /var/lib/postgresql/data/setupFinished ]; then
+    echo "### first run - setting up invidious database"
+    /usr/local/bin/docker-entrypoint.sh postgres &
+    sleep 10
+    until runuser -l postgres -c 'pg_isready' 2>/dev/null; do
+        >&2 echo "### Postgres is unavailable - waiting"
+        sleep 5
+    done
+    >&2 echo "### importing table schemas"
+    su postgres -c "/setup.sh" && touch /var/lib/postgresql/data/setupFinished
+    echo "### invidious database setup finished"
+    exit
+fi
+
+echo "running postgres /usr/local/bin/docker-entrypoint.sh $CMD"
+exec /usr/local/bin/docker-entrypoint.sh $CMD


docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:bionic
+
+RUN apt-get update && apt-get install -y git curl imagemagick librsvg2-bin gnupg && \
```
omarroth: Currently the clock CAPTCHA looks like this:
download
Which I suspect is due to missing dependencies for imagemagick, but I'm having trouble tracking it down. I've tried using `librsvg2-dev` instead, unfortunately still encountering it.

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    curl -sL "https://keybase.io/crystal/pgp_keys.asc" | apt-key add - && \
+    echo "deb https://dist.crystal-lang.org/apt crystal main" | tee /etc/apt/sources.list.d/crystal.list && \
+    apt-get update && apt-get install -y crystal libyaml-dev libxml2-dev && \
+    git clone https://github.com/omarroth/invidious
```
omarroth: Would it be possible to use the local repository instead of cloning it? I think that would make it easier to test changes, as well as making it easier for someone to run off a fork of this project.

setup.sh
```diff
@@ -1,8 +1,8 @@
...
-psql invidious < config/sql/channels.sql
-psql invidious < config/sql/videos.sql
-psql invidious < config/sql/channel_videos.sql
-psql invidious < config/sql/users.sql
+createdb -U kemal invidious
+# createuser -U kemal kemal
+psql -U kemal invidious < config/sql/channels.sql
+psql -U kemal invidious < config/sql/videos.sql
+psql -U kemal invidious < config/sql/channel_videos.sql
+psql -U kemal invidious < config/sql/users.sql
```
omarroth: It might make more sense for these commands to be run similar to how they are in the [Dockerfile](https://github.com/omarroth/invidious/pull/186/files#diff-ebacf6f6ae4ee68078bb16454b23247dR3) if possible, since the changes here break normal setup. If that's not possible, I'd move these changes into a separate file for docker to use.

docker/Dockerfile
flourgaz: There are several ways how the docker image can be constructed:
1. for production it should be built using a git clone from a tagged master revision
2. for dev it should be build using the local sources, which can be either
    - included in the image using dockerfile ADD instruction - so you'd need to rebuild the image after a change in the source
    - or mounted using volume - however this would also need some automatic way to recompile the sources, it's useful mainly for scripted languages. I'm not very familiar with compiled languages

I will change it to the ADD instruction, we'll see how well it works

setup.sh
flourgaz: this line in setup.sh: `createuser kemal` - wouldn't it create the user without password? Because then in config.yml we have user kemal with password kemal, so I'm surprised it's even working for you... For the automatic install I think I have to change one or the other...

omarroth: You're right. Looks like [here](https://www.postgresql.org/docs/10/static/app-createuser.html) when creating a user password authentication is disabled by default, so any password in `config.yml` appears to work. That's my mistake.

Your approach works fine for the docker installation, so I think I'll mark this as resolved. For the default installation, I think I'll change how the kemal user is created, so a user can change e.g. default password, and add that info to the README.

docker/Dockerfile
flourgaz: I was playing around with the dependencies and extras for imagemagick, but nothing worked, until I tried building it from ubuntu:xenial, but I'm hesitant to go that far back. I'll try to check what's changed between xenial imagemagick and the current bionic

setup.sh
flourgaz: I actually found another approach, where the only change in setup.sh will be

# createuser kemal
psql -c "CREATE USER kemal WITH PASSWORD 'kemal';"

the rest can stay the same, without "-U kemal". I'll push it to this PR soon

omarroth:
The solution you mentioned with imagemagick actually reminded me that I encountered the same issue when attempting to setup Invidious on DigitalOcean. Using Bionic, I tried building from source, using imagemagick 6.9, installing various other libraries, and ended up using Xenial rather than investigate further.

For reference, here are two different machines that render the CAPTCHA as expected:

```
$ uname -srm
Linux 4.4.0-134-generic x86_64

$ convert -version
convert -version
Version: ImageMagick 6.8.9-9 Q16 x86_64 2018-07-10 http://www.imagemagick.org
Copyright: Copyright (C) 1999-2014 ImageMagick Studio LLC
Features: DPC Modules OpenMP
Delegates: bzlib cairo djvu fftw fontconfig freetype jbig jng jpeg lcms lqr ltdl lzma openexr pangocairo png rsvg tiff wmf x xml zlib
```

```
$ uname -srvm
Linux 4.18.9-arch1-1-ARCH x86_64

$ convert -version
Version: ImageMagick 7.0.8-12 Q16 x86_64 2018-09-23 https://imagemagick.org
Copyright: © 1999-2018 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules OpenCL OpenMP 
Delegates (built-in): bzlib cairo fontconfig freetype gslib heic jbig jng jp2 jpeg lcms lqr ltdl lzma openexr pangocairo png ps raw rsvg tiff webp wmf x xml zlib
```

flourgaz:
yes, I'm pretty sure it's because of the "rsvg" mentioned among the delegates, because the docker bionic version doesn't have it, but the xenial does. Unfortunately I haven't found any other means of getting it there, than compiling imagemagick from sources with a --with-librsvg parameter during the configuration. It seems like the bionic version of the package simply doesn't come with it. I could compile it, but I really wish I wouldn't have to. As far as I read in the tutorials, the build deps are 500MB+

flourgaz:
Ok, I just found that the artful and bionic versions of imagemagick are explicitly built with `--without-rsvg`. But the arch linux package still has it. And this is the beauty of docker, I'll simply swap the base image and build it on arch linux ;-)

flourgaz:
finally got it working on arch, there are couple of basic packages needed for the build and clock needed a font, but it works fine now. And I bound the port to localhost:3000 so now you don't need to look for the IPAddress of the container

flourgaz:
Looks great.
Next step - production docker image automatically built on hub.docker.com. I'll look into it...

PR #271, opened 2018-12-23
Fixed subscription's xml feed

tor255:
yep, changed "xmlns:" to ""xmlns:yt":", because that apparently associates it with a namespace, according to `feedparser`


omarroth:
LGTM

Issue #110, opened 2018-08-14
Legacy annotation support


Discookie:
Before 2017, YouTube allowed creating annotations on top of videos. Some older videos still utilize these.

Example: https://invidio.us/watch?v=8VaFE6RU4bI

Discookie:
Does `https://www.youtube.com/annotations_invideo?video_id=` do this?
Seems to work for the [example](https://www.youtube.com/annotations_invideo?video_id=8VaFE6RU4bI).

omarroth:
The main issue is with displaying them. For VideoJS, there does not seem to be a plugin that is capable of displaying annotations the same way as YouTube.

Zero3K:
I found two plugins that might work:

https://github.com/contently/videojs-annotation-comments
https://github.com/brightcove/videojs-overlay

omarroth:
Ooh, I like how [videojs-annotation-comments](https://github.com/contently/videojs-annotation-comments). I'd like to hear thoughts on how YouTube's annotations could be improved.

elypter:
my first thought when looking at this was that it could be used to mark sponsoring or ads that are included in the video and potentially automatically skip them.

elypter:
thanks for doing this. it's really a shame that they are removing millions of man hours just so they can make the expierience worse.

glmdgrielson:
If need be, I can get you a few comparison images/videos so that it's easier to recreate.

omarroth:
I've actually done some work making recordings with annotations already, namely [this](https://www.youtube.com/watch?v=XwxBJEzgqWU) video, which seems pretty comprehensive. If there's any videos that use some features that aren't included or you think would help, please do record them!

glmdgrielson:
Ah. I was going to say that one of the channels I asked for is currently doing reuploads and I have a couple of screenshots that may be usable as references for how they should look. Do we have a recording of that one that won't break on the 15th?

omarroth:
YouTube deleted all annotations around 15:00 UTC as expected. An estimated 1.4 billion videos have been archived, and I'm working through everything right now to make it suitable for Invidious and other projects to use.

There's a list of videos available [here](https://archive.org/download/archived_annotations_video_ids.csv/archived_video_ids.csv). If anyone has backups that aren't in the list I would love to have them!

elypter:
[...]

PR #303, opened 2019-01-20
Add annotation player

glmdgrielson:
This addresses issue #110. Yay for adding annotations back!

This should do until the archive system gets revamped.


src/invidious/views/components/player.ecr
```diff
@@ -167,7 +167,6 @@ if (bpb) {
 <% end %>
 
 
-var video = player;
```
omarroth: `videojs-youtube-annotatations.js` relies on the player being called `video`. Without, it throws:
```
video is not defined
```
[here](https://github.com/afrmtbl/videojs-youtube-annotations/blob/master/dist/dist.js#L408).

glmdgrielson: Oh... oops!

omarroth:
I'll plan on adding an updated version of `videojs-youtube-annotations.js` before merging, since there's quite a few features missing that are included in the extension it's based on. 

The `/api/v1/annotations/:id` is temporary and will also change once everything has been uploaded to the Internet Archive.

https://dev.invidio.us is now pulling from this branch, so if annotations for a video are available from the temporary API, you should be able to view them.

src/invidious/views/components/player.ecr
omarroth: Updated dist.js with a fixed copy.

glmdgrielson:
So the `video` issue gas been resolved? What about the 4.3 one?

omarroth:
I expect that will be fixed in [afrmtbl/annotationlib](https://github.com/afrmtbl/annotationlib).

omarroth:
`<%= video.id %>` refers to the current video. For annotations, we want to find the video that's being linked, so we use the video ID included in the search parameters.

AgentGrady:
Several videos that had working annotations on invidious just a few days ago have suddenly stopped working. Why is this?

omarroth:
Mind posting the output of the console or linking to an example?

AgentGrady:
Ok, here are some examples

https://dev.invidio.us/watch?v=29-q7YnyUmY

https://dev.invidio.us/watch?v=FBeuzMM7HSc

https://dev.invidio.us/watch?v=gf5YT60DaqQ

AgentGrady:
Annotations have suddenly stopped working on several videos.

AgentGrady:
https://dev.invidio.us/watch?v=FBeuzMM7HSc&iv_load_policy=1

PR #309, opened 2019-01-21
Use random password for database user (#298)

### LINKED ISSUE #298: Security: Use random password while creating "kemal" user
At the moment the database user has always the same password if you do not change it in setup.sh and config manually. It should be some random string for better security.

Perflyst:
Fix #298


omarroth:
See discussion in #298.

PR #313, opened 2019-01-23
Fix installation guide, Add Upgrade information, Create and mention documentation

Perflyst:
- [ ] `config/config.yml` has `domain: ` which user should replace with his own domain. If there is a new release which changes content of config.yml there may be issues on `git pull && git checkout` with the new release.

- [x] At the moment the default log output is syslog. As mentioned in the chat room, `-o output.log` should be available. This can be pre-defined in `invidious.service` as soon as this exists.

- [x] There is no documentation about the configuration options yet, this should be added to a wiki page and linked in the README.md

- [x] "How to upgrade Invidious" is still not documented, best option is probably to base the setup on release tags.


omarroth:
LGTM :+1:

PR #322, opened 2019-01-26
Remove Extensions from README.md

Perflyst:
Moved to wiki: https://github.com/omarroth/invidious/wiki/Extensions

Documentation and user stuff should be in the wiki, only very basics and some informations in the readme


omarroth:
I would suggest putting a link to the Wiki in place of the extensions listed.

omarroth:
:+1:

PR #334, opened 2019-02-02
Color change in the links and several improvements in the comments

aaferrari:
This is a summary with the modifications:

* Change the color of the links in `assets/css/lighttheme.css` so that they are better distinguishable from the text. If you prefer you can omit this or use a different color.
* In the comments API add the following parameters:
  * **videoId**: Necessary for permalinks of comments.
  * **authorIsChannelOwner**: To highlight the name of the video uploader in the comments from them.
  * **creatorHeart**: Optional, only in the comments that received a heart from the uploader. This in turn contains the fields **creatorThumbnail** and **creatorName**.
  * **isEdited**: To indicate if a comment was edited.
* With the comments I did the following:
  * The text that indicates how much time has passed since its publication now shows the complete date when you hover the cursor over it. The text that formats the date was added in the `translate` function because it [depends a lot on the country](https://en.wikipedia.org/wiki/Date_format_by_country), but it would be missing some way of translate the days of the week and months according to the language of the user in a similar way to [this](http://babel.pocoo.org/en/latest/dates.html).
  * Along with the time of publication add a link to see it also on Youtube (this could also be implemented for Invidious if it lets show a specific comment before the rest).
  * The name of the person who uploaded the video is highlighted from the rest.¹
  * It is indicated if a comment received a heart from the person who uploaded it¹ and if it was modified.
* Due to the previous changes 4 additional strings were added in `locales/en-US.json`.

In the following screenshots you can better appreciate all the changes made:
comment example1
comment example2
comment example3

1. This can be adjusted from the classes added in `assets/css/default.css`.


assets/css/default.css
```diff
@@ -1,3 +1,43 @@
+.channel-owner {
+  background-color: #008BEC;
+  color: beige;
+  border-radius: 9px;
+  padding: 1px 6px;
+}
+
```
omarroth: Hovering over the channel owner makes it difficult to see.

image

I would consider adding a `:hover` with the colors used by the subscribe button, something that looks something like:
```css
.channel-owner:hover {
  background-color: #a0a0a0;
  color: rgba(35, 35, 35, 1);
}
```

assets/css/default.css
```diff
@@ -1,3 +1,43 @@
+.channel-owner {
+  background-color: #008BEC;
+  color: beige;
```
omarroth: I would try and use colors already present in this file, although I wouldn't consider it a deal-breaker and I'll still be happy to merge. Here, you might consider using `#fff`.

assets/css/lighttheme.css
```diff
@@ -4,6 +4,6 @@ a:active {
 }
 
 a {
-  color: #303030;
+  color: #0093ff;
```
omarroth: Would something like `#606060` be better here? I also think links should be more noticeable, but this makes everything on the page a bright blue, so it's difficult to distinguish important information. Having only some links be this color (only links in the description, for example) would be better here.

aaferrari: I'm also not convinced by the color `#0093ff` for all the links, I changed it to `#61809b` that I think achieves a good balance between making the links more noticeable but without being too bright. What do you think? Otherwise I have no problem in leaving it as it was and highlight only the links of certain sections (such as description or comments).

dimqua:
@aaferrari did you test your changes also with dark theme?

omarroth:
LGTM. :+1:

assets/css/lighttheme.css
omarroth: I missed this comment before merging. If you'd like to open a new PR or a new issue I'd be happy to continue discussion there.

dimqua:
Awesome!

@aaferrari I suggest you to help omarroth with https://github.com/omarroth/invidious/issues/118, if you have time and interest. It'd be really appreciated.

PR #339, opened 2019-02-03
Update README.md

tmiland:
Add manual commands to Debian and Ubuntu install instructions.

Got a pubkey error on Ubuntu. Added instructions from: https://crystal-lang.org/reference/installation/on_debian_and_ubuntu.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,12 @@ $ docker-compose build
 $ sudo pacman -S shards crystal imagemagick librsvg postgresql
 
 # Ubuntu or Debian
+# First you have to add the repository to your APT configuration. For easy setup just run in your command line:
 $ curl -sSL https://dist.crystal-lang.org/apt/setup.sh | sudo bash
-$ sudo apt update
+# That will add the signing key and the repository configuration. If you prefer to do it manually, execute the following commands:
+$ curl -sL "https://keybase.io/crystal/pgp_keys.asc" | sudo apt-key add -
+$ echo "deb https://dist.crystal-lang.org/apt crystal main" | sudo tee /etc/apt/sources.list.d/crystal.list
+$ sudo apt-get update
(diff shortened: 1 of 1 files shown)


dimqua:
Shouldn't this information be moved to the wiki?

tmiland:
> Shouldn't this information be moved to the wiki?

Probably. 

> The wiki is probably a better place for this. `setup.sh` executes `apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54` which should work on any ubuntu and debian.

Wasn't setup.sh removed? It's not in my repos. 🙂

Perflyst:
I am talking about this `setup.sh`: `curl -sSL https://dist.crystal-lang.org/apt/setup.sh | sudo bash`

tmiland:
> I am talking about this `setup.sh`: `curl -sSL https://dist.crystal-lang.org/apt/setup.sh | sudo bash`

Ah, okay. Well, then, that's the main reason i added the above to the readme, in case someone else would get the same error. 


```
Err:18 https://dist.crystal-lang.org/apt crystal InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 09617FD37CC06B54
Reading package lists... Done 
W: GPG error: https://dist.crystal-lang.org/apt crystal InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 09617FD37CC06B54
E: The repository 'https://dist.crystal-lang.org/apt crystal InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

This is from -->  `setup.sh`: `curl -sSL https://dist.crystal-lang.org/apt/setup.sh | sudo bash`

omarroth:
LGTM :+1:

PR #350, opened 2019-02-09
Just a couple of adjustments

aaferrari:
In this occasion only change the colors of the links by the previous one in the icon beside the title, the recommended videos and the playlist.
Personally I think it looks better that way instead of leaving all the links of the same color.
Also add a tooltip to the icon to switch to the audio mode for usability (especially for first time Invidious users).

### DIFF
--- a/assets/css/lighttheme.css
+++ b/assets/css/lighttheme.css
@@ -1,9 +1,16 @@
 a:hover,
 a:active {
-  color: #167ac6;
+  color: #167ac6 !important;
 }
 
 a {
   color: #61809b;
   text-decoration: none;
+}
+
+/* All links that do not fit with the default color goes here */
+a > .icon, 
+.pure-u-md-1-5 > .h-box > a[href^="/watch?"],
+.playlist-restricted > ol > li > a {
+  color: #303030;
 }
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -35,11 +35,11 @@
 <h1>
     <%= HTML.escape(video.title) %> 
     <% if params[:listen] %>
-    <a href="/watch?<%= env.params.query %>&listen=0">
+    <a title="Video/audio mode" href="/watch?<%= env.params.query %>&listen=0">
         <i class="icon ion-ios-videocam"></i>
     </a>
     <% else %>
-    <a href="/watch?<%= env.params.query %>&listen=1">
+    <a title="Audio only mode" href="/watch?<%= env.params.query %>&listen=1">
         <i class="icon ion-ios-volume-high"></i>
     </a>
     <% end %>
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -39,7 +39,7 @@
         <i class="icon ion-ios-videocam"></i>
     </a>
     <% else %>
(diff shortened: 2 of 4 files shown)


assets/css/lighttheme.css
```diff
@@ -1,9 +1,16 @@
...
   color: #61809b;
   text-decoration: none;
 }
+
+/* All links that do not fit with the default color goes here */
+a > .icon, 
+.pure-u-md-1-5 > .h-box > a[href^="/watch?"],
+.playlist-restricted > ol > li > a {
+  color: #303030;
+}
```
omarroth: I'm a little confused as to what is happening here. Would it make more sense to have only video links be the exception, and have `#303030` be the default color? Or I'm likely misunderstanding how the styling is applied.

aaferrari: I have been using Invidious with the changes I made a week ago and it seems to me that the colors in the links of the playlists and related videos don't look good compared to those of the description, the comments or the videos that appear in the other pages (like the main page, the search, the channels, etc), that's why I only revert the change in those sections while the rest keeps the color `#61809b`.
If you have doubts you can try this adjustment in the style for a few days with an extension like [Stylus](https://github.com/openstyles/stylus) before making the final merge and if you are not convinced you can propose the changes you think appropriate.

src/invidious/views/watch.ecr
```diff
@@ -35,11 +35,11 @@
 <h1>
     <%= HTML.escape(video.title) %> 
     <% if params[:listen] %>
-    <a href="/watch?<%= env.params.query %>&listen=0">
+    <a title="Video/audio mode" href="/watch?<%= env.params.query %>&listen=0">
```
omarroth: I'd probably just call this "Video mode", with audio being implied.

omarroth:
The styling LGTM. The only comment I had was on the title for "Video/audio mode". Once that's resolved I'll consider this ready to merge.

src/invidious/views/watch.ecr
aaferrari: Done, also take the opportunity to make some minor fixes (details are in the commit).

src/invidious/views/watch.ecr
```diff
@@ -35,11 +35,11 @@
 <h1>
     <%= HTML.escape(video.title) %> 
     <% if params[:listen] %>
-    <a href="/watch?<%= env.params.query %>&listen=0">
+    <a title="<%=translate(locale, "Video mode")%>" href="/watch?<%= env.params.query %>&listen=0">
         <i class="icon ion-ios-videocam"></i>
     </a>
     <% else %>
-    <a href="/watch?<%= env.params.query %>&listen=1">
+    <a title="<%=translate(locale, "Audio only mode")%>" href="/watch?<%= env.params.query %>&listen=1">
```
omarroth: This should be "Audio mode" or the locale should be "Audio only mode". Personally prefer "Audio mode" but whichever you prefer is fine as long as it's consistent.

aaferrari: I didn't notice that before sending the commit, I thought I had left it as "Audio mode" both in the locale and in watch.ecr, if you correct that detail for me it would be fine.

omarroth:
LGTM :+1:

PR #353, opened 2019-02-10
Add screenshots

### LINKED ISSUE #347: Screenshot
A screenshot would be nice to encourage users to try it out ! Thanks !

Perflyst:
Closes #347


README.md
```diff
@@ -34,6 +34,11 @@ Onion links:
 
 [Alternative Invidious instances](https://github.com/omarroth/invidious/wiki/Invidious-Instances)
 
+## Screenshots
+<img src="/screenshots/invidious_popular.png?raw=true" height="140" width="280"> <img src="/screenshots/invidious_video.png?raw=true" height="140" width="280"> <img src="/screenshots/invidious_preferences.png?raw=true" height="140" width="280">
+
+<img src="/screenshots/invidious_popular_dark.png?raw=true" height="140" width="280"> <img src="/screenshots/invidious_video_dark.png?raw=true" height="140" width="280"> <img src="/screenshots/invidious_preferences_dark.png?raw=true" height="140" width="280">
+
```
omarroth: These need to be relative paths, looking at the [NewPipe README](https://github.com/TeamNewPipe/NewPipe/blob/dev/README.md) you can see there's also links to the images themselves, here's an example of what they might look like:

```markdown
[<img src="screenshots/invidious_popular.png?raw=true" height="140" width="280">](screenshots/invidious_popular.png?raw=true)
[<img src="screenshots/invidious_video.png?raw=true" height="140" width="280">](screenshots/invidious_video.png?raw=true)
[<img src="screenshots/invidious_preferences.png?raw=true" height="140" width="280">](screenshots/invidious_preferences.png?raw=true)
[<img src="screenshots/invidious_popular_dark.png?raw=true" height="140" width="280">](screenshots/invidious_popular_dark.png?raw=true)
[<img src="screenshots/invidious_video_dark.png?raw=true" height="140" width="280">](screenshots/invidious_video_dark.png?raw=true)
[<img src="screenshots/invidious_preferences_dark.png?raw=true" height="140" width="280">](screenshots/invidious_preferences_dark.png?raw=true)
```

Perflyst: Well, they use markdown, I use HTML which is allowed in .md files. Anyway, I will change it.

omarroth:
@arthurlutz hasn't given his thoughts yet, but I would like to move forward with this. 

The changes I suggested w.r.t. markdown allow you to click on the images, which I think is a nice feature. The features mentioned in the README are the ones I would like to see screenshots of if possible. Subscription feed, dark mode, comments, preferences, are the ones that seem most immediately relevant.

I think if people are still curious what the site looks like to use, there's several [places](https://github.com/omarroth/invidious/wiki/Invidious-Instances) where they can try it out, so I don't think you need many screenshots. I would probably only show features you need an account for, since that requires a larger investment from users.

arthurlutz:
LGMT !

dimqua:
> I would probably only show features you need an account for, since that requires a larger investment from users.

Agreed. Besides, I think no one needs the same screenshots in both dark and light themes. They should be as different as possible.

omarroth:
Pushed a couple minor changes, if it looks good to you I'll merge.

Perflyst:
Looks good.

PR #355, opened 2019-02-10
Add MusicPiped

dimqua:


omarroth:
It's an interesting project and I'll keep my eye on it. I'm currently a little confused however as to the current status of the project.

There was some mention of dropping Invidious support in deep-gaurav/MusicPiped#33 and I'm unsure if @deep-gaurav is still considering doing so. 

Invidious also isn't mentioned in the [Credits](https://github.com/deep-gaurav/MusicPiped#credits) section of the README, but that could just be out of date. Invidious endpoints appear to be used internally though.

I don't mind merging this, but I'd appreciate any clarification.

deep-gaurav:
@omarroth  No, I'm not dropping support for invidious it was mainly considered due to lack of region setting earlier which thanks to you is fixed.

Although yes, I would like to mention that development is slow as I'm very busy right now

omarroth:
Absolutely understandable. Let me know if you have any other issues and I wish you the best with your project.

LGTM :+1:

PR #366, opened 2019-02-17
Language fixes

comradekingu:


locales/en-US.json
```diff
@@ -76,24 +76,24 @@
...
-  "Import/Export data": "Import/Export data",
+  "Import/export data": "Import/export data",
   "Manage subscriptions": "Manage subscriptions",
   "Watch history": "Watch history",
   "Delete account": "Delete account",
   "Save preferences": "Save preferences",
   "Subscription manager": "Subscription manager",
   "`x` subscriptions": "`x` subscriptions",
-  "Import/Export": "Import/Export",
+  "Import/export": "Import/export",
```
comradekingu: This is a headline?

omarroth:
Most of these look good. The ones I'm confused about are:
- `Log in` vs `Login`
- `Sign out` vs `Log out`
- `Invalid` vs `Wrong`/`Erroneous`
- `Turned off` vs `Disabled`
- `Turned on` vs `Enabled`

They seem quite arbitrary to me. Are these changes more understandable for other English speakers, or is there some other reason why they are better?

locales/en-US.json
omarroth: It appears in `Data preferences`, here:
image

Other locales have a similar capitalization to `Import/export`, so what's proposed here LGTM.

comradekingu:
@omarroth
"Log in" is a verb, "Login" is a noun. The latter is in use to that effect.
Same for "Sign out" and "Log out", in one tense you would get an opposed outsigning, or outlogging.
"Log in/out" is now used consistently. There is one redundant "Sign in" still present, which may help a user that otherwise doesn't understand "log in".
Turned on/off is in my view more simple and familiar than the more computer-centric en-/dis-able.

dimqua:
I think `Invalid` is more common word than `Erroneous`.

GauthierPLM:
In case this is merged, it would be nice to update the keys of other locales too to avoid breaking them.

Zero3K:
Another fix to add is when its only been a year since a video was added to Youtube, it should show "Shared 1 year ago" instead of "Shared 1 years ago".

omarroth:
Sorry for not following up on this earlier. I'd be happy to merge most of this (with the exceptions I mentioned above) and I can add the relevant fixes to the other locales.

The rationale for "Login" vs. "Log in" seems like it could be either way. The justification for "Login" could easily be "Click here to go to the login page" vs. "Click here to log in".

And similarly with the rest of these it seems mostly like a difference in opinion. If it's more commonly understandable then it seems reasonable to change it, but that doesn't seem to be the case here.

omarroth:
@Zero3K  The problem with "Shared `x` ago" is caused by #373. The issue is that different languages have different rules for pluralization, so for example Russian (as @dimqua kindly explained to me) has logic that looks like this (for pluralizing "subscribers"):
```crystal
case sub_count
when .ends_with? "1" && !.ends_with? "11"
  "подписчик_"
when .ends_with? "2","3","4" && !.ends_with? "12","13","14"
  "подписчика"
else
  "подписчиков"
end
```

Support for that hasn't been added yet, and I would not expect it to be added in this PR.

comradekingu:
@omarroth 
"Click here to log in", is always "log in". That is the action, and it happens on the "login page". Where one can opt for either, is a button that says "login", alluding to the place it takes you, or "log in", describing the action. As English is more oriented towards engagement and action, and it is less of an arbitration, I opted for the latter.

There is a rationale to all of these, either to save space (info/information), or avoid errors in translation(e-mail/email).

omarroth:
Although I understand the reasoning behind this PR, I'd like to avoid modifying phrases as much as possible, since it creates more work for translators that could be spent elsewhere. I'd like to reduce work for them as much as possible, and I would prefer not to merge changes that significantly affect the accuracy of the translation.

These changes would need to be added to the other locale files to avoid breakage as mentioned by @GauthierPLM, and the code would also need to be updated to use the new translations.

comradekingu:
@omarroth I am a translator, and have been for quite a while. Nobody has ever complained about good changes being made. Translators want to see good language, and this is an improvement thereof. Since this is on Weblate, it will just be fuzzy, so nothing will break. I can unfuzzify the stuff to do with spelling for other languages.

locales/en-US.json
```diff
@@ -30,19 +30,19 @@
   "History": "History",
   "Previous page": "Previous page",
   "An alternative front-end to YouTube": "An alternative front-end to YouTube",
-  "JavaScript license information": "JavaScript license information",
+  "JavaScript license info": "JavaScript license info",
```
omarroth: I'd like to have this match the [official template for LibreJS], which uses:
```html
<a href="/about/javascript" rel="jslicense">JavaScript license information</a>
```

locales/en-US.json
```diff
@@ -135,15 +135,15 @@
...
   "Playlist does not exist.": "Playlist does not exist.",
   "Could not pull trending pages.": "Could not pull trending pages.",
   "Hidden field \"challenge\" is a required field": "Hidden field \"challenge\" is a required field",
   "Hidden field \"token\" is a required field": "Hidden field \"token\" is a required field",
-  "Invalid challenge": "Invalid challenge",
-  "Invalid token": "Invalid token",
-  "Invalid user": "Invalid user",
+  "Erroneous challenge": "Invalid challenge",
+  "Erroneous token": "Invalid token",
+  "No such user": "No such user",
```
omarroth: Appears the translation doesn't match the updated text, should be:
```
  "Erroneous challenge": "Erroneous challenge",
  "Erroneous token": "Erroneous token",
  "No such user": "No such user",
```

omarroth:
My apologies for not responding sooner. I've had a chance to take a fresh look at this and suggested a couple changes. Everything else looks ready to merge.

As mentioned I'd like to avoid any breakage so be sure the other locales use the updated text, as well as any spots in the code. If you would like I'll take a look through and commit to this branch with the necessary changes.

I've made you an admin on weblate. Thanks for being so patient.

omarroth:
Updated the other locale files and code. Let me know if I've missed anything. Otherwise I'll consider this ready to merge :+1:

comradekingu:
@omarroth 
Thank you for looking into this.
Looks good to me, apart from possibly making
"youtube": "", YouTube and "reddit": "", Reddit
As per https://en.wikipedia.org/wiki/YouTube and https://en.wikipedia.org/wiki/Reddit

PR #372, opened 2019-02-20
Fixed some localisation

eutampieri:
Yesterday I was tired so I missed a few strings.
Sorry for having committed twice.


omarroth:
No worries. You can also use [Weblate](https://hosted.weblate.org/projects/invidious/translations/) if that's easier for you. LGTM.

PR #387, opened 2019-03-02
A tiny fix for Russian translation

dimqua:
Before:

After:

The default is `10em`.


omarroth:
Doesn't appear to be working for me, would it make more sense to have something like:
```css
.pure-control-group label {
  word-wrap: normal;
}
```

PR #403, opened 2019-03-07
Add alternate link with rss feed to playlist page

em92:
Palemoon and old versions of firefox show feed icon in url bar (see img, right side) if there is `<link rel="alternate" type="application/rss+xml"`
http://i.imgur.com/DW5ygKC.png
This makes more convenient to get playlist feed link.

I didn't compile/deploy invidious in my pc, so I didn't test it. So feel free to edit this PR.

Offtopic: project is amazing! Keep up good work.


omarroth:
LGTM :+1:. I'll also add alt links for channels and subscriptions in a subsequent commit.

PR #410, opened 2019-03-12
Responsive Side Menu

tmiland:
A layout example with a side menu that hides on mobile.

https://purecss.io/layouts/side-menu/

Did a mock-up of a side menu. I kind of miss it, especially on desktop/laptop where some of the buttons are hard to hit (small surface area). 

Not done anything for light/dark mode, just thought I'd throw the idea out here. 

Screenshot at 16-06-08


tmiland:
Check it out: https://invidiou.sh 🤓

Perflyst:
Looks good but it shouldn't be opened by default. You also cannot hide it!

tmiland:
> Looks good but it shouldn't be opened by default. You also cannot hide it!

Thanks.

That's default behaviour.

>  /*
> Hides the menu at `48em`, but modify this based on your app's needs.
> */

omarroth:
Looks very nice. Unfortunately it appears to have bullet points in Chrome:
image.

How would you like this PR to be handled? It's definitely a nice idea. I'm currently not sure how well it would be integrated into the rest of the site. Since it has links to subscriptions, trending, etc. would it make sense to remove the feed menu along the top? Also as mentioned, it should be possible to hide it. Using:

```diff
diff --git a/assets/css/side-menu.css b/assets/css/side-menu.css
index 7572752..ca514ac 100644
--- a/assets/css/side-menu.css
+++ b/assets/css/side-menu.css
@@ -207,7 +207,7 @@ small screens.
 /*
 Hides the menu at `48em`, but modify this based on your app's needs.
 */
-@media (min-width: 48em) {
+@media (min-width: 200em) {
 
     .header,
     .content {
```
Makes it hidden by default, although if possible it would make sense to remove the `@media` for the given styling.

src/invidious/views/template.ecr
```diff
@@ -28,118 +29,205 @@
...
+            <%= translate(locale, "History") %>
           </a>
-        </div>
-        <div class="pure-u-1-3">
-          <a title="<%= translate(locale, "Preferences") %>" href="/preferences?referer=<%= env.get?("current_page") %>" class="pure-menu-heading">
-            <i class="icon ion-ios-cog"></i>
+        </li>
+        <li class="pure-menu-item menu-item-divided">
+          <a title="<%= translate(locale, " Preferences") %>" href="/preferences?referer=
+            <%= env.get?("current_page") %>" class="pure-menu-link">
```
omarroth: This should be `href="/preferences?referer=<%= env.get?("current_page") %>"`, since otherwise the given URL contains the leading spaces, so it links to `/preferences?referer=             %2Ffeed%2Fsubscriptions`. If you want, you can still spread the attributes over multiple lines, something like:
```ecr
<a title="<%= translate(locale, " Preferences") %>"
href="/preferences?referer=<%= env.get?("current_page") %>"
class="pure-menu-link">
```

src/invidious/views/template.ecr
```diff
@@ -28,118 +29,205 @@
...
+            <div class="pure-u-1-4">
+              <a title="<%= translate(locale, " Preferences") %>" href="/preferences?referer=
+                <%= env.get?("current_page") %>" class="pure-menu-heading">
+                <i class="icon ion-ios-cog"></i>
+              </a>
+            </div>
+            <div class="pure-u-1-4">
+              <a href="/signout?referer=<%= env.get?(" current_page") %>&token=
+                <%= env.get?("token") %>&challenge=
+                <%= env.get?("challenge") %>" class="pure-menu-heading">
```
omarroth: See above.

src/invidious/views/template.ecr
```diff
@@ -28,118 +29,205 @@
...
+                <% if notification_count > 0 %>
+                <%= notification_count %> <i class="icon ion-ios-notifications"></i>
+                <% else %>
+                <i class="icon ion-ios-notifications-outline"></i>
+                <% end %>
+              </a>
+            </div>
+            <div class="pure-u-1-4">
+              <a title="<%= translate(locale, " Preferences") %>" href="/preferences?referer=
+                <%= env.get?("current_page") %>" class="pure-menu-heading">
```
omarroth: See above.

src/invidious/views/template.ecr
```diff
@@ -28,118 +29,205 @@
...
+                <% if env.get?("preferences").try &.as(Preferences).dark_mode %>
+                <i class="icon ion-ios-sunny"></i>
+                <% else %>
+                <i class="icon ion-ios-moon"></i>
+                <% end %>
+              </a>
+            </div>
+            <div class="pure-u-1-3">
+              <a title="<%= translate(locale, " Preferences") %>" href="/preferences?referer=
+                <%= env.get?("current_page") %>" class="pure-menu-heading">
```
omarroth: See above.

tmiland:
> Looks very nice. Unfortunately it appears to have bullet points in Chrome:
> image.
> 
> How would you like this PR to be handled? It's definitely a nice idea. I'm currently not sure how well it would be integrated into the rest of the site. Since it has links to subscriptions, trending, etc. would it make sense to remove the feed menu along the top? Also as mentioned, it should be possible to hide it. Using:
> 
> ```diff
> diff --git a/assets/css/side-menu.css b/assets/css/side-menu.css
> index 7572752..ca514ac 100644
> --- a/assets/css/side-menu.css
> +++ b/assets/css/side-menu.css
> @@ -207,7 +207,7 @@ small screens.
>  /*
>  Hides the menu at `48em`, but modify this based on your app's needs.
>  */
> -@media (min-width: 48em) {
> +@media (min-width: 200em) {
>  
>      .header,
>      .content {
> ```
> 
> Makes it hidden by default, although if possible it would make sense to remove the `@media` for the given styling.

I'm not exactly sure. The content of the menu is just a mock-up.

My thoughts though are to have the various links gathered in the menu, like youtube. 

Not sure how to handle the above reviews, do i just mark them as resolved, and it will be merged?

I have added the above suggestions to the css, and Removed bullet points in Chrome.

Pushed it to the site just now. 🙂

src/invidious/views/template.ecr
tmiland: Seems like this was the Atom "beautifier" who broke this. Should be fixed now.

Perflyst:
I wonder if you can design this so that we have a bit more youtube like layout. The videos are currently very big at the front page. If you make the videos a bit smaller you could expand it on the startpage but hide it everywhere else by default.

It also overwrites the `Invidious` on the top. Maybe you can really take a look at youtube.com how they expand and hide it.

Your side menu is also dark design while the startpage is white.

tmiland:
> I wonder if you can design this so that we have a bit more youtube like layout. The videos are currently very big at the front page. If you make the videos a bit smaller you could expand it on the startpage but hide it everywhere else by default.
> 
> It also overwrites the `Invidious` on the top. Maybe you can really take a look at youtube.com how they expand and hide it.
> 
> Your side menu is also dark design while the startpage is white.

Thanks for the feedback @Perflyst 

Check out the latest changes. :smile:

dimqua:
IMHO, if you're going to add `Preferences` link to the side menu, it should be removed from the top menu (or vice versa).

PR #412, opened 2019-03-14
French translation updated, some translation restored

TheFrenchGhosty:
I restored some of the lines that were edited in https://github.com/omarroth/invidious/pull/380 (since we were okay about those lines @omarroth ) and updated it by translating the new lines.

(This time I didn't forget to add spaces before the punctuation (I'm too used to write English)


omarroth:
LGTM :+1:

PR #436, opened 2019-03-25
French translation updated - More consistency

TheFrenchGhosty:
I added more consistency in the translation especially on the sentences "Written Like That": no more: "Lire Automatiquement" and "Lire automatiquement la vidéo suivante" the next line.

I also edited some of the lines that were badly written, and fixed the translation of "Default captions" since "Fallback captions" is not used anymore

PS : In case someone else ever contribute to the French translation : The lines that are not translated (IE: The translation is written in English) are the lines that are not used / not supposed to be seen.


omarroth:
LGTM :+1:

PR #451, opened 2019-03-27
French translation updated - New words translated, even more consistency

TheFrenchGhosty:
The new words have been translated, and I added even more consistency to the translation by using some of the translation used on youtube itself.


omarroth:
LGTM :+1:

Issue #277, opened 2019-01-05
[Suggestion] Autoselect the search field when loading the page


elypter:

omarroth:
I'm not opposed to this, but is there a reason this should be preferred over the default behavior? YouTube doesn't appear to do this, and I'm having trouble 

YouTube and Github both have a `/` hotkey to select the search bar, so maybe it would be good to have this as an option available in Invidious.

elypter:
i dont know why youtube doesnt do this but just because they have a lot of money that doesnt mean their ux is always the best. the popularity of this site is the best proof. maybe they value looks over function so much that they dont want a blinking cursor, maybe they want to viewers into their suggested content, maybe they want to discourage the classic desktop expierience, maybe its just a lack of thought. 

i think as long as the keyboard is not used for anything else without having to use the mouse first there is no real reason against it. on search engines this is the default behavior too and search is a very frequently used function on invidio.us as well judging from my own expierience.

elypter:
related to #513

still relevant in combination with #449 /  #996

SamantazFox:
bump

markozajc:
This actually gives me some trouble; I usually press 'space' to start video playback, since I have autoplay disabled. Autofocus on the search field instead redirects the keypress there, which is a bit annoying. It would be great if this was optional/disabled on video pages.

Issue #501, opened 2019-04-18
Add shortcuts (same as on YouTube)


Perflyst:
YouTube has a lot of shortcuts, you can find a list [here](https://support.google.com/youtube/answer/7631406)

omarroth:
Invidious already supports quite a few of these, `spacebar`, `k`, `j`, `l`, `f`, `m`, other player shortcuts should be fairly straightforward. Invidious also supports changing playback rate with `Shift+<` and `Shift+>` which does not appear to be listed.

Several others also appear not to be listed that appear with `Shift+?` on the YT player.

Perflyst:
Didn't noticed that as the shortcuts only work if the player is "focused". If I have multiple windows opened and click back to the browser window which is opened than shortcuts dont  work.

Also noticed that l and j only jumps 5 seconds, not 10.
An interesting shortcut is `Shift` and a number (0, 1-9) to skip the a specifc timestamp in the video

alexgleason:
My partner and I switched to Invidious on our HTPC. She keeps smashing spacebar to pause during a conversation, and says "dammit" when it scrolls to the comments.

Would be awesome if spacebar paused the video regardles of whether it's focused! Thank you.

Perflyst:
YouTube also does not pause the video and scrolls to the comments if not focused. This is normal behaviour, at least for me.

alexgleason:
Hmm I think maybe it's because on YouTube the video stays fullscreen between videos. We've gotten used to leaving it not fullscreen since the autoplay feature causes the page to refresh. Maybe this is why the video stays focused on YouTube but not Invidious.

omarroth:
I think on YouTube the player is focused by default. Likely would make sense for Invidious to behave similarly.

practik:
> I think on YouTube the player is focused by default. Likely would make sense for Invidious to behave similarly.

I believe on YouTube it's only focused after the video has started playing, so no autoplay = no keyboard shortcuts. The ability to have shortcuts enabled with autoplay disabled was one of my favorite things about Iridium (https://github.com/ParticleCore/Iridium/wiki/Features#shortcuts_always_active), and I'd love to see it become a part of Invidious as well.

nikolayhg:
The YouTube shortcuts work no matter whether the focus is on the video or on the title, description or comments - which of course is very user friendly. Would be nice, if invidious behaves the same.

omarroth:
Implemented by #678.

Issue #449, opened 2019-03-27
[Feature Request] SearchEngine-like page


TheFrenchGhosty:
Having a page that looks like a search engine, including only a search box and some basic stuff might be a good idea.

Example (Searx) : https://i.imgur.com/i18iPGd.jpg

omarroth:
See also #295.

PopeRigby:
Why would you do that exactly?

omarroth:
You can now do this now by setting "Default home" to "" in your preferences. You can also remove the feed menu by setting all entries in "Feed menu" to "".

TheFrenchGhosty:
Amazing, thanks for that.

One more thing, can you add a way to have it centered vertically (like SearX or every other search engine)?

PR #513, opened 2019-04-28
Autofocus search input

jorgesumle:
Because I don't like wasting my time using the mouse to focus on it.

### DIFF
--- a/src/invidious/views/template.ecr
+++ b/src/invidious/views/template.ecr
@@ -38,7 +38,7 @@
                 <div class="pure-u-1 pure-u-md-12-24 searchbar">
                     <form class="pure-form" action="/search" method="get">
                         <fieldset>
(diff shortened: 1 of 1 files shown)


omarroth:
Would appear to close #277.

As mentioned in the linked issue I would very much prefer for this to be implemented as a shortcut (see #501), as it is on YouTube.

jorgesumle:
A shortcut means one more keystroke, which means more wasted time, and very few people would actually use it because many don't know it exists. With autofocus, you don't have to select the field before you start typing to search a video. As a compromise for those who don't like that, a configuration setting could be implemented.

omarroth:
I don't like this as the default behavior because it's assuming the user's intent. 

The only services I can find that have the search bar autofocused by default are search engines, where it makes much more sense to assume a user wants to search. Even then, only the landing page has the search bar auto-focused.

There's an open issue #449 for having a search page as the default landing page, where it would make sense to have it auto-focused.

> very few people would actually use it because many don't know it exists

GitHub provides a *hint* as a `/` icon, so I could see that as a good option here as well.

jorgesumle:
OK, I understand.

PR #592, opened 2019-06-13
Add Eepsite

ghbjklhv:
Add Eepsite to README.md


Perflyst:
This is not an eepsite from omarroth, I would not include it in the readme

omarroth:
I agree it makes sense to keep a list of instances in the README. Maybe it would be good to make the link to alternative instances more prominent?

Perflyst:
Yes, I would prefer a more prominent link to alternative instances.

omarroth:
Added "Invidious Instances" header to README with 6783bf9903aaa810e626864fab73be7ee29d0f0f.

PR #598, opened 2019-06-17
let docker listen to 127.0.0.1 by default

tleydxdy:


omarroth:
Thanks! I'm a bit confused though, `docker-compose up` should automatically listen on `127.0.0.1`. Am I mistaken here or is there another reason to make this explicit?

psvenk:
On my system, `docker-compose up` already listens on `127.0.0.1` or on `localhost`—both point to Invidious, and this patch does not change anything. However, [this article](https://www.maketecheasier.com/localhost-and-127-0-0-1/) suggests that `localhost` is not necessarily mapped to `127.0.0.1` (if `/etc/hosts` is missing the line `127.0.0.1 localhost`). So, I am guessing that on @tleydxdy's system, this line is missing from the `/etc/hosts` file, and docker listens on `localhost` (which is not mapped to `127.0.0.1`) by default. I would imagine that this would not be an issue if one types in `localhost:3000` as recommended in the README. If the OS doesn't even recognize `localhost` at all, then any software that assumes that `localhost` exists will inevitably break anyway, and the only difference would be that Invidious does not assume that `localhost` exists.

tleydxdy:
my debian system listens to 0.0.0.0 which is kinda creepy, and thus this pr

psvenk:
I just saw that this pull request makes Invidious listen on 127.0.0.1, and on no other addresses. This became a problem for me, as I needed to access Invidious over my LAN for testing purposes but Invidious would refuse connections directed at my computer's local IP address. Is there any way to simultaneously address the concerns of @tleydxdy and me?

Perflyst:
Just revert this change locally? ;)

psvenk:
I did revert this change locally. I brought this up because I figured that others may also be facing the same issue and that changing the default setting might be beneficial depending on how many people are affected and in what capacity.

tleydxdy:
it's better to lock things down by default and open it if you want that imo.

PR #620, opened 2019-07-04
Add translation of Simplified Chinese

outloudvi:
Here is the interface translation of `zh-CN` (or say, `zh-hans`). Reviews and suggestions welcomed.


omarroth:
LGTM. Very much appreciated, thanks!

PR #654, opened 2019-07-31
Fix engagement for zero-view videos

leonklingele:
Division by zero resulted in 'NaN'.

Fixes https://github.com/omarroth/invidious/issues/653.


omarroth:
LGTM, thanks!

PR #655, opened 2019-07-31
Add Travis CI and pin dependencies

leonklingele:


leonklingele:
.. currently only checking if the project builds fine to detect broken commits.

leonklingele:
Added build status to `README.md`.

omarroth:
When creating a new project, the Crystal boilerplate includes this `.travis.yml` by default:

```yaml
language: crystal

# Uncomment the following if you'd like Travis to run specs and check code formatting
# script:
#   - crystal spec
#   - crystal tool format --check
```

Would it make more sense to use that here?

See also [the relevant page in Crystal docs](https://crystal-lang.org/reference/guides/writing_shards.html).

leonklingele:
Now also running unit tests and code formatter.

leonklingele:
Job passing: https://travis-ci.org/leonklingele/invidious

PTAL @omarroth

.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: crystal
+
+crystal:
+  - latest
+  - nightly
```
omarroth: Mind providing the reasoning for also testing on nightly? It seems like it could misreport a build as failing if breaking changes are introduced in nightly.

leonklingele: If the CI job fails on `nightly` (e.g. because some Crystal API has been removed / changed behavior), there might be a solution to support both `latest` _and_ `nightly` (e.g. by using some other future-proof API). If the CI job fails on `nightly` because the whole `nightly` build env is broken (when Crystal devs have screwed up), the failure could simply be ignored until a fixed `nightly` is available. I assume testing on `nightly` makes it easier to support upcoming Crystal releases and improve overall code quality. I'm new to Crystal though and don't know how frequently breaking changes are introduced to the language.

omarroth: You may consider using `crystal build src/invidious.cr --error-on-warnings` in that case instead of two builds on `latest` and `nightly`.

Otherwise looks good. If you'd like to mark as resolved I'll consider this ready to merge :+1:.

leonklingele: Building with `--error-on-warnings` now and removed `nightly`.

leonklingele:
CI failed: https://travis-ci.org/leonklingele/invidious/builds/567181117#L219

```
$ shards update
Fetching https://github.com/will/crystal-pg.git
Fetching https://github.com/crystal-lang/crystal-db.git
Fetching https://github.com/crystal-lang/crystal-sqlite3.git
Error resolving db (~> 0.5.0, ~> 0.6.0)
The command "shards update" failed and exited with 1 during .
```

Latest `master` is broken as well. Do we need to pin some dependency to fix this?

leonklingele:
I've opened https://github.com/will/crystal-pg/pull/179 which might fix the issue.

leonklingele:
@omarroth what about pinning all our dependencies?

```
diff --git a/shard.yml b/shard.yml
index 3e2b3d0..b1b500d 100644
--- a/shard.yml
+++ b/shard.yml
@@ -11,10 +11,13 @@ targets:
 dependencies:
   pg:
     github: will/crystal-pg
+    version: ~> 0.17.0
   sqlite3:
     github: crystal-lang/crystal-sqlite3
+    version: ~> 0.12.0
   kemal:
     github: kemalcr/kemal
+    version: ~> 0.25.2

 crystal: 0.29.0

```

omarroth:
Pinning dependencies would be an excellent addition to this PR.

leonklingele:
Updated.

omarroth:
LGTM

leonklingele:
Just for the records: Now that dependencies are pinned to a specific version, we need to regularly check for new releases of the dependencies. In such a case, we might need to bump the dependency's version if the new version includes, for example, security-related changes.
I will take care of this using an automated fashion (https://github.com/leonklingele/github-release-checker), but it would be good if I'm not the only one :)

PR #424, opened 2019-03-21
Centered the "Play" button on the video thumbnail

Leroux47:
If the Play arrow is centered, it will be more accessible and create a more Youtube-like experience.

PR #661, opened 2019-08-01
Center video player play button

leonklingele:
Before:

before

After:

after


omarroth:
As mentioned in #424 I would prefer for this to be added as a preference. If you'd like I can make the appropriate changes to this PR before merging.

leonklingele:
Which of the layouts would you want to use as a default? If you prefer the current style (i.e. button off-center), I'd rather close this PR instead of beginning to clutter the settings with such cosmetic options and provide some *monkey user script instead.

omarroth:
I would prefer to keep the current layout as default for the reasons mentioned [here](https://github.com/omarroth/invidious/pull/424#issuecomment-475457393). If you'd like to implement this as a userscript instead I'd be happy to add it to [Extensions](https://github.com/omarroth/invidious/wiki/Extensions).

omarroth:
Closing in favor of #680.

PR #662, opened 2019-08-01
docker: fix port forwarding

leonklingele:
Previous version resulted in ERR_CONNECTION_REFUSED.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       dockerfile: docker/Dockerfile
     restart: unless-stopped
     ports:
-      - "127.0.0.1:3000:3000"
+      - "3000:3000"
     depends_on:
       - postgres
 


omarroth:
It looks like this reverts #598. Is there a solution that solves both the mentioned issue and the issue addressed by @tleydxdy?

leonklingele:
@omarroth unfortunately, I don't think there is a way. I was trying to use it on a Mac, which doesn't natively support running Docker but only through a VM (`boot2docker`). Binding to `127.0.0.1:3000` will bind to the local interface of that VM, not to the interface of the host (Mac). I've not found a way to set up proper port forwarding rules inside VirtualBox to circumvent this.
Anyway, I will now simply apply this patch locally only as I too prefer to be as restrictive as possible and never bind to "public" interfaces.

PR #665, opened 2019-08-01
docker: various improvements to Dockerfile

leonklingele:
This includes the following changes:

- Use multi-stage build to run application in an optimized environment, see
  https://docs.docker.com/develop/develop-images/multistage-build/
- Run application on alpine instead of archlinux to further reduce image size
- Build Crystal application with --release for improved runtime performance
- Run application as non-root user for better security, see
  https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
- Only rebuild Docker layers when required

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,15 +1,28 @@
-FROM archlinux/base
-
-RUN pacman -Sy --noconfirm shards crystal imagemagick librsvg \
-    which pkgconf gcc ttf-liberation glibc
-# base-devel contains many other basic packages, that are normally assumed to already exist on a clean arch system
-
-ADD . /invidious
-
+FROM alpine:latest AS builder
+RUN apk add -u crystal shards libc-dev \
+    yaml-dev libxml2-dev sqlite-dev sqlite-static zlib-dev openssl-dev
 WORKDIR /invidious
+COPY ./shard.yml ./shard.yml
+RUN shards update && shards install
+COPY ./src/ ./src/
+# TODO: .git folder is required for building – this is destructive.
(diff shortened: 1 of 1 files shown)


leonklingele:
Please test this extensively.

Image size of `invidious_invidious` decreased by over 96% (from 1.13GB down to 38.6MB).

leonklingele:
Added missing runtime dependency `imagemagick` (required for CAPTCHAs).

omarroth:
Alpine doesn't appear to include any default fonts, so the CAPTCHA looks like this:
image

It's enough to include `opensans-ttf` or an alternative font:
```diff
diff --git a/docker/Dockerfile b/docker/Dockerfile
index 0c30616..c9fa636 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17 +17 @@ FROM alpine:latest
-RUN apk add -u imagemagick
+RUN apk add -u imagemagick ttf-opensans
```
Which appears to fix the issue:
image

leonklingele:
Thanks! Updated.

omarroth:
LGTM :+1:

leonklingele:
@omarroth I'd really like to remove the `.git` folder from the Docker build env, see https://github.com/omarroth/invidious/pull/665/files#diff-ebacf6f6ae4ee68078bb16454b23247dR8-R10
This requires to rebuild the Crystal application for every new commit, even if the commit doesn't touch anything related to Crystal.
Do you know of a way to inject variables at build time, similar to Go's `-ldflags "-X main.version=1.2.3"` (https://stackoverflow.com/a/11355611)?

leonklingele:
Ignore what I just said. Rebuilding is required to update the version information shown on the web UI.

PR #675, opened 2019-08-07
Add Postgres health check

Dragnucs:


docker-compose.yml
```diff
@@ -7,6 +7,8 @@ services:
     restart: unless-stopped
     volumes:
       - postgresdata:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
```
leonklingele: You might want to configure the following directives as well, otherwise some defaults might be used:

- `interval`
- `timeout`
- `retries`
- `start_period`

See https://docs.docker.com/compose/compose-file/#healthcheck

Dragnucs: Default values are used when those directives are missing. Currently, from what I seen and tested, they are fair enough for Postgres and most software.

omarroth:
Thanks!

PR #679, opened 2019-08-08
Change font family to better native selection

Dragnucs:
This set the used font to a better native selection. No font needs to be downloaded. The rendered font will depend on the browser and platform.


omarroth:
Thanks!

PR #697, opened 2019-08-14
shard: track dependencies

leonklingele:
Commit the whole ./lib/ folder which stores the Crystal dependencies.
This has a few benefits:

- Allows to build the project without a connection to the Internet
  to retrieve dependencies.
- Makes the project resistant against dependency re-tags which might
  include malicious code.


leonklingele:
How to verify that I'm not trying to inject code:

```sh
$ cd $(mktemp -d)
$ git clone https://github.com/omarroth/invidious && cd invidious/
$ git fetch origin pull/697/head:leonklingele/shards-commit-vendor
$ git checkout leonklingele/shards-commit-vendor
$ rm -rf ./lib/ shard.lock
$ shards update ; shards install
$ git status
On branch leonklingele/shards-commit-vendor
nothing to commit, working tree clean
```

But don't trust me on that ;)

leonklingele:
Build will pass once https://github.com/omarroth/invidious/pull/698 is merged. Sorry for the trouble.

leonklingele:
@omarroth would you be willing to accept and merge an updated version of this?

omarroth:
Probably not. Currently [lsquic.cr](https://github.com/omarroth/lsquic.cr) is quite bloated (~30MB IIRC) which is an issue I'd like to resolve before merging something like this.

saltycrys:
I don't think tracking the dependencies will ever be possible.

Even if lsquic.cr removes the binary from the repo and instead downloads it in a shards `postinstall` step the `shards` command will still need to be run to download it. So building without an internet connection isn't possible without a large binary in the repo.

To protect against dependency re-tags it would be possible to pin dependencies to commits but that would be annoying to update and not worth it in my opinion.

PR #698, opened 2019-08-14
docker: use alpine:edge base image for building

leonklingele:
This fixes currently failing Docker builds.
kemalcr/kemal in version 0.26.0 requires Crystal 0.30.0 which is not
yet available on Alpine 3.10 (previously used as the Docker base image).


omarroth:
:+1:

PR #704, opened 2019-08-18
js: add support to detect alt, meta and control key in keydown handler

leonklingele:
This fixes a quite severe user experience issue where pressing the
'alt', 'meta' and/or 'ctrl' key along with one of the supported keys
(e.g. 'f' to enter video fullscreen mode) would overwrite the default
browser behavior. In the case of 'f+meta' we would enter fullscreen
mode, and not open the browser search panel as one might expect.

This change is required to stay consistent with the way YouTube
handles keydown events.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -327,8 +327,13 @@ window.addEventListener('keydown', e => {
     let action = null;
 
     const code = e.keyCode;
-    const key = e.key;
-    switch (key) {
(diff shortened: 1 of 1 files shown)


leonklingele:
Updated to also detect the `alt` and `ctrl` keys.

omarroth:
:+1:

Issue #712, opened 2019-08-19
fastforward and rewind broken in ps4 browser


Djang007:
Sometimes when I'm away from my PC I like to use invidious on my PS4 to watch HD videos.  

I use the PS4's standard browser because the official YouTube app sucks and invidious lets me watch region locked videos.  

Lately I've noticed that something has changed and I can no longer use the D-pad on the PS4 controller to fast-forward and rewind videos whenever I watch them in the ps4 browser. Is anyone else getting the same problem on their ps4?

omarroth:
Would you mind testing on https://dev.invidio.us to confirm that #713 fixes the issue for you?

Djang007:
Sorry for the late reply, I've been busy. I checked to see if the fix worked at https://dev.invidio.us and the issue still persist. 

I did notice however that I could use the D-pad controls when manually clicking on the play bar icon. Whatever is causing the bug it seems to happen when going into full screen mode.  Can anyone with a ps4 see if they have the same issue? That way we would be able to tell if it's a bug or something on my end causing the issue.

TheFrenchGhosty:
Reopening.

As reported in the previous message #713 doesn't fix this issue.

PR #713, opened 2019-08-19
js: add support to detect media keys in keydown handler

leonklingele:
See [0] for all the relevant codes.

[0]: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#Multimedia_keys

Fixes a regression introduced in e6b4e1268945777c5d07dfca4362a1af23f6d970.
Fixes https://github.com/omarroth/invidious/issues/712.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -228,11 +228,24 @@ function set_time_percent(percent) {
     player.currentTime(newTime);
 }
 
+function play() {
+    player.play();
+}
+
+function pause() {
+    player.pause();
+}
+
+function stop() {
+    player.pause();
+    player.currentTime(0);
+}
+
 function toggle_play() {
     if (player.paused()) {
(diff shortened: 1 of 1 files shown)


leonklingele:
Please note that the patch is untested as I don't have access to a device with media keys.

@omarroth could you please deploy this change to the Invidious dev Instance so @Djang007 can test it?

omarroth:
Testing locally LGTM. Deployed on https://dev.invidio.us if @Djang007 would like to confirm.

omarroth:
Although this doesn't close #712, it does work as expected with media keys, so if you'd like I'm happy to merge as is.

leonklingele:
Rebased. Feel free to merge this @omarroth! :)

TheFrenchGhosty:
Merged.

PR #727, opened 2019-08-28
removed trailing space "Family friendly?"

leela52452:


omarroth:
The trailing space is used as formatting:
image

This change also breaks handling with other locales, since it's referred to in code with the trailing space.

PR #729, opened 2019-08-29
Provide db user in migrate-db-3646395.sh

Infinisil:
Database user was missing from this single command in the migration scripts.


omarroth:
Thanks!

PR #733, opened 2019-08-30
change docker-compose to api-only

greentornado:
Modify docker-compose file to start with indivious api only

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,7 @@
 FROM alpine:edge AS builder
 RUN apk add -u crystal shards libc-dev \
-    yaml-dev libxml2-dev sqlite-dev sqlite-static zlib-dev openssl-dev
+    yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,21 +1,12 @@
...
       context: .
       dockerfile: docker/Dockerfile
     restart: unless-stopped
     ports:
       - "127.0.0.1:3000:3000"
-    depends_on:
-      - postgres
 
 volumes:
   postgresdata:
```
omarroth: This should also be removed, otherwise it will create a default network, etc. for unneeded `postgresdata`.

.gitignore
```diff
@@ -7,3 +7,5 @@
 /invidious
 /sentry
 shard.lock
+invidious.o
+invidious.dwarf
```
omarroth: This seems like it should go in it's own PR.

omarroth:
Thanks! Left a couple comments, otherwise LGTM :+1:.

greentornado:
Yeah I made changes followed to your comments @omarroth

omarroth:
Thanks!

PR #734, opened 2019-08-30
fix docker build

tleydxdy:


omarroth:
LGTM. I'm a bit confused as to why static objects aren't included in either `openssl` or `openssl-dev`, since it's listed as a sub-package in the Alpine repository.

PR #743, opened 2019-09-08
Forward query string parameters from Atom feeds

girst:
Any query string parameters (except private's token=) are appended to
/watch URLs.  This allows e.g. passing &raw=1&listen=1 to a playlist of
music and use an rss reader like newsboat as a media player.

All three feeds--channels, playlists, subscriptions--are supported.

Feedback welcome!

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -2630,6 +2630,8 @@ get "/feed/channel/:ucid" do |env|
 
   ucid = env.params.url["ucid"]
 
+  params = HTTP::Params.parse(env.params.query["params"]? || "")
+
   begin
     channel = get_about_info(ucid, locale)
   rescue ex
@@ -2690,7 +2692,7 @@ get "/feed/channel/:ucid" do |env|
       end
 
       videos.each do |video|
-        video.to_xml(host_url, channel.auto_generated, xml)
+        video.to_xml(host_url, channel.auto_generated, params, xml)
       end
     end
   end
@@ -2721,6 +2723,8 @@ get "/feed/private" do |env|
   page = env.params.query["page"]?.try &.to_i?
   page ||= 1
 
+  params = HTTP::Params.parse(env.params.query["params"]? || "")
+
   videos, notifications = get_subscription_feed(PG_DB, user, max_results, page)
   host_url = make_host_url(config, Kemal.config)
 
@@ -2734,7 +2738,7 @@ get "/feed/private" do |env|
       xml.element("title") { xml.text translate(locale, "Invidious Private Feed for `x`", user.email) }
 
       (notifications + videos).each do |video|
-        video.to_xml(locale, host_url, xml)
+        video.to_xml(locale, host_url, params, xml)
       end
     end
   end
@@ -2747,6 +2751,8 @@ get "/feed/playlist/:plid" do |env|
 
   plid = env.params.url["plid"]
 
+  params = HTTP::Params.parse(env.params.query["params"]? || "")
+
   host_url = make_host_url(config, Kemal.config)
   path = env.request.path
 
@@ -2757,10 +2763,10 @@ get "/feed/playlist/:plid" do |env|
   document.xpath_nodes(%q(//*[@href]|//*[@url])).each do |node|
     node.attributes.each do |attribute|
       case attribute.name
-      when "url"
-        node["url"] = "#{host_url}#{URI.parse(node["url"]).full_path}"
-      when "href"
-        node["href"] = "#{host_url}#{URI.parse(node["href"]).full_path}"
+      when "url", "href"
+        full_path = URI.parse(node[attribute.name]).full_path
+        query_string_opt = full_path.starts_with?("/watch?v=") ? "&#{params}" : ""
+        node[attribute.name] = "#{host_url}#{full_path}#{query_string_opt}"
       end
     end
   end
--- a/src/invidious/channels.cr
+++ b/src/invidious/channels.cr
@@ -41,13 +41,15 @@ struct ChannelVideo
     end
   end
 
-  def to_xml(locale, host_url, xml : XML::Builder)
+  def to_xml(locale, host_url, query_params, xml : XML::Builder)
+    query_params["v"] = self.id
+
     xml.element("entry") do
       xml.element("id") { xml.text "yt:video:#{self.id}" }
       xml.element("yt:videoId") { xml.text self.id }
       xml.element("yt:channelId") { xml.text self.ucid }
       xml.element("title") { xml.text self.title }
-      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}")
+      xml.element("link", rel: "alternate", href: "#{host_url}/watch?#{query_params}")
 
       xml.element("author") do
         xml.element("name") { xml.text self.author }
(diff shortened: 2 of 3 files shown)


src/invidious.cr
```diff
@@ -2701,7 +2701,7 @@ get "/feed/private" do |env|
 
   env.response.content_type = "application/atom+xml"
 
-  token = env.params.query["token"]?
+  token = env.params.query.delete("token")
```
omarroth: You should avoid modifying `env.params.query` since it's used elsewhere (namely for logging). This change for example turns logs from:

```
200 GET /feed/private?token=XXXXX
```
to
```
200 GET /feed/private
```

src/invidious.cr
```diff
@@ -2753,14 +2753,15 @@ get "/feed/playlist/:plid" do |env|
   client = make_client(YT_URL)
   response = client.get("/feeds/videos.xml?playlist_id=#{plid}")
   document = XML.parse(response.body)
+  query_string = env.params.query.map { |s| "&#{s}" }.join
```
omarroth: I don't think this does quite what you expect.

`.map` in this case should accept two arguments: `k` and `v`. Since `env.params.query` is a `Hash(String, Array(String))`, `s` in this case is actually both arguments combined into a tuple. The above code generates:
```crystal
"&{\"param1\", \"value1\"}&{\"param2\", \"value2\"}"
```

Calling `env.params.query.to_s` directly (or preferably as `&#{env.params.query}` when building `node["url"]` below) should produce the result you want.

src/invidious/channels.cr
```diff
@@ -41,13 +41,14 @@ struct ChannelVideo
...
 
-  def to_xml(locale, host_url, xml : XML::Builder)
+  def to_xml(locale, host_url, query_params, xml : XML::Builder)
     xml.element("entry") do
       xml.element("id") { xml.text "yt:video:#{self.id}" }
       xml.element("yt:videoId") { xml.text self.id }
       xml.element("yt:channelId") { xml.text self.ucid }
       xml.element("title") { xml.text self.title }
-      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}")
+      query_string = query_params.map { |s| "&#{s}" }.join
```
omarroth: See above.

src/invidious/search.cr
```diff
@@ -1,11 +1,12 @@
...
-  def to_xml(host_url, auto_generated, xml : XML::Builder)
+  def to_xml(host_url, auto_generated, query_params, xml : XML::Builder)
     xml.element("entry") do
       xml.element("id") { xml.text "yt:video:#{self.id}" }
       xml.element("yt:videoId") { xml.text self.id }
       xml.element("yt:channelId") { xml.text self.ucid }
       xml.element("title") { xml.text self.title }
-      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}")
+      query_string = query_params.map { |s| "&#{s}" }.join
+      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}#{query_params}")
```
omarroth: See above.

omarroth:
Thank you!

I think it makes sense for this to have its own parameter, e.g. `params`, so it doesn't conflict with existing fields used for RSS endpoints (`token`, `max_results`, `page`, etc). Maybe as a URI-escaped field, similar to the `&referer=%2Fwatch...` field used elsewhere.

src/invidious.cr
girst: Oh, I did not know that! Maybe we can call this a *happy accident*: In my opinion, user tokens should not be logged. 

If that's not acceptable, I'll copy env.params.query before modifying it.

girst:
>I think it makes sense for this to have its own parameter, e.g. params, so it doesn't conflict with existing fields used for RSS endpoints (token, max_results, page, etc). Maybe as a URI-escaped field, similar to the &referer=%2Fwatch... field used elsewhere.

Since query parameters are only appended to /watch URLs, I don't think there is much (if any) potential for overlapping params with feed URLs (which are lists). Having it like it is now seems more logical to me. 
I think it is a good idea though to remove any parameters that are understood by the feed, as a page=2 parameter won't make any sense on a watch url.

src/invidious.cr
omarroth: The reason I would like this feature to use a separate parameter is because it removes any need to handle unsupported values (`token`, `page`...) by modifying `env.params.query`.

For example, something like:

```
/feed/private?token=a&params=speed%3D4%26local%3Dtrue
```

Could be consumed like:

```crystal
params = HTTP::Params.parse(env.params.query["params"]? || "")
video.to_xml(locale, host_url, params, xml)
```

girst: good arguments. implemented!

src/invidious/channels.cr
```diff
@@ -41,13 +41,13 @@ struct ChannelVideo
...
 
-  def to_xml(locale, host_url, xml : XML::Builder)
+  def to_xml(locale, host_url, query_params, xml : XML::Builder)
     xml.element("entry") do
       xml.element("id") { xml.text "yt:video:#{self.id}" }
       xml.element("yt:videoId") { xml.text self.id }
       xml.element("yt:channelId") { xml.text self.ucid }
       xml.element("title") { xml.text self.title }
-      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}")
+      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}&#{query_params}")
```
omarroth: This actually can be made a little simpler. Something like:

```diff
diff --git a/src/invidious/channels.cr b/src/invidious/channels.cr
index 6004476..cce9831 100644
--- a/src/invidious/channels.cr
+++ b/src/invidious/channels.cr
@@ -44,0 +45,2 @@ struct ChannelVideo
+    query_params["v"] = self.id
+
@@ -50 +52 @@ struct ChannelVideo
-      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}&#{query_params}")
+      xml.element("link", rel: "alternate", href: "#{host_url}/watch?#{query_params}")
@@ -59 +61 @@ struct ChannelVideo
-          xml.element("a", href: "#{host_url}/watch?v=#{self.id}") do
+          xml.element("a", href: "#{host_url}/watch?#{query_params}") do
, href: "#{host_url}/watch?#{query_params}") do
```

Which will also fix the dangling `&` when `params` is empty.

src/invidious/search.cr
```diff
@@ -1,11 +1,11 @@
 struct SearchVideo
-  def to_xml(host_url, auto_generated, xml : XML::Builder)
+  def to_xml(host_url, auto_generated, query_params, xml : XML::Builder)
     xml.element("entry") do
       xml.element("id") { xml.text "yt:video:#{self.id}" }
       xml.element("yt:videoId") { xml.text self.id }
       xml.element("yt:channelId") { xml.text self.ucid }
       xml.element("title") { xml.text self.title }
-      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}")
+      xml.element("link", rel: "alternate", href: "#{host_url}/watch?v=#{self.id}&#{query_params}")
```
omarroth: Same here.

omarroth:
Missing `params` here:

https://github.com/omarroth/invidious/blob/553d52a45e018bbfc4948e8a61411d9449cbbab4/src/invidious/channels.cr#L50

and here:

https://github.com/omarroth/invidious/blob/553d52a45e018bbfc4948e8a61411d9449cbbab4/src/invidious/search.cr#L22

Left a couple other comments, otherwise should be ready to merge :+1:.

src/invidious/channels.cr
girst: good idea, implemented!

omarroth:
LGTM, thanks!

girst:
Thanks, I really appreciate your patience ;)

PR #752, opened 2019-09-13
[Fix][Docker] Update Package Repository for Install

gnomus:
Fixing a Typo in `docker/Dockerfile`

We want apk to update it's Repository before installing Packages instead of the "prefer upgrade" Flag.

I presumed this is a Typo. If not `-uU` is also possible.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:edge AS builder
-RUN apk add -u crystal shards libc-dev \
+RUN apk add --no-cache crystal shards libc-dev \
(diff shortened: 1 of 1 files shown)


leonklingele:
`--no-cache` is even better: https://github.com/gliderlabs/docker-alpine/blob/1278f26dcac9e84c78c7cc2afdf2a6c16a44d5a0/docs/usage.md#disabling-cache

gnomus:
> `--no-cache` is even better: https://github.com/gliderlabs/docker-alpine/blob/1278f26dcac9e84c78c7cc2afdf2a6c16a44d5a0/docs/usage.md#disabling-cache

good point, saved 2MB :)

omarroth:
Thanks!

PR #765, opened 2019-09-23
docker,travis: fail build on any warning

leonklingele:


leonklingele:
CI will fail due to some problems with Crystal 0.31.0 as described in https://github.com/omarroth/invidious/pull/764.
Merging https://github.com/omarroth/invidious/pull/764 and retriggering a CI build should be enough to make it green again.

omarroth:
Unfortunately looks like this and #766 need to be rebased against master in order to pass CI.

They both LGTM though so I don't mind merging as is.

leonklingele:
Rebased.

omarroth:
LGTM :+1:

PR #789, opened 2019-10-02
Make most files REUSE compliant

### LINKED ISSUE #501: Add shortcuts (same as on YouTube)

Perflyst:
This makes invidious mostly REUSE compliant, not fully because I do not really know from which are all these javascript files compiled from. Maybe you can help here.

Closes #501


Perflyst:
I see now that this is just annoying, if you really want it cherry pick otherwise /close

PR #817, opened 2019-10-12
French Translation updated - Rewording and corrections

TheFrenchGhosty:
Mainly reword the autoplay and play next functions so that they are now finally understandable.

Major rewording for some other lines to increase translation quality.

**Problems coming from Weblate that are now fixed:**

Fix "Hidden field \"challenge\" is a required field" that used guillemets instead of quotation mark.

Revert the change made to "Clear watch history?" to keep the consistency of the wording used by other lines.


omarroth:
Looks great, thanks!

PR #830, opened 2019-10-17
French Translation updated, custom playlists update, enhancements and corrections

TheFrenchGhosty:
Translate the lines added with the custom playlists update

Translate the lines containing questions asking for deletion using context (in a consistent way)

Fix some typos that where introduced with the rewording done in #817

Camel case lines completely rewritten in a normal way (only the lines supposed to use it are using it now)

Massive enhancement for some lines

(Due to the number of commits I think squashing them is the way to go)


omarroth:
Looks great, thank you!

PR #839, opened 2019-10-19
Add protocol to the cryptocurrencies

TheFrenchGhosty:
Most if not all cryptocurrency wallets handle the protocol associated to the cryptocurrency they support.

This is why I think it might be a good idea to use the protocol associated to the cryptocurrencies used for Invidious.

This PR should (hopefully) do this.


omarroth:
Thanks!

PR #851, opened 2019-10-26
Replace languages ISO 639-1 code with the languages name

TheFrenchGhosty:
The languages ISO 639-1 code isn't always the easiest thing to understand especially for people using something other than the Latin alphabet.

The languages names were sourced from: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes and from some Wikipedia pages about the languages.


omarroth:
Thanks!

The 639-1 codes are used for the `&hl=` parameter, which this would break. I think it makes sense to instead have something like:

```crystal
LOCALE_MAP = {
  "en-US" => "English",
  # ...
}
```

Which would be used as a translation text in `preferences`, like:
```crystal
<div class="pure-control-group">
    <label for="locale"><%= translate(locale, "Language: ") %></label>
    <select name="locale" id="locale">
        <% LOCALE_MAP.each_key do |iso, option| %>
            <option value="<%= iso %>" <% if preferences.locale == option %> selected <% end %>><%= translate(LOCALES[iso], option) %></option>
        <% end %>
    </select>
</div>
```

If you'd like I can make the necessary changes to this PR.

TheFrenchGhosty:
Oh yes, I didn't think about url parameters.

I will try to do it later.

TheFrenchGhosty:
I don't have enough time or motivation to do it, so I'm closing this PR.

PR #896, opened 2019-11-11
Fixed double quotes in meta description

### LINKED ISSUE #893: Description has unescaped " in meta tag value, it breaks HTML
image

needs to be escaped. maybe just replace " characters? i'm not familiar with crystal lang so I'm not sure how to do HTML escape

sh4dowb:
fixes #893


omarroth:
Thanks!

PR #925, opened 2019-11-21
Introduce CSS Automatic mode.

### LINKED ISSUE #867: CSS Automatic Dark/Light Mode
Support for automatic switching between light and dark mode with the system's respective modes (macOS, iOS, etc.). Seems to be able to be implemented with "prefers-color-scheme" in CSS.

Any thoughts?

kmcopper:
Fixes #867. Allows OSX in auto mode and manual switching to work properly and without refresh. Ignore if a specific setting has been requested.

### DIFF
--- a/assets/js/themes.js
+++ b/assets/js/themes.js
@@ -28,6 +28,27 @@ window.addEventListener('load', function () {
     update_mode(window.localStorage.dark_mode);
 });
 
+
+var darkScheme = window.matchMedia('(prefers-color-scheme: dark)');
+var lightScheme = window.matchMedia('(prefers-color-scheme: light)');
+
+darkScheme.addListener(scheme_switch);
+lightScheme.addListener(scheme_switch);
+
+function scheme_switch (e) {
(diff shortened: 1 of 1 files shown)


assets/js/themes.js
```diff
@@ -48,7 +69,7 @@ function update_mode (mode) {
 		// If preference for light mode indicated
 		set_mode(false);
 	}
-    else if (document.getElementById('dark_mode_pref').textContent === '' && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+    else if (document.getElementById('dark_mode_pref').textContent === '') {
```
omarroth: Is there a specific reason for removing this line? Currently, this will set the default to dark mode in all cases where no preference is indicated (regardless of `prefers-color-scheme`).

kmcopper: Mostly because it serves no purpose with the new watch event. But I can add it back if you still want it.

omarroth:
Looks great! Thank you!

PR #938, opened 2019-11-26
Proper fix for docker build

tleydxdy:
return to static linking.

It installs openssl for the system, and copies the boringssl libraries to local directory for use by lsquic.cr
Because docker build will crash when running, I wasn't able to test if it causes problem.


tleydxdy:
ye, seems to be a bug in the compiler. Build sometimes succeed, sometimes not.

omarroth:
Thanks!

PR #949, opened 2019-12-06
Verify download, fix invidious file permission

tleydxdy:


docker/Dockerfile
```diff
@@ -29,11 +30,12 @@ RUN apk add --no-cache librsvg ttf-opensans
 WORKDIR /invidious
 RUN addgroup -g 1000 -S invidious && \
     adduser -u 1000 -S invidious -G invidious
-COPY ./assets/ ./assets/
-COPY ./config/config.yml ./config/config.yml
-COPY ./config/sql/ ./config/sql/
-COPY ./locales/ ./locales/
+COPY --chown=invidious ./assets/ ./assets/
```
leonklingele: Why do all those files need to owned by `invidious`?

tleydxdy: I think it is because if the file is owned by root, the invidious user might not be able to read it. and the docker will fail.

tleydxdy: e.g. if only owner can read

docker/Dockerfile
```diff
@@ -30,10 +31,11 @@ WORKDIR /invidious
 RUN addgroup -g 1000 -S invidious && \
     adduser -u 1000 -S invidious -G invidious
 COPY ./assets/ ./assets/
-COPY ./config/config.yml ./config/config.yml
+COPY --chown=invidious ./config/config.yml ./config/config.yml
```
leonklingele, line 34: I'd prefer to have this owned by root, any reason invidious needs write permissions to its config file?

docker/Dockerfile
```diff
@@ -30,10 +31,11 @@ WORKDIR /invidious
...
+RUN sed -i 's/host: \(127.0.0.1\|localhost\)/host: postgres/' config/config.yml
```
leonklingele, line 35: 👍

docker/Dockerfile
```diff
@@ -1,17 +1,18 @@
...
 COPY ./shard.yml ./shard.yml
 RUN shards update && shards install
 RUN apk add --no-cache curl && \
     curl -Lo /etc/apk/keys/omarroth.rsa.pub https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/omarroth.rsa.pub && \
     curl -Lo boringssl-dev.apk https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/boringssl-dev-1.1.0-r0.apk && \
     curl -Lo lsquic.apk https://github.com/omarroth/lsquic-alpine/releases/download/2.6.3-r0/lsquic-2.6.3-r0.apk && \
-    tar -xf boringssl-dev.apk && \
-    tar -xf lsquic.apk
-RUN mv ./usr/lib/libcrypto.a ./lib/lsquic/src/lsquic/ext/libcrypto.a && \
+    apk update && apk verify boringssl-dev.apk lsquic.apk
```
leonklingele, line 12: Do we need to delete some caches after running `apk update`?

leonklingele: Oh, and while you're at it.. can you please combine the `RUN apk add` and `RUN tar -xf` layers and remove the apks once they're unpacked?

docker/Dockerfile
tleydxdy: invidious will write to the config if you change them in the web ui as admin.

leonklingele: Is this something we want to support? The config file is stored in a container after all and is reverted once the container gets recreated.

leonklingele: btw, where can I find the admin UI? ;)

tleydxdy: for example I use it to enable registration temporarily when needed. also the invidious errors out with "looks like you found a bug".

docker/Dockerfile
tleydxdy: done

psvenk:
@tleydxdy Your branch does not build on my machine (with `docker-compose up`), but rebasing it on `omarroth:master` fixed the issue. In case you're unfamiliar, running these commands should rebase your changes:
```
git checkout patch-1
git remote add upstream https://github.com/omarroth/invidious.git
git fetch upstream
git rebase upstream/master
git push -f
```
The rebase should proceed cleanly (with no conflicts).

docker/Dockerfile
```diff
@@ -1,19 +1,19 @@
...
-RUN apk add --no-cache crystal shards libc-dev \
-    yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
-    sqlite-static zlib-static openssl-libs-static
 WORKDIR /invidious
 COPY ./shard.yml ./shard.yml
-RUN shards update && shards install
-RUN apk add --no-cache curl && \
+RUN apk add crystal shards libc-dev \
+    yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
+    yaml-static sqlite-static zlib-static openssl-libs-static && \
```
psvenk, line 6: It seems that `yaml-dev` and `sqlite-dev` are not necessary (from my testing with `docker-compose`).
```suggestion
RUN apk add crystal shards libc-dev \
    libxml2-dev zlib-dev openssl-dev \
    yaml-static sqlite-static zlib-static openssl-libs-static && \
```

tleydxdy: strange, I would think linking would need header files.

psvenk: I think it's because the YAML and SQLite libraries are only statically linked and so the dynamic libraries in `yaml-dev` and `sqlite-dev` are not used.

tleydxdy: I mean yaml.h etc. or am I misunderstanding how linking works?

tleydxdy:
I remember `docker-compose` was changed to not detach because there was builds that was faulty when ran. ref: https://github.com/omarroth/invidious/commit/c80c5631f0483c32c3b86bad3584c349a61b4b92

I think the proper way is to figure out why it wasn't connecting properly

docker/Dockerfile
psvenk: > I mean yaml.h etc. or am I misunderstanding how linking works?

If I understand correctly `yaml.h` is the source code header file which can be compiled to a `.so` for dynamic linking or a `.a` for static linking.

psvenk:
> I think the proper way is to figure out why it wasn't connecting properly

Sounds good; the `docker-compose -d` and `sleep` solution is anyway hacky at best.

I am still confused why running `docker-compose up` without `-d` worked in the first place&nbsp;&mdash;&nbsp;from reading the Travis CI build logs it seems that `docker-compose up` went into the background and returned an exit code once the database became ready, but this is not the behavior that I have expected and observed locally (the server keeps running in the foreground).

Is there any way to run a program (`docker-compose up`) in the foreground until a certain output is emitted ("database system is ready to accept connections" or "database setup finished") and then to disown the process?

docker/Dockerfile
```diff
@@ -1,19 +1,19 @@
 FROM alpine:edge AS builder
-RUN apk add --no-cache crystal shards libc-dev \
-    yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
-    sqlite-static zlib-static openssl-libs-static
 WORKDIR /invidious
 COPY ./shard.yml ./shard.yml
-RUN shards update && shards install
-RUN apk add --no-cache curl && \
+RUN apk add crystal shards libc-dev \
```
leonklingele, line 4: This now requires all packages to be reinstalled once `shard.yml` changes, is this a good idea?

tleydxdy: oops, didn't thought about that. fixed

docker/Dockerfile
```diff
@@ -1,19 +1,19 @@
...
 WORKDIR /invidious
 COPY ./shard.yml ./shard.yml
-RUN apk add crystal shards libc-dev \
+RUN apk add --no-cache curl crystal shards libc-dev \
     yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
-    yaml-static sqlite-static zlib-static openssl-libs-static && \
-    shards update && shards install && \
-    apk add curl && \
+    yaml-static sqlite-static zlib-static openssl-libs-static
+COPY ./shard.yml ./shard.yml
```
psvenk, line 7: Minor nitpick: this line is duplicated in line 3.

docker/Dockerfile
```diff
@@ -1,27 +1,24 @@
...
-    tar -xf boringssl-dev.apk && \
-    tar -xf lsquic.apk
-RUN mv ./usr/lib/libcrypto.a ./lib/lsquic/src/lsquic/ext/libcrypto.a && \
-    mv ./usr/lib/libssl.a ./lib/lsquic/src/lsquic/ext/libssl.a && \
-    mv ./usr/lib/liblsquic.a ./lib/lsquic/src/lsquic/ext/liblsquic.a
+    apk verify --no-cache boringssl-dev.apk lsquic.apk && \
+    tar -xf boringssl-dev.apk usr/lib/libcrypto.a usr/lib/libssl.a && \
+    tar -xf lsquic.apk usr/lib/liblsquic.a && \
+    mv usr/lib/* lib/lsquic/src/lsquic/ext && \
+    rm -r usr boringssl-dev.apk lsquic.apk /root/.cache
```
psvenk, line 15: Thanks for all of the work. Would it be possible to move these lines above `COPY ./shard.yml ./shard.yml` so that these commands can be run from cache even if `shard.yml` is changed?

tleydxdy: they actually overwrite the files shard generates. so if shard.yml changes we need to update and therefore need to overwrite them again.

leonklingele: I'm sorry for nitpicking again, but what about

```patch
@@ -3,16 +3,19 @@ RUN apk add --no-cache curl crystal shards libc-dev \
     yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
     yaml-static sqlite-static zlib-static openssl-libs-static
 WORKDIR /invidious
+RUN curl -Lo /etc/apk/keys/omarroth.rsa.pub https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/omarroth.rsa.pub && \
+    curl -Lo ./boringssl-dev.apk https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/boringssl-dev-1.1.0-r0.apk && \
+    curl -Lo ./lsquic.apk https://github.com/omarroth/lsquic-alpine/releases/download/2.6.3-r0/lsquic-2.6.3-r0.apk && \
+    apk verify --no-cache ./boringssl-dev.apk ./lsquic.apk && \
+    tar -xf ./boringssl-dev.apk ./usr/lib/libcrypto.a ./usr/lib/libssl.a && \
+    tar -xf ./lsquic.apk ./usr/lib/liblsquic.a && \
+    mkdir ./ext/ && \
+    mv ./usr/lib/* ./ext/ && \
+    rm -r ./usr/ /etc/apk/keys/omarroth.rsa.pub ./boringssl-dev.apk ./lsquic.apk /root/.cache
 COPY ./shard.yml ./shard.yml
 RUN shards update && shards install && \
-    curl -Lo /etc/apk/keys/omarroth.rsa.pub https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/omarroth.rsa.pub && \
-    curl -Lo boringssl-dev.apk https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/boringssl-dev-1.1.0-r0.apk && \
-    curl -Lo lsquic.apk https://github.com/omarroth/lsquic-alpine/releases/download/2.6.3-r0/lsquic-2.6.3-r0.apk && \
-    apk verify --no-cache boringssl-dev.apk lsquic.apk && \
-    tar -xf boringssl-dev.apk usr/lib/libcrypto.a usr/lib/libssl.a && \
-    tar -xf lsquic.apk usr/lib/liblsquic.a && \
-    mv usr/lib/* lib/lsquic/src/lsquic/ext && \
-    rm -r usr boringssl-dev.apk lsquic.apk /root/.cache
+    mv ./ext/* ./lib/lsquic/src/lsquic/ext/ && \
+    rm -r ./ext/
 COPY ./src/ ./src/
 # TODO: .git folder is required for building – this is destructive.
 # See definition of CURRENT_BRANCH, CURRENT_COMMIT and CURRENT_VERSION.
```

(untested). That way we don't need to `curl` again if `shared.yml` is changed. Also I updated to using `./` when targeting files in the current working directory for reasons of consistency.

leonklingele: Oh, and what happened to `rm -rf /var/cache/apk/*`? I've never seen a cleanup of `/root/.cache` when working with `apk`.

docker/Dockerfile
```diff
@@ -1,27 +1,24 @@
...
 COPY ./src/ ./src/
 # TODO: .git folder is required for building – this is destructive.
 # See definition of CURRENT_BRANCH, CURRENT_COMMIT and CURRENT_VERSION.
 COPY ./.git/ ./.git/
 RUN crystal build ./src/invidious.cr \
     --static --warnings all --error-on-warnings \
-# TODO: Remove next line, see https://github.com/crystal-lang/crystal/issues/7946
```
leonklingele, line 23: Nice :+1:

.travis.yml
```diff
@@ -27,8 +27,5 @@ jobs:
       install:
         - docker-compose build
       script:
-        - docker-compose up
-        - sleep 15 # Wait for cluster to become ready, TODO: do not sleep
-        - HEADERS="$(curl -I -s http://localhost:3000/)"
-        - STATUS="$(echo $HEADERS | head -n1)"
-        - if [[ "$STATUS" != *"200 OK"* ]]; then echo "$HEADERS"; exit 1; fi
+        - docker-compose up -d
+        - while curl -Isf http://localhost:3000; do sleep 1; done
```
psvenk, line 31: Should we not add these lines after that?
```
        - HEADERS="$(curl -I -s http://localhost:3000/)"
        - STATUS="$(echo $HEADERS | head -n1)"
        - if [[ "$STATUS" != *"200 OK"* ]]; then echo "$HEADERS"; exit 1; fi
```

psvenk: (Sorry, I'm not yet accustomed to GitHub's "review" system)

tleydxdy: -f checks if the code is a failure or not. saving all the hacks.

docker/Dockerfile
tleydxdy: `--no-cache` and `rm -rf /var/cache/apk/*` has the same effect, `/root/.cache` is the share cache I found that doesn't need to be there, so I included it in the rm

tleydxdy: the change looks good tho, will add it in.

leonklingele:
Thanks for your patience! ❤️ I'd merge this now but let's wait for @omarroth, just in case.. ;)

omarroth:
LGTM and works great :tada:. Thank you all!

tleydxdy:
I've realized that maybe getting the key and the binary from the same place may defeat the point. but I can't find a better way. just to put this here.

leonklingele:
You could try something like the following (untested):

```patch
diff --git a/docker/Dockerfile b/docker/Dockerfile
index d0e4827..51e3425 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,9 +3,11 @@ RUN apk add --no-cache curl crystal shards libc-dev \
     yaml-dev libxml2-dev sqlite-dev zlib-dev openssl-dev \
     yaml-static sqlite-static zlib-static openssl-libs-static
 WORKDIR /invidious
-RUN curl -Lo /etc/apk/keys/omarroth.rsa.pub https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/omarroth.rsa.pub && \
+RUN curl -Lo omarroth.rsa.pub https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/omarroth.rsa.pub && \
     curl -Lo boringssl-dev.apk https://github.com/omarroth/boringssl-alpine/releases/download/1.1.0-r0/boringssl-dev-1.1.0-r0.apk && \
     curl -Lo lsquic.apk https://github.com/omarroth/lsquic-alpine/releases/download/2.6.3-r0/lsquic-2.6.3-r0.apk && \
+    echo "d986b7bb3a59116b51bae968450a4c84f0962b7577ae34c2bb8a35def9ecf096 ./omarroth.rsa.pub" | sha256sum -c - && \
+    mv ./omarroth.rsa.pub /etc/apk/keys/omarroth.rsa.pub && \
     apk verify --no-cache boringssl-dev.apk lsquic.apk && \
     tar -xf boringssl-dev.apk usr/lib/libcrypto.a usr/lib/libssl.a && \
     tar -xf lsquic.apk usr/lib/liblsquic.a && \
```

The code and this message however is still controlled by GitHub after all, so that doesn't make much sense.

PR #975, opened 2020-01-04
Change embed code

jorgesumle:
With the actual embedding code I get this error from [html5validator](https://github.com/svenkreiss/html5validator): `The "frameborder" attribute on the "iframe" element is obsolete. Use CSS instead.` This aims to fix that.


jorgesumle:
I also get `Attribute "type" not allowed on element "iframe" at this point.`, but I haven't fixed that.

omarroth:
Thanks!

PR #1011, opened 2020-02-01
Remove invalid and useless HTML from embed player

jorgesumle:
Sorry, I forgot to do that on #975 

The removed code is useless and non-standard (see https://stackoverflow.com/questions/4800227/why-did-youtube-put-a-type-attribute-in-iframe-for-embedded-video)


omarroth:
:+1:

PR #1016, opened 2020-02-04
Add support to read config from environment variable

leonklingele:
Try to read app config from the "INVIDIOUS_CONFIG" environment variable.
If the variable is undefined, read config from config.yml file as before.

Required by https://github.com/omarroth/invidious/pull/1015 et al.


omarroth:
:+1:

unixfox:
Maybe good to document this new feature in the wiki page: https://github.com/omarroth/invidious/wiki/Preferences?

PR #1032, opened 2020-02-15
Verify token signature in constant time, Run cheap checks first in token validation process

leonklingele:
See commit messages for details

### DIFF
--- a/src/invidious/helpers/tokens.cr
+++ b/src/invidious/helpers/tokens.cr
@@ -1,3 +1,5 @@
+require "crypto/subtle"
+
 def generate_token(email, scopes, expire, key, db)
   session = "v1:#{Base64.urlsafe_encode(Random::Secure.random_bytes(32))}"
   PG_DB.exec("INSERT INTO session_ids VALUES ($1, $2, $3)", session, email, Time.utc)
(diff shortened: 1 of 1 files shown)


leonklingele:
Please note that the second commit slightly changes (fixes?) behavior:
Before, the nonce would be updated in the DB (`db.exec("UPDATE nonces..`) even when the token contained an invalid signature. This is not longer the case and the call to the DB is only done when the token has been completely and successfully validated.

src/invidious/helpers/tokens.cr
```diff
@@ -78,32 +78,31 @@ def validate_request(token, session, request, key, db, locale = nil)
...
-    if nonce[1] > Time.utc
-      db.exec("UPDATE nonces SET expire = $1 WHERE nonce = $2", Time.utc(1990, 1, 1), nonce[0])
-    else
-      raise translate(locale, "Erroneous token")
-    end
+  if !Crypto::Subtle.constant_time_compare(token["signature"].to_s, sign_token(key, token))
+    raise translate(locale, "Invalid signature")
   end
 
   scopes = token["scopes"].as_a.map { |v| v.as_s }
```
leonklingele: The scope check could be made _after_ verifying the signature, what do you think @omarroth?

omarroth: Makes sense to me.

leonklingele:
Updated. ptal @omarroth

omarroth:
LGTM 👍

PR #1111, opened 2020-04-10
Add Hungarian translation

cstamas:
Adds HU translation.

Not all cleared up, but is good for beginning.


locales/hu-HU.json
```diff
@@ -0,0 +1,335 @@
+{
+    "`x` subscribers": "`x` feliratkozó",
+    "`x` videos": "`x` videó",
+    "`x` playlists": "`x` playlist",
```
cstamas, line 4: In general used "playlist" as foreign word is Hungarian as well, kinda "slang", while there is "official" `lejátszási lista` but personally I find it a bit awkward.

omarroth:
LGTM, thanks!

PR #1125, opened 2020-04-18
Make number of columns in feed customizable

### LINKED ISSUE #1018: Enhancement Request: Customisable/Selectable number of columns.
A maximum of 4 columns leave a lot to be desired when you have 150-200 subscriptions, they take up too much width real estate on my paltry 1080p monitor. I would love for the option of 8 or 10 videos per row.

psvenk:
Closes #1018. Add a new preference `feed_cols` and use it to determine the width of each video relative to the feed.

Please note that this only works when the value of `feed_cols` is a factor of 5 or 24, because of the way that Pure CSS's [grid system](https://purecss.io/grids/) works. If it would be beneficial to give the user a warning or to reject invalid inputs instead of blindly following the user's request (which would cause the number of columns to be effectively 1), I could also implement that.

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -414,5 +414,6 @@
     "location": "Location",
     "hdr": "HDR",
     "filter": "Filter",
-    "Current version: ": "Current version: "
+    "Current version: ": "Current version: ",
+    "Number of videos shown per row: ": "Number of videos shown per row: "
 }
--- a/locales/ar.json
+++ b/locales/ar.json
@@ -414,5 +414,6 @@
     "location": "",
     "hdr": "",
     "filter": "",
-    "Current version: ": "الإصدار الحالي: "
+    "Current version: ": "الإصدار الحالي: ",
+    "Number of videos shown per row: ": ""
 }
--- a/locales/bn_BD.json
+++ b/locales/bn_BD.json
@@ -353,5 +353,6 @@
     "Videos": "",
     "Playlists": "",
     "Community": "",
-    "Current version: ": ""
+    "Current version: ": "",
+    "Number of videos shown per row: ": ""
 }
--- a/locales/cs.json
+++ b/locales/cs.json
@@ -414,5 +414,6 @@
     "location": "umístění",
     "hdr": "HDR",
     "filter": "filtr",
-    "Current version: ": ""
+    "Current version: ": "",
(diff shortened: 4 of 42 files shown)


SamantazFox:
Wouldn't it be better to allow only a specific number of columns (let's say from 3 to 10) and then map that to pre-defined `pure-g` classes? Less hassle for the user or instance owner, and it makes sure that we won't have weird results.

In essence:
* `3` => `pure-g-*-1-3`
* `4` => `pure-g-*-1-4`
* `5` => `pure-g-*-1-5`
* `6` => `pure-g-*-1-6`
* `7` => `pure-g-*-3-24` (as this one must be approximated, maybe don't propose it?)
* `8` => `pure-g-*-1-8`

---

And also, **please** add your translation key to __ALL__ locales files. I spent a few days doing #1920, I don't want to do it again! And it also helps translators by showing them an "untranslated string" in weblate!

locales/en-US.json
```diff
@@ -383,5 +383,6 @@
     "Videos": "Videos",
     "Playlists": "Playlists",
     "Community": "Community",
-    "Current version: ": "Current version: "
+    "Current version: ": "Current version: ",
+    "Number of videos shown per row: ": "Number of videos shown per row: "
```
SamantazFox, line 387: Please add the following to all locales:
```
    "Number of videos shown per row: ": ""
```

psvenk:
@SamantazFox Thank you for your review. I just rebased my branch on the latest `master` and adapted to changes made over the past year, and I also added the translation key to the rest of the locale files. I'll test the code and respond to your other comment some time soon, after I sort out an issue with Docker and get Invidious to start properly.

syeopite:
@TheFrenchGhosty Could you please request a review from @saltycrys for this PR? Thanks

SamantazFox (approved):
For the locales part, this looks fine to me!

saltycrys (approved):

syeopite:
Just tested and it works perfectly! I just have one small concern. Since a column size of 7 isn't supported, would it be possible to add a warning about that?

psvenk:
Sorry for the delay.

@syeopite I'll try to get this done some time soon.

TheFrenchGhosty:
Bump

PR #1133, opened 2020-04-23
Consistent IDs for info section

mendel5:
Five of the eight info section stats had HTML IDs, e.g. genre, family_friendly, etc.
However the remaining three stats did not have IDs: views, likes and dislikes. Therefore I have added them so there is a consistent naming of the IDs in the info section.


omarroth:
Thanks!

mendel5:
Thank you, too! 🚀

PR #1137, opened 2020-04-25
Move from alpine:edge to alpine:latest

raycheung:
The alpine:edge fails to build.


leonklingele:
A fix is in https://github.com/omarroth/invidious/pull/949. Thanks for your support @raycheung!! 😊🚀

raycheung:
Thank you! #949 is a much better fix than this workaround. Closing this now.

Issue #1194, opened 2020-05-25
Improve error messages / error pages


girst:
It'd be great if error pages would have a link to the original youtube URL and/or a small selection of reliable other instances. This would allow the user to watch the video elsewhere.

Instances should probably have to opt-in to being listed on invidio.us (e.g. by asking for inclusion themselves with a PR/issue), since traffic to them would likely increase. maybe instances.invidio.us or uptime.invidio.us can be queried before suggesting an alternative instance.

Also, when the instance is blocked, visiting a /channel/ page returns `This channel does not exist.`, which is not true. this was fixed for /watch with 0e1d6aa85c40d071270092de0195a75f928e3506, but not for other pages.

exilexi:
Yep I’d love a link to the original youtube or other instances, or even better, other P2P formats ( #328)

SamantazFox:
Closing, as this has been fixed ("next steps" message, improved bug page, etc...)

PR #1206, opened 2020-05-29
"Watch on YouTube" link for error pages

recette-lemon:
Adding #1194.

### DIFF
--- a/src/invidious/views/error.ecr
+++ b/src/invidious/views/error.ecr
@@ -3,5 +3,9 @@
 <% end %>
 
 <div class="h-box">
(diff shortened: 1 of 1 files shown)


recette-lemon:
has problems with showing it in error messages that dont make sense, and one error message already links. ill fix later

GauthierPLM:
I believe it would make more sense to display it only if the page is a video or a channel page, based on the query params.

It might be a good idea to write this as a helper rather than in the template directly so not too much logic is in the view.

syeopite:
Superseded by #1940

PR #1218, opened 2020-06-06
Add RAM requirement

### LINKED ISSUE #1152: State the minimal requirement of RAM to build invidious in the README.md
I tried to do the setup of invidious as documented here:
https://github.com/omarroth/invidious#setup-invidious
on a virtual machine Debian 10 (buster) 32bits with 512GB of ram, it failed. 
There is a reference in #209 saying 2GB of ram is needed. That would be very nice to state it in the README.md file.
Cheers,

SuperSandro2000:
Closes #1152

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ $ docker-compose build
 
 ### Linux
 
-To manually compile invidious you need at least 2GB of RAM. If you have less please fall back to Docker.
+To manually compile invidious you need at least 2GB of RAM. If you have less you can setup SWAP to have a combined amount of 2 GB or fall back to Docker.
 
 #### Install dependencies
 
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ $ docker-compose build
 
 ### Linux
 
-To manually compile invidious you need at least 2GB of RAM. If you have less you can setup SWAP to have a combined amount of 2 GB or fall back to Docker.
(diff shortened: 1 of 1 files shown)


cloudrac3r:
A smaller amount of RAM also works if RAM + swap >= 2 GB. You might want to update this to reference swap.

SuperSandro2000:
Compiling on swap is slower and everyone should have 2GB of RAM nowdays.

cloudrac3r:
It's slower but it actually works, unlike 1 GB total.

If everyone should have 2 GB+, why is this PR necessary at all?

SuperSandro2000:
> It's slower but it actually works, unlike 1 GB total.

Docker also works on 1GB.

> If everyone should have 2 GB+, why is this PR necessary at all?

Then close the issue that requested this.

> And slow compiling is better than no compiling :)

This does not forbid you do to that.

gripped:
Productive would be editing the PR to state reality and not a falsehood. Some people might not know they can use swap as as workaround.
Discussing not arguing :)

README.md
```diff
@@ -73,7 +73,9 @@ $ docker volume rm invidious_postgresdata
 $ docker-compose build
 ```
 
-### Linux:
+### Linux
+
+To manually compile invidious you need at least 2GB of RAM. If you have less you can setup SWAP to have a combined amount of 2 GB or fall back to Docker.
```
Perflyst, line 78:
```suggestion
To manually compile invidious you need at least 2GB of RAM. If you have less you can setup SWAP to have a combined amount of 2 GB or use Docker instead.
```

PR #1220, opened 2020-06-06
Disable autocorrect in search input

matbia:
On mobile it's generally annoying and just gets in the way, especially when the autocorrect language is set to something else than English.


SuperSandro2000:
I think autocomplete should stay on so users are not forced to type in everything and spelling should probably also stay on so people know if they did a typo.

Perflyst:
I see that the same way as @SuperSandro2000, if you dont want auto-correct you can turn it off on your device too

PR #1227, opened 2020-06-08
Update PeerTubeify URL to new location

AliceDTRH:
On the URL specified it says: "Project 'Ealhad/peertubeify' was moved to 'Cha_deL/peertubeify'. Please update any links and bookmarks that may still have the old path."


omarroth:
Thanks!

Issue #578, opened 2019-06-07
accept: Too many open files (Errno)


tmiland:
Hi, have had some downtime recently (connection timeout), get this error:

`accept: Too many open files (Errno)
Jun 06 22:12:45 debian invidious[18277]:   from /usr/share/crystal/src/thread.cr:69:3 in 'accept?'
Jun 06 22:12:45 debian invidious[18277]:   from /usr/share/crystal/src/http/server.cr:419:13 in '->
Jun 06 22:12:45 debian invidious[18277]:   from /usr/share/crystal/src/fiber.cr:255:3 in '???'
Jun 06 22:12:45 debian invidious[18277]:   from ???
Jun 06 22:12:45 debian invidious[18277]: accept: Too many open files (Errno)
Jun 06 22:12:45 debian invidious[18277]:   from /usr/share/crystal/src/thread.cr:69:3 in 'accept?'
Jun 06 22:12:45 debian invidious[18277]:   from /usr/share/crystal/src/http/server.cr:419:13 in '->
Jun 06 22:12:45 debian invidious[18277]:   from /usr/share/crystal/src/fiber.cr:255:3 in '???'
Jun 06 22:12:45 debian invidious[18277]:   from ???`

When this happens, the invidious log-file is filled with info from the system-log.

Never seen this before... What gives?

omarroth:
I've encounter this as well, generally it's because of a high request for static files (`/js/watch.js`, `/css/default.css`, etc.), but haven't found it to be a serious issue. Log for me looks like:

```
Exception: Error opening file '/css/default.css' with mode 'r': Too many open files (Errno)
   from ...
```

This may also be crystal-lang/crystal#7843 since from the above log it doesn't appear any files are being opened.

tmiland:
Thanks!

Weird stuff. 

The instance comes back online after a restart. Not sure if this is nginx, openssl or crystal, but there might be a setting somewhere to increase the amount of open files limit probably (idk.), that i forgot to change after migrating the instance... Hard to remember everything 😂

I'll look more into it later. 👍

tmiland:
Suspect: `proxy_read_timeout 90;` in nginx reverse proxy

Increased to `proxy_read_timeout 3600;` to see if it makes any difference. 

Also added: 
```
nginx       soft    nofile  10000
nginx       hard    nofile  30000
```
to `/etc/security/limits.conf`

tmiland:
Woke up to this today:

[...]

PR #1330, opened 2020-08-09
Potential fix for issue #578

mintphin:
Had this issue on my instance. This fix was mentioned in the issue mentioned so I decided to implement it on the service file.


Perflyst:
This does not really fix the issue. It may be a workaround, but if you get enough requests, the actual limit will be reached.

Perflyst:
Feel free to add a page in the wiki though https://github.com/iv-org/invidious/wiki

PR #1335, opened 2020-08-13
Fix CI badge

szepeviktor:

PR #1336, opened 2020-08-14
Update lsquic.cr dependency to use iv-org version

GauthierPLM:


GauthierPLM:
I tried to understand why the CI fails, but I have no idea, and doing the same steps locally does not cause any issue…

Perflyst:
CI already failed before, this will maybe be fixed with an apt-get update before hand in the CI, see #1335

PR #1337, opened 2020-08-14
Ensure all crystal dependencies are installed before running CI build

GauthierPLM:
I give a try to fix the CI, but it seems it already has been fixed on the side of Travis, as the update and install steps I added is run without any effects. @Perflyst could you  run the CI on master without merging my PR to check?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ jobs:
       language: crystal
       crystal: latest
       before_install:
+        - sudo apt-get update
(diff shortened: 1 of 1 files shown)


GauthierPLM:
It seems the step that caused the issue in the CI (crystal_install) now correctly resolve the `libevent-dev` package. I don't know if Travis fixed their repository or something, but as it happens before the `before_install` step of our Travis config, I guess it was fixed on their side.

Perflyst:
I prefer the workaround from https://github.com/iv-org/invidious/pull/1335/files#diff-354f30a63fb0907d4ad57269548329e3R3

PR #1355, opened 2020-08-29
Use new youtube API to fetch channel videos

### LINKED ISSUE #1333: Videos tab of channel pages are empty on every instance (except invidio.us)
Example: 

https://invidious.snopyta.org/channel/UCXuqSBlHAE6Xw-yeJA0Tunw

https://invidious.ggc-project.de/channel/UCXuqSBlHAE6Xw-yeJA0Tunw

Invidio.us is the only instance not affected by this issue for whatever reason.

https://invidio.us/channel/UCXuqSBlHAE6Xw-yeJA0Tunw

Edit: according to the comments it's a new rate limit by youtube, invidio.us isn't affected by this because it uses caching.

Edit2: This issue is only present on *some* channels, and in the meantime RSS still work, this is a frontend issue I think.

afuous:
This fixes issue #1333 where channel pages no longer load any videos. The solution is described by girst here: https://github.com/iv-org/invidious/issues/1333#issuecomment-674525743.

I edited `src/helpers/helpers.cr` because the format that the new API returns results in was not handled by the existing code. The changes I made should make it easier to work with

### DIFF
--- a/src/invidious/channels.cr
+++ b/src/invidious/channels.cr
@@ -927,7 +927,6 @@ def get_channel_videos_response(ucid, page = 1, auto_generated = nil, sort_by = 
   if needs_v2
     url = produce_channel_videos_url(ucid, page, auto_generated: auto_generated, sort_by: sort_by, v2: true)
     response = YT_POOL.client &.get(url)
-    initial_data = JSON.parse(response.body).as_a.find &.["response"]?
   end
   response
 end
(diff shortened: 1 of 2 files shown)


zethexx:
Nice! Thank you. Will this be merged soon? If not, I don't have much free time on me right now so I was thinking of allocating some time in a few days to try out this patch.

raycheung:
Checked out and tested on my side, it works very well for me. Thank you!

imapotatoalso:
Thank you! It works for me :)

TheFrenchGhosty:
Merging at the request of @Perflyst

PR #1360, opened 2020-09-05
[Helm] add service type values and InitContainer to wait for Postgres

FlipEnergy:
The Service type really should be there to change the type of service and external port in Helm chart values. This does not change the default helm install behaviour in terms of Service type since Kubernetes default is clusterIP.

I also added a initContainer to check if postgres is up and ready to serve before starting Invidious.

### DIFF
--- a/kubernetes/templates/deployment.yaml
+++ b/kubernetes/templates/deployment.yaml
@@ -50,4 +50,5 @@ spec:
           httpGet:
             port: 3000
             path: /
+          initialDelaySeconds: 15
       restartPolicy: Always
--- a/kubernetes/templates/deployment.yaml
+++ b/kubernetes/templates/deployment.yaml
@@ -23,6 +23,13 @@ spec:
         runAsUser: {{ .Values.securityContext.runAsUser }}
         runAsGroup: {{ .Values.securityContext.runAsGroup }}
         fsGroup: {{ .Values.securityContext.fsGroup }}
+      initContainers:
+        - name: wait-for-postgresql
+          image: postgres
+          args:
+          - /bin/sh
+          - -c
+          - until pg_isready -h {{ .Values.config.db.host }} -p {{ .Values.config.db.port }} -U {{ .Values.config.db.port }}; do echo waiting for database; sleep 2; done;
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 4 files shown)


TheFrenchGhosty:
@leonklingele Since you are the one that worked on this, it would be nice if you could review this PR.

FlipEnergy:
Also please consider setting up a helm repo for folks to add and download directly instead of cloning and packaging it themselves. You could use github pages to do this.

kubernetes/templates/deployment.yaml
```diff
@@ -23,6 +23,13 @@ spec:
         runAsUser: {{ .Values.securityContext.runAsUser }}
         runAsGroup: {{ .Values.securityContext.runAsGroup }}
         fsGroup: {{ .Values.securityContext.fsGroup }}
+      initContainers:
+        - name: wait-for-postgresql
+          image: postgres
+          args:
+          - /bin/sh
+          - -c
+          - until pg_isready -h {{ .Values.config.db.host }} -p {{ .Values.config.db.port }} -U {{ .Values.config.db.port }}; do echo waiting for database; sleep 2; done;
```
leonklingele, line 32: Shouldn't this be `-U {{ .Values.config.db.user }}` instead of `-U {{ .Values.config.db.port }}`?

kubernetes/values.yaml
```diff
@@ -13,6 +13,11 @@ autoscaling:
   maxReplicas: 16
   targetCPUUtilizationPercentage: 50
 
+service:
+  type: clusterIP
+  port: 3000
+  # loadBalancerIP:
```
leonklingele, line 19: Could you please remove the space here? Sorry! :see_no_evil:

leonklingele (commented):
Thank you so much for your pull request, Dennis! I'll see how we can provide pre-packaged Helm Charts.

kubernetes/templates/deployment.yaml
FlipEnergy: Oh yes. Surprised that my testing worked. Will push fix

kubernetes/values.yaml
FlipEnergy: The space after the "#"? Sure

leonklingele:
Dennis, you're awesome! :rocket: I will merge this later next week — Thank you for your contribution!

TheFrenchGhosty:
@leonklingele Can this be merged?

Issue #1342, opened 2020-08-19
Substitution of storyboard thumbnail HOST_URL fails on other domain names


resttime:
https://github.com/iv-org/invidious/blob/2ce82f3da59037d533c58c8c8f78bb81904850ae/src/invidious.cr#L3765

**Problem:** Instead of `https://i9.ytimg.com` storyboard thumbnails also come from `https://i.ytimg.com`. In that situation, substituting the HOST_URL will never occur and an incorrect path will be served.

**Suggestion:** Use a regexp instead of a string: `/https:\/\/i[0-9]*\.ytimg\.com/`

Afterwards there might be another fix required since 403 errors started happening instead that I haven't been able to find the cause.

I've created a corresponding fork with the commit here, though I can't consider the issue fully fixed without getting rid of the 403 errors: https://github.com/resttime/invidious/commit/c674476c811c1dc609507d8c65ffcf110e3ea256

raycheung:
I tried to do this: https://github.com/iv-org/invidious/pull/1367
and it fixes my issue. Please see if it help.

raycheung:
I dived a bit and found the https://github.com/iv-org/invidious/blob/master/src/invidious.cr#L5634 line differs from the link the browser is actually hitting...
There, it implements as `get "/sb/:id/:storyboard/:index"` but the browser is actually going to `/api/v1/storyboards/sb/:id/:storyboard/:index"`, resulting in some HTTP 500.

I tried to modify it to `get "/api/v1/storyboards/sb/:id/:storyboard/:index"`, and also update the headers as follows:
```
headers = HTTP::Headers{":authority" => "i.ytimg.com"}
```

Now I'm able to get HTTP 200, however the proxied response is always broken.
I tried to add a few lines to https://github.com/iv-org/invidious/blob/master/src/invidious/helpers/helpers.cr#L618 and write to a local tempfile, the tempfile is actually a valid webp image. I'm currently stuck at here.

saltycrys:
This can be closed @TheFrenchGhosty

PR #1367, opened 2020-09-08
Fix storyboard not loading images from i.ytimg.com

raycheung:
The `Content-Security-Policy` is not allowing the browser to load images from `i.ytimg.com`, which are images of the storyboard.


TheFrenchGhosty:
I think those pictures supposed to be proxy-ed

raycheung:
Honestly I agree. I wonder why googlevideo.com is actually whitelisted in the CSP header.

TheFrenchGhosty:
That's because proxy-ing video is optional, however invidious is supposed to proxy pictures.

raycheung:
Thank you @TheFrenchGhosty . I've investigated a bit about the issue, commented here: https://github.com/iv-org/invidious/issues/1342#issuecomment-690297350

Issue #674, opened 2019-08-06
How to unfreeze subscriptions?


cyberchief1984:
How do I unfreeze my subscription feed? I;m not using the official instance. Somebody froze it.

fliegentoeter:
Happens to me on the official instance on a daily basis. Seems like subspriptions are only updated twice a day or even less...

omarroth:
> I;m not using the official instance

Mind providing which instance you are using?

Otherwise I expect it's an issue with higher load. There's a couple things I have in mind to try that should fix the issue, although it may be a couple days before I can take a look at it.

cyberchief1984:
I am using https://invidious.13ad.de.

PureTryOut:
I'm using a private self-hosted one, I'm the only user, and I haven't seen an update to my subscriptions since last week. If I go to an individual channel the newest videos do show up however.

omarroth:
What does your `config.yml` look like?

PureTryOut:
```
channel_threads: 1
feed_threads: 1
db:
  user: ****
  password: ****
  host: ****
  port: ****
  dbname: ****
full_refresh: false
https_only: false
domain:
```

omarroth:
You may check that you're up-to-date, otherwise make sure that postgres isn't logging any errors. If you're seeing something like #619, then you should be able to fix it with `check_tables: true` in your config.

omarroth:
> I am using https://invidious.13ad.de.

Since I am not the administrator of that instance there's not much I can there unfortunately. For the official instance it's still something I'm looking into.

PureTryOut:
I'm definitely up to date, version 0.19.1.

Invidious didn't log any error besides the following:
```
INSERT has more expressions than target columns
```

I added `check_tables: true` to the config which seems to have at least fixed that warning/error, but it didn't update my subscriptions sadly.

PostgreSQL doesn't have any obvious errors either.

Gleddified:
Still an issue for me on the official instance. If it's any help, this video, uploaded over a day ago, is still not in my subsciptions feed:

https://www.invidio.us/watch?v=WVDQEoe6ZWY

schwukas:
Despite the merged fix I am unfortunately still having that same problem where videos are shown in my Youtube but not the Invidious feed. 

I do however get quite a lot of errors in my invidious container logs:
```
[...]

PR #1370, opened 2020-09-09
Fix sub refresh

omarroth:
Currently, when determining when a feed needs to be updated, the last time a subscription feed was viewed is used to determine whether or not it should be updated. This means a subscription feed will not be updated if a user views their subscription feed before new videos are pulled.

Instead, this PR uses whether or not a video was inserted into `channel_videos` to determine which feeds should be updated.

Fix for #674.

### DIFF
--- a/src/invidious/channels.cr
+++ b/src/invidious/channels.cr
@@ -97,6 +97,14 @@ struct ChannelVideo
       end
     end
   end
+
+  def to_tuple
+    {% begin %}
+      {
+        {{*@type.instance_vars.map { |var| var.name }}}
+      }
+    {% end %}
+  end
 end
 
 struct AboutRelatedChannel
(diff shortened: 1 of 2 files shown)


src/invidious/channels.cr
```diff
@@ -315,39 +310,22 @@ def fetch_channel(ucid, db, pull_all_videos = true, locale = nil)
...
-        break
-      end
-
+      break if count < 25
       page += 1
     end
 
     # When a video is deleted from a channel, we find and remove it here
-    db.exec("DELETE FROM channel_videos * WHERE NOT id = ANY ('{#{ids.map { |id| %("#{id}") }.join(",")}}') AND ucid = $1", ucid)
+    # db.exec("DELETE FROM channel_videos * WHERE NOT id = ANY ('{#{ids.map { |id| %("#{id}") }.join(",")}}') AND ucid = $1", ucid)
```
SuperSandro2000, line 328:
```suggestion
```

TheFrenchGhosty:
Merged.

PR #1371, opened 2020-09-09
Allow cookies from anti-captcha to be optional

omarroth:
Occasionally Anti-Captcha will return an empty array `[]` in the `cookies` field instead of a hash after a successful response. In this case it's still possible to continue without any cookies.


TheFrenchGhosty:
Merged.

PR #1372, opened 2020-09-09
Add exponential backoff to refresh_channels

tleydxdy:
In practice with the patch I usually see backoff to 2 hours when blocked, so it should improve recovery time. The lim_thread is to work with multi-threading, not sure if it's the best way to do it.

### DIFF
--- a/src/invidious/helpers/jobs.cr
+++ b/src/invidious/helpers/jobs.cr
@@ -3,15 +3,17 @@ def refresh_channels(db, logger, config)
 
   spawn do
     max_threads = max_channel.receive
+    lim_threads = max_threads
     active_threads = 0
     active_channel = Channel(Bool).new
+    backoff = 1.seconds
 
     loop do
       db.query("SELECT id FROM channels ORDER BY updated") do |rs|
         rs.each do
           id = rs.read(String)
 
-          if active_threads >= max_threads
+          if active_threads >= lim_threads
             if active_channel.receive
               active_threads -= 1
             end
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
TL;DR? What does it do? What does it fixes?

tleydxdy:
invidious pulls the channel pages every minutes, even after the IP was temporarily blocked. This makes the ban last longer (days in my case). This patch lets invidious wait progressively longer until the next pull. In practice I see my instance getting un-banned in a few hours.

TheFrenchGhosty:
@omarroth It would be nice if you could review this PR

tleydxdy:
rip, I possibly messed something up, getting database error
```
 there is no unique or exclusion constraint matching the ON CONFLICT specification 
```

Perflyst:
This might work on a private instance but what about public instances? There are thousands of requests to channels every minute. I expect the delay gets higher and higher and channels wont refresh at all anymore? Am I wrong?

tleydxdy:
if you visit the channel pages manually it wouldn't have any back off, since this pr is only changing the background refresh parts of the code.

src/invidious/helpers/jobs.cr
```diff
@@ -3,15 +3,17 @@ def refresh_channels(db, logger, config)
...
     active_channel = Channel(Bool).new
+    backoff = 1.seconds
 
     loop do
       db.query("SELECT id FROM channels ORDER BY updated") do |rs|
         rs.each do
           id = rs.read(String)
 
-          if active_threads >= max_threads
+          if active_threads >= lim_threads
```
matthewmcgarvey, line 16: If the idea of this is to start at max threads (say 5) and if an error occurs, force the job to drop to one thread, I think this needs to be changed from an if to a while loop. There's a lot going on in this so I could be wrong.

tleydxdy: I think you are correct

PR #1379, opened 2020-09-15
Remove on-hover tooltip from player

Rjevski:
Using the player on latest Safari, the tooltip appears and stays stuck for long even when switching to fullscreen which is annoying. You need to explicitly click anywhere to dismiss that stuck tooltip.

This doesn't seem to happen in Firefox so I am not sure whether this is a browser bug, but in any case I don't see any value in keeping this tooltip so maybe we can just remove it?


TheFrenchGhosty:
Can you explain what do you mean by tooltip?

And can you show us the issue?

Rjevski:
Here's a screenshot of what I mean - hovering over the video player would display the following (with the title of the video):

image

Perflyst (approved):
I dont see any value neither and if it is annoying in some browsers, drop it

PR #1387, opened 2020-09-25
Fix #1382 : Playlists missing first video

### LINKED ISSUE #1382: Playlists missing first video
When playlists are loaded on Invidious, playlists are missing the first video in the playlist.

To reproduce, go to any Invidious instance and load a playlist ([example](https://invidious.snopyta.org/playlist?list=PLSfegfOiC2Ayz10yngrfQoe-GZtdI3ZIM)), you'll notice how the first video in the playlist is missing.

GauthierPLM:
The index was set to index - 1, causing the first video to be shifted in fetch_playlist_videos#L448 (because of its index being -1 lower than it should) and thus not displayed on playlist page.

### DIFF
--- a/src/invidious/playlists.cr
+++ b/src/invidious/playlists.cr
@@ -483,7 +483,7 @@ def extract_playlist_videos(initial_data : Hash(String, JSON::Any))
         published:      Time.utc,
         plid:           plid,
         live_now:       live,
-        index:          index - 1,
+        index:          index,
       })
     end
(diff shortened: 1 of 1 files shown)


GauthierPLM:
@omarroth Do you have any idea why the index was set to -1? I couldn't find any use for it.

omarroth:
No, it looks like a typo. Your fix LGTM.

PR #1389, opened 2020-09-27
add config to decrypt on demand instead of polling

vhuynh3000:
Add a config option to get the decryption function on demand instead of polling every 1 minute. In docker environments, the lsquic.cr library has a memory leak when making quic connections. While this commit does not fix the memory issue, it works around it by making the connection when the decrypt function is needed so the memory usage does not increase while idle. It only increases when in use. 

This also decrease the amount of data used. Each call to youtube to get the decrypt function is over 1mb so over a 1 gb of data is download each day even when the invidious instance is not used.


unixfox:
Can someone test this PR? I'm very interested to include it on my instance.

saltycrys (changes requested):
Looks good overall but there is a lint error and a merge conflict.

Please run `crystal tool format` to fix the lint error and rebase onto master.

unixfox:
Any updates :)? I'm still interested into this PR.

Perflyst:
I tried to rebase but it got worse. There were a lot of (structural) changes

saltycrys (approved):

Issue #1361, opened 2020-09-06
Why is Invidious written in Crystal?


mendel5:
I am asking this question out of curiosity. The initial commit for Invidious can be found at https://github.com/iv-org/invidious/tree/26959020b732b5d729432294df6246307819e32f . Even back then the backend of Invidious (src) was written in Crystal (.cr) and the frontend (views) in Embedded Crystal (.ecr).  So Invidious was not refactored to use Crystal but rather has this been the case since the beginning.

What were the reasons to choose Crystal and not any other language? I am only trying to understand the reasoning of this decision.

Maybe you @omarroth can help us.

markozajc:
I'd assume that was due to familiarity with the language, which is also the reason I (and a lot of people that I know) use languages that they know the best, even if they're obscure.

mendel5:
> I'd assume that was due to familiarity with the language

That's what I am thinking as well.

Issue #1352, opened 2020-08-25
Alternative To Invidious


zaheerhakim:
Your service was amazing and it's sad that we will no longer be able to access it after the 1st of September, is there any other alternative to Invidious?

Scaledi:
the specific instance is shutting down, not the project.

zaheerhakim:
> the specific instance is shutting down, not the project. 

So I’ll be still able to use the website to watch YouTube videos?

diazepan:
> So I’ll still able to use the website to watch YouTube videos?

It's more like you have to use other instance. Here's a list
https://invidious-redirect.cadence.moe/

zaheerhakim:
> > So I’ll still able to use the website to watch YouTube videos?
> 
> 
> 
> It's more like you have to use other instance. Here's a list
> 
> https://invidious-redirect.cadence.moe/

Thank you 😃

Stay Safe!

TheFrenchGhosty:
> the specific instance is shutting down, not the project.

Yes, but no.

The instance is shutting down, yes, but the only developer is also quitting. If Invidious don't get a new developer the project is dead.



> > the specific instance is shutting down, not the project.
> 
> So I’ll still able to use the website to watch YouTube videos?

As long as Invidious doesn't get a new developer the project is dying, currently if YouTube change something and no one can fix it, the project will be dead.

zaheerhakim:
> > the specific instance is shutting down, not the project.
> 
> 
> 
> Yes, but no.
> 
> 
> 
> The instance is shutting down, yes, but the only developer is also quitting. If Invidious doesn't get a new developer the project is dead.
> 
> 
> 
> 
> 
> 
> 
> > > the specific instance is shutting down, not the project.
> 
> > 
> 
> > So I’ll still able to use the website to watch YouTube videos?
> 
> 
> 
> As long as Invidious doesn't get a new developer the project is dying, currently if YouTube change something and no one can fix it, the project will be dead.

Thank you. 

Understood. 

Hope everything goes well. 

Stay Safe!

SolAZDev:
> The instance is shutting down, yes, but the only developer is also quitting. If Invidious doesn't get a new developer the project is dead.
> As long as Invidious doesn't get a new developer the project is dying, currently if YouTube change something and no one can fix it, the project will be dead.

Begs the question; is this not about to become a community-developed project? If so, how come? I'm pretty sure Invidious could benefit from itt

Scaledi:
[...]

Issue #1350, opened 2020-08-24
[Discussion/Documentation] Replacing the Invidious backend / Current structure of Invidious/the Invidious codebase


TheFrenchGhosty:
This is something that I planned to ask in the Matrix room, but I think it's a good idea to have it available publicly.

@cloudrac3r : Since you are working on [Second](https://git.sr.ht/~cadence/Second), do you think it's possible to have it become the backend of Invidious? (meaning that we just keep the frontend of Invidious, making things easier to fix (since it's not Crystal / is using youtube-dl)), or do you think it won't be doable?

@omarroth : Speaking of replacing the backend: what is the structure of the backend/frontend/api currently? 

Is the frontend calling the API that will then call the backend to make request? Or is the frontend making call to the backend and the API is making call to the backend separately?

Also: Are the code of the 3 properly separated or is there (for example) some backend code in the frontend? In short: is it possible to even replace the backend "easily"?

cloudrac3r:
I have only looked briefly at the Invidious source, but I don't _think_ it would be easy to replace its backend with API calls while keeping the rest of its code intact. _I'll wait for omar to weigh in on this though._

I _think_ it would require a lot of modifications in Crystal and I don't know Crystal.

I _think_ it's more practical to create a brand new front-end that can be used with any API implementation. We already have a few of these like CloudTube, FreeTube, LapisTube, and more.

As I already said, it might be possible to make Invidious into a front-end only, but _I expect_ that the modifications would be significant enough that this would require a person who actually knows Crystal.

TheFrenchGhosty:
>I have only looked briefly at the Invidious source, but I don't think it would be easy to replace its backend with API calls while keeping the rest of its code intact. I'll wait for omar to weigh in on this though.

This is also what I concluded by looking at the layout of the source code.

>I think it's more practical to create a brand new front-end that can be used with any API implementation. We already have a few of these like CloudTube, FreeTube, LapisTube, and more.

The thing is that lots of people (including me) really like/love the Invidious frontend, other than bugfix it doesn't really need anything, and having it usable without the current backend would be really nice.

>I think it would require a lot of modifications in Crystal and I don't know Crystal.

Yeah this is why I said "In short: is it possible to even replace the backend "easily"?" 

>I think it's more practical to create a brand new front-end that can be used with any API implementation. We already have a few of these like CloudTube, FreeTube, LapisTube, and more.

[...]

Issue #1121, opened 2020-04-16
The "single file cluster" could use some refactoring?


grkek:
### Problem description

Hi, I was browsing through the chaos that is the main file of the invidious web app and since I have experienced the same problem with the chaotic route distribution I had to fork the Kemal framework and add additional features to it which lead me to creation of the [Grip](https://github.com/grip-framework/grip) framework.

### Suggestion

My suggestion is that the current code can be ported over to Grip with which you could benefit by having a file structure, better routing mediums, additional features like route grouping and object oriented distribution of your endpoints, pipes, etc.

omarroth:
PR welcome :)

grkek:
> PR welcome :)

Sounds lovely to me

TheFrenchGhosty:
Fixed in #1399 #1403 #1420 by @matthewmcgarvey

grkek:
@TheFrenchGhosty I don't see how this resolves the issue of a one single 2000k+ line file having all of the routes and logic.

TheFrenchGhosty:
@grkek Because this single file has been replaced by multiples one

matthewmcgarvey:
@grkek @TheFrenchGhosty it hasn't been fully pulled out. There's just sooo much code that I wouldn't dare do it in one PR.

PR #1399, opened 2020-10-06
Provide rough draft of better project organization

matthewmcgarvey:
Hi 👋  I've not used invidious much, but I'm seeing a lot of comments talking about the project being in need of more crystal devs. As a crystal dev, I feel like I can pitch in a bit. https://github.com/iv-org/invidious/issues/1352, https://github.com/iv-org/invidious/issues/1350, https://github.com/iv-org/invidious/issues/1361

As https://github.com/iv-org/invidious/issues/1121 pointed out, everything being in a single file and global is quite difficult to work with. Before making a huge leap to a new framework, though, I think the best thing to do is to organize everything just a bit. This PR adds two new patterns that I hope will help wrangle the code.

1. I've added an `Invidious::Routes` module that serves as controllers along with an `Invidious::Routing` module for the route macros. It's definitely not the best thing in the world but I can hope you see it going in the right direction.
2. I've also added an `Invidious::Jobs` module that can wrap all the async jobs being performed so that we can move away from global methods

I would hope that once these patterns are worked into Invidious that it might make the transition to a framework like [Grip](https://github.com/grip-framework/grip) or even [Lucky](https://luckyframework.org/) manageable.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -27,6 +27,8 @@ require "compress/zip"
 require "protodec/utils"
 require "./invidious/helpers/*"
 require "./invidious/*"
+require "./invidious/routes/**"
+require "./invidious/jobs/**"
 
 ENV_CONFIG_NAME = "INVIDIOUS_CONFIG"
 
@@ -196,11 +198,11 @@ if config.statistics_enabled
   end
 end
 
-popular_videos = [] of ChannelVideo
-spawn do
-  pull_popular_videos(PG_DB) do |videos|
-    popular_videos = videos
-  end
+Invidious::Jobs.register Invidious::Jobs::PullPopularVideosJob.new(PG_DB)
+Invidious::Jobs.start_all
+
+def popular_videos
+  Invidious::Jobs::PullPopularVideosJob::POPULAR_VIDEOS.get
 end
 
 DECRYPT_FUNCTION = [] of {SigProc, Int32}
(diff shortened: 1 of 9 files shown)


TheFrenchGhosty:
Amazing, is this PR finished? Can we assume everything will be working with it?

matthewmcgarvey:
@TheFrenchGhosty this PR can be considered finished and if it gets merged I can update the rest of the code. I did verify it worked locally

TheFrenchGhosty:
We are currently testing this PR. If everything is working we will merge it soon.

Thanks again.

Perflyst:
No issues so far, thank you!

SamantazFox:
> I would hope that once these patterns are worked into Invidious that it might make the transition to a framework like [Grip](https://github.com/grip-framework/grip) or even [Lucky](https://luckyframework.org/) manageable.

I see a lot of "REST" and "RESTful" on the pages you liked.
I'm a bit worried because this often means "a lot of javascript on the client". Is this your idea?

I appreciate a lot the fact that invidius runs super fine without javascript, so loosing that would be loosing its core idea.

matthewmcgarvey:
@SamantazFox I have no intention of moving towards a heavy javascript implementation so I think we're on the same page 😄. I provided those links just as a goal of moving this project to have a better structure than we have right now.

grkek:
@SamantazFox In Grip you can use the same rendering engine Kemal uses to avoid the JavaScript hell.

PR #1403, opened 2020-10-08
Extract RefreshChannelsJob

matthewmcgarvey:
Moved refresh_channels job into the new job pattern.

The only change is that I removed the `max_channel` variable. It was redundant when we can pull the max channels from the config.


TheFrenchGhosty:
@matthewmcgarvey If you have the time to resolve the conflict I would like to merge this

TheFrenchGhosty:
Thanks!

PR #1407, opened 2020-10-10
Major readme enhancements and fixes

TheFrenchGhosty:
Various enhancement to the readme, updating and fixing stuff.

Closes #1363

PR #1408, opened 2020-10-10
More readme enhancements and fixes

TheFrenchGhosty:
Follow up to #1407


TheFrenchGhosty:
@B0pol Agree, this is something I thought about after merging this PR, I will do it now.

PR #1412, opened 2020-10-16
Set preferences before conditional breaks in before_all

### LINKED ISSUE #1386: ERROR - http.server: Unhandled exception on HTTP::Handler Missing hash key: "preferences" (KeyError)
Am getting the following error on my hosted invidious instance server (invidious.site), instance was working fine till yesterday. Can someone help me understand the error and possible causes of this error to fix it? 

2020-09-24T20:02:52.978310Z  ERROR - http.server: Unhandled exception on HTTP::Handler
Missing hash key: "preferences" (KeyError)
  from ../../../usr/share/crystal/src/hash.cr:1030:9 in 'get'
  from src/invidious/views/template.ecr:2:17 in '->'
  from ../../../usr/share/crystal/src/primitives.cr:255:3 in 'call'
  from src/invidious/helpers/logger.cr:17:33 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
  from ../../../usr/share/crystal/src/http/server.cr:498:5 in '->'
  from ../../../usr/share/crystal/src/primitives.cr:255:3 in 'run'
  from ???

matthewmcgarvey:
Fixes #1386 

Invidious has a chunk of code that runs before every request https://github.com/iv-org/invidious/blob/989317e5d399b275f4ebf28878a8a558fa23bd55/src/invidious.cr#L248

On certain requests, the whole chunk of code does not run https://github.com/iv-org/invidious/blob/989317e5d399b275f4ebf28878a8a558fa23bd55/src/invidious.cr#L269-L278

I found in at least one of them that a template is rendered ([like this one](https://github.com/iv-org/invidious/blob/989317e5d399b275f4ebf28878a8a558fa23bd55/src/invidious.cr#L5439)) and every template attempts to pull the preferences from the env. The way the code worked with the early break from the `before_all`, the preferences was not set on the env which causes an error.

This change makes sure that the preferences are set on the env before the early break. My guess is that this error has been hiding a real error.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -246,12 +246,13 @@ spawn do
 end
 
 before_all do |env|
-  begin
-    preferences = Preferences.from_json(env.request.cookies["PREFS"]?.try &.value || "{}")
+  preferences = begin
+    Preferences.from_json(env.request.cookies["PREFS"]?.try &.value || "{}")
   rescue
-    preferences = Preferences.from_json("{}")
+    Preferences.from_json("{}")
   end
 
+  env.set "preferences", preferences
   env.response.headers["X-XSS-Protection"] = "1; mode=block"
   env.response.headers["X-Content-Type-Options"] = "nosniff"
   extra_media_csp = ""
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
Tested, everything seems to be working.

Thanks you!

Issue #1257, opened 2020-06-18
Version of lsquic and boringssl used


tuxor1337:
https://github.com/omarroth/invidious/blob/1f435522b4c469e0a2c34dfc41da27db93db2803/docker/Dockerfile#L11
For people who are manually compiling invidious without Docker, it would be extremely helpful to know the version of lsquic and boringssl that you use for your `lsquic.a`. Would it be possible to mention these in the README or a similar place?

Perflyst:
You can find it here: https://github.com/iv-org/lsquic.cr

tuxor1337:
I'm asking for the version number of the [LSQUIC](https://github.com/litespeedtech/lsquic) library that is used to generate the `lsquic.a` file. I can't find that information under the link you specified.

omarroth:
The version number is listed [here](https://github.com/iv-org/lsquic.cr/blob/master/src/lsquic/liblsquic.cr#L8-L10) which obviously isn't very easy to find.

The latest release of lsquic.cr and future releases will track the parent library. So for example [v2.18.1](https://github.com/iv-org/lsquic.cr/releases/tag/v2.18.1) is built against the current latest version of lsquic (v2.18.1). There are also build instructions included as release notes that should provide the information you need to build it yourself.

PR #1419, opened 2020-10-16
Move omar.yt/lsquic/liblsquic-v2.18.1.a to GitHub.

### LINKED ISSUE #1414: The Dockerfile curl a file hosted on omar.yt
The Dockerfile curl a file hosted on omar.yt: https://github.com/iv-org/invidious/blob/master/docker/Dockerfile#L11

Since omarroth left invidious this file needs to be moved somewhere else, preferably Github.

ChaseHall:
Closes #1414.
Think I got it right.

Edit: this will obviously error until you merge it. It should work then though.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,7 +8,7 @@ RUN shards update && shards install && \
     # https://github.com/omarroth/lsquic-alpine/blob/master/APKBUILD,
     # https://github.com/omarroth/lsquic.cr/issues/1#issuecomment-631610081
     # for details building static lib
-    curl -Lo ./lib/lsquic/src/lsquic/ext/liblsquic.a https://omar.yt/lsquic/liblsquic-v2.18.1.a
+    curl -Lo ./lib/lsquic/src/lsquic/ext/liblsquic.a https://github.com/iv-org/invidious/blob/master/docker/liblsquic-v2.18.1.a?raw=true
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
I looked a bit more, and it looks like this file is actually built from https://github.com/iv-org/lsquic.cr/releases/tag/v2.18.1 since we updated lsquic.cr, I'm guessing we need to rebuild it, and, I'm guessing it's a good idea to actually put the file in the release page of lsquic.cr.

I might be wrong, obviously but I think this is the correct thing to do.

I wonder why the release was on omar.yt and not github though @omarroth

FireMasterK:
Why not build lsquic.cr in the Dockerfile instead of using a precompiled binary? It should also make room for other architectures too technically

Perflyst:
That would need quite a lot resources.

I expect these two files are the same (already checked shasums, those are different though):
https://omar.yt/lsquic/liblsquic-v2.18.1.a
https://raw.githubusercontent.com/iv-org/lsquic.cr/master/src/lsquic/ext/liblsquic.a

Why I am thinking this? See https://github.com/iv-org/invidious/issues/1257#issuecomment-660658614

TheFrenchGhosty:
Closing, see: https://github.com/iv-org/invidious/issues/1414#issuecomment-710494532

PR #1420, opened 2020-10-17
[refactor] Finish converting jobs to new job setup

matthewmcgarvey:
All jobs are now in their own classes.


src/invidious.cr
```diff
@@ -3658,12 +3597,7 @@ get "/api/v1/stats" do |env|
     next error_message
   end
 
-  if statistics["error"]?
-    env.response.status_code = 500
-    next statistics.to_json
-  end
-
-  statistics.to_json
```
matthewmcgarvey, line 3666: This chunk was removed because the only time the statistics hash has the error key is when statistics are not enabled which is handled right above this.

TheFrenchGhosty:
Thanks a lot!

PR #1428, opened 2020-10-22
Untrack config.yml and create config.example.yml

TheFrenchGhosty:
Currently Invidious include `config.yml` directly in the repo, a file that is used by the software, but that is also supposed to be edited by users.

This is a massive issue because it's actually incomplete, and editing it will cause issue for existing instance, more specifically, this issue:

```
error: Your local changes to the following files would be overwritten by merge:
	config/config.yml
Please commit your changes or stash them before you merge.
```

The best solution is to untrack this file, and instead have an example (named `config.example.yml`) that users are supposed to copy to `config.yml` (using `cp config/config.example.yml config/config.yml`) when deploying Invidious.

This PR is here to fixes that, sadly the issue mentioned before (error: Your local changes to the following files would be overwritten by merge: config/config.yml) will happen for this commit, but won't happen in the future.

**This PR will also enhance the `config.example.yml` massively, to include most/all of the content of the [Wiki configuration page](https://github.com/iv-org/invidious/wiki/Configuration) in it.**

---

One problem this cause is the fact that Docker uses the config.yml, so the dockerfile needs to be edited to include `cp config/config.example.yml config/config.yml` somewhere (I don't know what happen if the file exist though).

I'm not experienced with Docker enough to do it, so this PR won't be merged until someone else do it.

---

This may also break [Invidious-Updater](https://github.com/tmiland/Invidious-Updater) but I think it already stash the changes, I might be wrong though. @tmiland


FireMasterK:
In the Dockerfile, just make it copy from the config.sample.yml to config.yml.

Check out this line: https://github.com/iv-org/invidious/blob/master/docker/Dockerfile#L26

TheFrenchGhosty:
@FireMasterK What will happen if the instance isn't "new", and the file already exist, and was already edited?

FireMasterK:
That would make git throw an error, but docker installations shouldn't be affected by this as the configuration is done in an environment variable in docker-compose.yml

TheFrenchGhosty:
@FireMasterK Right, but this also means that we need to edit the docker-compose.yml to include every possible settings.

Or better, make the config.yml in a volume that's on `/home/invidious/`.

FireMasterK:
Yes, ideally that would be the best solution.

Perflyst:
The example configuration should contain all available configuration options.

FireMasterK:
This PR also seems to break travis, maybe we should an additional step to copy the file

TheFrenchGhosty:
@FireMasterK The PR is unfinished, but yeah, some more work is maybe needed

tmiland:
> This may also break [Invidious-Updater](https://github.com/tmiland/Invidious-Updater) but I think it already stash the changes, I might be wrong though. @tmiland

Any changes to the config file as it currently is, will break the script, yes.

I'm using sed to search and replace the user input, like password, db name, domain, port etc...

TheFrenchGhosty:
@tmiland I will finish this PR soon, then I will wait for you to get invidious-updater and for docker to get ready.

saltycrys:
What's the status on this?

TheFrenchGhosty:
Superseded by: #1557

PR #1440, opened 2020-10-25
Fix storyboards

saltycrys:
This pull request solves two issues:
- Storyboards are served from both `i.ytimg.com` and `i9.ytimg.com` but invidious currently only proxies the ones from `i.ytimg.com`
- Proxied storyboards currently don't set the Content-Type and are compressed

Note that this relies on having `config.domain` set, if it is unset storyboards will not load. For local testing it can be set to `localhost` or `0.0.0.0`.

Related issue #1342

CC @resttime @raycheung

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -3661,7 +3661,10 @@ get "/api/v1/storyboards/:id" do |env|
     end_time = storyboard[:interval].milliseconds
 
     storyboard[:storyboard_count].times do |i|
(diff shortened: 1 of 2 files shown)


raycheung:
Thank you! @saltycrys , it looks great and works well on my side!

resttime:
Awesome!

TheFrenchGhosty:
Thank you, merged!

Thanks for testing the PR @raycheung @resttime

Issue #1448, opened 2020-11-01
Playlist video count does not update accordingly


raycheung:
Issue:
The playlist video count is wrong on add / drop video.

Steps to reproduce:
- Create a playlist / pick an existing playlist
- Add a video with the drop down menu on the left
- Go to playlist page and observe the video count, it's not updated
- Remove a video from an existing playlist, the video count is also not updated

Perflyst:
Closed via https://github.com/iv-org/invidious/pull/1449

PR #1449, opened 2020-11-01
Fix stale playlist video counting on add/drop (#1448)

raycheung:
The return of `cardinality()` is not affected by the `array_append()`
and `array_remove()` of the same statement.


Perflyst:
Thank you

Issue #1451, opened 2020-11-01
Playlist starts from the beginning after the first play, disregard the sequence


raycheung:
Issue:
For a playlist with a few videos, play a video in the middle of it. When that video finishes, the beginning of the playlist is played next, instead of the immediate video according to the sequence.

Steps to reproduce:
- Go to playlist, select a playlist with multiple videos
- Click on a thumbnail of a video in the middle of the sequence to start playing
- After that video finishes, the beginning of the playlist is played, not the next in sequence

raycheung:
The playlist rendering misses the `index` (an offset count) in https://github.com/iv-org/invidious/blob/master/src/invidious/views/playlist.ecr#L91 when passing on to link generation at https://github.com/iv-org/invidious/blob/master/src/invidious/views/components/item.ecr#L55 .

I couldn't figure a way to pass the offset around with `rendered`.

TheFrenchGhosty:
@raycheung I just tested, and I actually don't have this issue.

I tried a playlist with 116 videos, played a video at around the middle, the following video was the next one in the playlist.

I also tried a playlist with 8 videos, same result.

Do you have an example playlist? Right no I can't reproduce the issue.

raycheung:
Let me test it over again, I doubt if it was my browser issue.

raycheung:
> 
> 
> @raycheung I just tested, and I actually don't have this issue.
> 
> I tried a playlist with 116 videos, played a video at around the middle, the following video was the next one in the playlist.
> 
> I also tried a playlist with 8 videos, same result.
> 
> Do you have an example playlist? Right now, I can't reproduce the issue.

I'm looking my local playlist,
```
{"title":"點播","description":"","privacy":"Public","videos":["WQzU0zVG4hg","kTzUwhPuIwg","8RcLFOCJiww","zxq8mp5LcBo","tsVXQwE_MuY","DXjN9aIn2yc","UL3rtnZB93s","gGcHv7NxLh0","UeBB_9R_FHY","ttFcm7-IBLQ","HjA8JMYrZoY","Qc6Hlg1wyfs","FjS50ATT1v0","LGz4CGazqNQ","8wB0S5nZMJw","3DGOlguqseQ","AP-jOWm7Fu4","RyoFapcT8o4","-olEFyZJ77g"]}
```

[...]

PR #1452, opened 2020-11-01
Work out the playlist continuation by video id (#1451)

raycheung:
Not the best but it tries to work out the continuation. However it always picks the first match even if the same video appears multiple times in the playlist.


TheFrenchGhosty:
Fixes: #1451

PR #1453, opened 2020-11-03
Extract /watch route from main file

matthewmcgarvey:
Related to #1121

Moving another route out of the global file. This is only a refactor, no behavior was changed.


Perflyst:
Thanks!

PR #1459, opened 2020-11-06
Extract embed routes into separate classes

matthewmcgarvey:
Related to #1121

Moving embed routes out of the global file. This is only a refactor, no behavior was changed.


Perflyst:
Thanks again! Can you please check the formatting? 

```
$ crystal tool format --check

formatting './src/invidious/routes/embed/show.cr' produced changes

The command "crystal tool format --check" exited with 1.
```

PR #1462, opened 2020-11-08
Fix missing videos tab on some channels

### LINKED ISSUE #1446: On some channels, videos tab is missing
On some channels, the videos tab, which is selected by default, doesn't appear in Invidious even though there is a videos tab on the Youtube channel. When this happens, the playlists tab is shown instead.

### Actual result
Example from the [Yewtube instance
](https://yewtu.be/channel/UCpa-Zb0ZcQjTCPP1Dx_1M8Q)

Invidious page:
image

Youtube page for reference:
image

### Expected result
The Invidious page should show a videos tab:
image

Johnquai:
Resolves #1446


TheFrenchGhosty:
Tested.

Working, thanks.

Issue #1253, opened 2020-06-17
Rename "" theme to "auto" 


Poopooracoocoo:
or "follow browser" or maybe something else?
![image](https://user-images.githubusercontent.com/51363864/84871127-7d913100-b0c3-11ea-8e54-17a01b7fa7c5.png)

TheFrenchGhosty:
Fixed in #1477

PR #1477, opened 2020-11-16
Fill empty preference values

saltycrys:
See https://github.com/iv-org/invidious/issues/1253

### DIFF
--- a/src/invidious/views/preferences.ecr
+++ b/src/invidious/views/preferences.ecr
@@ -68,7 +68,7 @@
                 <% preferences.comments.each_with_index do |comments, index| %>
                     <select name="comments[<%= index %>]" id="comments[<%= index %>]">
                         <% {"", "youtube", "reddit"}.each do |option| %>
-                            <option value="<%= option %>" <% if preferences.comments[index] == option %> selected <% end %>><%= translate(locale, option) %></option>
+                            <option value="<%= option %>" <% if preferences.comments[index] == option %> selected <% end %>><%= translate(locale, option.blank? ? "none" : option) %></option>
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
Tested and working, however I'm a bit scared that this will break stuff, so we'll wait for a proper code review.

saltycrys:
This is a visual change only.

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option:

> **value**
> The content of this attribute represents the value to be submitted with the form, should this option be selected. If this attribute is omitted, the value is taken from the text content of the option element.

The value attribute is never omitted so the text content is never submitted.

TheFrenchGhosty:
So if I understand correctly:

- this won't affect the existing cookies that store the settings
- this won't affect the URL parameters
- this won't affect the instance configuration (example: https://github.com/TheFrenchGhosty/PussTheCat.org-Configs/blob/master/Services/Invidious/config.yml#L32 where the second source is set to nothing)


An empty value will still act correctly in those or won't need to be updated with "none" instead of the empty value? Correct?

saltycrys:
Yes, correct. I went through and added visual, front-end only displays of empty preference values. The preference values themselves are still empty.

TheFrenchGhosty:
@saltycrys Okay, thanks. merging.

saltycrys:
"Fill empty preference values" was a bad title for this. What I was referring to was the value of the dropdowns on the preferences page, but with value being such a generic term and then talking about the value attribute of the HTLM option element it got too confusing.

A better title would have been "Improve display of empty dropdown values on preferences page".

Issue #945, opened 2019-12-03
[improvement] Apply dark theme (if applicable) _before_ remaining page load


merberich:
Currently, theme is applied to elements after the entire page loads. For slow connections or hosts (and in most cases mobile devices), this means the page spends a fair amount of time with the bright default theme, then enters dark theme. From a UX perspective, it would be nice to avoid the "flash" from default to dark.

haizrul:
I think https://yewtu.be use that on their website. Loads fast too.

GauthierPLM:
The issue is that currently, the theme is handled with JavaScript rather than CSS (which would solve the issue for people with browsers supporting the `prefers-color-scheme` media query).

I will give a look on how to handle this properly.

PR #1479, opened 2020-11-16
Apply dark theme immediately

saltycrys:
Themes are now controlled with a class on the body element.

If a preference is set the body element will have either "dark-theme"
or "light-theme" class. If no preference is set or the preference is
empty the class will be "no-theme".

"dark-theme" and "light-theme" are handled by darktheme.css and
lighttheme.css respectively.

"no-theme" is handled by default.css where depending on the value of
"prefers-color-scheme" the styles corresponding to "dark-theme" or
"light-theme" are applied.

Unfortunately this means that both themes are duplicated, once in the
theme .css and once in default.css.

Supersedes https://github.com/iv-org/invidious/pull/1478

Closes https://github.com/iv-org/invidious/issues/945


saltycrys:
Firefox has a `devtools.inspector.color-scheme-simulation.enabled` flag in `about:config` that is very helpful for testing this. It adds a color scheme toggle (paint brush button) at the top of the style column in the Inspector tab.

Locally saved preferences can be cleared using the Storage tab, Invidious uses cookies for all preferences and local storage for dark theme.

saltycrys:
I did some testing and found a bunch of issues:
- Embeds don't have dark theme applied
- Footer styles are overridden by dark theme styles
- Link button styles are overridden by dark theme styles

I'll fix all of these. I'll also move the theme styles into default.css next to the color scheme styles so everything is in one place.

saltycrys:
After testing:
- Video (logged in and logged out)
- Embed
- Trending page
- Channel
  - Main page
  - Playlist page
  - Community page
- Preferences page
- All data preferences subpages

I found two more issues:
- Navbar link colors overridden by light theme
- Channel owner link colors overridden by both themes

saltycrys:
I'd like to get this merged but it needs testing first, so I'm enlisting the help of all people that expressed interest in this in https://github.com/iv-org/invidious/issues/945:
- @merberich
- @haizrul
- @GauthierPLM 
- @elypter
- @OdinGitDat 
- @mehlon
- @hgoldstein
- @ReekyMarko
- @B0pol
- @urlordjames
- @Lamdarer

Could any of you help test this please?

merberich:
@saltycrys I could probably assist but don't have much experience with web dev toolchains. Is there a docker container I could run or an existing instance that has the changes? If so, or with sufficient instructions, I'd readily contribute some time.

saltycrys:
That's great, thank you for your time!

I don't think any instance has these changes but you could use docker-compose:
- clone my fork `git clone https://github.com/saltycrys/invidious`
- check out the `theme-flash` branch: `git checkout theme-flash`
- build and start cluster: `docker-compose up`
- visit `localhost:3000`

The theme switching and automatic theming based on `prefers-color-scheme` is tested and works well, but you can try it out yourself, see https://github.com/iv-org/invidious/pull/1479#issuecomment-729175208 for Firefox. Chrome probably has something similar.

What needs testing is that everything looks the same as before. The way themes work now is by setting a class on the body element, so all theming related CSS rules are now more specific than before. This can overwrite other unrelated rules.

I suggest opening the same page on the local instance and some other instance that doesn't have the new CSS and checking if they are different. If they are different in any way please note what is different, and if they are the same please note that, too.

GauthierPLM:
Hi,

I gave it a try and have not seen any issue on my side: manual transition (by changing the setting) works well as well as auto change based on prefers-color-scheme. The theme loading is now much much faster and efficient.

I have not tested every pages, I will try to do so a bit later, but I have not seen any issue yet.

merberich:
Sorry, things got a bit hectic on my end. I'll do a full test later today on Chrome and Firefox.

merberich:
Alright, I tested the landing, settings, and login pages on Chrome and Firefox, comparing to existing instances, and things looked fine on my end. This certainly resolves the initial issue I put up. Fantastic work :)

TheFrenchGhosty:
Merged, thanks!

PR #1482, opened 2020-11-20
Extract playlist routes from global file

matthewmcgarvey:
Related to https://github.com/iv-org/invidious/issues/1121

Playlist routes moved into `Invidious::Routes::Playlists`. The `/feed/playlists` was moved down to the feed section rather than put in that extracted file.


TheFrenchGhosty:
Thanks!

Issue #1488, opened 2020-11-25
Search Broken


FireMasterK:
![image](https://user-images.githubusercontent.com/20838718/100190182-b3a8cb80-2f13-11eb-8907-e5c096cbf92c.png)

https://invidious.kavin.rocks/search?q=bitcoin
https://invidious.snopyta.org/search?q=bitcoin

carverh:
Can confirm is broken on my instance as well.

dentistformyeye:
Is this due to a change in Youtube? Up to date instances worked just fine until late yesterday. Even an instance that's still on version 0.20.1-9995a8d shows a blank space, where there would be "nil assertion failed".
Video playback still works as usual.

TheFrenchGhosty:
@dentistformyeye 

>Is this due to a change in Youtube?

Yes, supposedly.

FireMasterK:
It's quite weird that these changes only broke Invidious, yt-dlc, and NewPipeExtractor work just fine.

LostWanita21:
Can confirm it affects both Yewtube and the Snopyta instances, I'm thinking it's something on Youtubes end since I'm getting similar search issues on Freetube.

qorg11:
Not working in my .onion (if that has something to do) instance

andrigamerita:
I can confirm the bug started occurring on my locally hosted instance (0.20.1-1973f93@master) as well, from about 45 hours ago. The strange thing I have experienced is the search was not entirely broken: it had like 10% chance of working and displaying results. As of now it seems to have stopped working completely.

Coaggly:
> I can confirm the bug started occurring on my locally hosted instance (0.20.1-1973f93@master) as well, from about 45 hours ago. The strange thing I have experienced is the search was not entirely broken: it had like 10% chance of working and displaying results. As of now it seems to have stopped working completely.

I'm also having a near identical experience.

TheFrenchGhosty:
Locking since it's becoming a "me too" issue. The fix ( #1491 ) will be merged soon

qorg11:
I updated invidious but my search is still not working...

Issue #1480, opened 2020-11-18
Search page result are the same


DavidHenryThoreau:
The differences between the first search page :
https://invidious.snopyta.org/search?q=diresta
and others pages are the same :
https://invidious.snopyta.org/search?q=diresta&page=2
https://invidious.snopyta.org/search?q=diresta&page=3
https://invidious.snopyta.org/search?q=diresta&page=4

girst:
here's how to solve this issue. for reference, here's the patch for s.gir.st:

https://git.gir.st/subscriptionfeed.git/commitdiff/ec4bfb8ab7a4c9cd492405b412705a42dffbaa3d

- the query parameter `q` has been renamed `search_query`
- the `page` query parameter has to be removed
- an offset `(page-1)*20` has to be given to the search_params object as field 9, type int64.

below is a patch against `fe73ecc`, not tested, since i don't run an instance.

CC: @Perflyst, @TheFrenchGhosty, @tirz (since you worked on search recently, hope that's ok).

```diff
--- search.cr.old	2020-11-29 02:14:58.078936252 +0100
+++ search.cr	2020-11-29 02:21:01.011372730 +0100
@@ -252,7 +252,7 @@
 def search(query, page = 1, search_params = produce_search_params(content_type: "all"), region = nil)
   return 0, [] of SearchItem if query.empty?
 
-  body = YT_POOL.client(region, &.get("/results?q=#{URI.encode_www_form(query)}&page=#{page}&sp=#{search_params}&hl=en").body)
+  body = YT_POOL.client(region, &.get("/results?search_query=#{URI.encode_www_form(query)}&sp=#{search_params}&hl=en").body)
   return 0, [] of SearchItem if body.empty?
 
   initial_data = extract_initial_data(body)
@@ -263,11 +263,12 @@
   return items.size, items
 end
 
-def produce_search_params(sort : String = "relevance", date : String = "", content_type : String = "",
+def produce_search_params(page : Int64 = 1, sort : String = "relevance", date : String = "", content_type : String = "",
                           duration : String = "", features : Array(String) = [] of String)
   object = {
     "1:varint"   => 0_i64,
     "2:embedded" => {} of String => Int64,
+    "9:varint"   => (page-1)*20,
   }
 
   case sort
@@ -439,10 +440,10 @@
       count = 0
     end
   else
-    search_params = produce_search_params(sort: sort, date: date, content_type: content_type,
[...]

PR #1490, opened 2020-11-26
fix: ytInitialData parsing with regex

tirz:
Fix https://github.com/iv-org/invidious/issues/1488

The issue was that our regex checked for spaces / new lines after `;`.
But by fixing this issue, we may break others parsing... (tested manually since I did not found any unit test)
So we have to enforce the regex to parse only between the brackets `{...}` and everything works fine \o/ !

Also remove the legacy code:
```
if initial_data.starts_with?("JSON.parse(\"")
```
Since it will always be false (because of `{...}`).

### DIFF
--- a/src/invidious/channels.cr
+++ b/src/invidious/channels.cr
@@ -775,38 +775,31 @@ def extract_channel_community_cursor(continuation)
   cursor
 end
 
-INITDATA_PREQUERY = "window[\"ytInitialData\"] = {"
-
 def get_about_info(ucid, locale)
-  about = YT_POOL.client &.get("/channel/#{ucid}/about?gl=US&hl=en")
-  if about.status_code != 200
-    about = YT_POOL.client &.get("/user/#{ucid}/about?gl=US&hl=en")
+  result = YT_POOL.client &.get("/channel/#{ucid}/about?gl=US&hl=en")
+  if result.status_code != 200
+    result = YT_POOL.client &.get("/user/#{ucid}/about?gl=US&hl=en")
   end
 
-  if md = about.headers["location"]?.try &.match(/\/channel\/(?<ucid>UC[a-zA-Z0-9_-]{22})/)
+  if md = result.headers["location"]?.try &.match(/\/channel\/(?<ucid>UC[a-zA-Z0-9_-]{22})/)
     raise ChannelRedirect.new(channel_id: md["ucid"])
   end
 
-  if about.status_code != 200
+  if result.status_code != 200
     error_message = translate(locale, "This channel does not exist.")
     raise error_message
   end
 
-  initdata_pre = about.body.index(INITDATA_PREQUERY)
-  initdata_post = initdata_pre.nil? ? nil : about.body.index("};", initdata_pre)
-  if initdata_post.nil?
-    about = XML.parse_html(about.body)
-    error_message = about.xpath_node(%q(//div[@class="yt-alert-content"])).try &.content.strip
-    error_message ||= translate(locale, "Could not get channel info.")
-    raise error_message
-  end
-  initdata_pre = initdata_pre.not_nil! + INITDATA_PREQUERY.size - 1
-
-  initdata = JSON.parse(about.body[initdata_pre, initdata_post - initdata_pre + 1])
-  about = XML.parse_html(about.body)
-
+  about = XML.parse_html(result.body)
   if about.xpath_node(%q(//div[contains(@class, "channel-empty-message")]))
     error_message = translate(locale, "This channel does not exist.")
+    raise error_message
+  end
+
+  initdata = extract_initial_data(result.body)
+  if initdata.empty?
+    error_message = about.xpath_node(%q(//div[@class="yt-alert-content"])).try &.content.strip
+    error_message ||= translate(locale, "Could not get channel info.")
     raise error_message
   end
 
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -598,7 +598,7 @@ def create_notification_stream(env, topics, connection_channel)
 end
 
 def extract_initial_data(body) : Hash(String, JSON::Any)
-  return JSON.parse(body.match(/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(?<info>\{.*?\});/m).try &.["info"] || "{}").as_h
+  return JSON.parse(body.match(/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(JSON\.parse\(")?(?<info>\{.*?\})("\))?;/m).try &.["info"] || "{}").as_h
 end
 
 def proxy_file(response, env)
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -598,7 +598,7 @@ def create_notification_stream(env, topics, connection_channel)
 end
 
 def extract_initial_data(body) : Hash(String, JSON::Any)
-  return JSON.parse(body.match(/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(?<info>\{.*?\});/).try &.["info"] || "{}").as_h
+  return JSON.parse(body.match(/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(?<info>\{.*?\});/m).try &.["info"] || "{}").as_h
 end
 
 def proxy_file(response, env)
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -598,12 +598,7 @@ def create_notification_stream(env, topics, connection_channel)
 end
 
 def extract_initial_data(body) : Hash(String, JSON::Any)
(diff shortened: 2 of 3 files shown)


hackerncoder:
This PR is up for testing on https://vid.encryptionin.space

saltycrys:
I like this approach better than mine. It makes the code more resistant to future YouTube changes.

I found that the JSON chunk can contain newlines, this can be fixed with the `m` flag. Also your regex currently doesn't match if `JSON.parse` is present.

How about this regex:
```regex
/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(JSON\.parse\(")?(?<info>\{.*?\})("\))?;/m
```

I also took this opportunity to move all `ytInitialData` parsing to the `extract_initial_data` helper function, see https://github.com/iv-org/invidious/pull/1491/commits/9ea20cca322ddf777b6f1194a88abfec27113686. If you want you can include this here.

tirz:
https://github.com/iv-org/invidious/commit/9ea20cca322ddf777b6f1194a88abfec27113686 is interesting. I will add it.

I got some errors if we do not enforce the brackets, so `{.*?\}` is mandatory.
About `JSON.parse("`, are you sure that this is still in used?

tirz:
@saltycrys I did like https://github.com/iv-org/invidious/commit/9ea20cca322ddf777b6f1194a88abfec27113686 with a little difference.

I changed:

```
  initdata_pre = about.body.index(INITDATA_PREQUERY)
  initdata_post = initdata_pre.nil? ? nil : about.body.index("};", initdata_pre)
  if initdata_post.nil?
    about = XML.parse_html(about.body)
    error_message = about.xpath_node(%q(//div[@class="yt-alert-content"])).try &.content.strip
    error_message ||= translate(locale, "Could not get channel info.")
    raise error_message
  end
  initdata_pre = initdata_pre.not_nil! + INITDATA_PREQUERY.size - 1

  initdata = JSON.parse(about.body[initdata_pre, initdata_post - initdata_pre + 1])
  about = XML.parse_html(about.body)

  if about.xpath_node(%q(//div[contains(@class, "channel-empty-message")]))
    error_message = translate(locale, "This channel does not exist.")
    raise error_message
  end
```

to:

```
  about = XML.parse_html(result.body)
  if about.xpath_node(%q(//div[contains(@class, "channel-empty-message")]))
    error_message = translate(locale, "This channel does not exist.")
    raise error_message
  end

  initdata = extract_initial_data(result.body)
  if initdata.empty?
    error_message = about.xpath_node(%q(//div[@class="yt-alert-content"])).try &.content.strip
    error_message ||= translate(locale, "Could not get channel info.")
    raise error_message
  end
```

Note the reverse order of `if initdata_post.nil?` and `if about.xpath_node(%q(//div[contains(@class, "channel-empty-message")]))` - also note that the value `about` was sometime renamed `result` since they was really different values with different type !

In sum up:
 - the xml `about` was shadowing the the string `about` (not the same type)
 - we were checking if the channel data was parse-able before checking if the channel really exist

hackerncoder:
The newest commit to this PR is up for testing on https://vid.encryptionin.space

saltycrys:
Looks good, `INITDATA_PREQUERY` can now be removed from `src/invidious/channels.cr`.

About `JSON.parse`: I've never seen it but better safe than sorry. What I have seen is newlines in the JSON chunk, so the `m` flag should be added to enable dots matching newlines. So `/[regex]/m`.

tirz:
- Used value `INITDATA_PREQUERY` removed.
 - `/[regex]/m` was already handled since this patch parse everything between `{...}` (and ending by `;`), the new line feed `\n` included. But I added it anyway, it will allow us to remove the strict `{...}` if `JSON.parse` is really needed.
 - I have some doubt about adding `JSON.parse` back since it look like an outdated code and their is no need to maintain it.

hackerncoder:
I finally got the newest commit to this PR is up for testing on https://vid.encryptionin.space

saltycrys:
The dot does not match newlines by default in PCRE (Perl Compatible Regular Expressions, the regex library Crystal uses).

You can try it yourself:
```shell
$ crystal eval 'puts "a\nb".match(/a.*b/).nil?; puts "a\nb".match(/a.*b/m)'
true
Regex::MatchData("a\nb")
```

I think `JSON.parse` should be included, but it's up to the maintainers.

I propose this regex for `JSON.parse` support:
```regex
/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(JSON\.parse\(")?(?<info>\{.*?\})("\))?;/m
```
which is your regex:
```regex
/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(?<info>\{.*?\});/m
```
with the info group wrapped in `(JSON\.parse\(")?` and `("\))?`.

saltycrys:
I'd like to have `JSON.parse` supported because it's the safer choice.

Best case: YouTube uses `JSON.parse` again and Invidious does not break and does not need to be updated.
Worst case: YouTube never uses `JSON.parse` again and Invidious has some unneeded code.

saltycrys:
I really like your simplification of `extract_initial_data` and just meant to include `JSON.parse` with it. Sorry about the confusion.

What I meant was:
```crystal
def extract_initial_data(body) : Hash(String, JSON::Any)
  return JSON.parse(body.match(/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(JSON\.parse\(")?(?<info>\{.*?\})("\))?;/m).try &.["info"] || "{}").as_h
end
```

The info group will have all of the JSON in any case without any special handling. No matter what follows after the data because the brackets take care of it.

TheFrenchGhosty:
PR has been tested and seems to be working, but there's one massive issue: the search results are the same on every page (tested with the search "ltt").

hackerncoder:
> PR has been tested and seems to be working, but there's one massive issue: the search results are the same on every page (tested with the search "ltt" and "pewdiepie").

Hasn't this been a problem for some time?

TheFrenchGhosty:
@hackerncoder 

>Hasn't this been a problem for some time?

I actually wasn't aware of that, but looks like you're right #1480

B0pol:
@TheFrenchGhosty it's an unrelated issue: #1480

tirz:
@saltycrys I reverted part of my revert :p
So `extract_initial_data` is a one liner with `JSON.parse("` on is regex.

@TheFrenchGhosty yes, I was trying to find this issue (https://github.com/iv-org/invidious/issues/1480).
This bug is another bug not related to the parsing of the regex :/

TheFrenchGhosty:
Should I merge it even with this issue?

hackerncoder:
@TheFrenchGhosty Yes, at least let us have some search.

tirz:
@TheFrenchGhosty yes you can, https://github.com/iv-org/invidious/issues/1480 need another PR with a totally different fix.

tirz:
@saltycrys ?

TheFrenchGhosty:
I'm merging this then, if you @saltycrys think it can be enhanced, feel free to open another PR.

saltycrys:
It cannot be enhanced, it is a fantastic one liner that solves all of our `ytInitialData` problems.

Well done @tirz and thank you!

TheFrenchGhosty:
@saltycrys Amazing then!

Big thanks to both you @saltycrys @tirz !

cy8aer:
Unfortunatelly with this patch no channel searching does not work anymore: Nil assertion failed

hackerncoder:
@cy8aer like this: https://vid.encryptionin.space/channel/UC3swwxiALG5c0Tvom83tPGg ?

cy8aer:
More like this:

grafik

cy8aer:
@hackerncoder If you mean the type of url: yes this type (sorry).

hackerncoder:
@Perflyst 

Commit b41ca72d2b85c8d6ee5186873bf4be4c331e3798 failed in travis.

My server running 1ba17a0e148b7cb52b34d756c945f5d8976b9913 is working fine, but the one @cy8aer is using which is running 99ba9874b11bb0a5b67854d338ce968e83c69c41 isn't.

hackerncoder:
This doesn't make any sense? I just updated to 99ba987 and mine is working fine, yet other instances don't work????

cy8aer:
My instance is vanilla. Are there are other factors that let this problem occur? With b41ca72d2b85c8 it worked fine.

tirz:
Interesting...
I tried to reproduce the issue on https://vid.encryptionin.space/ and https://invidious.snopyta.org/ : both of them updated to https://github.com/iv-org/invidious/commit/99ba9874b11bb0a5b67854d338ce968e83c69c41 but with no luck.
Everything is worked fine on my side :/
@cy8aer does the bug appear periodically?
Do you have a special case which may trigger the `Nil assertion failed`?
And does it affect ONLY the channels?

cy8aer:
It is only and all channels.

hackerncoder:
I seem to be able to reproduce this on all other instance, just not mine.

tirz:
https://github.com/iv-org/invidious/commit/b41ca72d2b85c8d6ee5186873bf4be4c331e3798 and https://github.com/iv-org/invidious/commit/99ba9874b11bb0a5b67854d338ce968e83c69c41 runs the sames code ^^'
It is one commit and one merge from 2 different person but with the same content... So I guess the bug is running periodically on the channels :/

cy8aer:
Hm, switched my instance from ipv4 to ipv6. Now it works. Looks like a local blocking.

Issue #1497, opened 2020-11-28
Missing error translations


saltycrys:
Missing translations:
  - `Login has been disabled by administrator.`
  - `Account has temporarily been disabled`
  - `Registration has been disabled by administrator.`
  - `Administrator has disabled this endpoint.`
  - `Title cannot be empty.`
  - `Invalid privacy setting.`
  - `User cannot have more than 100 playlists.`
  - `Playlist cannot have more than 500 videos`
  - `This playlist is private.`
  - `Video is unavailable.`
  - `Invalid parameters.`
  - `Could not extract continuation.`
  - `Could not extract community tab.`
  - `Could not extract channel info. Instance is likely blocked.`
  - `Could not extract channel JSON`
  - `Could not extract playlist info. Instance is likely blocked.`
  - `Could not extract playlist info`
  - `Could not extract author info`
  - `Invalid scope`
  - `Invalid signature`

saltycrys:
They are all missing from the locale files. I don't know how those are handled and if they can be edited directly or if they need to be updated through Weblate. If they can be edited directly I'll do so.

TheFrenchGhosty:
Blocked because of  #1426

syeopite:
#1426 has been resolved. Shouldn't this issue be unblocked?

unixfox:
It would be great to have some documentation about how to add new translations in the code.

SamantazFox:
Yes! Good idea!

PR #1498, opened 2020-11-28
Add backtraces to errors

saltycrys:
Error handling has been reworked to always go through the new `error_template`,
`error_json` and `error_atom` macros.
They all accept a status code followed by a string message or an exception
object. `error_json` accepts a hash with additional fields as third argument.

If the second argument is an exception a backtrace will be printed, if it is a
string only the string is printed. Since up till now only the exception message
was printed a new `InfoException` class was added for situations where no
backtrace is intended but a string cannot be used.

`error_template` with a string message automatically localizes the message.
Missing error translations have been collected in https://github.com/iv-org/invidious/issues/1497
`error_json` with a string message does not localize the message. This is the
same as previous behavior. If translations are desired for `error_json` they
can be added easily but those error messages have not been collected yet.

Uncaught exceptions previously only printed a generic message ("Looks like
you've found a bug in Invidious. [...]"). They still print that message
but now also include a backtrace.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -367,9 +367,7 @@ get "/search" do |env|
   begin
     search_query, count, videos = process_search_query(query, page, user, region: nil)
   rescue ex
-    error_message = ex.message
-    env.response.status_code = 500
-    next templated "error"
+    next error_template(500, ex)
   end
 
   env.set "search", query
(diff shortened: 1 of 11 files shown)


saltycrys:
I had some spelling mistakes and a lint error so I decided to do a force push.

Perflyst:
Appears to be working https://github.com/iv-org/invidious/issues/1503
Didn't noticed any other issues after this patch. Thank you!

Issue #1503, opened 2020-11-29
Video unavailable error


Hybjis:
![invidious](https://user-images.githubusercontent.com/75221845/100552171-539c8700-327d-11eb-9197-2a17cb4df3cb.png)

This error came when trying to watch a video on invidious.snopyta.org. I'm logged in if that matters

_Video unavailable (Exception)
  from src/invidious/videos.cr:1001:3 in 'fetch_video'
  from src/invidious/videos.cr:943:13 in 'get_video'
  from src/invidious/videos.cr:926:1 in 'get_video:region'
  from src/invidious/routes/watch.cr:53:15 in 'handle'
  from src/invidious.cr:299:1 in '->'
  from ../../../usr/share/crystal/src/primitives.cr:255:3 in '->'
  from ../../../usr/share/crystal/src/primitives.cr:255:3 in 'process_request'
  from lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:212:5 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:94:12 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:145:12 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:70:5 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:189:11 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/exception_handler.cr:8:7 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
[...]

Issue #1499, opened 2020-11-28
Change default admin email address


saltycrys:
Currently the default admin email address is `omarroth@protonmail.com`.

This is displayed in error messages:
>  Looks like you've found a bug in Invidious. Feel free to open a new issue <a href="https://github.com/iv-org/invidious/issues">here</a> or send an email to <a href="mailto:#{CONFIG.admin_email}">#{CONFIG.admin_email}</a>.

The default admin email address should be changed to one that belongs to a current maintainer so all bug reports end up in the issue tracker.

TheFrenchGhosty:
I'll open an email for Invidious that will be shared between the people that maintain Invidious.

FireMasterK:
Why not use something like https://fire.fundersclub.com/? Sending an email would create an issue here.

saltycrys:
Another option is to not include the admin email address in the error message. And add more useful information while we're at it.

The message could be changed to something like:
> Looks like you've found a bug in Invidious. Please open a new issue <a href="https://github.com/iv-org/invidious/issues">on GitHub</a> and include the following text in your message:
> <pre>```
> Date: 2020-01-23T00:00:00+0
> Route: /route/here
> Version: 0.20.1-hashhere @ master
> ```</pre>

Perflyst:
Sounds good to me

PR #1504, opened 2020-11-30
Improve error message

saltycrys:
The error message has been reworded and the issue template now includes the
date, route, version and backtrace. Closes https://github.com/iv-org/invidious/issues/1499

This PR also includes a fix for an oversight from my last PR:
YouTube returning an error is not a bug in Invidious, so it should not print
a backtrace. Closes https://github.com/iv-org/invidious/issues/1503


TheFrenchGhosty:
Thanks!

TheFrenchGhosty:
@saltycrys I just noticed that Travis actually complain about the formatting, I have no idea where the issue is.

Issue #1507, opened 2020-11-30
Improve error message 2!


B0pol:
https://github.com/iv-org/invidious/pull/1504 has improved the error message but it still can be improved.
The current system will flood the github issues with **long** crash reports.

An easy way to counter this is to use `<details>` and `<summary>` HTML tags ([relevant documentation](https://developer.mozilla.org/fr/docs/Web/HTML/Element/details)).

Example with a crash report from NewPipe:

## Exception
* __User Action:__ ui error
* __Request:__ App crash, UI failure
* __Content Country:__ FR
* __Content Language:__ fr-FR
* __App Language:__ fr_FR
* __Service:__ none
* __Version:__ 0.20.1
* __OS:__ Linux Android 9 - 28
<details><summary><b>Crash log </b></summary><p>

```
java.lang.RuntimeException: Error receiving broadcast Intent { act=org.schabi.newpipe.VideoDetailFragment.ACTION_PLAYER_STARTED flg=0x10 } in org.schabi.newpipe.fragments.detail.VideoDetailFragment$4@4feadd9
	at android.app.LoadedApk$ReceiverDispatcher$Args.lambda$getRunnable$0(LoadedApk.java:1401)
	at android.app.-$$Lambda$LoadedApk$ReceiverDispatcher$Args$_BumDX2UKsnxLVrE6UJsJZkotuA.run(Unknown Source:2)
	at android.os.Handler.handleCallback(Handler.java:873)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loop(Looper.java:193)
	at android.app.ActivityThread.main(ActivityThread.java:6718)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:491)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
Caused by: java.lang.IllegalStateException: Not allowed to start service Intent { cmp=org.schabi.newpipe/.player.MainPlayer }: app is in background uid UidRecord{1b3bf85 u0a590 CAC  bg:+2m34s765ms idle procs:1 seq(0,0,0)}
	at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1577)
	at android.app.ContextImpl.startService(ContextImpl.java:1532)
	at android.content.ContextWrapper.startService(ContextWrapper.java:664)
[...]

PR #1513, opened 2020-12-01
Improve error message 2

saltycrys:
**Switch to date based versioning scheme**

Since no new tags are created for releases the version has been frozen for some
time, with only the commit hash changing.

Versions based on the latest commit date make it much easier to identify them.

**Improve error message 2**

The long backtrace has been moved into a `<details>` HTML element, as suggested
by @B0pol. To make the error still visible it has been added to the top under
`Title:`. This also encourages informative issue titles.

Closes https://github.com/iv-org/invidious/issues/1507


TheFrenchGhosty:
>Since no new tags are created for releases the version has been frozen for some
time, with only the commit hash changing.

This is on purpose, Invidious is much better without them.

TheFrenchGhosty:
Tested, and after discussing it with @b0pol this is indeed what he was expecting.

Thank you! Merged.

PR #1519, opened 2020-12-01
Switch to GitHub Actions for CI

saltycrys:
Travis CI is slow, builds are first queued up forever and then the builds
themselves are slow, too.

GitHub Actions run pretty much immediately and are fast. Also separate jobs run
at the same time unlike on Travis CI.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+      
+      - name: Run tests
+        run: crystal spec
+      
+      - name: Run lint
+        run: |
+          if ! crystal tool format --check; then
+            crystal tool format
+            exit 1
+          fi
```
matthewmcgarvey, line 55: Why not just `crystal tool format --check`?

saltycrys: I wanted to add a diff to see what was formatted but forgot about it. I'll fix it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+      #
+      # --------------------------------------------------------------------------------
+      #
+      # running either `shards install` or `shards update` without a `shard.lock` file does the same thing,
+      # downloading shards into the `lib` folder and writing a `shard.lock` file
+      #
+      # if there is a `shard.lock` file `shards install` installs the versions from the lock file while
+      # `shards update` does a `shards install` that pretends there isn't a lock file present
+      - name: Update and install shards
+        run: shards update
```
saltycrys, line 45: @matthewmcgarvey I wanted to add caching for the shards and crystal itself but at least for the shards that's not possible without `shard.lock`. This file isn't tracked anymore and I don't know why. I think it should be, is that right?

Builds are already only three minutes though, so caching isn't really necessary.

matthewmcgarvey: I wouldn't expect a service to ignore that file. The git blame doesn't turn up any reasoning and it was done 3 years ago. I'm for adding it back.

saltycrys: Cool, I'll add it back and enable caching.

saltycrys:
I could not figure out how to cache Crystal itself, but the shards are cached now. Installing Crystal only takes seven seconds though so it's fine.

See https://github.com/saltycrys/invidious/actions for some example runs.

.github/workflows/ci.yml
```diff
@@ -10,39 +10,48 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: ./lib
+          key: shards-${{ hashFiles('shard.lock') }}
+      
+      - name: Install Shards
+        run: |
+          if ! shards check; then
+            shards install
+          fi
```
matthewmcgarvey, line 54:
```suggestion
        run: shards install
```

Should be good enough I think

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
+name: Invidious CI
+
+on: [push, pull_request]
```
matthewmcgarvey, line 3: This is just my opinion, but, I would normally have this be pushes to master only and all pull requests. With this running on all pushes and pull requests I believe it runs twice.

.github/workflows/ci.yml
saltycrys: `shards check` is instant while `shards install` fetches all dependencies, so with the check it's a couple seconds faster.

.github/workflows/ci.yml
saltycrys: How can this be fixed?

matthewmcgarvey: This instead
```yml
on:
  push:
    branches: "master"
  pull_request:
    branches: "*"
```

saltycrys: Great, thanks!

matthewmcgarvey (approved):
I'm happy with this 👍

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
+name: Invidious CI
+
+on:
+  push:
+    branches: "master"
```
TheFrenchGhosty, line 5: This should apply to every branch if possible (the api-only branch was used by some people)

saltycrys: Fixed. We should deprecate that branch though, it hasn't been updated since March. We can add a config option instead.

saltycrys:
Travis jobs for https://github.com/iv-org/invidious/pull/1525 have been running for three hours now.

.github/workflows/ci.yml
TheFrenchGhosty: Technically it has it's use, the problem is that I don't know how to maintain it.

TheFrenchGhosty:
> Travis jobs for #1525 have been running for three hours now.

Yeah I know, I'm not really experienced enough to review this, and @Perflyst seems busy. If he doesn't review that tomorrow, I'll trust you guys and merge this.

.github/workflows/ci.yml
saltycrys: The `api-only` branch is just the `master` branch with only `/api/*` and proxy related endpoints. Once the routes are all extracted it's easy to add an `api-only` config option that disables any front-end endpoints.

If you want to manually maintain the `api-only` branch until then you can merge the current master into it and then just delete all the front-end routes from `src/invidious.cr`

TheFrenchGhosty:
No news from Perflyst.

Merging, let's hope nothing break.

Thank you, for this by the way! Far better than Travis

TheFrenchGhosty:
Anything needed for it to be run on already opened PR? Or it's not possible?

saltycrys:
I don't know but a push might trigger it.

PR #1525, opened 2020-12-03
Fix `extract_polymer_config`

saltycrys:
The `ytInitialPlayerResponse` regex can now handle `var` and `window`
assignments.

The video streams can now be extracted from `player_response` and
`initial_data`.

This fixes the descriptions on videos and videos themselves. Videos are
technically broken right now, but work becasue of a fallback that goes through
embeds.

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -816,7 +816,7 @@ end
 
 def extract_polymer_config(body)
   params = {} of String => JSON::Any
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
Tested.

What you fixed is indeed working, but the Comments are still broken, would be nice to fix that too if you can.

If you want I can merge that on its own.

saltycrys:
I'll look into comments. They're separate from this though so you can merge this and I'll make another PR for the comments.

TheFrenchGhosty:
Great, thanks!

Issue #1442, opened 2020-10-25
Nil assertion failed (NilAssertionError)


B0pol:
Cannot get channel https://invidio.us/channel/UCITk7Ky4iE5_xISw9IaHqpQ

It's because you do not support redirect channels. The url you get on youtube is https://invidious.snopyta.org/channel/UCrxkwepj7-4Wz1wHyfzw-sQ. See https://github.com/TeamNewPipe/NewPipeExtractor/pull/440

charliebarber:
I would like to try and fix this. Do you have any more information about redirect channels?

B0pol:
Date: `2020-11-30T15:30:26Z`
Route: `/channel/UCITk7Ky4iE5_xISw9IaHqpQ`
Version: `0.20.1-78ae9ff @ master`
Backtrace: 
```
Nil assertion failed (NilAssertionError)
  from ../../../usr/lib/crystal/nil.cr:106:5 in 'not_nil!'
  from src/invidious/channels.cr:804:12 in 'get_about_info'
  from src/invidious.cr:2475:15 in '->'
  from ../../../usr/lib/crystal/primitives.cr:255:3 in '->'
  from ../../../usr/lib/crystal/primitives.cr:255:3 in 'call'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:189:11 in 'call'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call'
  from ../../../usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
[...]

PR #1526, opened 2020-12-03
Fix redirect channels

saltycrys:
Redirect channels may use JS to redirect now, instead of only a response header
as it used to be. This fix reads the channel to redirect to from `ytInitialData`.

Closes https://github.com/iv-org/invidious/issues/1442


TheFrenchGhosty:
Tested and working perfectly. Thanks!

PR #1529, opened 2020-12-04
Fix invidious version for old git versions

saltycrys:
The `%cs` format was only added to git in version 2.25 while `%ci` has been
around forever.

See https://github.com/iv-org/invidious/issues/1527


TheFrenchGhosty:
Thanks :)

PR #1538, opened 2020-12-05
Fix invidious version for docker image

Sonic-Y3k:
As Alpine-Linux per default uses the busybox date command, the -f option is not available and the build will fail.


saltycrys (approved):
Thank you!

TheFrenchGhosty:
Thanks!

PR #1539, opened 2020-12-06
Fix comments

saltycrys:
The YouTube headers are now always added for requests to YouTube.
Previously they were only added for requests going through QUIC.

The session token is now JSON decoded to unescape escaped Unicode characters.

The comment continuation protobuf has been updated and the request now goes
through the YouTube `pbj` JSON API.


TheFrenchGhosty:
Amazing! Thanks a ton!

spec/helpers_spec.cr
```diff
@@ -87,13 +87,13 @@ describe "Helper" do
...
 
-      produce_comment_continuation("29-q7YnyUmY", "").should eq("EiYSCzI5LXE3WW55VW1ZwAEByAEB4AEBogINKP___________wFAABgGMhUKACIPIgsyOS1xN1lueVVtWTAAKBQ%3D")
+      produce_comment_continuation("29-q7YnyUmY", "").should eq("EkMSCzI5LXE3WW55VW1ZyAEA4AEBogINKP___________wFAAMICHQgEGhdodHRwczovL3d3dy55b3V0dWJlLmNvbSIAGAYyFQoAIg8iCzI5LXE3WW55VW1ZMAAoFA%3D%3D")
```
TheFrenchGhosty, line 94: This PR made me learn that Invidious actually had tests... we need to change those videos ID to something else.

saltycrys: I think using copyrighted videos is fine in this context but to change the tests you can just put in new video IDs and let the tests fail, it will print what it got and what it expected. Then update the tests with the ctoken it expected.

PR #1540, opened 2020-12-06
Increase YouTube request timeout

saltycrys:
From 0.1 seconds to 2 seconds.

The default timeout is five seconds (`QUICPool` in `src/invidious/helpers/utils.cr`)
but it seems to have been chose randomly. If reads continue to time out the
timeout can be increased again.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -104,7 +104,7 @@ LOCALES = {
   "zh-TW" => load_locale("zh-TW"),
 }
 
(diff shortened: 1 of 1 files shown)


saltycrys:
The one time I don't compile everything breaks, serves me right for not testing locally.

@TheFrenchGhosty if you could squash these two commits that would be great.

TheFrenchGhosty:
Thanks!

saltycrys:
And thank you :)

Issue #1454, opened 2020-11-03
The README file in this repo has a bad link - [404:NotFound]


MrCull:
The README file in this repo has a bad link - [404:NotFound]

Status code [404:NotFound] - Link: https://github.com/iv-org/invidious/blob/master/screenshots/03_subscriptions.png?raw=true

This was found by an new experimental hobby project that I have just created: https://github.com/MrCull/GitHub-Repo-ReadMe-Dead-Link-Finder
If this has been in any way helpful then please consider giving the above Repo a Star.

TheFrenchGhosty:
Fixed in #1543

Issue #1331, opened 2020-08-09
You were added to awesome-humane-tech


aschrijver:
This is just a FYI issue to notify that you were added to the curated awesome-humane-tech in the 'Social media' category, and - if you like that - are now entitled to wear our badge:

[![Awesome Humane Tech](https://raw.githubusercontent.com/humanetech-community/awesome-humane-tech/main/humane-tech-badge.svg?sanitize=true)](https://github.com/humanetech-community/awesome-humane-tech)

By adding this to the README:

```markdown
[![Awesome Humane Tech](https://raw.githubusercontent.com/humanetech-community/awesome-humane-tech/main/humane-tech-badge.svg?sanitize=true)](https://github.com/humanetech-community/awesome-humane-tech)
```

https://github.com/humanetech-community/awesome-humane-tech

aschrijver:
For other readers. Downvotes probably related to wrong association of our community, caused by forum + website URL's being *.humametech.com

The Humane Tech Community is fully independent of Center for Humane Technology. I started the grassroots community as a volunteer, using their forum when they had no time for community building. Other than that we have our own website at https://humanetech.community

PR #1543, opened 2020-12-06
Add missing screenshot and update the older ones

schwukas:
I took the liberty to fix the missing link mentioned in #1454 and update all the existing screenshots since the frontend changed a little bit since they were taken. 

This would also be a good opportunity to add the _awesome-humane-tech_ badge mentioned in #1331 however there are quite a lot of :-1: which led me to believe the badge is not really wanted.


TheFrenchGhosty:
Thanks.

About the "awesome-humane-tech", I'm personally not against it but some people seems to be, so it needs to be discussed if possible.

Issue #1301, opened 2020-07-25
Fix Import youtube subscriptions link


protodrew:
it redirects to https://support.google.com/youtube/answer/6224202?hl=en

which is a depreciated solution.

redirect to https://www.youtube.com/subscription_manager instead

fazlerabbi37:
> redirect to https://www.youtube.com/subscription_manager instead

this seems to get redirected to https://www.youtube.com/feed/channels where I don't see any option to export sub list. any idea?

schwukas:
From my latest research on this it seems there is no way to export your subscriptions anymore.

fazlerabbi37:
Haven't tested it yet, but there is a new instruction for NewPipe which may be applicable for this as well? The new link there is: https://takeout.google.com/takeout/custom/youtube

schwukas:
Seems to work fine and produces output like this:

```
[ {
  "contentDetails" : {
    "activityType" : "all",
    "newItemCount" : 0,
    "totalItemCount" : 51
  },
  "etag" : "XK-zfi9HThstwtdK7I-lj77puOw",
  "id" : "C_9pstbkHjNjllqH3C0S1ERFUH5xidQoExsN8wRwc94",
  "kind" : "youtube#subscription",
  "snippet" : {
    "channelId" : "UCkX5ePSBJFiw8TzmxfzXx1A",
    "description" : "Go to www.c90adventures.co.uk for more C90 related adventure and stupidity.",
    "publishedAt" : "2018-05-25T21:20:54.022Z",
    "resourceId" : {
      "channelId" : "UCVqpNG1R72i21jh-nAxEk4A",
      "kind" : "youtube#channel"
    },
    "thumbnails" : {
      "default" : {
        "url" : "https://yt3.ggpht.com/ytc/AAUvwnj4GweWE0W7V7QZIcD5wWZv99Iwkfs_pxJ0KQ=s88-c-k-c0xffffffff-no-rj-mo"
      },
      "high" : {
        "url" : "https://yt3.ggpht.com/ytc/AAUvwnj4GweWE0W7V7QZIcD5wWZv99Iwkfs_pxJ0KQ=s800-c-k-c0xffffffff-no-rj-mo"
      },
      "medium" : {
        "url" : "https://yt3.ggpht.com/ytc/AAUvwnj4GweWE0W7V7QZIcD5wWZv99Iwkfs_pxJ0KQ=s240-c-k-c0xffffffff-no-rj-mo"
      }
    },
[...]

PR #1544, opened 2020-12-06
Fix youtube subscription import

schwukas:
This PR fixes the broken YouTube subscription import mentioned in #1301 or rather adapts the new format and structure for the parser. I also included a link to the instructions on how to export your subscriptions as a JSON file. This link is only temporary as it links to my fork but can be easily swapped for a URL to the documentation repo or similar before merging. 

There's probably a more elegant solution for parsing the JSON. Feel free to heavily criticise this, as I'm not at all familiar with Crystal. I promise I won't cry :slightly_smiling_face:

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -1492,9 +1492,10 @@ post "/data_control" do |env|
           end
         end
       when "import_youtube"
-        subscriptions = XML.parse(body)
-        user.subscriptions += subscriptions.xpath_nodes(%q(//outline[@type="rss"])).map do |channel|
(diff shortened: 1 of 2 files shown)


TheFrenchGhosty:
I personally can't test this since I don't use youtube, and don't use an Invidious account.

The link indeed needs to be swapped for a link to the documentation page (please open an PR in the documentation repo with the content if you can).

saltycrys (approved):

TheFrenchGhosty:
Since it's currently broken, I will assume that this will work and merge as is.

Thanks for the review @saltycrys and thanks for this PR @schwukas !

PR #1548, opened 2020-12-07
Remove unused steps from CI

saltycrys:
There is no way to cache Crystal unfortunately.
See https://github.com/oprypin/install-crystal/issues/2


TheFrenchGhosty:
Thanks!

PR #1549, opened 2020-12-07
Fix Invidious playlist continuation

saltycrys:
Playing a video in a Invidious playlist now correctly redirects to the next
video. The problem was that the offset was updated in the wrong place.

Supersedes https://github.com/iv-org/invidious/pull/1452

Closes https://github.com/iv-org/invidious/issues/1451


TheFrenchGhosty:
I can't this since I don't use an Invidious account. I will assume that this will work though.

Thanks

Issue #1471, opened 2020-11-12
Feature Request: Add embed link


ubone:
Add embed link to YT next to the watch on YT link ie https://www.youtube.com/embed/some-id

Obosobrio:
Gone. !!

——-
Warning : May contain misinformation. Recipients must take responsibility
for fact checking content.
——-

On Nov 12, 2020, at 09:15, ubone <notifications@github.com> wrote:

﻿
Add embed link to YT next to the watch on YT link ie https://www.youtube.com/embed/some-id

TheFrenchGhosty:
@Obosobrio What?

PR #1551, opened 2020-12-08
Add embed link to watch page

saltycrys:
This makes for a great fake fullscreen when watching playlists. A real fullscreen solution would be to load the next video on the same page, like YouTube does it.

Closes https://github.com/iv-org/invidious/issues/1471

### DIFF
--- a/assets/js/embed.js
+++ b/assets/js/embed.js
@@ -31,6 +31,11 @@ function get_playlist(plid, retries) {
                     player.on('ended', function () {
                         var url = new URL('https://example.com/embed/' + xhr.response.nextVideo);
 
+                        url.searchParams.set('list', plid);
+                        if (!plid.startsWith('RD')) {
+                            url.searchParams.set('index', xhr.response.index);
+                        }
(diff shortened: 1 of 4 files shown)


TheFrenchGhosty:
Okay, so I just tested it, it works great... 

But I think you didn't understand the feature request: basically they were asking for a link to the YouTube embed, not to the Invidious one, this is useful in case the video is blocked in the instance country, or if it's broken for *whatever* reason.

saltycrys:
Oh, in that case it doesn't close that issue.

I added the embed link to make it easy to navigate to the Invidious embed. Using the embed with a playlist makes for a great fake fullscreen. By that I mean for example F11 on Firefox.

TheFrenchGhosty:
This is a good idea to have both yeah. Should I merge this, or do you want to do the youtube embed at the same time?

saltycrys:
This can be merged.

TheFrenchGhosty:
Thanks, perfect!

PR #1552, opened 2020-12-08
do not proxy hls livestrem on supported browser

azhao-squareup:
### Issue ###

Currently, invidious proxies all livestream video data.  I have seen that the generally proxied hls livestream has poor performance and lag.  For some browsers, the browser can play hls natively, meaning we are not restricted by xhr same-origin restrictions.  Some browsers which support this:
* mobile browsers (safari, chrome, firefox  on ios and android)
* safari

### Change ###
This change allows the livestream to proceed without proxying if a supported browser is being used. If the browser is not supported, or the proxy option is selected in the options, then proxy will be used.
1. Unsupported browser - behavior should be unaffected.
1. Supported browser and proxy disabled - use native hls support and connect directly to youtube server.
1. Supported browser and proxy enabled - use native hls support and connect to proxy.

### testing ###
1. Tested on macos firefox and chrome - behavior is the same
1. Tested on macos safari - stream is not proxied unless proxy is selected in options
1. Tested on android firefox - stream is not proxied unless selected in options

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -41,7 +41,15 @@ var shareOptions = {
     embedCode: "<iframe id='ivplayer' width='640' height='360' src='" + embed_url + "' style='border:none;'></iframe>"
 }
 
+videojs.Hls.xhr.beforeRequest = function(options) {
+    if (options.uri.indexOf('local=true') === -1) {
+        options.uri = options.uri + '?local=true';
+    }
+    return options;
+};
+
 var player = videojs('player', options);
+
 
 if (location.pathname.startsWith('/embed/')) {
     player.overlay({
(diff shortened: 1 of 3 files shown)


TheFrenchGhosty:
I'm not really able to test that, so if someone reviewing this code think it's fine I'll merge it.

Perflyst:
The author has tested three different scenarios, so this should be ok to merge. If there are new issues afterwards with livestreams we will know soon.

saltycrys (approved):
Code looks fine.

TheFrenchGhosty:
Thanks, merging.

saltycrys:
This is breaking Docker installs, see https://github.com/iv-org/invidious/issues/1569.

@jksladjflkjsadflkjsadf why is `?local=true` being added to `/videoplayback` URLs?

PR #1557, opened 2020-12-09
Untrack config.yml

saltycrys:
`config/config.yml` has been untracked and moved to `config/config.example.yml`.

The Dockerfile has been updated to copy all `config/config.*` files and to try
to move `config/config.example.yml` to `config/config.yml`. If a user supplied
`config/config.yml` exists it is not overwritten.

The Dockerfile has also been updated to use `shard.lock` like the regular build.

Supersedes https://github.com/iv-org/invidious/pull/1428


TheFrenchGhosty:
So, if I understand correctly: it's the same as #1428 except it doesn't break CI, will work with existing instances, and doesn't contain? Every options (see https://github.com/iv-org/invidious/pull/1428/commits/c0aca1883c81cd3ebed14f5cee6d0eb8e834d862 )?

spec/helpers_spec.cr
```diff
@@ -12,7 +12,7 @@ require "../src/invidious/search"
 require "../src/invidious/trending"
 require "../src/invidious/users"
 
-CONFIG = Config.from_yaml(File.open("config/config.yml"))
+CONFIG = Config.from_yaml(File.open("config/config.example.yml"))
```
Perflyst, line 15: What is the point of this change? I guess if you want to install invidious you have to copy config.example.yml to config.yml?

saltycrys: > So, if I understand correctly: it's the same as #1428 except it doesn't break CI, will work with existing instances, and doesn't contain every options (see [c0aca18](https://github.com/iv-org/invidious/commit/c0aca1883c81cd3ebed14f5cee6d0eb8e834d862) )?
> 
> Basically we just need to add add everything to config.example.yml and it will work?

Yes, correct.

> How will docker environment variable work with that though?

It is unaffected, it will work the same as before.

saltycrys:
> So, if I understand correctly: it's the same as #1428 except it doesn't break CI, will work with existing instances, and doesn't contain every options (see [c0aca18](https://github.com/iv-org/invidious/commit/c0aca1883c81cd3ebed14f5cee6d0eb8e834d862) )?
> 
> Basically we just need to add add everything to config.example.yml and it will work?

Yes, correct.

> How will docker environment variable work with that though?

It is unaffected, it will work the same as before.

TheFrenchGhosty:
@saltycrys Thanks, ideally I want both the example + docker-compose file to include every configuration option possible before merging this (like https://github.com/iv-org/invidious/commit/c0aca1883c81cd3ebed14f5cee6d0eb8e834d862 but with every single option https://github.com/iv-org/documentation/blob/master/Configuration.md ). (I'll do that later)

saltycrys:
So do I have to do anything else to get this merged?

TheFrenchGhosty:
@saltycrys Nope, unless you want to do the boring job of including every configuration option in the example + docker-compose instead of me xD

spec/helpers_spec.cr
Perflyst: okay, I get it. spec contains the CI test.

PR #1565, opened 2020-12-10
Update README.md

freestylemaster:
already logged in


Perflyst:
If you import the table as "postgres" user the table is owned by the "postgres" user instead of "kemal" user.

Perflyst:
If you see this different please let us know.

Issue #646, opened 2019-07-30
Option to chose default dash quality


atahanacar:
The "auto" function is good, but not enough for all use cases. Adding a default quality option to dash would solve a lot of issues. For example, if I want to limit the quality to 480p because of, say, limited data, or to 720p because the monitor itself is 720p, "auto" allows me to do neither. This option would also be good for when "auto" gets buggy because of non-stable net speed.

rezad1393:
I need this too.
the non-dash options are limited but dash option doesnt have a preset quality that user can select.

unixfox:
bump still interested

PR #1572, opened 2020-12-13
Add DASH quality preference

saltycrys:
The options are `auto` (the current and default behavior), `best` and `worst`.

The UI is only updated once playback starts.

Closes https://github.com/iv-org/invidious/issues/646

### DIFF
--- a/src/invidious/views/preferences.ecr
+++ b/src/invidious/views/preferences.ecr
@@ -57,6 +57,17 @@
                 </select>
             </div>
 
+            <% if !CONFIG.disabled?("dash") %>
+                <div class="pure-control-group">
+                    <label for="quality_dash"><%= translate(locale, "Preferred dash video quality: ") %></label>
+                    <select name="quality_dash" id="quality_dash">
+                        <% {"auto", "best", "worst"}.each do |option| %>
+                            <option value="<%= option %>" <% if preferences.quality_dash == option %> selected <% end %>><%= translate(locale, option) %></option>
+                        <% end %>
+                    </select>
+                </div>
+            <% end %>
+
             <div class="pure-control-group">
                 <label for="volume"><%= translate(locale, "Player volume: ") %></label>
                 <input name="volume" id="volume" data-onrange="update_volume_value" type="range" min="0" max="100" step="5" value="<%= preferences.volume %>">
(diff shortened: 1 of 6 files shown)


LennyPenny:
would be cool if you could select something like "1440p" instead, where it would try and select 1440p by default, but fall back to lower qualities if the video does not support that resoltuion

saltycrys:
I added support for selecting a specific quality and rebased onto master.

@unixfox could you test this please?

Perflyst:
I am going to test this now

src/invidious/views/preferences.ecr
```diff
@@ -57,6 +57,17 @@
                 </select>
             </div>
 
+            <% if !CONFIG.disabled?("dash") %>
+                <div class="pure-control-group">
+                    <label for="quality_dash"><%= translate(locale, "Preferred dash video quality: ") %></label>
+                    <select name="quality_dash" id="quality_dash">
+                        <% {"auto", "best", "4320p", "2160p", "1440p", "1080p", "720p", "480p", "360p", "240p", "144p", "worst"}.each do |option| %>
```
Perflyst, line 64: All those options are available in the preferences. In your PR you say only auto best and worst should be available?

lenny-optiopay: I think he just implemented my comment and didnt update the description https://github.com/iv-org/invidious/pull/1572#issuecomment-748291375

LennyPenny: I think he just implemented what I was suggested in my comment (https://github.com/iv-org/invidious/pull/1572#issuecomment-748291375) and didn't update the description

saltycrys: > I think he just implemented what I was suggested in my comment ([#1572 (comment)](https://github.com/iv-org/invidious/pull/1572#issuecomment-748291375)) and didn't update the description

Correct. An updated description would be:

The options are `auto` (the current and default behavior), `best`, `worst`, `4320p`, `2160p`, `1440p`, `1080p`, `720p`, `480p`, `360p`, `240p` and `144p`.
If the target height is not available the closest lower height is selected.

The UI is only updated once playback starts.

rezad1393:
thanks for this.
now hopefully I don't have to worry too much about firefox buffering  whole video even before playing it.

PR #1584, opened 2020-12-16
Extract out User Preferences endpoints

matthewmcgarvey:
Related to https://github.com/iv-org/invidious/issues/1121

One day... this will be done... maybe


TheFrenchGhosty:
Thanks!

PR #1592, opened 2020-12-19
Remove "Top" feed option from preferences

saltycrys:
The Top feed used to be a feed based on YouTube ratings. Once YouTube removed
publicly available ratings the Top feed was removed from Invidious but the
option to display a link to it remained.

This change might cause the feed menu preferences to change in unexpected ways the first time preferences are updated.
- Instance admins should update the default feed menu preference to avoid any issues.
- If bug reports come in about the feed menu reordering itself they can be ignored.

### DIFF
--- a/src/invidious/routes/user_preferences.cr
+++ b/src/invidious/routes/user_preferences.cr
@@ -74,7 +74,7 @@ class Invidious::Routes::UserPreferences < Invidious::Routes::BaseRoute
     default_home = env.params.body["default_home"]?.try &.as(String) || CONFIG.default_user_preferences.default_home
 
     feed_menu = [] of String
-    5.times do |index|
+    4.times do |index|
       option = env.params.body["feed_menu[#{index}]"]?.try &.as(String) || ""
       if !option.empty?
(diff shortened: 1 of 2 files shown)


TheFrenchGhosty:
Thanks!

TheFrenchGhosty:
By the way, in my [config](https://github.com/TheFrenchGhosty/PussTheCat.org-Configs/blob/master/Services/Invidious/config.yml) I have "top_enabled" set to false, is this the same "top" you're removing here? Because the comments of the files says "disable the "top", showing most watched videos on this instance" meaning that this isn't a youtube feed...

Did this PR actually remove a feature?

saltycrys:
The feature was removed in https://github.com/iv-org/invidious/commit/4d4b6a2fa03b9de428a931729600c252ed801c36, this just updates the preferences page accordingly.

The `top_enabled` option doesn't exist anymore. A `popular_enabled` option should be added though.

TheFrenchGhosty:
Oh interesting I didn't know that! Thanks, I'm updating the config.

PR #1594, opened 2020-12-21
Improve logging

### LINKED ISSUE #385: Configure Log level
Add an option to set the log level. For example I do not want `GET` request logged.
Maybe following log levels:
Error, Warn, (Normal / Default), Debug

I do not know if there are already any debug logs which you may use somewhere for testing, if not just drop debug.

saltycrys:
Everything that gets logged now has a log level associated with it.

The log level can be set with the new `-l` or `--log-level` arguments.

The defaul log level is `debug` for now. There aren't many things that get
logged but if the logs get spammed in the future it can be set down to `info`.

Closes #385

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -107,7 +107,8 @@ LOCALES = {
 YT_POOL = QUICPool.new(YT_URL, capacity: CONFIG.pool_size, timeout: 2.0)
 
 config = CONFIG
-logger = Invidious::LogHandler.new
+output = STDOUT
+loglvl = LogLevel::Debug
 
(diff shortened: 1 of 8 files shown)


Perflyst:
Closes #385

saltycrys:
@Perflyst do you want me to set the default log level to `warn` instead? That wouldn't print requests.

Perflyst:
No, if debug matches the current logging debug by default is fine.

GauthierPLM:
This is a good improvement. Paired with more verbose logging (more informations returned), we could use that to improve the reliability of Invidious instances and finally solve these "Video not available" errors.

PR #1595, opened 2020-12-21
Improve storyboards

saltycrys:
Instead of limiting the width of storyboards through CSS it is now done in the
VTT directly.


Perflyst:
Thanks :+1:

Issue #1132, opened 2020-04-22
Read timed out


T0MuX:
Hi, I just installed Invidious on my server, following the instructions on this GitHub. It worked directly at the first start, but an issue keep occurring : Read timed out. This error shows up on the "TRENDING" page, and on many videos. Also, all thumbnails are missing on the "popular" page 

Instance URL : https://yt.mux.re

Instance version : 0.20.1-408f385 @ master

I find this error in invidious.log : 
```
Exception: Read timed out (IO::TimeoutError)
  from /usr/lib/crystal/io/evented.cr:120:45 in 'unbuffered_read'
  from /usr/lib/crystal/io/buffered.cr:241:5 in '???'
  from /usr/lib/crystal/io/buffered.cr:83:21 in 'read'
  from lib/pool/src/pool.cr:55:7 in 'checkout'
  from src/invidious/videos.cr:30:14 in 'fetch_video'
  from src/invidious/videos.cr:906:13 in 'get_video:region'
  from src/invidious.cr:5392:11 in '->'
  from lib/kemal/src/kemal/route.cr:12:9 in '->'
  from src/invidious/helpers/handlers.cr:255:3 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:189:11 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
  from /usr/lib/crystal/fiber.cr:255:3 in 'run'
  from ???
```

My nginx vhost : 
```
server {
    listen 443 ssl http2;
[...]

PR #1600, opened 2020-12-23
Close http clients after use

jksladjflkjsadflkjsadf:
The crystal http client maintains a keepalive connection to the other
server which stays alive for some time. This should be closed if the
client instance is not used again to avoid hogging resources

### DIFF
--- a/src/invidious/helpers/utils.cr
+++ b/src/invidious/helpers/utils.cr
@@ -101,6 +101,15 @@ def make_client(url : URI, region = nil)
   return client
 end
 
+def make_client(url : URI, region = nil, &block)
+  client = make_client(url, region)
+  begin
+    yield client
+  ensure
+    client.close
+  end
+end
+
 def decode_length_seconds(string)
   length_seconds = string.gsub(/[^0-9:]/, "").split(":").map &.to_i
   length_seconds = [0] * (3 - length_seconds.size) + length_seconds
@@ -361,7 +370,7 @@ def subscribe_pubsub(topic, key, config)
     "hub.secret"        => key.to_s,
   }
 
-  return make_client(PUBSUB_URL).post("/subscribe", form: body)
+  return make_client(PUBSUB_URL, &.post("/subscribe", form: body))
 end
 
 def parse_range(range)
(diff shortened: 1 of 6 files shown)


azhao-squareup:
this is finished now

src/invidious/helpers/utils.cr
```diff
@@ -101,6 +101,15 @@ def make_client(url : URI, region = nil)
   return client
 end
 
+def make_client(url : URI, region = nil)
```
saltycrys, line 104:
```suggestion
def make_client(url : URI, region = nil, &block)
```
I think it's a good idea to be explicit here.

saltycrys (changes requested):

saltycrys:
This is off-topic but sometimes [lsquic](https://github.com/iv-org/lsquic.cr) `get` requests don't finish, any idea why that might happen?

FireMasterK:
Why not just use 1 client for all requests? It doesn't make sense to create a client for every request/host...

unixfox:
I'm really interested into this PR, maybe this could help https://github.com/iv-org/invidious/issues/1132

jksladjflkjsadflkjsadf:
> This is off-topic but sometimes [lsquic](https://github.com/iv-org/lsquic.cr) `get` requests don't finish, any idea why that might happen?

I noticed that happening too, but I don't really have expeirence with crystal bindings or quic, so I have no idea right now.  I might take a look later.



> I'm really interested into this PR, maybe this could help solving #1132

Unfortunately this won't help with that, because the error is coming from the quic pool clients.


> Why not just use 1 client for all requests? It doesn't make sense to create a client for every request/host...

1 client is not enough, but this idea is what we are doing with the pooled request clients in YT_POOL.  But for the client to be reusable, it needs to be reused for the same host as the original, which is not applicable for the youtube cdn and some other uses.

saltycrys (approved):

src/invidious/helpers/utils.cr
saltycrys: This is resolved but I can't mark it as resolved because of the force-pushes.

jksladjflkjsadflkjsadf:
@TheFrenchGhosty are you able to merge this?

TheFrenchGhosty:
Thanks!

PR #1609, opened 2020-12-27
Add `popular-enabled` option

saltycrys:
This is similar to the removed `top-enabled` option but for the Popular feed.
The instance needs to be restarted if the feed was enabled.

Editing admin options on the preferences page is also fixed.

The handling of the feed pages now only happens in a single place.

Instead of redirecting:
  - The Top feed now displays a message that it was removed from Invidious.
  - The Popular feed now displays a message that it was disabled if it was.


saltycrys:
@TheFrenchGhosty while implementing this I discovered that changing admin preferences through the preferences page writes out `config.yml` with all preferences set. This could be used on a fresh instance to get a complete default `config.yml`.

To enable the admin preferences just add your username to the admins like so:
```yaml
admins:
  - username
```

Issue #1604, opened 2020-12-23
Video Mode button redirects to homepage


markozajc:
The 'Video Mode' button does not appear to work, but instead redirects to the homepage:
![image](https://user-images.githubusercontent.com/30626847/103042014-6eaaa000-4578-11eb-93c2-c7aa99b34ea1.png)
Curiously enough, the 'Audio Mode' button works just fine:
![image](https://user-images.githubusercontent.com/30626847/103042037-7e29e900-4578-11eb-9c11-300eef163105.png)
This issue seems to affect all videos. I would assume that it has something to do with https://github.com/iv-org/invidious/blob/eeeecf9763bc0ee0f34b87e67c7ce02d431a4b84/src/invidious/views/watch.ecr#L59, but it looks good to me.
Do note that this is not an issue with the listen=0 parameter itself, it still works fine.
Occurs on eeeecf9, latest master as of time of writing.

jksladjflkjsadflkjsadf:
Show/hide annotations also doesn't work for me.  Could be a related issue.

jksladjflkjsadflkjsadf:
Hi,

I investigated this issue and found the cause:
https://github.com/iv-org/invidious/blame/8df591e8d4fb46246e8da8f89d31659ae36da3c7/src/invidious/routes/watch.cr#L34

Here, we created a shallow copy of Http::params.  Then, we deleted the v= parameter from the shallow copy.  From the implementation of Http::params, we can see that it is a wrapper around the Hash object:
https://github.com/crystal-lang/crystal/blob/7f82f79bd1e3a7a1a73607e35c0662906736f1f8/src/http/params.cr#L99

When we created the shallow copy of Http::params, it used the same underlying instance of raw_params.  This means that when embed_params mutates the params, it also affects the env.params.query value.

To solve this, we need to properly copy the params before modifying them for embed_params.

#1551

saltycrys:
Sorry about that, fixed in https://github.com/iv-org/invidious/pull/1618.

markozajc:
Great, thanks!

PR #1618, opened 2020-12-29
Fix Video Mode Button

saltycrys:
The query params that get edited for `embed_params` are now deep copied instead
of shallow copied, preventing the originals from being changed.

Thanks to @jksladjflkjsadflkjsadf for investigating.

Closes https://github.com/iv-org/invidious/issues/1604


TheFrenchGhosty:
Thanks!

Issue #1569, opened 2020-12-11
fresh Docker install: proxy not working & feed/popular and feed/top empty


Slinky-Wrangle-Punch:
I just installed invidious using Docker. I use traefik as the reverse proxy for all my docker containers. Obeying the README I set `https_only: true` `domain: tube.mydomain.com` and `external_port: 443` The setup was easy and now I can watch videos on my own instance at tube.mydomain.com. But I do have two problems.

1. The feed/popular and feed/top are empty, but the the feed/trending does show videos. I cannot find any errors in the docker logs or the browser console regarding this.

2. If I want to proxy the videos or use dash (which uses proxy AFAIK) the videos do not play and the the banner `The media could not be loaded, either because the server or network failed or because the format is not supported.` occurs.

For this I found some errors in the browser console:

- `GET https://tube.mydomain.com/videoplayback?expire=160773[...]`  has `400 Bad Request` as status
- `VIDEOJS: ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported.`
- `Content Security Policy: The page’s settings blocked the loading of a resource at eval (“script-src”).`

Any ideas what's wrong? Did I miss any configurations hints from the README?

FireMasterK:
This seems to be affecting my instance (`invidious.kavin.rocks`) as well, the issue is that invidious adds an `?local=true` to the URL, thus breaking it.

The issue was introduced 11 days ago in https://github.com/iv-org/invidious/commit/0cb442d40ece54b40da1f8415d874c18a1f7f9b5.

The problematic line is here: https://github.com/iv-org/invidious/blob/28ca5b2b57fad510ed3f877bb6662b095e3c41c1/assets/js/player.js#L46

FireMasterK:
I'm don't think lines https://github.com/iv-org/invidious/blob/28ca5b2b57fad510ed3f877bb6662b095e3c41c1/assets/js/player.js#L44-L49
are necessary, I fail to understand why `?local=true` is being added to `/videoplayback` urls.

saltycrys:
The Top feed has been removed from Invidious. It relied on YouTube ratings and those are not public anymore.

The Popular feed is the latest video from the most subscribed to channels on the instance, so if there are no subscriptions the feed will be empty.

jksladjflkjsadflkjsadf:
[...]

PR #1619, opened 2020-12-29
do not add local to xhr when in videoplayback

tenpura-shrimp:
I previously added this logic to allow the frontend to automatically distignuish if native hls is possible and use proxy if not (we must always use proxy for xhr due to cors)  We do not need to have this for the videoplayback call it is only needed for the playlist call to get the proxied media urls.

I also thought this would only apply to hls because in the javascript it says videojs.Hls.xhr.beforeRequest but it seems to affect the dash as well.

On some systems, other people reported issues caused by this on the videoplayback endpoint so I make it only do it for the manifest.
#1569

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -42,7 +42,7 @@ var shareOptions = {
 }
 
 videojs.Hls.xhr.beforeRequest = function(options) {
-    if (options.uri.indexOf('local=true') === -1) {
+    if (options.uri.indexOf('videoplayback') === -1 && options.uri.indexOf('local=true') === -1) {
         options.uri = options.uri + '?local=true';
     }
     return options;


saltycrys (approved):
Code looks fine but this should be tested by @FireMasterK.

FireMasterK:
image

I made the following change using the devtools and it works. (since I can't test these changes)

I just realized something - the proxy setting has no effect on dash. `local=true` is always present in the dash manifest url, even with "Proxy Videos" disabled.
image

Perflyst:
That is expected behavior, proxy is required for dash.

FireMasterK:
That's because it's blocked by Invidious's `CSP`:
image
image

I think this is because the `CSP` doesn't allow for `videojs` to request the video with a `xhr` but allows it when it's classified as a `media` request.

FireMasterK:
As you can see here, it's being marked as a media request when not playing in dash:
image

PR #1622, opened 2020-12-30
Download liblsquic.a from iv-org/liblsquic-static

saltycrys:
This only affects Docker installs, regular builds still use the binary shipped
with `lsquic.cr`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,12 +4,7 @@ WORKDIR /invidious
 COPY ./shard.yml ./shard.yml
 COPY ./shard.lock ./shard.lock
 RUN shards install && \
-    # TODO: Document build instructions
-    # See https://github.com/omarroth/boringssl-alpine/blob/master/APKBUILD,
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
It doesn't seem to like something from this new file.

saltycrys:
Yeah, looks like it must be built on Alpine. This makes https://github.com/iv-org/lsquic-static useless, it's running on Ubuntu. Please delete it, I don't have the rights.

I'll try to make `lsquic-static-musl` instead.

TheFrenchGhosty:
@saltycrys so where a library is built affect it so much that it just breaks it? I've never seen that.

Perflyst:
alpine uses musl, ubuntu glibc. I guess that's why it breaks

saltycrys:
Yeah, I think that's the issue.

Both BoringSSL and LSQUIC itself are built with glibc on Alpine, too. So it must be an issue with some dependency.

Issue #179, opened 2018-09-23
More user-friendly search


omarroth:
There should be a form similar to YouTube so users can more easily use advanced search. Currently search filters have the advantage that they can be used anywhere, but they are much more opaque and less documented than they should be. An optional search form could also provide tooltips for text filters.

_Originally posted by @dimqua in https://github.com/omarroth/invidious/issues/126#issuecomment-417133791_

sebma:
@omarroth Do you plan to release that functionality soon ?

omarroth:
No work has been done on this currently.

SamantazFox:
Still relevant (related issues #1923 and #1969)

syeopite:
I'll be fixing this in the PR for #1950

SamantazFox:
Closed by #2956

PR #1629, opened 2021-01-03
Add ui for search sort and filter

tenpura-shrimp:
Invidious already supports search sort and filter using a specially formatted query, but there is no interface to expose this and the only way to know how to do it correctly is to read the source code

This change adds a ui for changing the search terms
Screen Shot 2021-01-02 at 4 39 51 PM


Perflyst:
Closes  #179

assets/js/search.js
```diff
@@ -0,0 +1,14 @@
...
+        target.innerHTML = '[ + ]';
+        body.style.display = 'none';
+    } else {
+        target.innerHTML = '[ - ]';
+        body.style.display = '';
+    }
+}
+
+document.getElementById('togglefilters').onclick = toggle_comments;
+document.getElementById('filters').style.display = 'none';
```
saltycrys, line 14: This should be done in CSS to avoid flashing during page load.

src/invidious/views/search.ecr
```diff
@@ -45,3 +134,4 @@
         <% end %>
     </div>
 </div>
+<script src="/js/search.js?v=<%= ASSET_COMMIT %>"></script>
```
saltycrys, line 137: This should be moved into `<head>`.

saltycrys (changes requested):
This is a great addition, I just have some nitpicks.

src/invidious/views/search.ecr
jksladjflkjsadflkjsadf: I'm not sure how to put this in the <head> because that's done in some common template for the head section.

I saw that watch.ecr is doing the same as here, but I don't see other places I can reference to where the js is included in the template.

assets/js/search.js
jksladjflkjsadflkjsadf: Hm I wanted to make it work with nojs also.  Should i do some stuff with noscript tag instead for this?

saltycrys: Yes, you could have a stylesheet in a `<noscript>` tag.

src/invidious/views/search.ecr
saltycrys: The part inside `<% content_for "header" do %>` gets put into `<head>`.

I didn't write `watch.ecr` but if I had I'd put the scripts into the header. I think having the definitions up top makes the code more readable. But since there's a precedent for putting it into the body it's up to you.

B0pol:
It's unusable on mobile:
image

assets/js/search.js
B0pol: then what's the reason to add this in JavaScript? If it works in CSS (and it can, https://teddit.net does it for comments), I see no reason to code twice the same thing, harder to maintain, harder to track…

jksladjflkjsadflkjsadf:
@saltycrys  this is ready to re-review

Updated the css for noscript and fixed the layout on mobile

Screen Shot 2021-01-10 at 12 12 23 PM

src/invidious/views/search.ecr
```diff
@@ -2,6 +2,102 @@
...
 
+<h3>
+    <a id="togglefilters" href="javascript:void(0)">[ + ]</a>
+    <%= translate(locale, "filter") %>
+</h3>
+
+<noscript>
+    <style>
+#filters {
+    display: flex !important;
```
saltycrys, line 13: Please indent this and get rid of the `!important`. I think if the `display: none` is added to `default.css` the styles here overwrite it.

saltycrys (changes requested):

tenpura-shrimp:
@saltycrys  This is updated for the requested change.

I don't know too much css so let me know of anything else.

saltycrys (approved):

PR #1631, opened 2021-01-03
Download liblsquic.a from iv-org/liblsquic-static-alpine

saltycrys:
This only affects Docker installs.
Regular builds still use the binary shipped with `lsquic.cr`.


TheFrenchGhosty:
It works, yay! Thanks!

PR #1636, opened 2021-01-04
Rework logging

saltycrys:
**Add CLI arguments to config file**

The log level can now be set with `log_level`.
The log file can now be set with `output` (also accepts `STDOUT`).

**Make logger a constant**

Instead of passing around `logger` there is now the global `LOGGER`.

**Adjust log verbosity**

The default log level has been changed from `debug` to `info`.
The `debug` log level is now more verbose. `debug` now gives a general overview
of what is happening (where implemented) while `trace` gives all available
details.

### DIFF
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -64,11 +64,13 @@ end
 class Config
   include YAML::Serializable
 
-  property channel_threads : Int32                 # Number of threads to use for crawling videos from channels (for updating subscriptions)
-  property feed_threads : Int32                    # Number of threads to use for updating feeds
+  property channel_threads : Int32 = 1             # Number of threads to use for crawling videos from channels (for updating subscriptions)
+  property feed_threads : Int32 = 1                # Number of threads to use for updating feeds
+  property output : String = "STDOUT"              # Log file path or STDOUT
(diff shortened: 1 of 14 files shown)


Perflyst:
Can you rebase on master? Afterwards I will test this PR

src/invidious/helpers/helpers.cr
```diff
@@ -64,10 +64,12 @@ end
 class Config
   include YAML::Serializable
 
-  property channel_threads : Int32                 # Number of threads to use for crawling videos from channels (for updating subscriptions)
-  property feed_threads : Int32                    # Number of threads to use for updating feeds
+  property channel_threads : Int32 = 1             # Number of threads to use for crawling videos from channels (for updating subscriptions)
+  property feed_threads : Int32 = 1                # Number of threads to use for updating feeds
+  property output : String = "STDOUT"              # Log file path or STDOUT
+  property log_level : LogLevel = LogLevel::Info   # Default log level, see src/invidious/helpers/logger.cr
```
Perflyst, line 70: An option like `log_level: info` would be better than `log_level: LogLevel::Info`

saltycrys:
Yes, I'll do it in a bit.

Also note that `LogLevel` is an enum so it serializes to int (when saving admin preferences) and can be deserialized from int (`log_level: 3` for `LogLevel::Info`) and strings (`log_level: info` or `log_level: Info`, not case sensitive).
I'll add a note in `helpers.cr`.

Perflyst (approved):

PR #1641, opened 2021-01-06
Add all available options to example config

Perflyst:


config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+# Define if accessed via HTTPS
 https_only: false
-domain:
+
+# Signing key for CSRF tokens
+# randomly generated by default
+#hmac_key: 
+
+# External domain
+domain: replace.me
```
TheFrenchGhosty, line 41: Should be empty for users running it locally.

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+
+# User login
+login_enabled: true
+
+# User registration
+registration_enabled: true
+
+# Instance statistics
+# Available on /api/v1/stats
+statistics_enabled: false
```
TheFrenchGhosty, line 62: More information should be added to explain what this shows, along with mentioning that it's useful to have for instances.invidio.us (and should be therefore enabled for public instances).

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+#external_port: 
+
+# Default preferences for
+# new and unregistered users
+default_user_preferences:
+  annotations: false
+  annotations_subscribed: false
+  autoplay: false
+  captions:
+  - ""
```
TheFrenchGhosty, line 79: ```dash_quality``` should be added.

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+  - ""
+  - ""
+  comments:
+  - youtube
+  - ""
```
TheFrenchGhosty, line 84: Reddit comments are enabled by default.

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+  continue: false
+  continue_autoplay: true
+  dark_mode: ""
```
TheFrenchGhosty, line 87: Set to false by default.

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+# HTTP host for internal webserver
+host_binding: 0.0.0.0
+
+# Pool for HTTP connections to YouTube
+pool_size: 100
+
+# Cookies for connection to YouTUbe
+#cookies: 
+
+# Anti-Captcha configuration for reCaptcha
```
TheFrenchGhosty, line 144: Should mention that both anti-captcha, and captcha-monsters are available as provider.

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+# Available on /api/v1/stats
+statistics_enabled: false
+
+# Instance admin account
+# account needs to exists
+admins: []
+
+# If running behind reverse proxy
+# an external port is required
+#external_port:
```
TheFrenchGhosty, line 70: Is 3000 by default, and should mention that this is critical.

TheFrenchGhosty (changes requested):
Why did you set some settings to their default value, but commented out some other settings, in my opinion commenting out the "optional" stuff, and keeping everything important uncommented would be better.

config/config.example.yml
```diff
@@ -1,11 +1,147 @@
...
+# randomly generated by default
+#hmac_key: 
+
+# External domain
+domain: replace.me
+
+# Use server-side notifications provided by YouTube
+# https://developers.google.com/youtube/v3/guides/push_notifications
+# Requires domain and hmac_key to be set
+use_pubsub_feeds: false
```
saltycrys, line 46: This does not require `hmac_key` to be set.

Might be good to point out that `domain: localhost` works, too.

config/config.example.yml
saltycrys: Might be good to point out that storyboards need this to be set and for local (non Docker) instances `domain: localhost` works.

config/config.example.yml
saltycrys: Also setting this to a number higher than zero sets the number of "threads" (fibers really) that are used for this. See https://github.com/iv-org/invidious/blob/168376b04671c6e362468267a3b07dd332b40702/src/invidious/jobs/subscribe_to_feeds_job.cr#L11-L13

There's a bug right now where setting this to `false` doesn't disable it, I'll fix that shortly.

config/config.example.yml
TheFrenchGhosty: Then I guess it should be set to localhost.

saltycrys: I don't think that works for Docker.

PR #1647, opened 2021-01-07
Remove omarroth mentions

### LINKED ISSUE #1621: Invidious still contain mention of omarroth in a lot of places
Invidious, obviously, will always be a project that exist because of Omarroth. 

However, Invidious should be updated to reflect his departure.

I noticed *some* places where either the old Github URL appears, his nickname, or his email address (example: https://github.com/iv-org/invidious/blob/master/src/invidious/users.cr#L430 )

This should be updated to mention the new Github URL, "invidious" instead of his nickname, the Github URL instead of his email address.

hackerncoder:
Fix #1621.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# Note: This is no longer updated and links to omarroths repo, which doesn't exist anymore.
+
 # 0.20.0 (2019-011-06)
 
 # Version 0.20.0: Custom Playlists
--- a/kubernetes/Chart.yaml
+++ b/kubernetes/Chart.yaml
@@ -9,9 +9,9 @@ keywords:
 - video
 - privacy
 home: https://invidio.us/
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# Note: This is no longer updated and links to omarroths repo, which doesn't exist anymore.
```
saltycrys, line 1: I though you meant removing email addresses in the changelog, updating repo links is fine.

saltycrys (approved):

CHANGELOG.md
hackerncoder: There are too many. Too many that link to the wiki which have to be rewritten to point to the documentation. I don't want to touch that right now.

TheFrenchGhosty:
Is that everything? Or just what you found?

TheFrenchGhosty (approved):

hackerncoder:
@TheFrenchGhosty What I found. But I did a grep recursive from the base of the repo for "omarroth" (and one just now that was also case insensitive) so it should be all.

TheFrenchGhosty:
Alright, thanks!

PR #1658, opened 2021-01-10
save host when using dash manifest

jksladjflkjsadflkjsadf:
In the case that the dash manifest is provided by youtube, we just chop the host off and then guess it later by using the mn and fvip query args

https://github.com/iv-org/invidious/blob/master/src/invidious.cr#L3486

I found that there are situations where that doesn't give the correct host, and we get a host not found error.  Instead, we should put the host in the query args like we do for the other proxy requests so the invidious instance hits exactly the host that was originally specified in the manifest.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -3190,7 +3190,8 @@ get "/api/manifest/dash/id/:id" do |env|
       url = url.rchop("</BaseURL>")
 
       if local
-        url = URI.parse(url).full_path
+        uri = URI.parse(url)
+        url = "#{uri.full_path}host/#{uri.host}/"
       end
 
       "<BaseURL>#{url}</BaseURL>"


saltycrys:
Please provide an example video.

jksladjflkjsadflkjsadf:
watch?v=0eaBxwuIbTU&quality=dash

<BaseURL>/videoplayback/expire/1610439920/ei/kAj9X-zXGYGx8gSqqZKgDQ/ip/[redacted]/id/d1e681c70b886d35/itag/251/source/youtube/requiressl/yes/mh/WE/mm/31,26/mn/sn-t0a7sn7d,sn-4g5ednsk/ms/au,onr/mv/m/mvi/1/pl/24/initcwndbps/721250/vprv/1/ratebypass/yes/mime/audio%2Fwebm/otfp/1/gir/yes/clen/931726/lmt/1610146703831255/dur/53.401/mt/1610418040/fvip/1/keepalive/yes/sparams/expire,ei,ip,id,itag,source,requiressl,vprv,ratebypass,mime,otfp,gir,clen,lmt,dur/sig/AOq0QJ8wRQIgGiLYbJmloiN327h1u-UC8sS-3mkf6ioVL2iUFwCgMJkCIQD9NSkI3aSgasWAx5vi0ulU1-Dk4MdMdILBi_RotGfcOQ%3D%3D/lsparams/mh,mm,mn,ms,mv,mvi,pl,initcwndbps/lsig/[redacted]/cmo/sensitive_content%3Dyes/</BaseURL>

jksladjflkjsadflkjsadf:
getting an error is intermittent but this video demonstrates the proxy manifest which does not include the original hostname provided by youtube.

saltycrys (approved):

PR #1674, opened 2021-01-16
Add container CI release

Perflyst:
We don't have access to the current "omarroth" namespace on docker and probably will never get access to this personal namespace. Since the instance page and the invidious-redirect container image are both pushed to quay already I suggest to push the official image to quay too.

The image will be available here https://quay.io/repository/invidious/invidious


Perflyst:
@TheFrenchGhosty please take a quick look

TheFrenchGhosty (changes requested):
the file should be named docker-release or quay-release, or something similar.

"release" imply Github release.

Otherwise looks good to me.

Perflyst:
Crap, Dockerfile is in a subfolder. I will fix it

PR #1676, opened 2021-01-17
Support "Database URL" in addition to existing DB-related parameters

Rjevski:
This PR adds support for using a `database_url` parameter (commonplace in other web frameworks, and recommended by the ["12-factor app" manifest](https://12factor.net/backing-services)) for configuring the DB instead of using separate parameters. This is to facilitate running Invidious in PaaS environments such as Heroku or its self-hosted alternatives ([Dokku](http://dokku.viewdocs.io/dokku/) in my case).

This parameter can be provided in the YAML config (and in the environment as `INVIDIOUS_DATABASE_URL` thanks to a separate PR that's already been merged). The existing `db` structure is also supported, so the change should be backwards-compatible. Note that if both are provided, then the `database_url` takes precedence over the `db` structure.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -49,7 +49,7 @@ PUBSUB_URL      = URI.parse("https://pubsubhubbub.appspot.com")
 REDDIT_URL      = URI.parse("https://www.reddit.com")
 TEXTCAPTCHA_URL = URI.parse("https://textcaptcha.com")
 YT_URL          = URI.parse("https://www.youtube.com")
-HOST_URL        = make_host_url(CONFIG, Kemal.config)
+HOST_URL        = make_host_url(Kemal.config)
 
 CHARS_SAFE         = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_"
 TEST_IDS           = {"AgbeGFYluEA", "BaW_jenozKc", "a9LDPn-MO4I", "ddFvjfvPnqk", "iqKdEhx-dD4"}
@@ -142,8 +142,6 @@ end
 OUTPUT = CONFIG.output.upcase == "STDOUT" ? STDOUT : File.open(CONFIG.output, mode: "a")
 LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
 
-config = CONFIG
-
 # Check table integrity
 if CONFIG.check_tables
   check_enum(PG_DB, "privacy", PlaylistPrivacy)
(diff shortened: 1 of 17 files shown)


src/invidious/helpers/helpers.cr
```diff
@@ -68,7 +68,7 @@ class Config
   property feed_threads : Int32 = 1                # Number of threads to use for updating feeds
   property output : String = "STDOUT"              # Log file path or STDOUT
   property log_level : LogLevel = LogLevel::Info   # Default log level, valid YAML values are ints and strings, see src/invidious/helpers/logger.cr
-  property db : DBConfig                           # Database configuration
+  property db_url : String = "postgres://kemal:kemal@localhost:5432/invidious" # Database configuration
```
Rjevski, line 71: There's probably a way to convert this to an URI directly here instead of calling `URI.parse` in a separate step?

saltycrys: You could subclass `URI` and make it YAML serializable.

- https://crystal-lang.org/api/0.35.1/Object.html#from_yaml(string_or_io:String%7CIO)-class-method
- Enum example:
  - https://github.com/crystal-lang/crystal/blob/5999ae29b/src/yaml/from_yaml.cr#L222
  - https://github.com/crystal-lang/crystal/blob/5999ae29b/src/yaml/to_yaml.cr#L109

Could also be added to `URI` itself.

kubernetes/values.yaml
```diff
@@ -45,12 +45,7 @@ postgresql:
 config:
   channel_threads: 1
   feed_threads: 1
-  db:
-    user: kemal
-    password: kemal
-    host: invidious-postgresql
-    port: 5432
-    dbname: invidious
+  db_uri: "postgres://kemal:kemal@invidious-postgresql:5432/invidious"
```
saltycrys, line 48: This should be `db_url`.

saltycrys (changes requested):
Please also update `docker-compose.yml`.

src/invidious/helpers/helpers.cr
saltycrys: Better example:
- https://github.com/iv-org/invidious/blob/ba148d749f3f3b745704c35bde2f72ea194b1ddb/src/invidious/helpers/helpers.cr#L93
- https://github.com/iv-org/invidious/blob/ba148d749f3f3b745704c35bde2f72ea194b1ddb/src/invidious/users.cr#L146

Perflyst:
This is a breaking change, since we don't have any kind of releases this will much likely break some instances during (auto) update

NiekertDev:
Why not leave the config as is, and make the code pick the environment variable DATABASE_URL  if set, otherwise fallback to config? When deploying Invidious in a PaaS environment, you don't even want to change your database url everytime you deploy a new instance. The use of environment variables is also advised by the 12 Factor app manifest.

src/invidious.cr
```diff
@@ -33,7 +33,7 @@ require "./invidious/jobs/**"
 CONFIG   = Config.load
 HMAC_KEY = CONFIG.hmac_key || Random::Secure.hex(32)
 
-PG_URL = URI.new(
+PG_URL = CONFIG.database_url || URI.new(
```
Rjevski, line 36: This doesn't actually work, and I've spent an hour on this trying various options such as moving the determination of the DB URI to a function with if/case statements and using bare if/case statements here.

The task at hand is simple, "if CONFIG.database_url is not nil then use that, otherwise if CONFIG.db is not nil then use that, otherwise raise" but I can't seem to get it to work. I'd appreciate any help here.

src/invidious/helpers/helpers.cr
```diff
@@ -64,11 +64,14 @@ end
...
-  property log_level : LogLevel = LogLevel::Info   # Default log level, valid YAML values are ints and strings, see src/invidious/helpers/logger.cr
-  property db : DBConfig                           # Database configuration
+  property channel_threads : Int32 = 1           # Number of threads to use for crawling videos from channels (for updating subscriptions)
+  property feed_threads : Int32 = 1              # Number of threads to use for updating feeds
+  property output : String = "STDOUT"            # Log file path or STDOUT
+  property log_level : LogLevel = LogLevel::Info # Default log level, valid YAML values are ints and strings, see src/invidious/helpers/logger.cr
+  property db : DBConfig? = nil                  # Database configuration with separate parameters (username, hostname, etc)
+
+  @[YAML::Field(converter: Preferences::URIConverter)]
+  property database_url : URI? = nil               # Database configuration using 12-Factor "Database URL" syntax
```
Rjevski, line 74: Made these nullable since we now allow either one of them to be provided (with the `database_url` taking precedence).

src/invidious/helpers/helpers.cr
```diff
@@ -170,6 +173,8 @@ class Config
         end
     {% end %}
 
+    raise "Either database_url or db.* is required" unless config.database_url || config.db
+
```
Rjevski, line 177: Guard against neither of those two being provided.

Rjevski:
I have updated the PR. Now the parameter is optional, and the `db` structure is still supported; thus this is no longer a breaking change. No updates to the Docker-related files are necessary.

@saltycrys I'm having trouble with a stupidly easy task actually determining which one of the `Config.database_url` or `Config.db` to use (since they are now both nullable). Would you be able to have a look when you have time and see if you can finish it off? I've implemented your suggestion of using a custom converter for (un)serializing the URI from the YAML directly.

saltycrys:
@Rjevski I made things work and added some comments, please clean everything up and overwrite my commit.

Rjevski:
Brilliant - thanks! I will test this and then submit a final commit.

saltycrys:
My pleasure. Please squash everything into one commit. Other than that I'm happy with this.

Rjevski:
One minute - I think something went horribly wrong during the rebase (the joys of managing multiple branches and git remotes)

Rjevski:
Ok - this still isn't working out. I will remake the changes manually in a separate PR from a whole separate repo and branch to avoid any issues. Git appears to be determined to ruin my day for some reason.

Issue #692, opened 2019-08-13
Dash not working in Safari


Duckbilled:
Dash on safari (macos) starts for a few seconds till it starts loading indefinitely.

omarroth:
Would you mind checking devtools to see if there are any errors being displayed? Can you reproduce this in another browser, or is it only an issue in Safari?

Uffje:
I’ve checked, it gives me the following error:

 Failed to load resources: the server responded with a status of 404 (not found)

I am not sure if I am looking at the correct log though. 

Unfortunately I have a problem with Firefox and can’t test it in an other browser, but I believe it used to work before.

Uffje:
Just realised FF works again and Invidious works as should in Firefox on Mac using Dash 1080p.

pawi:
Tested Chromium and the issue is only with Safari on macOS.

While a 30 second video play complete with Safari anything longer (at least > 3 min tested) stops playing after second 9 or 10. No specific errors found with Safari dev tools. It looks like it just doesn't load any further videoplayback chunks.

I use dash enabled and proxy video on my personal instance.

michaelkeuter:
I see exactly the same behaviour in the latest Safari version 13.0.4 on macOS 10.13.6.

michaelkeuter:
Update: the same happens on the iPad (v13.2.3) and it also happens with videos that have only audio:

https://invidio.us/watch?v=1D6Y3zdOSRI

imtbl:
I can confirm a similar behavior for both the latest iPadOS as well as the latest iOS using my self-hosted Invidious instance as well as the official one.

DASH works perfectly fine on Arch Linux( Brave), macOS (Brave) and Android (Brave and Kiwi), but on both my iPad and iPhone every video served with DASH stops after 9 seconds and seems like it’s buffering indefinitely (but never continues), regardless of the set video quality. I’ve also noticed that the audio only mode doesn’t load the video at all on iPad/iOS, it looks like it’s buffering straight away and never starts playing.

When not using DASH, all the videos play just fine on iPad/iOS, in any quality.

lucky13820:
I always thought it was my server is not powerful enough. Damn, really wish we can fix this problem.

Uffje:
I started using Firefox shortly after posting this and never really (just some design/feature choices) looked back. It still has some loading issues with DASH once in a while, in general a lot of things just work better.

lucky13820:
[...]

PR #1679, opened 2021-01-17
Fix DASH playback bug in Safari

Rjevski:
Fixes https://github.com/iv-org/invidious/issues/692 - works around upstream bug in Video.js.

This has been running successfully on my instance for over 2 weeks of daily usage - no issues to report.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -19,6 +19,11 @@ var options = {
             'playbackRateMenuButton',
             'fullscreenToggle'
         ]
+    },
+    html5: {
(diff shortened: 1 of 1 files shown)


saltycrys (approved):
Looks good to me but I haven't done any testing.

Perflyst:
Some testing with this option was done in https://github.com/videojs/video.js/issues/6762#issuecomment-662561367. I am going to merge this but we should track https://github.com/videojs/video.js/issues/6762 further and watch what they will do.

PR #1689, opened 2021-01-21
Build latest only on master, add commit sha tag

### LINKED ISSUE #1688: Tag container images with commit sha
We should tag additional container images to allow rollback. The latest tag always provides the latest commit but if something is broken you can't rollback.

Additionally we could build container images for each PR to allow easier testing with container deployments.

Perflyst:
Close #1688

### DIFF
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -45,6 +45,3 @@ jobs:
           file: docker/Dockerfile
           push: true
           tags: quay.io/invidious/invidious:${{ github.sha }}
-
-      - name: Image digest
-        run: echo ${{ steps.docker_build.outputs.digest }}
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - "master"
+  pull_request:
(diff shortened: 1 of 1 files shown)


Perflyst:
This patch builds the container image on master twice even if re-tagging would be sufficient. I didn't found an option to re-tag a container image with action template we are using.

Perflyst:
@FireMasterK what do you think?

FireMasterK:
I was thinking about adding a [set-output](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter) to determine the tags.

When it is a PR, we could have the output as
quay.io/invidious/invidious:sha
if it is a push trigger we could have the output as
quay.io/invidious/invidious:sha,quay.io/invidious/invidious:latest

although, I'm not sure how the expiry tag would work with my idea.

However, I have a different and easier idea that doesn't require building the image twice. The downside is that even the latest tag would have an expiry.

I've created a PR for my approach.

Perflyst:
> The downside is that even the latest tag would have an expiry.

Should be fine. If there is no new commit after 12 weeks. On the other hand we could rebuild on master every night to get security updates for packages in the base image.

PR #1691, opened 2021-01-22
Container sha without 2 builds

FireMasterK:
Thanks to Perflyst for doing most of the work.

### DIFF
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -45,6 +45,3 @@ jobs:
           file: docker/Dockerfile
           push: true
           tags: quay.io/invidious/invidious:${{ github.sha }}
-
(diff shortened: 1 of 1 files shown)


FireMasterK:
It's failing only because Secrets are not exposed when the action files are modified.

Perflyst:
From https://github.com/iv-org/invidious/pull/1689#issuecomment-765249359


> The downside is that even the latest tag would have an expiry.

Should be fine. If there is no new commit after 12 weeks I guess invidious is dead. On the other hand we could rebuild on master every night to get security updates for packages in the base image which also prevents expiration.

FireMasterK:
I've added a cron trigger as well to rebuild every night 👍

Perflyst (approved):

Issue #1677, opened 2021-01-17
Use environment variables for configuration


Rjevski:
I would prefer if Invidious used environment variables for config natively, to make it easier for Docker-based deploys as well as for PaaS platforms such as Heroku/Dokku. At the moment I am using the following workaround as a start script:

```bash
#!/usr/bin/env sh

export INVIDIOUS_CONFIG="
channel_threads: ${CHANNEL_THREADS:-1}
feed_threads: ${FEED_THREADS:-1}
db_url: $DATABASE_URL
full_refresh: ${FULL_REFRESH:-false}
https_only: ${HTTPS_ONLY:-false}
domain: $DOMAIN
hmac_key: $HMAC_KEY
use_pubsub_feeds: ${USE_PUBSUB_FEEDS:-false}
login_enabled: ${LOGIN_ENABLED:-false}
registration_enabled: ${REGISTRATION_ENABLED:-false}
statistics_enabled: ${STATISTICS_ENABLED:-false}
external_port: ${EXTERNAL_PORT:-443}
default_user_preferences:
  autoplay: ${DEFAULT_USER_PREFS_AUTOPLAY:-false}
  continue: ${DEFAULT_USER_PREFS_CONTINUE:-false}
  player_style: ${DEFAULT_USER_PREFS_PLAYER_STYLE:-invidious}
  local: ${DEFAULT_USER_PREFS_LOCAL:-true}
  default_home: ${DEFAULT_USER_PREFS_DEFAULT_HOME:-Trending}
  feed_menu: ${DEFAULT_USER_PREFS_FEED_MENU:-[\"Trending\"]}
  quality: ${DEFAULT_USER_PREFS_QUALITY:-dash}
"

exec ./invidious
```

I think it would be better if Invidious read those environment variables natively, and if there is still a need for a `config.yml` then a helper (either in Crystal or in shell) script could be provided that would do the reverse of the above - read the YAML file, populate the environment variables and then call the Invidious binary.

Related: https://github.com/iv-org/invidious/pull/1676

TheFrenchGhosty:
From what I know it's already a thing, see: https://github.com/iv-org/invidious/blob/master/docker-compose.yml#L25

Rjevski:
Yeah that's what I'm using now. I don't think it's good enough though - it requires you to format a YAML file at runtime. Ideally each setting would be its own environment variable.

unixfox:
Very good idea. Config files are difficult to deal with in the container ecosystem.

PR #1702, opened 2021-01-23
Config improvements

saltycrys:
**Add config environment variables**

The config file can now be specified with `INVIDIOUS_CONFIG_FILE`.
A YAML formatted string can still be passed with `INVIDIOUS_CONFIG`, replacing
the config file.

Additionally all options can now be specified as environment variables.
The syntax for variable names is `INVIDIOUS_` followed by the option name in
upper case. The values are parsed as YAML.

These new env vars only update the provided main configuration, but it is
possible to point the config file at the example config and then use env vars
for all config options:
```
INVIDIOUS_CONFIG_FILE=./config/config.example.yml \
INVIDIOUS_CHANNEL_THREADS=10 \
./invidious
```

**Make config a constant**

Instead of passing around `config` there is now the global `CONFIG`.

**Only start refresh jobs when necessary**

If `channel_threads` or `feed_threads` is set to zero the corresponding job is
now not started.

<hr>

Closes https://github.com/iv-org/invidious/issues/1677

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -49,7 +49,7 @@ PUBSUB_URL      = URI.parse("https://pubsubhubbub.appspot.com")
 REDDIT_URL      = URI.parse("https://www.reddit.com")
 TEXTCAPTCHA_URL = URI.parse("https://textcaptcha.com")
 YT_URL          = URI.parse("https://www.youtube.com")
-HOST_URL        = make_host_url(CONFIG, Kemal.config)
+HOST_URL        = make_host_url(Kemal.config)
 
 CHARS_SAFE         = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_"
 TEST_IDS           = {"AgbeGFYluEA", "BaW_jenozKc", "a9LDPn-MO4I", "ddFvjfvPnqk", "iqKdEhx-dD4"}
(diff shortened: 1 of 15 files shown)


Perflyst:
OT @FireMasterK looks like building an image for PR does not work because the secret is only available from with-in the repo :/

FireMasterK:
Look like they're no way to allow forks to access them too. (Safely)

Perflyst:
I get two errors:
```
[error] UpdateDecryptFunctionJob : Nil assertion failed
[error] BypassCaptchaJob: Missing hash key: "taskId"
```

saltycrys:
Please try with f1a7ee9 to see if the config constant is the problem.

Perflyst:
Just reverted and tested with https://github.com/iv-org/invidious/commit/f1a7ee997b5f6bc52388473dcd1b8043c5ceacff happens there too. Then I went back to master and it happens on master too so the errors are not related to this patch at all

saltycrys:
Seems like you're getting empty responses when doing web requests. Can you try with a machine that has a different internet connection?

Perflyst:
Works fine locally

Perflyst (approved):

Rjevski:
> If channel_threads or feed_threads is set to zero the corresponding job is
now not started.

Does this mean workers can now be offloaded to a separate process? Ie. run the main web process with channel_threads and feed_threads to 0, and run a process in the background with those values set to non-zero?

saltycrys:
> Does this mean workers can now be offloaded to a separate process? Ie. run the main web process with channel_threads and feed_threads to 0, and run a process in the background with those values set to non-zero?

Yes, but technically that was already possible.
What happened was that the jobs started and immediately stalled. The `RefreshChannelsJob` has some logging at startup that would always get logged, now it only happens with a non-zero config value.

PR #1715, opened 2021-01-27
Use large thumbnail for channels in search

### LINKED ISSUE #1705: [Feature request] We should use big channel thumbnails in the search
Youtube channel thumbnails have two sizes, big and small.  We should use the big size in the search results, because otherwise they look very pixellated.

tenpura-shrimp:
This uses a larger thumbnail resolution in the search results for channels, so it is not so pixellated.  This is a dynamic endpoint on yt side, so any resolution can be specified, but I use 176 because that is what yt uses.

Fixes #1705


Perflyst:
Would be interesting to know the size different of both images

Perflyst:
looks good to me, can you please rebase on master?

tenpura-shrimp:
rebased

PR #1718, opened 2021-01-29
Fix broken build - install crystal 35.1 in ci

tenpura-shrimp:
The crystal 36.0 release introduced several breaking changes which fail the build, including #1719.

The long term solution should be to update the code to be compatible with the crystal changes, but for the short term this resolves the failing build.


TheFrenchGhosty:
Thank you!

Issue #1675, opened 2021-01-17
Running Invidious Without Postgres


Rayqan:
Would you consider adding the ability to run Invidious without postgres? This would be extremely useful for small single user instances hosted on resource constrained machines.

TheFrenchGhosty:
This is definitely something I've been thinking about, for instances without accounts enabled this should be easily do-able. The caching should be done on plain files/ram.

DUOLabs333:
@TheFrenchGhosty I wasn't simply going to remove the word "Postgres", I was going to remove any code that depends on a database running. But yeah, I thought that if you disabled accounts and registration, it wouldn't make any calls to Postgres (or not nearly enough), so simply any lines that need Postgres can be removed. However, looking though the code  showed me that it would be a bit more complicated than that, as some of the functionality is tightly integrated with Postgres (though if you only use Youtube accounts, not local ones, Redis may be a good idea...)

TheFrenchGhosty:
@DUOLabs333 Ah makes sense.

Yeah basically all caching is done through Postgres, so a whole new handling is required to "get rid" of the Postgres requirement.

DUOLabs333:
Redis could work, no? It would also have the benefit of making it easier to set up?

syeopite:
I managed to patch Invidious to work without the database logic. Its quite hastily put together but if anyone's needing this feature it can be found over at my branch here:

https://github.com/syeopite/invidious/tree/no-database

It is done through a compile time flag so you'd need to build Invidious with the `-Dno_postgresql` parameter

`crystal build src/invidious.cr -Dno_postgresql`

UltraBlackLinux:
Would like to support this as well as the database server keeps corrupting for no apparent reason...

PR #1729, opened 2021-01-30
Support "Database URL" in addition to existing DB-related parameters

Rjevski:
This PR adds support for using a database_url parameter (commonplace in other web frameworks, and recommended by the "12-factor app" manifest) for configuring the DB instead of using separate parameters. This is to facilitate running Invidious in PaaS environments such as Heroku or its self-hosted alternatives (Dokku in my case).

This parameter can be provided in the YAML config (and in the environment as INVIDIOUS_DATABASE_URL thanks to a separate PR that's already been merged). The existing db structure is also supported, so the change should be backwards-compatible. Note that if both are provided, then the database_url takes precedence over the db structure.

Continued from [this PR](https://github.com/iv-org/invidious/pull/1676), which was trashed due to a screwup on my part and the easiest fix was just to start on a fresh branch.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -6,6 +6,8 @@ db:
   host: localhost
   port: 5432
   dbname: invidious
+# alternatively, the database URL can be provided directly - if both are set then the latter takes precedence
+# database_url: postgres://kemal:kemal@localhost:5432/invidious
 full_refresh: false
 https_only: false
 domain:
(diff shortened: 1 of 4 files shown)


saltycrys (approved):

tenpura-shrimp:
This might also solve #1675 but I'm not sure how postgres-specific the queries are already.

unixfox:
> This might also solve #1675 but I'm not sure how postgres-specific the queries are already.

No it doesn't, Invidious is using a postgres library called `pg`.

Perflyst (approved):

PR #1732, opened 2021-01-31
Unrequire `logger`

### LINKED ISSUE #1719: Migrate deprecated logger - broken build
The build is broken because the logger has been deprecated.  See https://github.com/crystal-lang/crystal/pull/9525

We should update the invidious logger wrapper to use the new logging functions.

saltycrys:
Crystal's `Logger` was required but never used in Invidious. Crystal 0.36.0
removed `Logger` in favor of `Log`.

Closes #1719.

### DIFF
--- a/src/invidious/helpers/logger.cr
+++ b/src/invidious/helpers/logger.cr
@@ -1,5 +1,3 @@
-require "logger"
-
 enum LogLevel
   All   = 0
   Trace = 1


FireMasterK:
Shouldn't #1718 be reverted then?

Perflyst:
I just tested but something is still broken with latest crystal, https://github.com/iv-org/invidious/runs/1800875655

```

In src/invidious/helpers/proxy.cr:76:7

 76 | @socket = nil
      ^
Error: instance variable @socket of HTTPClient was inferred to be Nil, but Nil alone provides no information
Error: Process completed with exit code 1.

saltycrys:
I think pinning the Crystal version is a good thing. It's already pinned in the Dockerfile and dependencies might require a specific version.

I'll look into upgrading Invidious for Crystal 0.36.0 though.

tenpura-shrimp:
Looks like this change in the release notes

https://crystal-lang.org/2021/01/26/crystal-0.36.0-released.html

> A second breaking-change that affects C bindings is that a Crystal nil value is no longer translated to a null pointer value. Read more at #9872.

saltycrys:
It's https://github.com/crystal-lang/crystal/pull/9543, see https://github.com/iv-org/invidious/pull/1734.

PR #1735, opened 2021-02-01
iOS audio mode double duration fix

138138138:
Fixes audio mode duration doubled in iPhone iOS browsers. The player will stop after reaching the real duration. 
iOS() checks both iOS and iPadOS. 
Only tested on iPhone iOS browsers. Testers needed for behavior of iPadOS and MacOS.


tenpura-shrimp:
Wow this is very strange.  I tested the bug on mac safari.

The actual bug isn't a videojs bug but actually a safari/youtube bug.  If you just browse over to the non-proxy url of the audio file, it'll show the wrong duration too.  Additionally, you see a whole bunch of incorrectly ranged partial-contents.

Apple engineering must be complete garbage these days.

saltycrys (approved):
LGTM but I haven't done any testing.

assets/js/player.js
```diff
@@ -547,3 +547,14 @@ window.addEventListener('keydown', e => {
 if (player.share) {
     player.share(shareOptions);
 }
+
+// Safari audio double duration fix
+player.on('loadedmetadata', function () {
+    if (navigator.vendor == "Apple Computer, Inc." && video_data.params.listen) {
```
tenpura-shrimp, line 553: Can you put the navigator.vendor check on the most outside, so that it won't add the loadedmetadata listener unless it's apple?

Perflyst:
I can apply this patch but cannot test it myself.

PR #1797, opened 2021-02-21
Use UTC for published date

### LINKED ISSUE #1796: [Bug] published date depends on instance location
**Describe the bug**

the `published` field in API depends on instance location

**Steps to Reproduce**

1. Use two different instances without the same timezone
2. Call any video and look for `published` field
3. Notice they are different

example:
- ytprivate.com (US) https://ytprivate.com/api/v1/videos/7PIMiDcwNvc -> published = 1566561600
- invidious.fdn.fr (FR) https://invidious.fdn.fr/api/v1/videos/7PIMiDcwNvc -> published = 1566597600

**Logs**

**Screenshots**

**Additional context**

Published date is gathered here
https://github.com/iv-org/invidious/blob/cbc7603248341839157d87a2addd0141281880f8/src/invidious/videos.cr#L519-L521

It uses Time::Location.local. 
It should use UTC: Time::Location::UTC
https://crystal-lang.org/api/0.36.1/Time/Location/Zone.html#constant-summary

B0pol:
Fixes #1796

Could someone test that if not based in CET timezone?
http://localhost:3000/api/v1/videos/MQNMWCJ1Tjg?fields=published should be 1593129600

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -517,7 +517,7 @@ struct Video
   end
 
   def published : Time
-    info["microformat"]?.try &.["playerMicroformatRenderer"]?.try &.["publishDate"]?.try { |t| Time.parse(t.as_s, "%Y-%m-%d", Time::Location.local) } || Time.local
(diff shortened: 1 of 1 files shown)


FireMasterK:
I can confirm that it shows `{"published":1593129600}`, which is correct.

saltycrys (approved):

TheFrenchGhosty:
Thanks @B0pol 

@FireMasterK thanks for testing this PR.

PR #1808, opened 2021-02-25
show how long ago stream started

tenpura-shrimp:
## Overview
When I start watching a stream already in progress, I want to know how much I missed.  This adds the time display like youtube

Screen Shot 2021-02-24 at 8 09 00 PM

## Change
Add a display of the time elapsed since the livestream premier if the livestream is still ongoing.

The previous live_now in video was using the wrong attribute and displays true for all livestreams.  This changes it to use the appropriate live_now which indicates if the livestream is currently ongoing.

## Testing
Tested appropriate behavior:
* ongoing livestream works as expected
* completed livestream does not display any change
* non-livestreams do not display any change

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -534,7 +534,8 @@ struct Video
   end
 
   def live_now
-    info["videoDetails"]["isLiveContent"]?.try &.as_bool || false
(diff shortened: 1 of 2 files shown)


saltycrys (approved):

TheFrenchGhosty:
Thank you @tenpura-shrimp

PR #1817, opened 2021-02-26
fix search

### LINKED ISSUE #1480: Search page result are the same

tenpura-shrimp:
Put search page in the super secret proto field

Fixes #1480

### DIFF
--- a/spec/helpers_spec.cr
+++ b/spec/helpers_spec.cr
@@ -65,15 +65,15 @@ describe "Helper" do
 
   describe "#produce_search_params" do
     it "correctly produces token for searching with specified filters" do
-      produce_search_params.should eq("CAASAhAB")
+      produce_search_params.should eq("CAASAhABSAA%3D")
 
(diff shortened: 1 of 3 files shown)


saltycrys (approved):

syeopite:
Works perfectly fine though it seems like sometimes a result from the first page makes it onto the next. I'm not sure if that's related to this though.

 

In the case of these two example its the "7 Riddles That Will Test Your Brain Power" video.

tenpura-shrimp:
Search results aren't consistent across queries.  You can try refreshing the page or searching a different instance and you will get different results.

To get no duplicates, we would probably need to use the continuation key functionality, which is much more complex.

TheFrenchGhosty:
Alright! Thanks @tenpura-shrimp 

Thanks for testing @syeopite

PR #1829, opened 2021-02-27
integrate legacy yt xml subscription import with new json sub import

SimonPhoenix96:
after @schwukas changed the youtube sub importer to support youtubes new way of exporting subs as json in this [commit]( https://github.com/iv-org/invidious/commit/62e8c091831ce2d2db7e928dd25cb55330d72323) , it wasn't possible to import old youtube sub exports (which were exported as  xml files).

i simply added an if conditional which checks the to be imported file format. if the file starts with "<opml" it'll use the legacy import if not then the new way of importing json files.

i did this because i deleted my google acc after my degooglefication process, as i only had my subs backed up as xml files i wasn't able to import these after the change. maybe there are many more out there such as myself that will benefit from this pr :)

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -760,7 +760,7 @@ post "/data_control" do |env|
           end
         end
       when "import_youtube"
-        if body[0..4] == "<opml" 
+        if body[0..4] == "<opml"
           subscriptions = XML.parse(body)
           user.subscriptions += subscriptions.xpath_nodes(%q(//outline[@type="rss"])).map do |channel|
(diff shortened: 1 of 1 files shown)


SimonPhoenix96:
no idea why the build check aren't going through, but it works on my machine :^)

Perflyst:
The linter failed. Run `crystal tool format` to format your code changes locally and push the changes.

SimonPhoenix96:
@Perflyst ah looks like that worked, thanks!

saltycrys (approved):

PR #1868, opened 2021-03-05
Publish docker images for Pull Requests

LennyPenny:
This should make the testing of PRs easier for people that run docker setups!

### DIFF
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - "master"
+  pull_request:
+    branches: "master"
   schedule:
     - cron: 0 0 * * *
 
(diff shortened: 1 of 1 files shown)


LennyPenny:
(the step fails because I obviously don't have access to the invidious quay.io repo)

Perflyst:
> (the step fails because I obviously don't have access to the invidious quay.io repo)

It will fail for anyone who does not have access to this repository.
We tried this in the past, https://github.com/iv-org/invidious/commit/e35345f1358d98fa36b8d3b8b5b420260a8aef10, this does not work. Did you get this working somewhere else with some custom settings @LennyPenny?

LennyPenny:
ah yea nvm the logic is flawed

it would be a very bad idea to let random people upload to the invidious quay account

LennyPenny:
the only way this could work is for people that can create branched on this repo to create pull requests from

PR #1899, opened 2021-03-17
Fix: functional hl / dark_mode / thin_mode params

Svallinn:
The `hl`, `dark_mode` and `thin_mode` URL query parameters should be overriding the preferences' options (or, in the case of direct use of the API, the `hl` parameter should be localizing some strings) but this isn't currently happening.

This PR fixes that behavior.


TheFrenchGhosty:
What was broken, what does this fix? Any associated issue?

saltycrys (approved):

PR #1912, opened 2021-03-21
Add missing filter ui locales

syeopite:
Seems like the "short" and "long" filter options are missing translations. I don't know enough about how the translations work to know if this change is necessary but I've added them into the English file.


TheFrenchGhosty:
Locales aren't just handled in the translation file. The actual code needs to be made to use translation strings.

syeopite:
I probably should've defended my point better in this PR back then. Basically, Invidious's code already calls the specific translation strings but since none are present in the locales, they don't get translated. The "short" and "long" strings you see are just the IDs used to fetch the translations which are used as a fallback when none is found.

SamantazFox:
Why did you merge master in this PR?

syeopite:
That was GitHub's web editor. I didn't see the point of editing it locally for such a minute change.

Issue #1916, opened 2021-03-22
[Enhancement] Don't translate locales name


SamantazFox:
**Is your enhancement request related to a problem? Please describe.**
Yes and no. I've noticed that currently, the different languages are translated in `locales/*.json`.
Imho, this seems useless and a waste of translation effort, as the principal use of these is to select UI languages in the preferences.

**Describe the solution you'd like**
As on Wikipedia, language names are not translated.

E.g: A russian speaking person will perfectly be able to read "russian" written in cyrillic.

TheFrenchGhosty:
Related: https://github.com/iv-org/invidious/pull/851

SamantazFox:
Closing, as this was fixed in #2576

Issue #571, opened 2019-06-01
ISO codes for languages


comradekingu:
Translation is duplicated among many projects, not all of which will be completed at the same rate.
Instead there is
https://hosted.weblate.org/projects/iso-codes/iso-639-3/
and
https://hosted.weblate.org/projects/iso-codes/iso-639-2/
fortunately on the same instance, which solves this problem.
Can they be used for Invidious?

omarroth:
Looks like they're provided as [PO files](http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html) which aren't compatible with what Invidious uses currently.

There's currently [some work](https://github.com/omarroth/gettext.cr) being done elsewhere that should allow them to be supported. Currently the main challenge is converting all existing locales to PO files, but definitely should be possible in future to use this to reduce work for translators.

SamantazFox:
@TheFrenchGhosty can you please remove the `stale` label? Thanks :)

PR #1920, opened 2021-03-23
Test & fix locales consistency

### LINKED ISSUE #1913: [Enhancement] Fix missing translation keys in other languages

There's currently a massive discrepancy in the amount of strings in the English translations when compared to every other language. Every new feature just expands this gap further. Afaik a translator on Weblate can't directly add any strings. A key has to be available first. 

Add all of the new English keys to the other languages. This way translations can begin on the recently added features like the new filtering ui added by #1629. Though this should probably be automated.

SamantazFox:
Add a test (spec) that checks if the key defined in the reference locale ("en-US") also exist in the other locales.
The goal is to fix #1913.

Current status:
- [x] ar (Arabic)
- [x] da (Danish)
- [x] de (German)
- [x] en-US (english US) **reference**
- [x] el (Greek)
- [x] eo (Esperanto)
- [x] es (Spanish)
- [x] fa (Persian)
- [x] fi (Finnish)
- [x] fr (French)
- [x] he (Hebrew)
- [x] hr (Hungarian)
- [x] id (Indonesian)
- [x] is (Icelandic)
- [x] it (Italian)
- [x] ja (Japanese)
- [x] nb-NO (Norwegian Bokmål)
- [x] nl (Dutch)
- [x] pl (Polish)
- [x] pt-BR (Portuguese, Brazil)
- [x] pt-PT (Portuguese, Portugal)
- [x] ro (Romanian)
- [x] ru (Russian)
- [x] sv (Swedish)
- [x] tr (Turkish)
- [x] uk (Ukrainian)
- [x] zh-CN (Simplified Chinese, China)
- [x] zh-TW (Traditional Chinese, Taiwan)


SamantazFox:
I may also fix #1916 and #851 in the meantime.

I see that there was also #571 (TL;DR use PO file format instead of json), but I think it's too much work to be included in this PR ^^

SamantazFox:
## Notes

- The PR #1629 added new translation strings for the search fields, but the key names ("year", "hour", "video", ...) seem to not correspond to the original string ("This year", "Last hour", ...) and it can be confusing for translators.

- This also raises the question of how translations need to be handled. Should we keep the current format (where the key is the original english string) or should we switch to short identifiers (like "last_video", "error_invalid_playlist") to ensure that locales are easy to maintain.

- Similarly to previous point, the question of singular/plural form need to be asked. The current system is bad, as we don't want a regex in the translation files. We have also to take into account that in some languages, the form for **zero**, **one**, **two**, and **more** of something can be different. [The Qt docs have a good article on the subject](https://doc.qt.io/archives/qq/qq19-plurals.html)

---

## Nota Bene

I didn't fix any of the issues above in this PR. Those are open questions, and this PR's goal is only to synchronize changes across files, so translators can fill the holes.

SamantazFox:
Done :D

@TheFrenchGhosty can you please mark this as "needs-review" and remove the "unfinished" tag?
I guess I also need to ping @saltycrys for code review, too.

SamantazFox:
Here we go. I should check my status 3 times before pushing ....

saltycrys:
This is great, another localization thing that could be done is finding strings that are supposed to get translated but aren't in the locale files. I made a list of some of them while adding error backtraces, see https://github.com/iv-org/invidious/issues/1497. Finding these could be done by grepping for calls to `translate` and `InfoException`.

SamantazFox:
@saltycrys I'll open a new issue with a list of all those problems. This PR is large enough like that, and I think that it's better to have separate PR for the different problems :)

saltycrys (approved):

Perflyst:
@TheFrenchGhosty do we need to change anything else on weblate before / after this gets merged?

Perflyst:
LGTM, couldn't find any issues after some testing. @TheFrenchGhosty please merge if weblate is ready or nothing needs to be changed.

TheFrenchGhosty:
@Perflyst I need to make weblate rebase from the repo. I'll do that later.

SamantazFox:
Do I need to rebase on current master?

SamantazFox:
rebased on master

PR #1921, opened 2021-03-24
Fix channel search API

Svallinn:
This PR fixes the functionality of the channels' search API
NOTE: `/browse_ajax` is deprecated and return 410s (Gone)


Svallinn:
My apologies. I only just realized I could run the tests myself instead of letting the CI do it -_-'

PR #1927, opened 2021-03-25
Add "Watch on Youtube" button next to audio mode Icon. 

### LINKED ISSUE #1833: [Enhancement] Small Button Next To Audio Icon For Youtube URL
Currently, for users who use Invidious simply for browsing or subscriptions, but choose to watch in something external like mpv using the original Youtube url, this is a bit tedious. If they wish to grab the original YouTube link they need to click onto the video, then right click the YouTube button to grab the link. Or manually edit the link.

My solution to make this easier, can come in two ways I imagine.
The first would be an option to rewrite all links as YouTube links, however this would only work if the user is using something such as Privacy Redirect. If they click the link, it redirects to the invidious link, if they right click it and copy it's the Youtube link.

The more ideal solution imo would come at the cost of adding an extra button, however it seems there is space. Right next to the headphones icon  for listen=1, could add a little YouTube icon that simply contains the original YouTube url. Users can then choose to grab the original Youtjbe url, or click the video title for the invidious url, without ever having to click into the video.

syeopite:
Closes #1833

### DIFF
--- a/src/invidious/views/components/item.ecr
+++ b/src/invidious/views/components/item.ecr
@@ -141,6 +141,9 @@
                 <b style="flex: 1;">
                     <a style="width:100%" href="/channel/<%= item.ucid %>"><%= item.author %></a>
                 </b>
+                <a title="<%=translate(locale, "Watch on YouTube")%>" href="https://youtube.com/watch?v=<%= item.id %>" style="margin-right: 5px;">
+                    <i class="icon ion-logo-youtube"></i>
+                </a>
                 <a title="<%=translate(locale, "Audio mode")%>" href="/watch?v=<%= item.id %>&amp;listen=1">
                     <i class="icon ion-md-headset"></i>
                 </a>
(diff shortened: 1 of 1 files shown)


src/invidious/views/components/item.ecr
```diff
@@ -141,7 +141,10 @@
                 <b style="flex: 1;">
                     <a style="width:100%" href="/channel/<%= item.ucid %>"><%= item.author %></a>
                 </b>
-                <a title="Audio mode" href="/watch?v=<%= item.id %>&amp;listen=1">
+                <a title="<%=translate(locale, "Watch on YouTube")%>" href="https://youtube.com/watch?v=<%= item.id %>" style="margin-right: 5px;">
```
Perflyst, line 144: youtube.com always redirects to www.youtube.com, so let's use www

saltycrys (approved):

TheFrenchGhosty:
Tested and merged.

Thanks @syeopite 

Issue #944, opened 2019-12-02
Promote the existence of other instances to clueless users on invidio.us and on other instances by default


Vagmer:
As is natural, many users of the official instance or even others may not be aware of the existence of alternatives instances, and that said instances may even function better for them. There's no indication that I can see of this concept on the Invidious site, so only if a visitor follows the link to github or reads about it elsewhere, they may know. Alleviating this can help user experience, split workload across instances and perhaps as a result help a little with instance YT banning/ratelimiting.

Possible suggestions:

1. Footer of every page is modified to include a one sentence description of how other Invidious instances exist and how experience on them might be better due to factors like geographical distance. This includes [a link to one of the instances lists](https://github.com/omarroth/invidious/wiki/Invidious-Instances).
2. Instance will try to perform a 'ping' action on visitors (limited to once every 0.5/1/12/24 hours etc) or have the visitor perform one to the instance (JS?). If the action is successful and its result is higher than a preset threshold in ms, for example 110ms, then **a header** will also be inserted to every page to let the user know of the same. This will appear above the video, video listing or channel listing. "Your connection to this Invidious instance seems to be slow. You may have a better experience on a different instance [<- links to an instances list]. Try to choose an instance geographically close to you.".  Mitigations:
	* this header includes a "Dismiss" link and a "Never show again" link at its end. "Dismiss" removes it from the current page. "Never show again" removes and sets a user preference (cookie/account) to disable whole feature # 2. Possible to decide to only show one link, the 2nd.
	* said user preference is accessible from the preferences page and can be supplied in 3 ways like other preferences: cookie/account/parameter in URL.
	* on the HTML-side this header includes some _unique_ `id` (or `class` possibly) on its parent element. This also allows foolproof easy hiding for users of extensions like uBlock Origin if they wish as well as easier potential interfacing for extensions/userscripts/userstyles.

IMO it would also be good to namedrop and link the Invidition extension on both # 1 and # 2. A sentence like  "Browser extensions such as [Invidition](https://addons.mozilla.org/en-US/firefox/addon/invidition/) can make it easier to use Invidious and to quickly switch between instances." Possible to also include note  "[instance name] is [not affiliated with/not responsible for] other instances or extensions."

Instance admin can configure the ping parameters (frequency, threshold) or choose to disable any of these features or subfeatures but they should probably be enabled by default. At any case should be enabled on invidio.us.

[...]

PR #1940, opened 2021-03-27
Add instance redirects!

### LINKED ISSUE #1873: [Feature request] Add UI button redirect to a healthier instance when watching a video on a particular instance
**Is your feature request related to a problem? 

It'd be great if there were a button or drop down menu to select another public instance.

For example,

For example, I would like to have a button on the interface for the video link below so I can choose a different instance if the snoypta instance is not cutting it that day for whatever reason:

https://invidious.snopyta.org/watch?v=2s9fM0CDAQ4

syeopite:
Add functions that redirect users to `redirect.invidious.io` or alternatively, if they choose so, redirect to a random update-to-date and healthy clear-net instance.

There's also a few front end additions.

The jet icon here is just a placeholder. I couldn't find any v4 ion icon that fits.

There's also a search redirect for when searching is broken. The filtering UI has been hidden to provide prominence to that message.
 redirect-3

The error template has also been enhanced to contain a link to Youtube, to switch Invidious instance, and to refresh the page.

error msg

Anyhow, this feature works by fetching an instance list from [api.invidious.io](https://api.invidious.io) so the load on that server would increase if merged. 

Closes #1873
Supersedes #1206


syeopite:
Related #944

TheFrenchGhosty:
>Anyhow, this feature works by fetching an instance list from api.invidious.io so the load on that server would increase if merged.

This isn't really a problem, the server has enough bandwidth to handle the requests, so this is fine.

src/invidious/routes/misc.cr
```diff
@@ -35,4 +35,15 @@ class Invidious::Routes::Misc < Invidious::Routes::BaseRoute
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
     rendered "licenses"
   end
+
+  def redirect(env)
```
saltycrys, line 39: Please change this to be generic, something like https://github.com/iv-org/invidious/blob/c9a316ad353cd1ee38c34d1aff0759765b37b2b3/src/invidious/views/template.ecr#L63

saltycrys (changes requested):

unixfox:
Is this really a good idea? What if it redirects to an instance that is down? api.invidious.io doesn't do a health check so that can happen.

Wouldn't it be better to redirect to https://redirect.invidious.io instead? This way the user can choose whatever instance he wants instead of random one and has a clear list of all the existing public instances.

TheFrenchGhosty:
Mmh I kinda agree with @unixfox here.

It would also simplify the code a lot.

syeopite:
> Is this really a good idea? What if it redirects to an instance that is down? api.invidious.io doesn't do a health check so that can happen.

Good point though the chances of running into an dead instance are pretty slim. I instance hop frequently and I've only ever encountered an instance that is down once. It's far more likely to reach a outdated instance instead. The latter of which can be solved if we just compared the versions to the current one.

Personally I prefer this version more than the only redirect to [https://redirect.invidious.io](https://redirect.invidious.io). Its less work on the users end and even allows for site routes to pass through for JS disabled users.

saltycrys (approved):

Perflyst:
How do we continue with this one? I also prefer redirect.invidious.io...

> Personally I prefer this version more than the only redirect to https://redirect.invidious.io. Its less work on the users end and even allows for site routes to pass through for JS disabled users.

redirect.invidious.io works without javascript too (not completely though)

syeopite:
Compromise and allow it to be configurable?

syeopite:
There we go. All of these concerns has been addressed.

unixfox:
What happens if api.invidious.io is down? Does it skip the rendering of the link?

syeopite:
`redirect.invidious.io` would be used instead.

syeopite:
I'll like to add to my previous argument about potentially redirecting to instances that are down. `api.invidious.io` currently refreshes its instance list every 5 minutes; any instance that is down for longer than that would quickly be removed from the API as well.

unixfox:
> I'll like to add to my previous argument about potentially redirecting to instances that are down. `api.invidious.io` currently refreshes its instance list every 5 minutes; any instance that is down for longer than that would quickly be removed from the API as well.  

Not if the API starts to freeze, which often happens and need a restart.

TheFrenchGhosty:
@unixfox The api has a cronjob that restart it daily

unixfox:
> @unixfox The api has a cronjob that restart it daily

If the API happens to freeze before the restart for whatever reason, this system will still offer instances that could be down until the API is restarted.

Unless this system rely on a reliable healthcheck service like uptime robot, there will be cases where an instance that is down could be offered to the user.

syeopite:
Even if one got redirected to an instance that is down —  the chance of which is already extremely low — how is it any different from clicking on an instance that is down on redirect.invidious.io? In both cases you'll have to go back to the original page and pick a new instance or in the case of this PR click on the switch instance button again

unixfox:
> Even if one got redirected to an instance that is down — the chance of which is already extremely low — how is it any different from clicking on an instance that is down on redirect.invidious.io? In both cases you'll have to go back to the original page and pick a new instance or in the case of this PR click on the switch instance button again

On redirect.invidious.io, multiple instances get presented to you. With your system, only one is given.
If an instance is down from redirect.invidious.io then the user will probably understand a bit better that he can try another one because he saw multiple instances available.
Whereas with your system just one link is given and the user may not understand directly that it can come back to the previous page THEN refresh the page for getting another instance.

syeopite:
Good point, though you exactly don't need to refresh the page to head to another instance. Just going back to the previous page would be enough to get reconnected to another instance as any connection to the `/redirect` endpoint would cause a redirect.

PR #1942, opened 2021-03-28
Add multiple video route aliases. 

### LINKED ISSUE #1736: https://www.youtube.com/watch/videoid support
can this be added to the support?
like this?
https://invidious.kavin.rocks/watch/ZvWn5xBflUs

syeopite:
Adds support for:

- /watch/
- /shorts/
- /e/
- /v/
- /w/

as possible video routes. 

Closes #1736

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -308,9 +308,17 @@ end
 Invidious::Routing.get "/", Invidious::Routes::Misc, :home
 Invidious::Routing.get "/privacy", Invidious::Routes::Misc, :privacy
 Invidious::Routing.get "/licenses", Invidious::Routes::Misc, :licenses
-Invidious::Routing.get "/watch", Invidious::Routes::Watch
+
+Invidious::Routing.get "/watch", Invidious::Routes::Watch, :handle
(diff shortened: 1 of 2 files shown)


saltycrys (changes requested):
All of these redirect to `/watch?q=` so it would be better to add a `Invidious::Routes::Watch :redirect` route for
- `/w/:id`
- `/v/:id`
- `/e/:id`
- `/shorts/:id`
- `/watch/:id`
- 
and leave the main watch route and 404 handler untouched. Basically the inverse of the embed routes.

saltycrys (approved):

Perflyst:
@TheFrenchGhosty do we need to change anything else on weblate before / after this gets merged?

Perflyst (approved):
Works great, thanks for your contribution!

B0pol:
what about URL parameters? They are just removed with this PR if using one of these short URLs

syeopite:
> what about URL parameters? They are just removed with this PR if using one of these short URLs

Oops, I forgot about url parameters. #2011 should fix it

Issue #1673, opened 2021-01-16
[Bug] Unexpected char '<' at 1:1 (JSON::ParseException) - On videos


M86xKC:
When trying to open a video it give the following error, it is worth noting that this only happens on certain videos ([this is one of them](https://youtube.com/watch?v=VQOdmckqNro)):

steps to reproduce:
open: `https://<instance name>/watch?v=VQOdmckqNro`
observe the resulted exception

```
Title: `Unterminated string at 1:34677 (JSON::ParseException)`
Date: `2021-01-16T19:14:42Z`
Route: `/watch?v=VQOdmckqNro`
Version: `2021.01.03-28dac81 @ master`
<details>
<summary>Backtrace</summary>
<p>
   
Unterminated string at 1:34677 (JSON::ParseException)
  from ../../../usr/share/crystal/src/json/lexer.cr:371:7 in 'raise'
  from ../../../usr/share/crystal/src/json/lexer/string_based.cr:63:8 in 'next_token'
  from ../../../usr/share/crystal/src/json/parser.cr:85:19 in 'parse_value'
  from ../../../usr/share/crystal/src/json/parser.cr:85:19 in 'parse_value'
  from ../../../usr/share/crystal/src/json/parser.cr:107:11 in 'extract_polymer_config'
  from src/invidious/videos.cr:965:10 in 'get_video:region'
  from src/invidious/routes/watch.cr:61:15 in 'handle'
  from ../../../usr/share/crystal/src/primitives.cr:255:3 in '->'
  from ../../../usr/share/crystal/src/primitives.cr:255:3 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from ../../../usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:189:11 in 'call'
[...]

PR #1953, opened 2021-03-30
Update regex expressions to handle unexpected '};'

syeopite:
Fixes: #1673

Tested with the following ids:
<details>

O4r9D2jI0_w

VQOdmckqNro

3AF3sYnvj5o

BFlF9gGnyro 

BUU7WFd8XLU

channel/UCRZY6MGMG1Fmzt50rYalTwQ

</details>

I wasn't able to reproduce the error in any of the videos linked in duplicate issues. However, this does fix the error in the original video and any other content with "};".

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -818,7 +818,7 @@ end
 
 def extract_polymer_config(body)
   params = {} of String => JSON::Any
-  player_response = body.match(/(window\["ytInitialPlayerResponse"\]|var\sytInitialPlayerResponse)\s*=\s*(?<info>{.*?});/m)
+  player_response = body.match(/(window\["ytInitialPlayerResponse"\]|var\sytInitialPlayerResponse)\s*=\s*(?<info>{.*?});\s*var\s*meta/m)
     .try { |r| JSON.parse(r["info"]).as_h }
 
   if body.includes?("To continue with your YouTube experience, please fill out the form below.") ||
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -678,7 +678,7 @@ def create_notification_stream(env, topics, connection_channel)
 end
 
 def extract_initial_data(body) : Hash(String, JSON::Any)
-  return JSON.parse(body.match(/(window\["ytInitialData"\]|var\s*ytInitialData)\s*=\s*(?<info>\{.*?\});/mx).try &.["info"] || "{}").as_h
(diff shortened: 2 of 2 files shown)


src/invidious/videos.cr
```diff
@@ -818,7 +818,7 @@ end
 
 def extract_polymer_config(body)
   params = {} of String => JSON::Any
-  player_response = body.match(/(window\["ytInitialPlayerResponse"\]|var\sytInitialPlayerResponse)\s*=\s*(?<info>{.*?});/m)
+  player_response = body.match(/(window\["ytInitialPlayerResponse"\]|var\sytInitialPlayerResponse)\s*=\s*(?<info>{.*?});\s*var\s*meta/m)
```
SamantazFox, line 821: I have noticed that this extra javascript (`var meta = ...`) is not always present in the script.
I can't tell what triggers it (I've tried multiple combinations of URL arguments, with no luck).

syeopite: Huh that's strange. Its always been present on my end.

saltycrys (approved):
The most robust way to handle this would be to put a full JSON parser in there so only the starting point needs to be found. That would be terrible for performance probably.

But as long as this works that's not necessary.

SamantazFox:
> The most robust way to handle this would be to put a full JSON parser in there so only the starting point needs to be found. That would be terrible for performance probably.

Or, as proposed in #1981, find a direct API providing the required `ytInitialData` (and `ytInitialPlayerResponse` for videos)...

saltycrys:
> Or, as proposed in #1981, find a direct API providing the required `ytInitialData` (and `ytInitialPlayerResponse` for videos)...

That is fantastic! @TheFrenchGhosty put some bounties on that please.

Perflyst:
`watch?v=VQOdmckqNro` works on my instance without this patch so I can't really test it

TheFrenchGhosty:
Merged.

@syeopite to receive your reward either send your BTC address in a comment (preferred), or contact us directly, with a proof that this is indeed you.

syeopite:
bc1qpgyc5hkcwg4l4a28j274g4tufym07e0as7pwwk

TheFrenchGhosty:
A 20$ bounty has been rewarded to @syeopite for this PR.

Transaction (caa3a36040c086642a0262048f95b0c5d84009e87ddc34acf65b6928895f3fdb):

Amount sent: 0.000405221 BTC (~16.46€ or ~20$)

Mining fees (10 sat/byte): 0.0000481 BTC (~1.95€)

PR #1911, opened 2021-03-20
Fix long playlists (100+ videos)

SamantazFox:
This should fix issue #1883.
I've also made some changes in `src/invidious/channels.cr` because both channel and playlist videos require a call to `[YT]/youtubei/v1/browse`, so I made a common wrapper (it will ease future updates if the key or client version strings change).

PR #1967, opened 2021-04-01
Fix missing last page in playlists

SamantazFox:
Bug introduced by #1911: Playlists with a video count is not a perfect multiple of 100 were missing their last page.
This is due to not taking into account the remainder when counting pages.

Thanks to Baobab on IRC for catching this one (They don't have a github account)!


TheFrenchGhosty:
Thank you!

Issue #957, opened 2019-12-10
Revert back to http connection


tleydxdy:
Now that using quic will be banned too, I think we should go back to using http, as it's less dependency, and less fragile (on Alpine at least). or offer it as a build option?

unixfox:
How do you know that the quic workaround is banned by Google? I do agree that it may fix #917 but if the workaround is still working why reverting to a state where almost every invidious would stop working due to the Google captcha?

haizrul:
Use https://anti-captcha.com will solve the captcha problem.

tleydxdy:
@unixfox because people are being banned?

unixfox:
@tleydxdy I didn't experience any ban at all (my instance is yewtu.be).
The only thing that I did is to deny all the API endpoints except from the ones that the web interface uses. This most likely reduced the amount of requests to Google but I didn't experience any ban **yet**.

haizrul:
> @tleydxdy I didn't experience any ban at all (my instance is yewtu.be).
> The only thing that I did is to block all the API endpoints except from the ones that the web interface uses. This most likely reduced the amount of requests to Google but I didn't experience any ban **yet**.
> I'm pretty sure that if Invidious wasn't using the quic workaround my instance would have been banned a long time ago ([from my experience with dealing with reCaptcha on Searx](https://github.com/asciimoo/searx/issues/729)).

Hi sir, may i know how to block all the API endpoints except for web uses like you said? I want to implement it on my instance too. Please help.

unixfox:
> > @tleydxdy I didn't experience any ban at all (my instance is yewtu.be).
> > The only thing that I did is to block all the API endpoints except from the ones that the web interface uses. This most likely reduced the amount of requests to Google but I didn't experience any ban **yet**.
> > I'm pretty sure that if Invidious wasn't using the quic workaround my instance would have been banned a long time ago ([from my experience with dealing with reCaptcha on Searx](https://github.com/asciimoo/searx/issues/729)).
> 
> Hi sir, may i know how to block all the API endpoints except for web uses like you said? I want to implement it on my instance too. Please help.

I just used the [status](https://caddyserver.com/v1/docs/status) parameter of Caddy like this way:
```
status 403 {
       /api/v1/videos
       /api/v1/channels
       /api/v1/search
       /api/v1/mixes
}
```

haizrul:
I use Debian 10, can you advice what file should i edit?

unixfox:
[...]

PR #1975, opened 2021-04-03
add option to disable quic

tenpura-shrimp:
The primary use of quic is to avoid certain google capchas (with other potential benefits).  But, for small instances this isn't necessary and also brings with it a whole bunch of new complexity.

related discussion: #957

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -95,7 +95,7 @@ LOCALES = {
   "zh-TW" => load_locale("zh-TW"),
 }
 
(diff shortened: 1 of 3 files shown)


unixfox:
More than half of a day of testing on yewtu.be and everything works as fine as it would with quic.

This doesn't reduce the memory usage and CPU usage at all though. I still get a lot of memory used on Invidious processes. And slowdowns when loading webpages on Invidious still exist.

saltycrys (changes requested):
Please rename `QUICPool`.

saltycrys (approved):

PR #1976, opened 2021-04-04
Fix footer position to bottom of page

syeopite:


hackerncoder:
360x740 (Galaxy S9/S9+)

Current: 
image


PR: 
image


&nbsp;

1280x720:

Current: 
image

PR: 
image

hackerncoder (approved):

hackerncoder:
Looks good on an empty page: 
image

And on a page with videos:
image

TheFrenchGhosty:
Tested, it's almost perfect except that on mobile it make it require scrolling to see everything, even though it's empty on most of the page.

This need to be fixed.

syeopite:
There we go. All but the tiniest screen sizes should have the entire footer visible now. 

microscopic screen sizes
Normal phone screen

src/invidious/views/template.ecr
```diff
@@ -106,6 +106,7 @@
 
             <%= content %>
 
+            <div id="placeholder" style="padding: 6em;"></div>
```
saltycrys, line 109: Please remove the placeholder and add the "padding" to the footer itself. It should be possible to change the footer's `margin` to a `padding` and add the placeholder's `padding` as `margin-top`.

While the footer is modified we should also change the `<div class="footer">` (which should have been a `<div id="footer">` to begin with) into `<footer>`.

saltycrys (changes requested):

src/invidious/views/template.ecr
syeopite: Without the placeholder element things quickly overlap even after transferring the css.

placeholder

saltycrys: Oh, I overlooked the `position: absolute` and thought the placeholder was between the content and the footer, sorry.

Please use flexbox instead of absolute positioning.

saltycrys (approved):

Issue #1625, opened 2020-12-31
default_home set to an empty value uses the default value


TheFrenchGhosty:
If ```default_home``` is set to an empty value using ```default_home: []``` Invidious will crash when starting.

The config I use: https://github.com/TheFrenchGhosty/PussTheCat.org-Configs/blob/master/Services/Invidious/config.yml

saltycrys:
The value of `default_home` is expected to be a string and `[]` is an empty array.
Does it crash with an empty string like `default_home: ""`?

TheFrenchGhosty:
@saltycrys I just tried and it indeed doesn't crash, however it assume it's set to popular, and redirect to popular.

Any reason why it doesn't work with ```[]``` but ```feed_menu``` does?

saltycrys:
Crystal is a strongly typed language, which means that all possible types of a value must be known at compile time.

When Invidious loads values from the config file it parses them into the types defined in the `Config` class: https://github.com/iv-org/invidious/blob/b19524d56a49c01ec72deb33bfb423ac25721ffc/src/invidious/helpers/helpers.cr#L64-L116

`default_home` is in `default_user_preferences` which has a type of `ConfigPreferences`: https://github.com/iv-org/invidious/blob/b19524d56a49c01ec72deb33bfb423ac25721ffc/src/invidious/helpers/helpers.cr#L82

The `ConfigPreferences` struct: https://github.com/iv-org/invidious/blob/b19524d56a49c01ec72deb33bfb423ac25721ffc/src/invidious/helpers/helpers.cr#L25-L62

`default_home`'s type is `String`: https://github.com/iv-org/invidious/blob/b19524d56a49c01ec72deb33bfb423ac25721ffc/src/invidious/helpers/helpers.cr#L45

`feed_menu`'s type is `Array(String)` (note that an empty array is an array of strings with no elements): https://github.com/iv-org/invidious/blob/b19524d56a49c01ec72deb33bfb423ac25721ffc/src/invidious/helpers/helpers.cr#L46

<hr>

`|` means `or` so `use_pubsub_feeds` can be either `Bool` or `Int32`: https://github.com/iv-org/invidious/blob/b19524d56a49c01ec72deb33bfb423ac25721ffc/src/invidious/helpers/helpers.cr#L74

[...]

PR #1977, opened 2021-04-04
Change empty home page to search engine like layout

### LINKED ISSUE #996: [Feature Request] Proper Search-engine-like page
After #449 it is now possible to have a page without anything other than a search box:

However the layout of the page follow the layout of every other page without the unwanted content and doesn't actually looks like a search engine page.

searx

The missing things for this to be complete are: search box centered, Invidious displayed above it, footer at the bottom.

Basic sketch of the result:

result

syeopite:
Search engine page

It does look a bit cluttered without #1976 though

Closes #996


TheFrenchGhosty:
Nice! How to enable it exactly?

syeopite:
Just set the default homepage in preferences to none

TheFrenchGhosty:
You sure? https://github.com/iv-org/invidious/issues/1625

syeopite:
Huh didn't know that was an issue. Seems like setting default_home to none in the configuration file is broken but not within the UI so just do it there.

GauthierPLM:
Nice improvement. Looking forward to use it.

TheFrenchGhosty:
Tested and works perfectly, except for the footer on mobile, but I'm guessing if https://github.com/iv-org/invidious/pull/1976 get fixed it will be fine.

saltycrys (changes requested):
This is very nice!

> Huh didn't know that was an issue. Seems like setting the default homepage to none in the configuration file is broken but not within the UI

Setting it to an empty string causes problems, see https://github.com/iv-org/invidious/pull/1980#issuecomment-814083378.

When setting it to none from the UI it sets it to `default_home: "none"` which works because of the else here: https://github.com/iv-org/invidious/blob/f20d420aa74a55ae80f91bba00aa9d472bd74f1e/src/invidious/routes/misc.cr#L7-L26

<hr>

The option should be renamed in the preferences:
https://github.com/iv-org/invidious/blob/198dfffaeb91c03889440a9f8141657bd53734d6/src/invidious/views/preferences.ecr#L142
https://github.com/iv-org/invidious/blob/198dfffaeb91c03889440a9f8141657bd53734d6/src/invidious/views/preferences.ecr#L214

syeopite:
> The option should be renamed in the preferences:

Could you elaborate on what you mean by rename? I assume it should be changed to something along the lines of "search"?

saltycrys:
> Could you elaborate on what you mean by rename? I'm assuming that you mean it should be changed to something that reflects the function of the repurposed page.

Yes. "Search" is fine.

saltycrys (approved):
This is fine and it can get merged, I just have some nitpicks.

- The page isn't empty anymore so it should be renamed
- The check in `template.ecr` to see if the search bar should be rendered should be moved into the `templated` macro

The `templatd` macro is in https://github.com/iv-org/invidious/blob/30a2c28d74114e717828d87a90647324f7512048/src/invidious/helpers/macros.cr#L51-L53
It could be changed to
```crystal
macro templated(filename, template = "template", navbar_search = true)
  navbar_search = {{navbar_search}}
  render "src/invidious/views/#{{{filename}}}.ecr", "src/invidious/views/#{{{template}}}.ecr"
end
```
which would be called with
```crystal
templated "empty", navbar_search: false
```
it could also just compare the filename. The advantage is that the search page can then be used from anywhere and doesn't rely on any preference values or the route.

TheFrenchGhosty:
Amazing PR, thanks for doing it @syeopite

PR #1980, opened 2021-04-04
Allow default_home config to be an empty string

### LINKED ISSUE #1625: default_home set to an empty value uses the default value

syeopite:
Fixes #1625 
This PR makes it possible to set an empty homepage through the configuration file.

Based on @saltycrys's [typing explanation here](https://github.com/iv-org/invidious/issues/1625#issuecomment-753240245).

### DIFF
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -42,7 +42,7 @@ struct ConfigPreferences
   property player_style : String = "invidious"
   property quality : String = "hd720"
   property quality_dash : String = "auto"
-  property default_home : String = "Popular"
+  property default_home : String? = "Popular"
   property feed_menu : Array(String) = ["Popular", "Trending", "Subscriptions", "Playlists"]
   property related_videos : Bool = true
   property sort : String = "published"
--- a/src/invidious/users.cr
+++ b/src/invidious/users.cr
@@ -68,7 +68,7 @@ struct Preferences
   property quality : String = CONFIG.default_user_preferences.quality
   @[JSON::Field(converter: Preferences::ProcessString)]
   property quality_dash : String = CONFIG.default_user_preferences.quality_dash
-  property default_home : String = CONFIG.default_user_preferences.default_home
+  property default_home : String? = CONFIG.default_user_preferences.default_home
   property feed_menu : Array(String) = CONFIG.default_user_preferences.feed_menu
   property related_videos : Bool = CONFIG.default_user_preferences.related_videos
 


TheFrenchGhosty:
What syntax is it expecting now for it to work?

syeopite:
Empty string.

TheFrenchGhosty:
```default_home: ""```

or

```default_home: []```

or

```default_home: ```

syeopite:
First one

TheFrenchGhosty:
Tested, it's working perfectly.

Thanks @syeopite

saltycrys:
Setting `default_home: ""` (to an empty string) was always allowed, the problem is that `Object.from_yaml` uses the default value when encountering an empty string. If the property is nillable and an empty string is encountered it uses `nil`. This is a Crystal bug, I opened a bug report at https://github.com/crystal-lang/crystal/issues/10606.

@TheFrenchGhosty 
> `default_home: ""`

That's an empty string.

> `default_home: []`

That's an empty array.

> `default_home: `

That's a null.

Maybe [this](https://yaml.org/spec/1.2/spec.html#id2805071) can help.

TheFrenchGhosty:
@saltycrys So why does this PR works then?

saltycrys:
As explained in my comment:
- Calling `Object.from_yaml` with an empty string value (YAML: `default_home: ""`) combined with a **non-nillable** string property that **has a default value** (Crystal: `property default_home : String = "Popular"`) causes the **default value** to be used, which in this case is "Popular".
So in this case YAML `default_home: ""` behaves the same as YAML `default_home: "Popular"`.
https://github.com/iv-org/invidious/blob/378c0d049e1064f18b2dbcd8d00c67a42da970f6/src/invidious/helpers/helpers.cr#L45
- Calling `Object.from_yaml` with an empty string value (YAML: `default_home: ""`) combined with a **nillable** string property, **regardless of default value** (Crystal: `property default_home : String? = "Popular"` note the `?` question mark) causes **nil** be used.
So in this case YAML `default_home: ""` behaves the same as YAML `default_home: null`.
https://github.com/iv-org/invidious/blob/ae353cef2e3013bc1d995d3c607e7ef2b76f7ffe/src/invidious/helpers/helpers.cr#L45

This behavior is a bug in Crystal and will be fixed in the next Crystal version.
<hr>

The code that checks the value of `default_home` compares it to known strings and if no known string is found the default empty page is used: https://github.com/iv-org/invidious/blob/ae353cef2e3013bc1d995d3c607e7ef2b76f7ffe/src/invidious/routes/misc.cr#L7-L26

PR #1005, opened 2020-01-27
Redesign videojs player controls for better mobile usage

Ruin0x11:
Closes #912.

This PR changes the player controls so the seek bar always fits the width of the screen. This makes seeking far easier in portrait mode on mobile devices. The controls are also larger for easier touch input.

The theme is based on the [city](https://github.com/videojs/themes/tree/master/city) theme provided by Video.js.

Previous:

New:

PR #2008, opened 2021-04-11
Improve player controls for mobile devices 

### LINKED ISSUE #1579: Control bar broken on mobile
With subtitles | Without subtitles
---- | -----
 | 

Please always put fullscreen button on portrait mode on mobile, it's only available on landscape mode.

I think speed button should be ditched, then it would be

With subtitles | Without subtitles
---- | ----
subtitles ; video settings ; fullscreen | share ; video settings ; fullscreen

syeopite:
Supersedes #1005 
Closes #1579

Screenshot_20210411-050427 cleaned

This PR also includes [videojs-mobile-ui](https://github.com/mister-ben/videojs-mobile-ui), enabling support for double tab forwarding/rewinding and on-screen pause/play toggle.

In addition to the overhauled mobile player UI this PR also contains a revamped Youtube player style option. 

yt-style


syeopite:
As a side note, I think we should allow users to configure whether they want the duration, remaining time or both.

PR #2011, opened 2021-04-12
Preserve url params in video route aliases

syeopite:


TheFrenchGhosty:
What does this fixes exactly?

SamantazFox:
> What does this fixes exactly?

See https://github.com/iv-org/invidious/pull/1942#issuecomment-816164829

PR #1678, opened 2021-01-17
Make DB migrations idempotent.

Rjevski:
Fixes #1772

This is more of a discussion than anything else - the provided code will work and maybe should be merged, but I don't think it's a complete solution.

Currently we don't seem to have a good system for database migrations. The scripts in `config/migrate-scripts` seem to be written manually, hardcode a database name and username, expect to be ran on the local machine and don't use the SQL files in `config/sql` (so it's one extra thing that needs to be manually kept in sync). They're also not aware of the current database state, so an automated deployment pipeline can't tell whether it needs to run them (or if they were already applied) and the SQL in there doesn't seem to be idempotent which means it will fail because those tables are already there.

I don't think implementing our own migration system is the way to go. If Crystal has something "standard" for it I suggest we use that, otherwise we could consider [Alembic](https://alembic.sqlalchemy.org/en/latest/)?

The reason behind this is that I'm trying (well, I've succeeded - with workarounds and hacks) to deploy Invidious on a PaaS such as Dokku. The PaaS deploy process allows you to run a "post-release" command which is where database migrations should go. Most web frameworks automatically detect which migrations are applied so it's safe to run the command on every deploy even if no migrations are pending; unfortunately Invidious does not have this. This change makes the migrations idempotent (by using `IF NOT EXISTS` on the create statements) so any subsequent runs will not error out.

PR #2043, opened 2021-04-28
Configuration fixes for Docker and docker-compose

sumdog:
The current docker-compose file indicates the entire configuration for invidious can be placed inside a single environment variable, but I couldn't get configuration changes to take effect. Unless I'm missing something, it seems like the program's entrypoint and crystal configuration loading does support pulling in the entire configuration from an environment variable (maybe it did at one time?)

I've adjusted the docker-compose file to define individual configuration variables and allow for custom usernames and password for the database initialization. I currently use a `sed` command in startup to adjust the SQL files, which is kinda hackey, and it seems like this might be addressed in #1678.

The configmap via Kubernetes seems to correctly support injecting a configuration file, but might break if this PR's startup replaces that configuration file on startup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,12 +5,10 @@ services:
     restart: unless-stopped
     volumes:
       - postgresdata:/var/lib/postgresql/data
-      - ./config/sql:/config/sql
-      - ./docker/init-invidious-db.sh:/docker-entrypoint-initdb.d/init-invidious-db.sh
     environment:
-      POSTGRES_DB: invidious
-      POSTGRES_PASSWORD: kemal
-      POSTGRES_USER: kemal
+      POSTGRES_DB: invidiousdb
+      POSTGRES_PASSWORD: R3p1aceW1thStr0ngPa33word
+      POSTGRES_USER: invidious
     healthcheck:
       test: ["CMD", "pg_isready", "-U", "postgres"]
   invidious:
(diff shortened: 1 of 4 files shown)


unixfox:
> The configmap via Kubernetes seems to correctly support injecting a configuration file, but might break if this PR's startup replaces that configuration file on startup.

Maybe we should just ditch the configmap and use environment variables for the helm chart.

sumdog:
> 
> Maybe we should just ditch the configmap and use environment variables for the helm chart.

I added f9891dbeef041d4f5d00dbb7b9e7ab8fe7dd2a04 which will check to see if the config file exists and skip overwriting it. This should allow to keep the existing helm chart working for now.

saltycrys:
Loading config is already possibly from a YAML file, a single env var containing YAML and multiple env vars to override specific preferences. See https://github.com/iv-org/invidious/blob/75e5b49c3ae684096c8074bada846414d2b02b88/src/invidious/helpers/helpers.cr#L124-L197

The logic is:
- Load base
  - If there is a YAML config in `INVIDIOUS_CONFIG` (what the docker-compose uses) load that
  - If not check `INVIDIOUS_CONFIG_FILE` for a config file name (defaulting to `config/config.yml`) and load that
- Override
  - Check all config key env vars (`INVIDIOUS_` + upcased preference key) and use the value to override

PR #2058, opened 2021-05-01
limit feeds and delete materialized views

tenpura-shrimp:
The materialized views causes excessive load on the db for large instances, because it is constantly requerying and recalculating the feeds.

Omar added these views because for some users, doing the join for the feed on the fly was too expensive for the database.  I found this was likely happening because there was too much in the feed for a single user to query and sort.  The feed includes all past videos from all subscriptions, so it continues to grow unlimitedly.

As a mitigation, this change limits feeds to 1 month which should put a reasonable upper bound on the number of videos the database needs to process to generate the feed.  Another contributor suggested to modify the limit based on the number of subscriptions the user has.  I can try that in a future revision but I think this is a good start to see if the approach is reasonable..

To instance admins:  you MUST run the migration or you will have a very bad time.


unixfox:
In testing on YewTube. First impressions: seems to reduce the load, at least from postgres.

unixfox:
3 days after I can see a very significant drop of CPU usage from postgres:

(I deployed this PR on the 2nd of May at around 11AM).

I haven't encountered any issue or frozen subscriptions. Rock solid to this day.

saltycrys (approved):
Please add a way to see older subscribed videos, maybe a time frame input somewhere.

unixfox:
@saltycrys 
> Please add a way to see older subscribed videos, maybe a time frame input somewhere.

What about allowing to set the duration (could be 1 monh by default) in the config file instead?

syeopite:
If @tenpura-shrimp doesn't continue with this PR perhaps we should just merge it as is. There really isn't any need to access subscribed videos that are older than a month considering it's the subscription *feed*.

SamantazFox:
@syeopite we need to make sure that the database can be migrated automagically, otherwise there is going to be a lot of broken instances.

saltycrys:
> What about allowing to set the duration (could be 1 month by default) in the config file instead? Would that be a good trade off?

Sounds good, but the default should be very long instead of very short. That way small instances used by only a couple of people can have the full functionality while big instances can have better performance.

unixfox:
This needs a rebase, if I've time I'll do it if not someone else can do it thanks.

unixfox:
Recreated in https://github.com/iv-org/invidious/pull/2469 as the author deleted its github repository.

PR #2062, opened 2021-05-05
Set correct permissions in Dockerfile

Maykin-99:
When having a `umask` of `xx7` (e.g. `027`) on the host machine then the directories `assets`, `config` and `locales` don't become readable to the `invidious` user inside the Docker container since the `COPY` commands result in files owned by `root` with the same file permissions like on the host (`640` in my case).

By adding `--chown=invidious` to the `COPY` command we ensure the `invidious` user can read these files.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,12 +18,12 @@ RUN apk add --no-cache librsvg ttf-opensans
 WORKDIR /invidious
 RUN addgroup -g 1000 -S invidious && \
     adduser -u 1000 -S invidious -G invidious
-COPY ./assets/ ./assets/
+COPY --chown=invidious ./assets/ ./assets/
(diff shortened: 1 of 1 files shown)


Perflyst (approved):

leonklingele:
I guess this also gives write access to the `invidious` user, right? Then this should be reverted.

Maykin-99:
> I guess this also gives write access to the `invidious` user, right? Then this should be reverted.

Yes it does. Is this a problem?

Alternatively I could adjust the commit and add the following near the end:

```Dockerfile
RUN chmod o+rX -R ./assets ./config ./locales
```

Perflyst:
@leonklingele It is reverted / improved now but I am still interested why this might be an issue.

PR #2072, opened 2021-05-09
Revert "Set correct permissions in Dockerfile"

Maykin-99:
Reverts iv-org/invidious#2062


Maykin-99:
The 2nd commit makes the `assets`, `config` and `locales` directories and their content only readable to `invidious` instead of owned by `invidious`.

Issue #25, opened 2018-07-20
Add thin mode


omarroth:
Currently, the largest amount of [content](http://www.webpagetest.org/result/180720_Q1_2ee7f7b0ab9afb73fe79e5d3c492e514/1/breakdown/) on the homepage (by byte size) is images, taking up more than 95% of the bandwidth used on the homepage. The site remains more than functional without them, so it should be possible for users to do so by adding an option in `/preferences`.

PR #2083, opened 2021-05-14
config_example.yaml: document all possible options

SamantazFox:
Fixes https://github.com/iv-org/documentation/issues/46

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -713,6 +713,15 @@ default_user_preferences:
   ##
   #volume: 100
 
+  ##
+  ## Allow 360° videos to be played.
+  ##
+  ## Note: This feature requires a WebGL-enabled browser.
+  ##
+  ## Accepted values: true, false
+  ## Default: true
+  ##
+  #vr_mode: true
 
   # -----------------------------
   #  Subscription feed
@@ -781,3 +790,23 @@ default_user_preferences:
   ## Default: false
   ##
   #local: false
+
+  ##
+  ## Show the connected user's nick at the top right.
+  ##
+  ## Accepted values: true, false
+  ## Default: true
+  ##
+  #show_nick: true
+
+  ##
+  ## Automatically redirect to a random instance when the user uses
+  ## any "switch invidious instance" link (For videos, it's the plane
+  ## icon, next to "watch on youtube" and "listen"). When set to false,
+  ## the user is sent to https://redirect.invidious.io instead, where
+  ## they can manually select an instance.
+  ##
+  ## Accepted values: true, false
+  ## Default: false
+  ##
+  #automatic_instance_redirect: false
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -146,9 +146,12 @@ https_only: false
 ## optimized. You can read more about QUIC at the link below:
 ## https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-31
 ##
-## Note: you should try both options and see what is the best for
-## your instance. Also, some maintainers reported that using QUIC
-## may reduce the occurence of captcha challenges from youtube.
+## Note: you should try both options and see what is the best for your
+## instance. In general QUIC is recommended for public instances, and
+## not recommended for private one. Your mileage may vary.
+##
+## Note 2: Using QUIC prevents some captcha challenges from appearing.
+## See: https://github.com/iv-org/invidious/issues/957#issuecomment-576424042
 ##
 ## Accepted values: true, false
 ## Default: true
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -19,7 +19,7 @@ db:
 ## Database configuration using a single URI. This is an
 ## alternative to the 'db' parameter above. If both forms
 ## are used, then only database_url is used.
-## This setting is MANDATORY if 'db' is not set.
+## This setting is MANDATORY, unless 'db' is used.
 ##
 ## Note: The 'database_url' setting allows the use of UNIX
 ## sockets. To do so, remove the IP address (or FQDN) and port
@@ -66,6 +66,9 @@ db:
 ## invidious know about it. This is used to craft absolute URLs
 ## to the instance (e.g in the API).
 ##
+## Note: This setting is MANDATORY if invidious is behind a
+## reverse proxy.
+##
 ## Accepted values: 1-65535
 ## Default: <none>
 ##
@@ -90,7 +93,9 @@ domain:
 
 ##
 ## Tell Invidious that it is behind a proxy that provides only
-## HTTPS, so all links must use the https:// scheme.
+## HTTPS, so all links must use the https:// scheme. This
+## setting MUST be set to true if invidious is behind a
+## reverse proxy.
 ##
 ## Accepted values: true, false
 ## Default: false
@@ -176,18 +181,10 @@ https_only: false
 ## Logging Verbosity. This is overriden if "-l LEVEL" or
 ## "--log-level=LEVEL" are passed on the command line.
 ##
-## Accepted values:
-##   LogLevel::All    (Log everything)
-##   LogLevel::Trace
-##   LogLevel::Debug
-##   LogLevel::Info
-##   LogLevel::Warn
-##   LogLevel::Error
-##   LogLevel::Fatal
-##   LogLevel::Off    (Disable log output)
-## Default: LogLevel::Info
+## Accepted values: All, Trace, Debug, Info, Warn, Error, Fatal, Off
+## Default: Info
 ##
-#log_level: LogLevel::Info
+#log_level: Info
 
 
 # -----------------------------
@@ -208,8 +205,8 @@ https_only: false
 ##   - Software name ("invidious") and version+branch (same data as
 ##     displayed in the footer, e.g: "2021.05.13-75e5b49" / "master")
 ##   - The value of the 'registration_enabled' config (true/false)
-##   - Number of currently connected users
-##   - Average Number of connected users (last month & last 6 months)
+##   - Number of currently registered users
+##   - Number of users who registered in the last month / last 6 months
 ##   - Timestamp of the last server restart
 ##   - Timestamp of the last "Channel Refresh" job execution
 ##
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -333,6 +333,9 @@ feed_threads: 1
 ## Enable/Disable the polling job that keeps the decryption
 ## function (for "secured" videos) up to date.
 ##
+## Note: This part of the code is currently broken, so changing
+## this setting has no impact.
+##
 ## Accepted values: true, false
 ## Default: true
 ##
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -673,7 +673,7 @@ default_user_preferences:
 
 
   # -----------------------------
-  #  Video player settings
+  #  Video playback settings
   # -----------------------------
 
   ##
@@ -698,7 +698,7 @@ default_user_preferences:
   #quality_dash: auto
 
   ##
-  ## Video playback speed.
+  ## Default video playback speed.
   ##
   ## Accepted values: 0.25, 0.5, 0.75, 1.0, 1.25, 1.5, 1.75, 2.0
   ## Default: 1.0
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -55,7 +55,11 @@ db:
 ##
 ## Port to listen on for incoming connections.
 ##
-## Accepted values: 1024-65535
+## Note: Ports lower than 1024 requires either root privileges
+## (not recommended) or the "CAP_NET_BIND_SERVICE" capability
+## (See https://stackoverflow.com/a/414258 and `man capabilities`)
+##
+## Accepted values: 1-65535
 ## Default: 3000
 ##
 #port: 3000
@@ -147,8 +151,8 @@ https_only: false
 ## https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-31
 ##
 ## Note: you should try both options and see what is the best for your
-## instance. In general QUIC is recommended for public instances, and
-## not recommended for private one. Your mileage may vary.
+## instance. In general QUIC is recommended for public instances. Your
+## mileage may vary.
 ##
 ## Note 2: Using QUIC prevents some captcha challenges from appearing.
 ## See: https://github.com/iv-org/invidious/issues/957#issuecomment-576424042
@@ -380,13 +384,19 @@ feed_threads: 1
 #banner:
 
 ##
-## Subscribe to channels using PubSubHub.
+## Subscribe to channels using PubSubHub (Google PubSubHubbub service).
+## PubSubHub allows Invidious to be instantly notified when a new video
+## is published on any subscribed channels. When PubSubHub is not used,
+## Invidious will check for new videos every minute.
 ##
-## Notes:
+## Note: This setting is recommended for public instances.
+##
+## Note 2:
+##  - Requires a public instance (it uses /feed/webhook/v1)
 ##  - Requires 'domain' and 'hmac_key' to be set.
-##  - Setting this parameter to any number greater than zer will
-##    enable channel subscriptions via PubSubHub, but will limit
-##    the amount of concurrent subscriptions.
+##  - Setting this parameter to any number greater than zero will
+##    enable channel subscriptions via PubSubHub, but will limit the
+##    amount of concurrent subscriptions.
 ##
 ## Accepted values: true, false, a positive integer
 ## Default: false
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -95,7 +95,7 @@ domain:
 ## Tell Invidious that it is behind a proxy that provides only
 ## HTTPS, so all links must use the https:// scheme. This
 ## setting MUST be set to true if invidious is behind a
-## reverse proxy.
+## reverse proxy serving HTTPs.
 ##
 ## Accepted values: true, false
 ## Default: false
@@ -206,7 +206,8 @@ https_only: false
 ##     displayed in the footer, e.g: "2021.05.13-75e5b49" / "master")
 ##   - The value of the 'registration_enabled' config (true/false)
 ##   - Number of currently registered users
-##   - Number of users who registered in the last month / last 6 months
+##   - Number of registered users who connected in the last month
+##   - Number of registered users who connected in the last 6 months
 ##   - Timestamp of the last server restart
 ##   - Timestamp of the last "Channel Refresh" job execution
 ##
@@ -256,6 +257,19 @@ https_only: false
 
 ##
 ## List of usernames that will be granted administrator rights.
+## A user with administrator rights will be able to change the
+## server configuration options listed below in /preferences,
+## in addition to the usual user preferences.
+##
+## Server-wide settings:
+##   - popular_enabled
+##   - captcha_enabled
+##   - login_enabled
+##   - registration_enabled
+##   - statistics_enabled
+## Default user preferences:
+##   - default_home
+##   - feed_menu
 ##
 ## Accepted values: an array of strings
 ## Default: [""]
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -5,8 +5,8 @@
 #########################################
 
 ##
-## Database configuration with separate parameters. This
-## setting is mandatory, unless database_url is used.
+## Database configuration with separate parameters.
+## This setting is MANDATORY, unless 'database_url' is used.
 ##
 db:
   user: kemal
@@ -18,8 +18,8 @@ db:
 ##
 ## Database configuration using a single URI. This is an
 ## alternative to the 'db' parameter above. If both forms
-## are used, then only database_url is used. Mandatory if
-## 'db' is not set.
+## are used, then only database_url is used.
+## This setting is MANDATORY if 'db' is not set.
 ##
 ## Note: The 'database_url' setting allows the use of UNIX
 ## sockets. To do so, remove the IP address (or FQDN) and port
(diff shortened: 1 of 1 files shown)


config/config.example.yml
```diff
@@ -1,13 +1,745 @@
...
+##
+#popular_enabled: true
+
+##
+## Enable/Disable statstics.
+##
+## Accepted values: true, false
+## Default: false
+##
+#statistics_enabled: false
```
Perflyst, line 211: From https://github.com/iv-org/invidious/pull/1641#discussion_r553423403

More information should be added to explain what this shows, along with mentioning that it's useful to have for instances.invidio.us (and should be therefore enabled for public instances).

SamantazFox: Done.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+##
+#popular_enabled: true
+
+##
+## Enable/Disable statstics (available at /api/v1/stats).
+## The following data is available:
+##   - Software name ("invidious") and version+branch (same data as
+##     displayed in the footer, e.g: "2021.05.13-75e5b49" / "master")
+##   - The value of the 'registration_enabled' config (true/false)
+##   - Number of currently connected users
```
TheFrenchGhosty, line 211: Registered, not connected.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+##   - Average Number of connected users (last month & last 6 months)
```
TheFrenchGhosty, line 212: Registered, not connected.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
-# alternatively, the database URL can be provided directly - if both are set then the latter takes precedence
-# database_url: postgres://kemal:kemal@localhost:5432/invidious
-full_refresh: false
-https_only: false
+
+##
+## Database configuration using a single URI. This is an
+## alternative to the 'db' parameter above. If both forms
+## are used, then only database_url is used.
+## This setting is MANDATORY if 'db' is not set.
```
TheFrenchGhosty, line 22: "This setting is MANDATORY if 'db' is commented out"*

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+##
+## When the invidious instance is behind a proxy, and the proxy
+## listens on a different port than the instance does, this lets
+## invidious know about it. This is used to craft absolute URLs
+## to the instance (e.g in the API).
+##
+## Accepted values: 1-65535
+## Default: <none>
+##
+#external_port:
```
TheFrenchGhosty, line 72: Mandatory if the instance is reverse proxied

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+#host_binding: 0.0.0.0
+
+##
+## Domain name under which this instance is hosted. This is
+## used to craft absolute URLs to the instance (e.g in the API).
+##
+## Accepted values: a fully qualified domain name (FQDN)
+## Default: <none>
+##
 domain:
```
TheFrenchGhosty, line 89: Mandatory if the instance is served behind a domain name.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+
+##
+## Tell Invidious that it is behind a proxy that provides only
+## HTTPS, so all links must use the https:// scheme.
+##
+## Accepted values: true, false
+## Default: false
+##
+https_only: false
```
TheFrenchGhosty, line 98: Mandatory if the instance is served via HTTPS.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+## Size of the HTTP pool used to connect to youtube. Each
+## domain ('youtube.com', 'ytimg.com', ...) has its own pool.
+##
+## Accepted values: a positive integer
+## Default: 100
+##
+#pool_size: 100
+
+##
+## Enable/Disable the use of QUIC (HTTP/3) when connecting
```
TheFrenchGhosty, line 133: Should include a basic explanation of the point of using quic

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+## The following data is available:
+##   - Software name ("invidious") and version+branch (same data as
+##     displayed in the footer, e.g: "2021.05.13-75e5b49" / "master")
+##   - The value of the 'registration_enabled' config (true/false)
+##   - Number of currently connected users
+##   - Average Number of connected users (last month & last 6 months)
+##   - Timestamp of the last server restart
+##   - Timestamp of the last "Channel Refresh" job execution
+##
+## Warning: This setting MUST be set to true if you plan to run
```
TheFrenchGhosty, line 216: It currently mustn't, but I think we should make it a prerequisite

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+##
+#statistics_enabled: false
+
+
+# -----------------------------
+#  Users and accounts
+# -----------------------------
+
+##
+## Allow/Forbid Invidious (local) account creation. Invidious
```
TheFrenchGhosty, line 231: Why "local"? It forbids all accounts creation.

\+ 

Accounts, not account

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+## Note: this is a basic captcha challenge that doesn't
+## depend on any third parties.
+##
+## Accepted values: true, false
+## Default: true
+##
+#captcha_enabled: true
+
+##
+## List of usernames that will be granted administrator rights.
```
TheFrenchGhosty, line 261: Should explain what "administrator rights" allows (I personally don't know)

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+##  - This setting is overriden if "-f THREADS" or
+##    "--feed-threads=THREADS" are passed on the command line.
+##
+## Accepted values: a positive integer
+## Default: 1
+##
+feed_threads: 1
+
+##
+## Enable/Disable the polling job that keeps the decryption
```
TheFrenchGhosty, line 310: Should explain what this mean, and what setting it to false does.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+## Subscribe to channels using PubSubHub.
+##
+## Notes:
+##  - Requires 'domain' and 'hmac_key' to be set.
+##  - Setting this parameter to any number greater than zer will
+##    enable channel subscriptions via PubSubHub, but will limit
+##    the amount of concurrent subscriptions.
+##
+## Accepted values: true, false, a positive integer
+## Default: false
```
TheFrenchGhosty, line 363: I think it's true by default.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+  ##
+  ## Enable/Disable dark mode.
+  ##
+  ## Accepted values: true, false
+  ## Default: <none>
+  ##
+  #dark_mode:
+
+  ##
+  ## Enable/Disable thin mode (no video thumbnails).
```
TheFrenchGhosty, line 512: It's more than no video thumbnails. There was an issue where omarroth explained it.

It's a bad feature and should be removed anyway.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+  ##
+  ## Accepted values: true, false
+  ## Default: false
+  ##
+  #thin_mode: false
+
+  ##
+  ## List of feeds available on the home page.
```
TheFrenchGhosty, line 520: "feeds" is a bad way to call them

TheFrenchGhosty (changes requested):

TheFrenchGhosty:
I saw: "```dash_quality``` should be added." from https://github.com/iv-org/invidious/pull/1641#discussion_r553423700

and it's indeed missing here, meaning that potentially more options are missing.

config/config.example.yml
SamantazFox: So it isn't an average either, right?

config/config.example.yml
SamantazFox: Wouldn't it be better with a form like this (to be coherent between the two)? 
https://github.com/SamantazFox/invidious/blob/patch-1/config/config.example.yml#L9

config/config.example.yml
TheFrenchGhosty: I just checked, and yes, it's the exact number.

config/config.example.yml
TheFrenchGhosty: Yes, this is even better.

config/config.example.yml
SamantazFox: So, should I change it?

config/config.example.yml
SamantazFox: local, as in accounts that belong to this instance (AFAIK, those aren't shared across instances).

SamantazFox: and "account" is singular because it's a general form.

config/config.example.yml
TheFrenchGhosty: Nope, I'll change the instance list document to include it as a prerequisite.

Great "idea" by the way (if that was supposed to be an idea xD), thanks.

config/config.example.yml
TheFrenchGhosty: >local, as in accounts that belong to this instance (AFAIK, those aren't shared across instances).

It assume that it doesn't forbid the google accounts feature

>and "account" is singular because it's a general form.

I feel like "accounts" is more correct, however I'm not a native english speaker, so I'm maybe wrong.

config/config.example.yml
```diff
@@ -1,13 +1,757 @@
...
+## Accepted values: a filesystem path or 'STDOUT'
+## Default: STDOUT
+##
+#output: STDOUT
+
+##
+## Logging Verbosity. This is overriden if "-l LEVEL" or
+## "--log-level=LEVEL" are passed on the command line.
+##
+## Accepted values:
```
unixfox, line 179: Accepted values are somewhat incorrect, it's those instead:
```
##   All    (Log everything)
##   Trace
##   Debug
##   Info
##   Warn
##   Error
##   Fatal
##   Off    (Disable log output)
```

unixfox (changes requested):

config/config.example.yml
SamantazFox: Ah, yeah, I misunderstood the code!

SamantazFox:
> I saw: "`dash_quality` should be added." from [#1641 (comment)](https://github.com/iv-org/invidious/pull/1641#discussion_r553423700)
> 
> `dash_quality` was never documented, it's indeed missing here, meaning that potentially more options are missing.

There is no such option as `dash_quality`. There is `quality_dash`, which is documented:
https://github.com/iv-org/invidious/blob/master/src/invidious/helpers/helpers.cr#L25-L63

config/config.example.yml
SamantazFox: I took the terms currently used by invidious. What should we name that?

config/config.example.yml
SamantazFox: It's false ^^ https://github.com/iv-org/invidious/blob/master/src/invidious/helpers/helpers.cr#L25-L63

TheFrenchGhosty:
@SamantazFox You're right, my bad, it's indeed quality_dash, I typed the wrong thing back then.

I'm indeed using quality_dash on my instance: https://github.com/PussTheCat-org/PussTheCat.org-Configs/blob/master/Services/Invidious/config.yml#L29

config/config.example.yml
SamantazFox: You said on Matrix that it was better with stats on, so I took that into account ^^

config/config.example.yml
```diff
@@ -208,8 +205,8 @@ https_only: false
 ##   - Software name ("invidious") and version+branch (same data as
 ##     displayed in the footer, e.g: "2021.05.13-75e5b49" / "master")
 ##   - The value of the 'registration_enabled' config (true/false)
-##   - Number of currently connected users
-##   - Average Number of connected users (last month & last 6 months)
+##   - Number of currently registered users
+##   - Number of users who registered in the last month / last 6 months
```
TheFrenchGhosty, line 209: It's the number of registered users who connected in the last month/6 months.

config/config.example.yml
SamantazFox: > It assume that it doesn't forbid the google accounts feature

It doesn't, yeah. Registration is only for invidious accounts. You can still connect to Google.

> I feel like "accounts" is more correct, however I'm not a native english speaker, so I'm maybe wrong.

It's the same in french: "Activer/Désactiver la création de compte utilisateur" <- singular

config/config.example.yml
SamantazFox: Yep! I'll have to read the code a bit, tho, because I don't know either x)

config/config.example.yml
SamantazFox: Uhhh, it's broken at the moment, so maybe we should see that later? ^^>

config/config.example.yml
SamantazFox: Heck xD

config/config.example.yml
TheFrenchGhosty: >It doesn't, yeah. Registration is only for invidious accounts. You can still connect to Google.

Oh... this isn't a good thing.

>It's the same in french: "Activer/Désactiver la création de compte utilisateur" <- singular

I guess...

config/config.example.yml
TheFrenchGhosty: That's a good question... I have no idea.

config/config.example.yml
SamantazFox: > Oh... this isn't a good thing.

Why? `login_enabled: false` forbids all types of login (Google/Invidious), and `enable_registration: false` disables all types of registration (i.e Invidious, because it's the only one available).

config/config.example.yml
syeopite: They're called "feeds" by Youtube as well.

SamantazFox: So, should we keep the naming like it is actually?

config/config.example.yml
SamantazFox: What sould I do? Keep it like that until we remove the code, remove that from the example config, or complete the description anyway?

config/config.example.yml
TheFrenchGhosty: Yeah I guess...

config/config.example.yml
TheFrenchGhosty: Disabling registration should disable all registration of new users, the problem is that now, if I understand what your saying, people can still register with a Google account. Meaning that the only way to fully disable registration, is to disable login... meaning existing user aren't able to login. It's a problem.

config/config.example.yml
TheFrenchGhosty: Remove it from the example config, no one should be using it.

config/config.example.yml
```diff
@@ -137,6 +139,16 @@ https_only: false
...
 ## Enable/Disable the use of QUIC (HTTP/3) when connecting
 ## to the youtube API and websites ('youtube.com', 'ytimg.com').
+## QUIC's main advantages are its lower latency and lower bandwidth
+## use, compared to its predecessors. However, the current version
+## of QUIC used in invidious is still based on the IETF draft 31,
+## meaning that the underlying library may still not be fully
+## optimized. You can read more about QUIC at the link below:
+## https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-31
+##
+## Note: you should try both options and see what is the best for
```
TheFrenchGhosty, line 149: `## Note: you should try both options and see what is the best for your instance, but in general QUIC is recommended on public instances, and not recommended for private one. Your mileage may vary.`

TheFrenchGhosty (changes requested):

config/config.example.yml
SamantazFox: > Disabling registration should disable all registration of new users, the problem is that now, if I understand what your saying, people can still register with a Google account. Meaning that the only way to fully disable registration, is to disable login... meaning existing user aren't able to login. It's a problem.

Ah, I should have been more explicit about it: you can't register a Google account from invidious itself. You have to do so on Google's website. So we have no impact on it from config ^^

config/config.example.yml
SamantazFox: Sure!

config/config.example.yml
TheFrenchGhosty: >you can't register a Google account from invidious itself. You have to do so on Google's website. So we have no impact on it from config

Of course, but the data of an user connecting via google is still stored on the server. 

Unless the google login just "brings" the user information to invidious and an actual registration is required?

config/config.example.yml
SamantazFox: I've always had QuIC enabled on my private instance. It doesn't seem to be an issue...

TheFrenchGhosty: >It doesn't seem to be an issue...

From what I know it's doesn't cause any issue, but I heard that it's where the memory leaks come from, so avoid it is preferable if it's possible. As I said, I haven't done any testing.

unixfox: Disabling QUIC tend to reduce the frequency at which you are going to restart the instance, but it does not solve the memory or crashing issues.
In the end you will just have to restart Invidious less frequently during the day.

config/config.example.yml
TheFrenchGhosty: I guess mark it as currently broken or remove it then.

TheFrenchGhosty (changes requested):

config/config.example.yml
saltycrys: Thin mode is about removing images in general, like video thumbnails or user avatars on comments, stuff like that. The rationale is to offer a version of Invidious that uses little bandwith, see https://github.com/iv-org/invidious/issues/25.

It is a perfectly fine feature that people might want to use, please readd it.

SamantazFox: @saltycrys yeah, we discussed about it on matrix and it's planned that I should revert that commit. I just haven't got the time to work on that ^^

config/config.example.yml
SamantazFox: Should I keep the "Some maintainers reported that using QUIC may reduce the occurence of captcha challenges requested by youtube.", part?

config/config.example.yml
SamantazFox: > Of course, but the data of an user connecting via google is still stored on the server.

Ah, yes.

config/config.example.yml
TheFrenchGhosty: @SamantazFox If it's true... sure. I have no idea if this is though (ping @unixfox )

unixfox: > Should I keep the "Some maintainers reported that using QUIC may reduce the occurence of captcha challenges requested by youtube.", part?

There are two CAPTCHA like explained here: https://github.com/iv-org/invidious/issues/957#issuecomment-576424042. Invidious just bypass the first one when QUIC is being used, it doesn't reduce the occurrence of CAPTCHA challenges for the second CAPTCHA.

SamantazFox: ok. I'll reword the thing and add a link to that specific message.

TheFrenchGhosty (approved):

config/config.example.yml
```diff
@@ -1,13 +1,815 @@
...
+## to the youtube API and websites ('youtube.com', 'ytimg.com').
+## QUIC's main advantages are its lower latency and lower bandwidth
+## use, compared to its predecessors. However, the current version
+## of QUIC used in invidious is still based on the IETF draft 31,
+## meaning that the underlying library may still not be fully
+## optimized. You can read more about QUIC at the link below:
+## https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-31
+##
+## Note: you should try both options and see what is the best for your
+## instance. In general QUIC is recommended for public instances, and
```
unixfox, line 150: I would just say instead: `In general QUIC is recommended for public instances.` and that's it. QUIC is fine for private instances, I don't see any harm... Invidious is still buggy when not using QUIC.

config/config.example.yml
```diff
@@ -1,13 +1,815 @@
...
+## Default: <none>
+##
+#banner:
+
+##
+## Subscribe to channels using PubSubHub.
+##
+## Notes:
+##  - Requires 'domain' and 'hmac_key' to be set.
+##  - Setting this parameter to any number greater than zer will
```
unixfox, line 387: typo with `zer`?

config/config.example.yml
```diff
@@ -1,13 +1,815 @@
...
+## Accepted values: any string. HTML is accepted.
+## Default: <none>
+##
+#banner:
+
+##
+## Subscribe to channels using PubSubHub.
+##
+## Notes:
+##  - Requires 'domain' and 'hmac_key' to be set.
```
unixfox, line 386: It requires the instance to be publicly accessible. For example, if I set `domain` to `example.org` then `example.org` need to serve Invidious.
Also, `external_port` is required too because Invidious build the URL from the domain and the external port.

config/config.example.yml
```diff
@@ -1,13 +1,815 @@
...
+## custom banner displayed at the top of every page. This can
+## used for instance announcements, e.g.
+##
+## Accepted values: any string. HTML is accepted.
+## Default: <none>
+##
+#banner:
+
+##
+## Subscribe to channels using PubSubHub.
```
unixfox, line 383:
```
## Subscribe to channels using PubSubHub on the Google PubSubHubbub service.
## PubSubHub allow Invidious to be instantly notified of new videos of subscribed channels instead of waiting for Invidious to check for new videos every 1 minute.
```
You may reword the sentences, though.

unixfox (changes requested):

config/config.example.yml
SamantazFox: Ah, yep, should be `zero`

config/config.example.yml
SamantazFox: external port is only required if the proxy port is not 80 nor 443.

config/config.example.yml
```diff
@@ -1,13 +1,815 @@
...
+#########################################
+
+# -----------------------------
+#  Network (inbound)
+# -----------------------------
+
+##
+## Port to listen on for incoming connections.
+##
+## Accepted values: 1024-65535
```
unixfox, line 58: If ran as root or with `setcap 'cap_net_bind_service=+ep' /path/to/program` (see https://stackoverflow.com/questions/413807/is-there-a-way-for-non-root-processes-to-bind-to-privileged-ports-on-linux), then any port under 1024 are accepted too.

unixfox (changes requested):

unixfox (approved):

PR #2102, opened 2021-05-20
Spelling: HD, 3D, HDR

comradekingu:


SamantazFox:
This is not required. those are the internal value, they are mapped to "real" text string when using `translate()` (line 61).

If you still want to change the internal values (a.k.a translation keys) make sur to also edit every single locale file!

SamantazFox:
Closing because that kind of issues will be entirely solved when switching to GNU Gettext files (See #2285)

PR #2108, opened 2021-05-21
fix comments not loading

### LINKED ISSUE #2092: [Bug] Comments are broken
**Describe the bug**
Comment section suddenly will not load. It will only show a rotating arrow icon for a second or two (or not at all) but will not load the comments themselves. I have tried refreshing and going to all instances listed to no avail. Chrome has not updated at all, the bug came out of the blue and has lasted for over a day.

**Steps to Reproduce**
1) Click on any video with enabled comments.
2) Scroll down to comment section.

**Logs**
N/A

**Screenshots**
Screenshot 2021-05-18 7 28 25 PM
Screenshot 2021-05-18 7 28 31 PM

**Additional context**
 - Browser (if applicable): Chrome
 - OS (if applicable): ChromeOS

B0pol:
(using a workaround: like count is always 0)

I did not use "" like https://github.com/iv-org/invidious/issues/2092#issuecomment-845753246 used, because using a string may break apps using the API until the real fix comes, whereas using an integer will provide a false value but won't break anything

fixes #2092 ,
I'll open a new issue for like count = 0 when this PR is merged

### DIFF
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -185,7 +185,7 @@ def fetch_youtube_comments(id, db, cursor, format, locale, thin_mode, region, so
               json.field "published", published.to_unix
               json.field "publishedText", translate(locale, "`x` ago", recode_date(published, locale))
 
-              json.field "likeCount", node_comment["likeCount"]
+              json.field "likeCount", 0
               json.field "commentId", node_comment["commentId"]
               json.field "authorIsChannelOwner", node_comment["authorIsChannelOwner"]
 


TheFrenchGhosty:
Discussed in Matrix, having dirty workaround in the code is a bad idea, it's better to wait for a proper fix.

Instance owners are free to use this PR themselves if they really want comments and can't wait for the proper fix.

TheFrenchGhosty:
Superseded by https://github.com/iv-org/invidious/pull/2110

PR #2110, opened 2021-05-21
Fix the parsing of likes in the comments

### LINKED ISSUE #2092: [Bug] Comments are broken

unixfox:
Fixes #2092

Based on the idea of @SamantazFox, I parse the label of the `accessibilityData` object in order to retrieve the number of likes:
```json
"accessibilityData":{
    "accessibilityData":{
        "label":"Like this comment along with 8,510 other people"
    }
}
```

I'm using a regex to find every digit in the string then combining every match, in this case `8` and `510` into one string (this is needed because [the regex function doesn't support the global match](https://github.com/crystal-lang/crystal/issues/2223)).
This is in my opinion pretty reliable because I don't see YouTube later adding numbers into the sentence.

I could add a fallback in case this break and instead parse the data from `voteCount`:
```json
"voteCount":{
    "accessibility":{
        "accessibilityData":{
        "label":"8.5K likes"
        }
    },
    "simpleText":"8.5K"
},
```
This would be acceptable for the HTML results but not for results from the API in JSON. So that's why I don't really see the point of doing that.

### DIFF
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -185,12 +185,14 @@ def fetch_youtube_comments(id, db, cursor, format, locale, thin_mode, region, so
               json.field "published", published.to_unix
               json.field "publishedText", translate(locale, "`x` ago", recode_date(published, locale))
 
-              json.field "likeCount", node_comment["likeCount"]
+              comment_action_buttons_renderer = node_comment["actionButtons"]["commentActionButtonsRenderer"]
+
+              json.field "likeCount", comment_action_buttons_renderer["likeButton"]["toggleButtonRenderer"]["accessibilityData"]["accessibilityData"]["label"].as_s.scan(/\d/).map(&.[0]).join.to_i
               json.field "commentId", node_comment["commentId"]
               json.field "authorIsChannelOwner", node_comment["authorIsChannelOwner"]
 
-              if node_comment["actionButtons"]["commentActionButtonsRenderer"]["creatorHeart"]?
-                hearth_data = node_comment["actionButtons"]["commentActionButtonsRenderer"]["creatorHeart"]["creatorHeartRenderer"]["creatorThumbnail"]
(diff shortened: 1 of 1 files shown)


RouterCFW:
I fuckin love you for fixing this

SamantazFox:
I could recommend to create an intermediate variable for `node_comment["actionButtons"]["commentActionButtonsRenderer"]` in the meantime, so you can also clean up a bit the code below :)

unixfox:
> I could recommend to create an intermediate variable for `node_comment["actionButtons"]["commentActionButtonsRenderer"]` in the meantime, so you can also clean up a bit the code below :)

Good idea, I modified the code by following your advice.

SamantazFox (approved):

TheFrenchGhosty:
Merged.

@unixfox to receive your reward either send your BTC address in a comment (preferred), or contact us directly, with a proof that this is indeed you.

TheFrenchGhosty:
A 20$ bounty has been rewarded to @unixfox for this PR.

Transaction (b3e617210624ed398afa999f7bb8ecdb23bc470f720ee5c6ad41c2de7edab367):

Amount sent: 0.00027 BTC (~8.29€ or ~10$)

Mining fees (40.9 sat/byte): 0.00005961 BTC (~1.83€)

PR #2111, opened 2021-05-22
Add the ability to listen on UNIX sockets

unixfox:
Kemal allows native ability to listen on UNIX sockets: https://kemalcr.com/cookbook/unix_domain_socket/. This pull request add the ability to configure the UNIX sockets on Invidious.

On some environments, especially Docker, this can improve the requests throughput up to 40%: https://blog.myhro.info/2017/01/benchmarking-ip-and-unix-domain-sockets-for-real! A very small Invidious instance won't notice this gain of performance but on a high demand instance with a lot of requests this can improve the latency and the amount of requests that an instance can serve.

On a bonus note: This improves the privacy because UNIX sockets are technically not forwarded on the network, so tools like tcpdump won't work in order to silently log any HTTP requests between invidious and the reverse proxy.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -3917,6 +3917,15 @@ add_context_storage_type(Preferences)
 add_context_storage_type(User)
 
 Kemal.config.logger = LOGGER
-Kemal.config.host_binding = Kemal.config.host_binding != "0.0.0.0" ? Kemal.config.host_binding : CONFIG.host_binding
-Kemal.config.port = Kemal.config.port != 3000 ? Kemal.config.port : CONFIG.port
-Kemal.run
+
+Kemal.run do |config|
+  if CONFIG.bind_unix
+    if File.exists?(CONFIG.bind_unix.not_nil!)
+      File.delete(CONFIG.bind_unix.not_nil!)
+    end
+    config.server.not_nil!.bind_unix CONFIG.bind_unix.not_nil!
+  else
+    config.host_binding = config.host_binding != "0.0.0.0" ? config.host_binding : CONFIG.host_binding
+    config.port = config.port != 3000 ? config.port : CONFIG.port
+  end
+end
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -98,6 +98,7 @@ class Config
   property force_resolve : Socket::Family = Socket::Family::UNSPEC # Connect to YouTube over 'ipv6', 'ipv4'. Will sometimes resolve fix issues with rate-limiting (see https://github.com/ytdl-org/youtube-dl/issues/21729)
   property port : Int32 = 3000                                     # Port to listen for connections (overrided by command line argument)
   property host_binding : String = "0.0.0.0"                       # Host to bind (overrided by command line argument)
+  property bind_unix : String? = nil                               # Make Invidious listening on UNIX sockets - Example: /tmp/invidious.sock
   property pool_size : Int32 = 100                                 # Pool size for HTTP requests to youtube.com and ytimg.com (each domain has a separate pool of `pool_size`)
   property use_quic : Bool = true                                  # Use quic transport for youtube api
 


SamantazFox:
To avoid adding another config option, that would require more documentation and confuse people, wouldn't it be better to rename `host_binding` to something like `listen` (similar to many other programs) and allow both IPv4/IPv6 and unix socket to be specified here?

FireMasterK:
I feel like properly documenting this would be a better idea instead.

I feel this way because IP stacks have a port in addition, which Unix sockets do not have.

SamantazFox:
> I feel like properly documenting this would be a better idea instead.
> 
> I feel this way because IP stacks have a port in addition, which Unix sockets do not have.

Yeah, but having a lot of mutually exclusive options not going to make documentation easier (Trust me, I'm behind #2083). The postgreSQL config already has two forms, and that's not intuitive (most projects I know, if not all, only have one form or the orther, but not both). Having more cases like that is going to make things worse.

unixfox:
> To avoid adding another config option, that would require more documentation and confuse people, wouldn't it be better to rename `host_binding` to something like `listen` (similar to many other programs) and allow both IPv4/IPv6 and unix socket to be specified here?

I get what you are trying to say, allow the ability to set both an IP address with a port or a unix socket in a same YAML field called `listen`. Am I right? If so, the issue is that I don't really want to do the fun part of detecting if it's an IP address or path that is passed to the field.

UNIX sockets are already a niche use case and for most of the people their setup will still work as expected. If one wants to configure UNIX socket, I think he will be tech-savvy enough to understand the differences between `host_binding` and `bind_unix`. I do consider that the field `bind_unix` is maybe not a correct name and should be renamed to a more appealing field.

SamantazFox:
> I get what you are trying to say, allow the ability to set both an IP address with a port or a unix socket in a same YAML field called `listen`. Am I right? If so, the issue is that I don't really want to do the fun part of detecting if it's an IP address or path that is passed to the field.

Simply that?
```
if CONFIG.listen.startsWith('/')
  # UNIX socket
else
  # Network interface
end
```

unixfox:
> > I get what you are trying to say, allow the ability to set both an IP address with a port or a unix socket in a same YAML field called `listen`. Am I right? If so, the issue is that I don't really want to do the fun part of detecting if it's an IP address or path that is passed to the field.
> 
> Simply that?
> 
> ```
> if CONFIG.listen.startsWith('/')
>   # UNIX socket
> else
>   # Network interface
> end
> ```

Won't work if you are on Windows, paths on Windows start with a letter of the drive.

SamantazFox:
> Won't work if you are on Windows, paths on Windows start with a letter of the drive.

Do Windows support unix sockets? And also, is invidious able to run natively on Windows (without a Linux docker container)?

unixfox:
> > Won't work if you are on Windows, paths on Windows start with a letter of the drive.
> 
> Do Windows support unix sockets? And also, is invidious able to run natively on Windows (without a Linux docker container)?

It's possible, but it doesn't really work the same way: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/. Just forget what I said.

Anyway, I really don't want to do the fun part of parsing the IP address from the port if we remove the "port" field too. Example: `0.0.0.0:3000` or `[::]:3000`.
And also we will need to do a backward compatibility with the existing fields "host_binding" and "port".

SamantazFox:
> Anyway, I really don't want to do the fun part of parsing the IP address from the port if we remove the "port" field too. Example: `0.0.0.0:3000` or `[::]:3000`.

Imo we should still keep the current port config, to avoid code clutter.

unixfox:
> > Anyway, I really don't want to do the fun part of parsing the IP address from the port if we remove the "port" field too. Example: `0.0.0.0:3000` or `[::]:3000`.
> 
> Imo we should still keep the current port config, to avoid code clutter.

That will make things even more confusing then, having the ability to set an IP address but not the port on a field just called `listen`... I think if we don't move forward and still keep the `port` field then I would prefer to add the ability to set a unix socket on the `host_binding` field instead then. At least we don't have to do backward compatibility.

I'm still wondering if `bind_unix` is not the optimal solution because in this case we don't break any potential weird configuration and no need for parsing the field in order to detect a path or an IP address.

FireMasterK:
> Do Windows support unix sockets?

They surprisingly do! Although, the included curl version doesn't support them.

I may have a solution of solving this:
```
if(value.endsWith(".sock")) {
     // bind to a socket here
} else {
      String ip = value.substring(0, value.lastIndexOf(":"));
      String  port = value.substring(value.lastIndexOf(":") + 1);
     // bind to IP here
}
````

idk how this can be done in Crystal tho

**Comment by unixfox** [comment]
> > Do Windows support unix sockets?
> 
> They surprisingly do! Although, the included curl version doesn't support them.
> 
> I may have a solution of solving this:
> 
> ```
> if(value.endsWith(".sock")) {
>      // bind to a socket here
> } else {
>       String ip = value.substring(0, value.lastIndexOf(":"));
>       String  port = value.substring(value.lastIndexOf(":") + 1);
>      // bind to IP here
> }
> ```
> 
> idk how this can be done in Crystal tho

But what if someone wants to name is file with something else than ending with `.sock`? That's not a proper solution.

**Comment by FireMasterK** [comment]
> But what if someone wants to name his file with something else than ending with .sock? That's not a proper solution.

```
if(value.contains(File.separator) || !value.contains(":")) { // this is os independant because File.separator is `\` on windows and `/` on linux
    // bind to a socket here
}
```

This should do the trick.

SamantazFox:
> ```
> if(value.contains(File.separator) || !value.contains(":")) { // this is os independant because File.separator is `\` on windows and `/` on linux
>     // bind to a socket here
> }
> ```
> 
> This should do the trick.

`value.contains(":")` will match IPv6 addresses, so that won't work.

FireMasterK:
> value.contains(":") will match IPv6 addresses (and !value.contains(":") will match IPv4 addresses), so that won't work.

`!value.contains(":")` does not matches both `0.0.0.0:3000` and `[::]:3000` because they both contain `:`?

I'm not sure what you meant here 😄

SamantazFox:
> `!value.contains(":")` does not matches both `0.0.0.0:3000` and `[::]:3000` because they both contain `:`?
> 
> I'm not sure what you meant here 😄

Oh, my bad, I stayed on the idea of having only IP addresses, without a port :X

syeopite (changes requested):
Looks good to me.  Just go ahead and merge bind_unix and host_binding like what everyone else wants here.

SamantazFox:
Same as @syeopite. What you can possibly do is check the input against the regexes below, in that order, and if none matches, then this is a unix socket.

Regex: `\[(?<ipv6>([0-9A-Fa-f]{0,4}:){2,}[0-9A-Fa-f]{0,4}\](?<port>:\d+)?`

=> matches IPV6 addresses, with or without a port number (`[::]`, `[::]:3000`, `[dead:beef::1]`, `[c0:ffee::42]:666`)

Regex: `(?<ipv4>(\d\.){3}\d)(?<port>:\d+)?`

=> matches IPV4 addresses, with or without a port number (`0.0.0.0`, `0.0.0.0:3000`, `127.0.0.1`, `10.42.0.2:666`)

unixfox:
bump

Caian:
I'm willing to continue the development of this feature because I found podman networking struggles a lot in rootless mode, especially when streaming videos. Sockets are a nice solution to get better performance if Invidious is behind a proxy.

My idea for the configuration is to keep `host_binding` (let's not break existing configurations), instead use `port=0` to switch between `host_binding` and `socket_binding=<path-to-socket>`.

This idea comes from [the way Nextcloud is configured to talk with Redis](https://help.nextcloud.com/t/redis-unix-socket/129153).

As for the name `socket_binding`, I think it keeps consistency with `host_binding` and, when using docker, sockets are pretty much available on Windows as well, so it's not a Unix thing anymore.

What do you think?

unixfox:
Good idea for the naming `socket_binding`.

About `port=0`, you don't need to do such thing. You can just check if `socket_binding` is set, if it is then use it exclusively over `host_binding`.

Caian:
> Good idea for the naming `socket_binding`.
> 
> About `port=0`, you don't need to do such thing. You can just check if `socket_binding` is set, if it is then use it exclusively over `host_binding`.

I agree. I was trying to go for a configuration that's ambiguity-free and wouldn't allow 2 valid configurations set at the same time, but as you mentioned in a previous post: users of this feature are probably tech-savvy enough to understand.

PR #2112, opened 2021-05-23
Replace lsquic.cr

syeopite:
Replaces lsquic.cr in Invidious with a interface to [invidious-quic-proxy](https://github.com/syeopite/Invidious-quic-proxy)

All of Invidious's features should be tested with this before merging. The dockerfile also needs to be changed for this but I'm not knowledgeable enough to do so.

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -16,10 +16,6 @@ shards:
     git: https://github.com/jeromegn/kilt.git
     version: 0.4.0
 
-  lsquic:
-    git: https://github.com/iv-org/lsquic.cr.git
-    version: 2.18.1-1
-
   pg:
     git: https://github.com/will/crystal-pg.git
     version: 0.23.1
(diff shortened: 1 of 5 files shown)


unixfox:
Benchmarks between vanilla Invidious and Invidious with quic proxy need to be done in order to understand if this new project bring any improvement or not.

syeopite:
Due to how the connection pool works in Invidious, the invidious-quic-proxy instance should be running before Invidious!

saltycrys (commented):
First and foremost: if this improves performance or stability then this should be used.

But I am against introducing a dependency on an entire language just because none of the Invidious maintainers can write proper LSQUIC Crystal bindings. Have you tried reaching out to the Crystal community? Surely some other Crystal developers want to use QUIC.

I have stated in the past that lsquic.cr sometimes fails to complete a request and stalls out fibers, causing subscriptions to not update. However thanks to the config option that disables QUIC I have found that this is not true, the problem happens regardless so it must be a problem with Crystal somehow.

PR #2085, opened 2021-05-14
Bring protodec to the Invidious organization

TheFrenchGhosty:
Omarroth seems to have "left" open source development completely and archived protodec: https://github.com/omarroth/protodec

Therefore, I forked it to the organization: https://github.com/iv-org/protodec and copied the release: https://github.com/iv-org/protodec/releases/tag/v0.1.3

This PR makes Invidious use this fork.

PR #2005, opened 2021-04-10
Update to crystal 1.0.0

tenpura-shrimp:
required PRs:
https://github.com/omarroth/protodec/pull/1
https://github.com/iv-org/lsquic.cr/pull/8

#1934 
#1933

PR #2115, opened 2021-05-24
Upgrade to crystal 1.0.0

syeopite:
This PR is based on #2005 with the cookie errors fixed. #2085 should be merged before this one as well.

The only dependency that needs to be bumped here is lsquic.cr but the status of #2112 would determine if this is necessary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install Crystal
         uses: oprypin/install-crystal@v1.2.4
         with:
-          crystal: 0.36.1
+          crystal: 1.0.0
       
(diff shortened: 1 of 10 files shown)


unixfox:
Can you rebase this PR?

unixfox:
I think you can switch to lsquic `v2.18.1-2` and this should be good to be merged.

unixfox:
> We want to decode `env.request.cookies["PREFS"]`, not the whole expression! That's why CI doesn't pass:
> 
> ```
> Preferences.from_json(
>   URI.decode_www_form(env.request.cookies["PREFS"])?
>     .try &.value
>   || "{}"
> )
> ```

Can you create a PR in order to fix that?

PR #2124, opened 2021-05-29
Fix storyboard when proxied with an external port

raycheung:
Say if it's `http://host:port` internally and proxied to
`https://domain:external_port`, the storyboard URL was rendered as
`https://domain:port`.


SamantazFox:
Nice catch :D

SamantazFox (approved):

Issue #811, opened 2019-10-09
 Video unavailable


simmstein:
Hello, 

**Versions**

* Invidious version:  0.19.1-e03b4b7 (sha256 of the compiled file 271291b36749ee7bae51c3834d08da8eb8bc7e4842525489c56a3399ef0b5d17)
* OS: Debian 10 (amd64)
* Crystal version: 0.31.1 [0e2e1d067] (2019-09-30)
* Database: Postgresql 9.6+181+deb9u2
* Logs: empty (no error)

**What happened**

I click on a video's thumbnail and the message "Video unavailable." is shown.

**Expected**

Play the video clicked.

**Information**

The video is played using invidio.us (0.19.1-e03b4b7 master).

simmstein:
From the server:

```
$ youtube-dl "https://www.youtube.com/watch?v=XXXXXXXXX"
[youtube] -OZ--RQcaY4: Downloading webpage
[youtube] -OZ--RQcaY4: Downloading video info webpage
WARNING: unable to download video info webpage: HTTP Error 429: Too Many Requests
WARNING: unable to download video info webpage: HTTP Error 429: Too Many Requests
WARNING: unable to download video info webpage: HTTP Error 429: Too Many Requests
WARNING: unable to download video info webpage: HTTP Error 429: Too Many Requests
WARNING: unable to download video info webpage: HTTP Error 429: Too Many Requests
Traceback (most recent call last):
  File "/usr/bin/youtube-dl", line 11, in <module>
    load_entry_point('youtube-dl==2019.1.17', 'console_scripts', 'youtube-dl')()
  File "/usr/lib/python3/dist-packages/youtube_dl/__init__.py", line 477, in main
    _real_main(argv)
  File "/usr/lib/python3/dist-packages/youtube_dl/__init__.py", line 467, in _real_main
    retcode = ydl.download(all_urls)
  File "/usr/lib/python3/dist-packages/youtube_dl/YoutubeDL.py", line 2002, in download
    url, force_generic_extractor=self.params.get('force_generic_extractor', False))
  File "/usr/lib/python3/dist-packages/youtube_dl/YoutubeDL.py", line 793, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python3/dist-packages/youtube_dl/extractor/common.py", line 508, in extract
    ie_result = self._real_extract(url)
  File "/usr/lib/python3/dist-packages/youtube_dl/extractor/youtube.py", line 1672, in _real_extract
[...]

PR #2129, opened 2021-06-05
Add workflow to lock inactive closed issues

syeopite:


TheFrenchGhosty:
Make it 30 days instead of 60. It's more than enough.

syeopite:
Done. Please let me know what else needs to be done before merging.

.github/workflows/lock.yml
```diff
@@ -0,0 +1,20 @@
...
+  schedule:
+    - cron: "0 */12 * * *"
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ github.token }}
```
SamantazFox, line 13: shouldn't that be `${{ secrets.GITHUB_TOKEN }}`, like in [here](https://github.com/iv-org/invidious/blob/master/.github/workflows/stale.yml#L15)?

syeopite: Apparently not. Accordingly to the documentation on lock-threads the value must be `${{ github.token }}`. Maybe I've just interpreted it wrong but [take a look at it here if you're interested](https://github.com/marketplace/actions/lock-threads)

SamantazFox: Ah, right, my bad ^^

.github/workflows/lock.yml
```diff
@@ -0,0 +1,20 @@
...
+          issue-lock-inactive-days: '30'
+          pr-lock-inactive-days: '30'
+          issue-lock-reason: 'resolved'
+          pr-lock-reason: 'resolved'
+
+          issue-lock-comment: 'This issue has been automatically locked since there has not been any recent activity after it was closed. Please open a new issue for related bugs.'
+          pr-lock-comment: 'This pull request has been automatically locked since there has not been any recent activity after it was closed. Please open a new pull request for related changes.'
```
TheFrenchGhosty, line 20: Better wording:

`This pull request has been automatically locked since there has not been any activity in it in the last 30 days. If you want to tell us about needed or wanted changes or if problem related to this code are discovered, feel free to open an issue or a new pull request.`

TheFrenchGhosty (changes requested):

TheFrenchGhosty:
After this is done feel free to merge it.

SamantazFox:
Okay, it works really well xD I got 120+ emails in a row!

Rjevski:
Has there been any prior discussion behind this change, or what's the reasoning for this? What's the advantage of doing this?

syeopite:
Yes. It was discussed briefly in the Matrix server. The main reason is to prevent people from commenting on ancient threads and notifying all of it's participants. 

Take #811 for example, a new comment on there would notify ~ 30 different people to a long dead thread that has been resolved an year ago.

However, the current situation isn't ideal either. While the back catalogue of issues/pulls are still being processed  it's going to be spamming the notifications. Not much we can do about that though.

Issue #1756, opened 2021-02-10
[Feature request] Monthly release


GauthierPLM:
**Is your feature request related to a problem? Please describe.**
The newest release of invidious is over a year old and master evolved a lot since.
For new users or people willing to follow Invidious' evolution without checking the commit history, having no release for such a long period of time is very confusing.

There is also no "reference" version to pull and install and people rely on the master branch that might have bugs introduced by a recent commit or pre-requites that are not clearly stated (for example, master requires Crystal 0.36 but I did not see any mention of it)

**Describe the solution you'd like**
A more regular release cycle would help people to keep track of the changes and upgrade their instance on a regular basis.
I'd suggest a monthly release cycle, with release named as YEAR.MONTH.BUGFIX.

The release would contain a generated change log with the list of PRs and issues fixed during the past month.

A week prior to the release, only fixes would be accepted in master to avoid introducing last-minute changes that might create issues and let instances based on master report issues and bug that would be high priority.

Version could also be a requirement to create new issues so the contributors can keep a better track of the issues and whether they were fixed or not since the release.

TheFrenchGhosty:
Currently this requires far too much management, and sadly, I (and I'm guessing @Perflyst ) don't have enough time for that.

I'm not against it, but we need (partial, or better, complete) automation for this.

Maybe in the future, or if someone know a Github actions that is smart enough, and is ready to open a PR for that... sure!

hackerncoder:
I've automated the creating of a branch, named YYYY.MM.01, on the 25th (was supposed to be 23, no idea why I used 25) then on the first of each month a tag and release is made.

My thought right now was to delete branches that are two months old, but if someone thinks this is wrong, do tell me. Or any other thoughs for that matter.

GauthierPLM:
The strategy seems good. If the tag references a commit that exists on master, it will be kept anyway, and release will have a copy of the source code, so nothing is lost even tho the branch has been deleted.

PR #2154, opened 2021-06-13
Automate monthly release

hackerncoder:
This is for #1756.

Every 25th of the month, a new branch is created, for feature freeze. Then on the first a new tag is made from the status at that time on that branch.

Fixes should be backported (right word?) from master to the next release in that week.


hackerncoder:
Marking as draft as I have two things I want to do: Make the tag better (probably show a list of commits) and figure out what to do with the branches, currently they'll just pile up.

TheFrenchGhosty:
Why branches and not just tags?

hackerncoder:
Feature freeze where fixes can be applied. As the original issue stated:

> A week prior to the release, only fixes would be accepted in master to avoid introducing last-minute changes that might create issues and let instances based on master report issues and bug that would be high priority.

I thought it should happen in a different branch. If you think that is wrong, do tell me.

syeopite (changes requested):
Please standardize the indents to two spaces like all of the other workflow files.

syeopite (approved):
Looks good to me. The whole branches vs tags things needs to be discussed however but for now, assuming we're sticking with branches, could you maybe add something to automatically delete feature freeze branches older than say 5 months? If we don't have something to automatically remove those ancient versions then the branch tree is going to get cluttered fast.

syeopite:
Also, could you limit the amount of force pushes you're doing? We should be able to reduce this PR into one commit via the merge and squash version on GitHub anyways.

hackerncoder:
> Looks good to me. The whole branches vs tags things needs to be discussed however but for now, assuming we're sticking with branches, could you maybe add something to automatically delete feature freeze branches older than say 5 months? If we don't have something to automatically remove those ancient versions then the branch tree is going to get cluttered fast.

Done.

> Also, could you limit the amount of force pushes you're doing? We should be able to reduce this PR into one commit via the merge and squash version on GitHub anyways.

Alright.

SamantazFox:
Quick remark: wouldn't bi-monthly be better? We're doing a lot of merges, especially recently, so "last month" release is going to be pretty old at some point.

hackerncoder:
And wouldn't bi-monthly be even more outdated?

TheFrenchGhosty (changes requested):
Okay, so I finally got around to reviewing this PR.

>Every 25th of the month

Why the 25th? Why not the 30th? Why not the 1st? Why an arbitrary "25th"?

>a new branch is created, for feature freeze. Then on the first a new tag is made from the status at that time on that branch.

... why have a branch on the 25th and a release on the 1st, what's the point, why not just have a release on the 1st? If I understand correctly you want us to maintain an outdated version with fixes... for 1 month? What's the point?

>Fixes should be backported (right word?) from master to the next release in that week.

We're not maintaining an outdated version.

>and figure out what to do with the branches, currently they'll just pile up.

Don't have branches, because they are completely useless.

>Feature freeze where fixes can be applied.

No, invidious won't have a branch that don't get features but just fixes. It's useless micromanagement for something completely useless.

And to add to that: Invidious works in a rolling release model, we didn't need releases, there isn't any point in it in my opinion... other than major version bump for breaking changes... but since it's currently using a date-based release name it's won't work.

TheFrenchGhosty:
@hackerncoder No need to close it, releases are a *maybe* (surely) a good idea (like I said for "major version bump for breaking changes") but it mustn't add more micromanagement, and from what I can see https://github.com/hackerncoder/invidious/releases/tag/2021.07.01 it works correctly, and can be used as a base for a "normal" version numbering.

PS: I know my review was a bit "agressive", I went to the point because I really don't understand the *why* most of it is needed, however please, don't dump all this work.

hackerncoder:
As it is currently, this PR is crap. If you want automated releases, I'll do it, but I don't think it should be based on a schedule (per your review).

> but since it's currently using a date-based release name it's won't work.

I'll also need to know how you want the release name to be.

----

Why: I implemented the issue #1756 + my own thoughts. 

> Why the 25th? Why not the 30th? Why not the 1st? Why an arbitrary "25th"?

"(was supposed to be 23, no idea why I used 25)" because "A week prior to the release, only fixes would be accepted in master to avoid introducing last-minute changes that might create issues and let instances based on master report issues and bug that would be high priority."

> for 1 month?

"in that week"

TheFrenchGhosty:
@hackerncoder 

For the release name I guess 1.XX, and if there's breaking change we bump to 2.XX (though a bump to 2.00 might look more than what it actually is, we need to discuss it) - for an example of software that does it that way (kinda) see [Komga](https://github.com/gotson/komga/releases)

>"(was supposed to be 23, no idea why I used 25)" because "A week prior to the release, only fixes would be accepted in master to avoid introducing last-minute changes that might create issues and let instances based on master report issues and bug that would be high priority."

Yeah, it really doesn't work for Invidious.

saltycrys:
Invidious currently uses date based versions: https://github.com/iv-org/invidious/blob/5a8825d01682def020acfd2baf95a44b94790f6f/src/invidious.cr#L55

SamantazFox:
bump

hackerncoder:
I am leaving the organization and cleaning up stuff.

PR #2167, opened 2021-06-17
Use alternate instance button vs watch on YouTube

RiversideRocks:
Encourages people to use the redirect.invidious.io site other than using YouTube when an instance is broken/ratelimited.


syeopite:
There's already a PR for this in #1940 :) though `Use Alternate Instance` sound so much nicer compared to what I choose lol

RiversideRocks:
Thanks for letting me know, I'll check that out.

PR #2174, opened 2021-06-18
Workflow improvements

syeopite:
This PR shortens the length before PRs are marked as stale (See commit for rationale) and also prevents CI from running when only files that doesn't affect Invidious or its workflows are changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,19 @@ on:
       - "api-only"
   pull_request:
     branches: "*"
+    paths-ignore:
+      - "*.md"
+      - LICENCE
+      - TRANSLATION
+      - invidious.service
+      - .git*
+      - .editorconfig
+
+      - screenshots/*
+      - assets/**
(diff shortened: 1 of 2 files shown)


SamantazFox:
can you also put `assets/` and `config/` in the exclude list?

syeopite:
Done.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,19 @@ on:
...
   pull_request:
     branches: "*"
+    paths-ignore:
+      - "*.md"
+      - LICENCE
+      - TRANSLATION
+      - invidious.service
+      - .git*
+      - .editorconfig
+
```
SamantazFox, line 17: Does Github support the extra space?

syeopite: Yep it does. I'm pretty sure GitHub supports all valid yaml.

TheFrenchGhosty (approved):

Issue #2131, opened 2021-06-05
How to check if anti-captcha works?


TROMsite:
I have used the instructions from here https://github.com/iv-org/documentation/blob/master/How-to-setup-Anti-Captcha.md but in the https://anti-captcha.com/clients/reports/stats/spendings I see no activity as if nothing is connected to anti-captcha.com. Is there an easy way to properly check if it works? Cheers!

Perflyst:
Is your instance actually blocked by google with a captcha? You can check with for example this curl

```
$ curl -s https://www.youtube.com/watch?v=QmyhcjpsF6E -4
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>302 Moved</TITLE></HEAD><BODY>
<H1>302 Moved</H1>
The document has moved
<A HREF="https://www.google.com/sorry/index?continue=https://www.youtube.com/watch%3Fv%3DQmyhcjpsF6E&amp;q=EgRf2BjmGJKaoO4FIhkA8aeDSwTc2PRqzI476Aa-2iHKmGCVxucTMgFy">here</A>.
</BODY></HTML>
```

TROMsite:
This is the output:
```
curl -s https://www.youtube.com/watch?v=QmyhcjpsF6E -4
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>302 Moved</TITLE></HEAD><BODY>
<H1>302 Moved</H1>
The document has moved
<A HREF="https://www.google.com/sorry/index?continue=https://www.youtube.com/watch%3Fv%3DQmyhcjpsF6E&amp;q=EgSnVkitGKWig4YGIhAeeym8f-YLWSqy0mwL2XevMgFy">here</A>.
</BODY></HTML>
```

unixfox:
So you are indeed blocked. If anti-captcha is misconfigured, you should see logs that start with `BypassCaptchaJob:`. Unless you turned off all the logs, if that's the case I advise you to set `log_level: Warn` in the config.yml.

Another way to check if anti-captcha doesn't work is if you see "instance blocked" messages appearing on some Invidious pages like on a video page.

TROMsite:
> Another way to check if anti-captcha doesn't work is if you see "instance blocked" messages appearing on some Invidious pages like on a video page.

I see that at times. So I do not understand what I am doing wrong. This is how my config looks like:
[...]

PR #2183, opened 2021-06-19
update video URL for recaptcha detection

unixfox:
It's a temporary fix for https://github.com/iv-org/invidious/issues/2131.


syeopite (approved):

TheFrenchGhosty:
I don't understand the issue, can you explain?

syeopite:
It seems like google is selectively allowing the video we use for testing if we should call anti-captcha to bypass it . This effectively means that anti captcha will never get called as Invidious thinks it isn't blocked.

Issue #1177, opened 2020-05-18
Add Bidi Support


ahangarha:
Hi

The current UI doesn't support bidirectional text. Therefore, if there be RTL text, it is shown in LTR direction which is problematic not only in terms of being aligned in wrong direction, but also being messed up if text be mixed with LTR characters. Look at the examples below:

Wrong direction:
![image](https://user-images.githubusercontent.com/11241315/82226395-b363be00-993b-11ea-9f61-28d84569ded2.png)

Right direction:
![image](https://user-images.githubusercontent.com/11241315/82226931-5ddbe100-993c-11ea-9075-a1852f41a6d3.png)

The change is made by adding `dir="auto"` to text containing elements such as headings, paragraphs and links. If needed, `text-align` is also set to `start`.

I have made a simple Firefox addons called [Add Bidi Support](https://addons.mozilla.org/en/firefox/addon/add-bidi-support/) which does the trick on all websites and so far, it works great on 99% of cases and in case of invidious, it works perfectly fine. You may check it yourself. There are plenty of screenshots showing the effects.

I suggest you apply the same method. I know where to apply but I am not sure if I can build the project to check the effect before sending PR.

I am willing to help in this regard.

ahangarha:
Resolved in #2196

PR #2196, opened 2021-06-24
Add bidi support

ahangarha:
Adding these styles is the easiest way to add bidi (bidirectional text) support without or with least side effect. I have tested it by adding it manually on yewtu.be instance and so far it works very well.

Related issue: #1177

---

Edit by @SamantazFox: putting a quick note in here, as requested by @unixfox, for people who don't know what "bidi" is.

Computers support two kinds of text flow:
 - "LTR": from left to right, like most western languages
 - "RTL": from right to left, like Arabic, Persian and Hebrew

"bidi support" means that a system supports text that goes in one direction, the other, or both (for instance, URLs, which are LTR, inside of a paragraph of RTL text).

### DIFF
--- a/src/invidious/views/channel.ecr
+++ b/src/invidious/views/channel.ecr
@@ -28,7 +28,9 @@
 </div>
 
 <div class="h-box">
-    <p><span style="white-space:pre-wrap"><%= channel.description_html %></span></p>
+    <div id="descriptionWrapper">
+        <p><span style="white-space:pre-wrap"><%= channel.description_html %></span></p>
+    </div>
 </div>
 
 <div class="h-box">
--- a/src/invidious/views/channel.ecr
+++ b/src/invidious/views/channel.ecr
@@ -21,7 +21,7 @@
         </div>
     </div>
     <div class="pure-u-1-3" style="text-align:right">
-        <h3>
+        <h3 style="text-align:right">
             <a href="/feed/channel/<%= channel.ucid %>"><i class="icon ion-logo-rss"></i></a>
         </h3>
     </div>
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -521,5 +521,4 @@ p,
 #descriptionWrapper,
 #description-box {
 	unicode-bidi: plaintext;
-	text-align: start;
 }
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -492,11 +492,6 @@ hr {
 }
 
 /* Description Expansion Styling*/
-#description-box {
-  display: flex;
-  flex-direction: column;
-}
-
 #descexpansionbutton {
   display: none
 }
@@ -511,7 +506,7 @@ hr {
   height: 100%;
 }
 
-#descexpansionbutton + label {
+#descexpansionbutton ~ label {
   order: 1;
   margin-top: 20px;
 }
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -521,4 +521,5 @@ p,
 #descriptionWrapper,
 #description-box {
 	unicode-bidi: plaintext;
+	text-align: start;
 }
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -493,8 +493,7 @@ hr {
 
 /* Description Expansion Styling*/
 #description-box {
-  display: flex;
-  flex-direction: column;
+  display: block;
 }
 
 #descexpansionbutton {
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -282,6 +282,21 @@ input[type="search"]::-webkit-search-cancel-button {
   }
 }
 
+
+/*
+ * Video "cards" (results/playlist/channel videos)
+ */
+
+.video-card-row { margin: 15px 0; }
+
+.flexible { display: flex; }
+.flex-left  { flex: 1 1 100%; flex-wrap: wrap;   }
+.flex-right { flex: 1 0 max-content; flex-wrap: nowrap; }
+
+p.channel-name { margin: 0; }
+p.video-data   { margin: 0; font-weight: bold; font-size: 80%; }
+
+
 /*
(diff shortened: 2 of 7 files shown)


ahangarha:
There is one issue and that is when description is enough short. As the result, the description text goes directly inside  `description-box` and then these styles cannot affect the content. There is a need to modify the source code and make sure description text is always inside a wrapper which can be referenced and styled properly.

cloudrac3r (approved):
Looks good to me!

ahangarha:
Let me see if I can find where to change the template. Locally I could wrap the description text in another div with id of `descriptionWrapper` and it worked well.

cloudrac3r (changes requested):
Actually, does not work when `description-box` is flex (which is all the time) and when the description text is inside `description-box` directly (some of the time, including this video: https://yewtu.be/watch?v=mRoqAz9Jepw).

I will try to change it to `block` and fix the resulting layout issues.

cloudrac3r:
I can't push to the PR branch, so here's a patch with my fix, which you can apply using `git am` :

```
From 507ad4dd8a5ef80cc92278b1be050d7b981a76be Mon Sep 17 00:00:00 2001
From: Cadence Ember <cadence@disroot.org>
Date: Thu, 24 Jun 2021 21:05:06 +1200
Subject: [PATCH] Change description-box from flex to block

I also make minor changes to the surroundings so that the same layout
and functionality as before is preserved.
---
 assets/css/default.css        |  7 +------
 src/invidious/views/watch.ecr | 10 +++++-----
 2 files changed, 6 insertions(+), 11 deletions(-)

diff --git a/assets/css/default.css b/assets/css/default.css
index de295501..808df295 100644
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -492,11 +492,6 @@ hr {
 }
 
 /* Description Expansion Styling*/
-#description-box {
-  display: flex;
-  flex-direction: column;
-}
-
 #descexpansionbutton {
   display: none
 }
@@ -511,7 +506,7 @@ hr {
   height: 100%;
 }
 
-#descexpansionbutton + label {
+#descexpansionbutton ~ label {
   order: 1;
   margin-top: 20px;
 }
diff --git a/src/invidious/views/watch.ecr b/src/invidious/views/watch.ecr
index 91e03725..8ea83384 100644
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -30,11 +30,11 @@
 we're going to need to do it here in order to allow for translations.
  -->
 <style>
-#descexpansionbutton + label > a::after {
+#descexpansionbutton ~ label > a::after {
     content: "<%= translate(locale, "Show more") %>"
 }
 
-#descexpansionbutton:checked + label > a::after {
+#descexpansionbutton:checked ~ label > a::after {
     content: "<%= translate(locale, "Show less") %>"
 }
 </style>
@@ -249,12 +249,12 @@ we're going to need to do it here in order to allow for translations.
                     <%= video.description_html %>
                 <% else %>
                     <input id="descexpansionbutton" type="checkbox"/>
-                    <label for="descexpansionbutton" style="order: 1;">
-                        <a></a>
-                    </label>
                     <div id="descriptionWrapper">
                         <%= video.description_html %>
                     </div>
+                    <label for="descexpansionbutton">
+                        <a></a>
+                    </label>
                 <% end %>
             </div>
 
-- 
2.25.1
```

cloudrac3r (approved):

SamantazFox (approved):

SamantazFox:
I've added a limit on the description width, so when LTR and RTL text coexist, they won't be far apart, especially in channel/playlists descriptions and large screen (behavior reproduced on a video description):

image

This fix is not perfect, but it will do the job while we work on an UI/UX rewrite to better support bidi text and accessibility.

[fix-description_width.patch.txt](https://github.com/iv-org/invidious/files/6709171/fix-description_width.patch.txt)

src/invidious/views/channel.ecr
```diff
@@ -21,7 +21,7 @@
         </div>
     </div>
     <div class="pure-u-1-3" style="text-align:right">
-        <h3>
+        <h3 style="text-align:right">
```
SamantazFox, line 24: you'll have to add that change to the other channel pages, too (`community.ecr` and `playlists.ecr`). And also remove the duplicate `style` on the `<div>` right above :)

ahangarha: I kept the alignment on div since I thought if there be more elements in it, it should be able to handle it properly.

I checked playlist. it didn't need since it doesn't use `h3`. There is inconsistency!

Cannot find when community view gets used.

ahangarha: Though I couldn't find community page, I added the style to relevant element there as well.

SamantazFox: > I checked playlist. it didn't need since it doesn't use `h3`. There is inconsistency!

It's `playlists.ecr`, with a `s` ! The file `playlist.ecr` shows a single playlist, where the file `playlists.ecr` show all the playlists that a channel has.

> Cannot find when community view gets used.

See here at the bottom? this is the "community" tab. This is also here that you find the channel playlists :)

image

SamantazFox: I know, this is not very obvious (both the files naming and the links position) but we're working on it ^^

SamantazFox:
@ahangarha I've added the few changes that seem to fix the alignment on all video "cards" (search results, playlists, channel videos, etc...). The whole changes in this PR are available for testing at https://test.invidious.io/ !

ahangarha:
So far everything looks fine

SamantazFox:
@ahangarha Thanks a lot for your PR and the great help on testing :)

ahangarha:
Thank you for accepting the PR and working hard to fix side effects.

I am happy I could have little share in improvement of this amazing project.

PR #2201, opened 2021-06-25
futureproof comment avatars

pinchese:
i was injecting custom css into the site that made the avatars round, and noticed comment avatars looked a little odd

i opened dev tools and siffed through the html, and noticed that the image was being padded,
when it would look nicer if the element used margin instead of padding

with padding:
image

with proposed changes (margin instead of padding):
image


syeopite (approved):

syeopite:
Thanks @pinchese

Issue #2193, opened 2021-06-23
[Security Issue] Channel names don't get sanitized for HTML


Davr1:
**Describe the bug**
Channel names don't get sanitized for Html. While I myself don't have any experience with XSS, I feel like this could be used for more than just injecting a picture of a big rat.

**Screenshots**
![firefox - 2021 23 06 - 17 50 46 - 1903x967](https://user-images.githubusercontent.com/42148912/123144840-6d412e00-d45c-11eb-9a53-aabef6d60beb.png)
(channel:UCNQLe71USVnxc7kjFyc-Jiw)

SamantazFox:
Indeed, that's quite a problem! Thanks for the report!

BlackBunnyHunter:
![afbeelding](https://user-images.githubusercontent.com/8104256/123274193-98328d00-d503-11eb-9453-ed92244bed72.png)
Big Chungus

TheFrenchGhosty:
A 20$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 20$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

TheFrenchGhosty:
@Davr1 I would like to deeply thank you a lot for reporting this problem!

For reporting this (massive) security issue I and most of the organization members would like to reward you with 20$ (in BTC) If you want to receive your reward feel free to send me your BTC address or post it directly in the PR.

Every payment will be, as usual, reflected in the [transparency repository](https://github.com/iv-org/transparency), in our monthly financial reports.

Davr1:
@TheFrenchGhosty `bc1qzunu09k7a74vh525ee0xz8y92u72yyyxzk6p55` this is my first time dealing with crypto, so I hope I don't mess up something lol

TheFrenchGhosty:
@Davr1 Sorry I got busy and completely forgot to send your reward.

Is the address still good?

syeopite:
Ping @Davr1

TheFrenchGhosty:
Ping @Davr1

Davr1:
@TheFrenchGhosty Sorry, I have also completely forgotten about this, and I didn't notice the github emails. Yes, the address is still good.

PR #2059, opened 2021-05-03
[WIP] Overhaul of Invidious UI

syeopite:
This PR is **far** from finished. It's pretty much only here for completion purposes and to make sure I don't mangle up the commit history too much.

## TODO
- [ ] Add all channel tabs (6/7)
- [ ] Overhaul mobile design 
- [ ] Trending localization #331
- [ ] Overhaul preference page
- [ ] Create special page for video game channels
- [ ] Fix livestream UI
- [ ] Download refactor #1563 
- [ ] Better video keybind support #974
- [ ] General improvements 
- [ ] CSS clean up
- [ ] Complete localization rewrite to use GNU Gettext
- [ ] Enhance UI for bidirectional text

And likely a lot more. These are just what I have currently planned.

## Closes
Closes #483
Closes #2032
Closes #1969 
Closes #1923
Closes #529
Closes #1950

This PR also uses #1985 so that'll have to get finished and merged before this one.

PR #2205, opened 2021-06-25
Partial (and temporary) fix for age restricted videos

### LINKED ISSUE #2189: [Bug] Age Restricted Videos do not play
Videos that have been marked age-restricted don't play
Any video with age-restriction doesn't work. All you can see is the thumbnail. I've tested this on all public instances.

1. Go to any age-restricted video
2. See error

Even the play button doesn't work
image

**Additional context**
 - Browser: Latest Windows version of Chrome
 - OS: Win10 x64

syeopite:
Fixes #2189

This is a temporarily fix for age restricted videos based on what many other projects such as yt-dlp have implemented. However, this only fixes a large portion of age restricted videos but not all of them. Furthermore, it seems like this fix also either exposes or creates issues regarding  captions on a subset of age gated videos. 

If this PR is still compatible with the bounty, I'll like to claim it. Please don't use my previous address though.

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -786,7 +786,7 @@ struct Video
   end
 end
 
-class Caption
+struct Caption
   property name
   property languageCode
   property baseUrl
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -989,9 +989,15 @@ def fetch_video(id, region)
 
   # Try to pull streams from embed URL
   if info["reason"]?
-    embed_page = YT_POOL.client &.get("/embed/#{id}").body
-    sts = embed_page.match(/"sts"\s*:\s*(?<sts>\d+)/).try &.["sts"]? || ""
-    embed_info = HTTP::Params.parse(YT_POOL.client &.get("/get_video_info?html5=1&video_id=#{id}&eurl=https://youtube.googleapis.com/v/#{id}&gl=US&hl=en&sts=#{sts}").body)
+    required_parameters = URI::Params.new({
+      "video_id" => [id],
+      "eurl"     => ["https://youtube.googleapis.com/v/#{id}"],
+      "html5"    => ["1"],
+      "c"        => ["TVHTML5"],
+      "cver"     => ["6.20180913"],
+    })
+
+    embed_info = HTTP::Params.parse(YT_POOL.client &.get("/get_video_info?#{required_parameters}", headers: HTTP::Headers{"x-youtube-client-version" => "6.20180913"}).body)
 
     if embed_info["player_response"]?
       player_response = JSON.parse(embed_info["player_response"])
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -425,9 +425,9 @@ struct Video
         json.array do
           self.captions.each do |caption|
             json.object do
-              json.field "label", caption.name.simpleText
+              json.field "label", caption.name
               json.field "languageCode", caption.languageCode
-              json.field "url", "/api/v1/captions/#{id}?label=#{URI.encode_www_form(caption.name.simpleText)}"
+              json.field "url", "/api/v1/captions/#{id}?label=#{URI.encode_www_form(caption.name)}"
             end
           end
         end
@@ -706,8 +706,12 @@ struct Video
   def captions : Array(Caption)
     return @captions.as(Array(Caption)) if @captions
     captions = info["captions"]?.try &.["playerCaptionsTracklistRenderer"]?.try &.["captionTracks"]?.try &.as_a.map do |caption|
-      caption = Caption.from_json(caption.to_json)
-      caption.name.simpleText = caption.name.simpleText.split(" - ")[0]
+      name = caption["name"]["simpleText"]? || caption["name"]["runs"][0]["text"]
+      languageCode = caption["languageCode"].to_s
+      baseUrl = caption["baseUrl"].to_s
+
+      caption = Caption.new(name.to_s, languageCode, baseUrl)
+      caption.name = caption.name.split(" - ")[0]
       caption
     end
     captions ||= [] of Caption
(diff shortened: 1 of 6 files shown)


unixfox:
> Furthermore, it seems like this fix also either exposes or creates issues regarding captions on a subset of age gated videos.

This is still better than not being able to load the video at all though.

I'm testing this PR on YewTube.

unixfox:
I tested with this age-restricted video: https://yewtu.be/watch?v=fzhHXt7NKUs. I know you told us that captions could break, but I didn't think the video wouldn't load at all.

```
Title: `Missing JSON attribute: simpleText
  parsing CaptionName at line 1, column 350
  parsing Caption#name at line 1, column 343 (JSON::SerializableError)`
```

I think we should do something about this error, skipping the ability to try fetching the captions?

syeopite:
Theoretically it should be a fairly simple fix. "Normal" captions are sent in this format:
`{"languageCode" => "fr", "languageName" => {"simpleText" => "French"}}`

But for some age restricted videos it becomes 

`{"languageCode" => "fr", "languageName" => {"runs" => [{"text" => "French"}]}}`

The code for parsing it is in [videos.cr](https://github.com/iv-org/invidious/blob/master/src/invidious/videos.cr)

https://github.com/iv-org/invidious/blob/5ba9a1f87d876464f5e7cf72777797c9772aebca/src/invidious/videos.cr#L785-L789

I'm just not too familiar with how `JSON::Serializable` works though.

pukkandan:
Whatever age-gate video doesn't work with this method didn't use to work previously either. See https://github.com/yt-dlp/yt-dlp/pull/328#issuecomment-864117168

src/invidious/videos.cr
```diff
@@ -989,9 +989,15 @@ def fetch_video(id, region)
...
   if info["reason"]?
-    embed_page = YT_POOL.client &.get("/embed/#{id}").body
-    sts = embed_page.match(/"sts"\s*:\s*(?<sts>\d+)/).try &.["sts"]? || ""
-    embed_info = HTTP::Params.parse(YT_POOL.client &.get("/get_video_info?html5=1&video_id=#{id}&eurl=https://youtube.googleapis.com/v/#{id}&gl=US&hl=en&sts=#{sts}").body)
+    required_parameters = URI::Params.new({
+      "video_id" => [id],
+      "eurl"     => ["https://youtube.googleapis.com/v/#{id}"],
+      "html5"    => ["1"],
+      "c"        => ["TVHTML5"],
+      "cver"     => ["6.20180913"],
```
SamantazFox, line 997: I don't think that putting each parameter into an array is required.

syeopite: Yeah that would be cleaner but unfortunately, the [`URI::Params.new`](https://crystal-lang.org/api/1.0.0/URI/Params.html#new(raw_params:Hash(String,Array(String)))-class-method) constructor requires a type of `Hash(String, Array(String)))`. I'll go ahead and switch to [`URI::Params.encode`](https://crystal-lang.org/api/1.0.0/URI/Params.html#encode(hash:Hash(String,String%7CArray(String)))-class-method) which does allow for just.

SamantazFox:
@syeopite can you please wrap the comment above the HTTP request?

SamantazFox:
I'm going to deploy that on https://test.invidious.io tomorrow. Currently #2196 is live there, then I'll have to fix #2193, then I have to backup the DB and try #1678, and finally I can go with this PR :)

syeopite:
I could go ahead and fix up #2193 if you'd like me to. I've been cleaning up #2059 for an initial merge anyways so adding the required sanitization to that shouldn't be too difficult.

SamantazFox:
> I could go ahead and fix up #2193 if you'd like me to. I've been cleaning up #2059 for an initial merge anyways so adding the required sanitization to that shouldn't be too difficult.

If you want to fix #2193, don't touch to `src/invidious/views/components/item.ecr`, as it will be fixed with #2196!

unixfox:
We really should look into fixing this bug:

> ```
> Title: `Missing JSON attribute: simpleText
>   parsing CaptionName at line 1, column 350
>   parsing Caption#name at line 1, column 343 (JSON::SerializableError)`
> ```

syeopite:
There we go, the caption parsing should be fixed now.

unixfox:
> There we go, the caption parsing should be fixed now.

Yes, it is working perfectly, thank you! I just applied your patch.

We will wait a couple of days for testing purposes and if we don't encounter any issue this is probably good to be merged.

SamantazFox:
> We will wait a couple of days for testing purposes and if we don't encounter any issue this is probably good to be merged.

@unixfox You've deployed in on yewtu.be?

unixfox:
> > We will wait a couple of days for testing purposes and if we don't encounter any issue this is probably good to be merged.
> 
> @unixfox You've deployed in on yewtu.be?

Yes that's why I did set this PR to "in-testing".

SamantazFox:
> Yes that's why I did set this PR to "in-testing".

Ok, nice :)

neurodiverseEsoteric:
test.invidious.io still doesn't play: https://www.youtube.com/watch?v=KgmoMO66uPg

syeopite:
Because this PR still isn't deployed on https://test.invidious.io. But it is running on yewtu.be, and the video works perfectly there.

neurodiverseEsoteric:
Ah ok, but it's not possible to download from yewtu.be

unixfox:
@syeopite Do you think you could trigger this bypass age restriction only when we detect it? With something like this:
```
if info["reason"]?.try &.as_s.includes?("confirm your age")
```

We don't exactly know if this can bypass the same other restricted videos (not age restricted videos) like the old method with the embed.

syeopite:
Are there even any videos that are restricted besides age gated ones? The only other I know of would be the videos that are geoblocked but those are handled elsewhere.

unixfox:
> Are there even any videos that are restricted besides age gated ones? The only other I know of would be the videos that are geoblocked but those are handled elsewhere.

I don't know either, but that's the issue, we can't test that this won't break any ability to watch restricted videos (not age or geo restricted) because we have no dataset. That's why it's safer to keep the old method as a last measure of trying to bypass a restriction.

syeopite:
Fair enough. The old method has been readded as a fallback in 57bb8c610a5489a7159a29004fdd56ec8bff50e5

syeopite:
Merge status?

unixfox:
For me it's safe to merge! It is working great.

src/invidious/videos.cr
```diff
@@ -782,18 +786,19 @@ struct Video
   end
 end
 
-struct CaptionName
-  include JSON::Serializable
+class Caption
```
SamantazFox, line 789: you should use a `struct` insteed of a `class` for such short-lived data. The former is stored on the stack, where as the latter is stored in momory, which is slower to access.

unixfox:
@SamantazFox good to be merged?

SamantazFox:
@unixfox Yep! It's looking good to me!

Issue #2191, opened 2021-06-23
[Enhancement] Improve privacy / make Invidious less detectable


unixfox:
**Is your enhancement request related to a problem? Please describe.**
YouTube is actively trying to restrict us. We should try to blend in with the crowd.

**Describe the solution you'd like**
- [ ] Randomize the user agent. Searx is already doing that: https://github.com/searx/searx/blob/master/searx/data/useragents.json
- [ ] Use a Firefox user agent. Google control chromium/chrome, so it can more easily detect some features that Invidious doesn't support compared to a real chromium/chrome browser.
- [ ] Randomize CONSENT cookie ID
- [ ] Randomize some headers (which ones?)
- [ ] Use a newer version of QUIC: related #1994

Anything else?

SamantazFox:
"Randomize CONSENT cookie ID" was done in #3893

unixfox:
Added more things to look like more of a browser.

firsturdiev:
Any plans to move to HTTP/3? I think this issue is more actual nowadays

unixfox:
@firsturdiev http3 will probably not going to help since youtube detect invidious based on other factors. It's entirely possible that with SABR we will be introducing an HTTP3 proxy though.

I'm closing this issue since all of this is outdated and doesn't make sense anymore with the current states.

PR #2207, opened 2021-06-25
Change consent to pending to avoid youtube tracking

unixfox:
This is a PR that is part of fixing https://github.com/iv-org/invidious/issues/2191.

We don't have to accept the YouTube consent, we can set it to pending with a random number after, and it works.

The major drawback of doing that is that we need to append `ucbcb=1` to every request that requires the consent to be either pending (`PENDING`) or accept (`YES`).

I tried to cover all the needed URLs, but I probably missed some of them. So that's why the PR is still a draft and I need help for finding all the part of the code that I would have missed.

### DIFF
--- a/src/invidious/helpers/utils.cr
+++ b/src/invidious/helpers/utils.cr
@@ -10,7 +10,7 @@ def add_yt_headers(request)
   request.headers["x-youtube-client-name"] ||= "1"
   request.headers["x-youtube-client-version"] ||= "2.20200609"
   # Preserve original cookies and add new YT consent cookie for EU servers
-  request.headers["cookie"] = "#{request.headers["cookie"]?}; CONSENT=YES+"
+  request.headers["cookie"] = "#{request.headers["cookie"]?}; CONSENT=PENDING+#{Random.new.rand(900)}"
   if !CONFIG.cookies.empty?
     request.headers["cookie"] = "#{(CONFIG.cookies.map { |c| "#{c.name}=#{c.value}" }).join("; ")}; #{request.headers["cookie"]?}"
   end
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -449,7 +449,7 @@ get "/modify_notifications" do |env|
     headers = HTTP::Headers.new
     headers["Cookie"] = env.request.headers["Cookie"]
 
-    html = YT_POOL.client &.get("/subscription_manager?disable_polymer=1", headers)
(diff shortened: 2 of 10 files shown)


src/invidious/helpers/utils.cr
```diff
@@ -10,7 +10,7 @@ def add_yt_headers(request)
   request.headers["x-youtube-client-name"] ||= "1"
   request.headers["x-youtube-client-version"] ||= "2.20200609"
   # Preserve original cookies and add new YT consent cookie for EU servers
-  request.headers["cookie"] = "#{request.headers["cookie"]?}; CONSENT=YES+"
+  request.headers["cookie"] = "#{request.headers["cookie"]?}; CONSENT=PENDING+#{Random.new.rand(900)}"
```
SamantazFox, line 13: Watch out, the random number should be >100 and <= 999!
[See what the peeps at youtube-dl did](https://github.com/TeamNewPipe/NewPipeExtractor/pull/600/files#diff-143031f296abdf44c2e3c0c434a9c3ed3e07db6c323b1b96aa57c6dcb618bdbeR717)

B0pol: I don't understand how that happened but it can be over 1000, it is often over 100 000

https://github.com/TeamNewPipe/NewPipeExtractor/blob/3966178979972da43f6dcbc90170058605be052b/extractor/src/test/resources/org/schabi/newpipe/extractor/kiosk/generated_mock_1.json#L10

https://github.com/TeamNewPipe/NewPipeExtractor/blob/3966178979972da43f6dcbc90170058605be052b/extractor/src/test/resources/org/schabi/newpipe/extractor/services/youtube/extractor/channel/gronkh/generated_mock_1.json#L10

and so on…

SamantazFox: > I don't understand how that happened but it can be over 1000, it is sometimes over 100 000

Oh, okay. I've never seen it being over 3 digits on youtube

SamantazFox: We should go with `Random.new.rand(100..999)` for now, so we can merge this PR and change that later, if needed...

SamantazFox:
Can you please rebase this changes on `master`?

unixfox:
Youtube doesn't block the real HTML page anymore, so we probably may remove the CONSENT cookie at all:
image

syeopite:
Closable?

unixfox:
Like I said on matrix if you really want to close it then feel free to do it. It doesn't hurt to let it open.

PR #2221, opened 2021-06-27
Use random user agent on each request

syeopite:
This PR attempts to resolve the user agent part of #2191.

On startup Invidious will now generate 10 random user agents out of a bunch of different components,  a random one would then be picked on each request.

The generated agents should all be fairly common allowing us to blend in more easily.

### DIFF
--- a/config/user-agents.json
+++ b/config/user-agents.json
@@ -1,0 +1,14 @@
+{
+"safari": {
+    "os": ["Macintosh; Intel Mac OS X 10_15_7", "Macintosh; Intel Mac OS X 10_15_6", 
+           "Macintosh; Intel Mac OS X 11_3", ""],
+    "versions": ["14.1.1", "14.1", "14.0.3"]
+    },
+
+"firefox": {
+    "os": ["Macintosh; Intel Mac OS X 10.15'", "Macintosh; Intel Mac OS X 10.14",
+            "Windows NT 10.0; Win64; x64", "X11; Ubuntu; Linux x86_64",
+            "X11; Linux x86_64", "Windows NT 10.0", "Macintosh; Intel Mac OS X 11.3"],
+    "versions": ["88.0 ", "89.0", "90.0", "78.0"]
+    }   
+}
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -66,6 +66,7 @@ SOFTWARE = {
   "branch"  => "#{CURRENT_BRANCH}",
 }
 
+UA_LIST = prepare_random_user_agents()
 YT_POOL = YoutubeConnectionPool.new(YT_URL, capacity: CONFIG.pool_size, timeout: 2.0, use_quic: CONFIG.use_quic)
 
 # CLI
--- a/src/invidious/helpers/random-user-agents.cr
+++ b/src/invidious/helpers/random-user-agents.cr
@@ -1,26 +1,9 @@
-# Using different browsers would allow us to disguise our traffic even more.
-# However, due to the widely different version and operating system values as well as
-# their frequency on different OSes we're going to need a separate dataset
-# for everything.
-
-AGENT_COMPONENTS = {
-  "safari" => {
-    "os"       => ["Macintosh; Intel Mac OS X 10_15_7", "Macintosh; Intel Mac OS X 10_15_6"],
-    "versions" => ["14.1.1", "14.1", "14.0.3"],
-  },
-
-  "firefox" => {
-    "os" => ["Macintosh; Intel Mac OS X 10.15'", "Macintosh; Intel Mac OS X 10.14",
-             "Windows NT 10.0; Win64; x64", "X11; Ubuntu; Linux x86_64",
-             "X11; Linux x86_64"],
-    "versions" => ["88.0 ", "89.0"],
-  },
-}
+AGENT_COMPONENTS = JSON.parse(File.read("config/user-agents.json")).as_h
 
 private def generate_user_agent
   agent_component = AGENT_COMPONENTS.keys.sample(1)[0]
-  os = AGENT_COMPONENTS[agent_component]["os"].sample(1)[0]
-  version = AGENT_COMPONENTS[agent_component]["versions"].sample(1)[0]
+  os = AGENT_COMPONENTS[agent_component]["os"].as_a.sample(1)[0]
+  version = AGENT_COMPONENTS[agent_component]["versions"].as_a.sample(1)[0]
   base = "Mozilla/5.0 "
 
   case agent_component
--- a/src/invidious/helpers/random-user-agents.cr
+++ b/src/invidious/helpers/random-user-agents.cr
@@ -1,0 +1,46 @@
+# Using different browsers would allow us to disguise our traffic even more.
+# However, due to the widely different version and operating system values as well as
+# their frequency on different OSes we're going to need a separate dataset
+# for everything.
+
+AGENT_COMPONENTS = {
+  "safari" => {
+    "os"       => ["Macintosh; Intel Mac OS X 10_15_7", "Macintosh; Intel Mac OS X 10_15_6"],
+    "versions" => ["14.1.1", "14.1", "14.0.3"],
+  },
+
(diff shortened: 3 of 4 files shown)


SamantazFox:
>  However, due to the widely different version and operating system values as well as their frequency on different OSes we're going to need a separate dataset for everything.

We can get some from here: https://developers.whatismybrowser.com/useragents/explore/

syeopite:
Should I change the implementation to scraping whatismybrowser.com on startup then?

unixfox:
> Should I change the implementation to scraping whatismybrowser.com on startup then?

No, that's a very bad idea to depend on this external service. If we start with just a randomized list of user agents that we frequently update through automation, it's already a good start.

syeopite:
I'll go ahead and extract the hard coded user agents into a JSON file then

SamantazFox:
@syeopite I'm wondering why you put everything in a json file, instead of a Hash structure in the source code. If the idea is to easily add new UA strings, the current implementation may still require a recompilation of invidious (base strings are hard-coded, so adding an opera/vivaldi/chromium UA is not possible without recompilation).

If you decide to keep the json file, can you please make it relocatable with an environment variable, similarly to `config.yaml`? https://github.com/iv-org/invidious/blob/0db23f925225f50ed574c69abbb0bd5de58bf709/src/invidious/helpers/helpers.cr#L128-L133

syeopite:
The main reason is to allow for the user agents to get updated through automation.

> the current implementation may still require a recompilation of invidious

No, it shouldn't. The base strings has been static for quite awhile now. If it does change in the future, we can update it though a commit.

> ... opera/vivaldi/chromium UA is not possible without recompilation)....

That's intended. See #2191

> can you please make it relocatable with an environment variable... 

I don't really see the point of this. The user agent file shouldn't really be configured by the end user. It's mostly just a file Invidious uses internally to load up user agents.

SamantazFox:
> Could you possibly elaborate a bit more on this? I don't think I understand the point of this. Imho, the user agent file shouldn't really be configured by the end user. It's mostly just a file Invidious uses internally to load up user agents.

The environment variable is meant for instance administrators, so they can deploy the binary in a more "conventional" way (i.e main binary and assets in `/usr/share`, config and systemd file in `/etc`).

config/user-agents.json
```diff
@@ -0,0 +1,14 @@
...
+           "Macintosh; Intel Mac OS X 11_3", ""],
+    "versions": ["14.1.1", "14.1", "14.0.3"]
+    },
+
+"firefox": {
+    "os": ["Macintosh; Intel Mac OS X 10.15'", "Macintosh; Intel Mac OS X 10.14",
+            "Windows NT 10.0; Win64; x64", "X11; Ubuntu; Linux x86_64",
+            "X11; Linux x86_64", "Windows NT 10.0", "Macintosh; Intel Mac OS X 11.3"],
+    "versions": ["88.0 ", "89.0", "90.0", "78.0"]
+    }
```
SamantazFox, line 13:
```suggestion
  "os": {
    "macos_1": [
      "Macintosh; Intel Mac OS X 10_15_6", 
      "Macintosh; Intel Mac OS X 10_15_7",
      "Macintosh; Intel Mac OS X 11_3"
    ],
    "macos_2": [
      "Macintosh; Intel Mac OS X 10.14",
      "Macintosh; Intel Mac OS X 10.15",
    ],
    "windows_xp": [
      "Windows NT 5.1"
      "Windows NT 5.1; WOW64"
    ],
    "windows_vista": [
      "Windows NT 6.0"
      "Windows NT 6.0; WOW64"
    ],
    "windows_7": [
      "Windows NT 6.1"
      "Windows NT 6.1; WOW64"
      "Windows NT 6.1; Win64; x64"
    ],
    "windows_8": [
      "Windows NT 6.2; WOW64"
      "Windows NT 6.2; Win64; x64"
      "Windows NT 6.3; WOW64"
    ],
    "windows_10": [
      "Windows NT 10.0; WOW64"
      "Windows NT 10.0; Win64; x64"
    ],
    "linux": [
      "X11; Ubuntu; Linux x86_64",
      "X11; Linux x86_64",
    ],
  },
  "browsers": {
    "safari": [
      {
        "base": "Mozilla/5.0 (<os>) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/<version> Safari/605.1.15"
        "os": ["macos_1"],
        "versions": ["14.1.1", "14.1", "14.0.3"]
      }
    ],
    "firefox": [
      {
        "base": "Mozilla/5.0 (<os>; rv:<version>) Gecko/20100101 Firefox/<version>",
        "os": ["macos_2", "windows_vista", "windows_7", "windows_8", "windows_10", "linux64"],
        "versions": ["78.0", "88.0 ", "89.0", "90.0", "91.0"],
      }
    ],
    "opera": [
      {
        "base": "Opera/9.80 (<os>) <version>",
        "os": ["windows_xp", "windows_vista", "windows_7", "windows_8"],
        "versions": [
          "Presto/2.9.168 Version/11.50",
          "Presto/2.12.388 Version/12.14",
          "Presto/2.12.388 Version/12.16",
          "Presto/2.12.388 Version/12.16",
          "Presto/2.12.388 Version/12.17",
          "Presto/2.12.388 Version/12.17",
          "Presto/2.12.388 Version/12.17",
          "Presto/2.12.388 Version/12.18",
          "Presto/2.12.407 Version/12.51",
        ]
      },
      {
        "base": " Mozilla/5.0 (<os>) AppleWebKit/537.36 (KHTML, like Gecko) <version>",
        "os": ["windows_xp", "windows_vista", "windows_7", "windows_8", "windows_10"],
        "versions": [
          "Chrome/66.0.3359.170 Safari/537.36 OPR/53.0.2907.99",
          "Chrome/66.0.3359.170 Safari/537.36 OPR/53.0.2907.99",
          "Chrome/69.0.3497.100 Safari/537.36 OPR/56.0.3051.52",
          "Chrome/78.0.3904.108 Safari/537.36 OPR/65.0.3467.78",
          "Chrome/79.0.3945.130 Safari/537.36 OPR/66.0.3515.44",
          "Chrome/79.0.3945.130 Safari/537.36 OPR/66.0.3515.72",
          "Chrome/80.0.3987.132 Safari/537.36 OPR/67.0.3575.97",
        ]
      }
    ]
```

SamantazFox: The structure is a bit more complex, but allows more flexibility.

syeopite: Not sure if using agents from XP, Vista and 7 are a good idea. Those systems are far too dated. Also, I thought we're not using chromium-based browsers? Both Opera and Chrome should be removed.

SamantazFox: > Not sure if using agents from XP, Vista, 7 or 8 are a good idea. Those systems are far too dated.

According to whatismybrowser, those are still relatively frequent (of course, not at the level of more modern UAs). A weighting mechanism could be a good thing, yeah.

> Also, I thought we're not using chromium-based browsers? Both Opera and Chrome should be removed in that case.

Chromium should be fine, as it doesn't have the proprietary bits of chrome that Google could detect.

syeopite: > According to whatismybrowser, those are still relatively frequent

Source? I haven't dug deeper into [their site yet but it really seems like they just put `Very Common` for all user agents](https://developers.whatismybrowser.com/useragents/explore/hardware_type_specific/computer/8). Instead I think [the dynamic data on this site probably more trustworthy](https://techblog.willshouse.com/2012/01/03/most-common-user-agents) which puts Win7 and Win8 at around ~0.2% - 0.3%. 

>  Chromium should be fine, as it doesn't have the proprietary bits of chrome that Google could detect...

Ping  @unixfox. Wdyt?

syeopite: VR data likely isn't returned for some of the ancient user agents above.

SamantazFox: > VR data likely isn't returned for some of the ancient user agents above.

Well, then we should only keep UAs for NT>= 6.1 (win 7) (if possible, otherwise only keep NT >=10)

PR #2222, opened 2021-06-28
Prevent YouTube throttling

### LINKED ISSUE #2194: [Bug] YouTube throtle speed on some videos

Not sure which category to chose, it's not a bug nor an enhancement.
I run a personal instance at home and it's really slow. I has gotten way worse recently, to the point where loading a video when clicking "play" (if it loads at all) takes ~10 sec and when it loads eventually, it plays at 144p instead of 1080p dash (default setting of the instance).

I got a 1Gpbs fiber at home so it definitely shouldn't be slow, the VM is using 500MB out of 2GB available and the CPU is <1% when playing a video. https://yewtu.be with dash enabled loads instantly and buffers the video very quickly. Did I miss-configure something or is google limiting me directly?

putara:
Hi there, I'm trying to fix #2194.

Though this is unfinished and a bit unstable, the parch seems working on PC/RPI4.
(I'm testing on a docker instance. #2147 needs `RUN apk add --no-cache make` to build this patch.)

Unfortunately it now requires a real JS interpreter (duktape) because the decryption code was too complicated without it.

ytdl-org/youtube-dl#29326 says
> This does change with different player versions, so youtube-dl will need to extract this for every video that it fetches and then modify the n parameter as such.

but it's not changed so often, and fetching it for every video is pretty expensive. This needs more investigating.

Anyway, I implemented two level caches to prevent refetching it for every video. The decryption function is called dozens times on page load. (you can see `cache #1` and `cache #2` in the debug log)

```
invidious_1  | 2021-06-28 05:54:47 UTC [debug] decrypt_n: 1LaKH5etJoE fn = cha n = RFcl_9VK-CZhPMH89IYrX -> Fcw3Gl6gxDsX_A
invidious_1  | 2021-06-28 05:54:47 UTC [debug] decrypt_n: 1LaKH5etJoE fn = (cache #1) n = RFcl_9VK-CZhPMH89IYrX -> Fcw3Gl6gxDsX_A
invidious_1  | 2021-06-28 05:54:48 UTC [debug] decrypt_n: 1LaKH5etJoE fn = (cache #2) n = Fbc5fFBLqpj0fJOfAj4Gr -> 48_sUCeKYn_LHg
invidious_1  | 2021-06-28 05:54:48 UTC [debug] decrypt_n: 1LaKH5etJoE fn = (cache #1) n = Fbc5fFBLqpj0fJOfAj4Gr -> 48_sUCeKYn_LHg
invidious_1  | 2021-06-28 05:54:48 UTC [debug] decrypt_n: 1LaKH5etJoE fn = (cache #1) n = Fbc5fFBLqpj0fJOfAj4Gr -> 48_sUCeKYn_LHg
invidious_1  | 2021-06-28 05:54:48 UTC [debug] decrypt_n: 1LaKH5etJoE fn = (cache #1) n = Fbc5fFBLqpj0fJOfAj4Gr -> 48_sUCeKYn_LHg
   :
```

### DIFF
--- a/src/invidious/helpers/signatures.cr
+++ b/src/invidious/helpers/signatures.cr
@@ -6,14 +6,14 @@ struct DecryptFunction
   @decrypt_function = [] of {SigProc, Int32}
   @decrypt_time = Time.monotonic
 
-  @playerjs_decrypt_n_function = {} of String => String
+  @decrypt_n_function = ""
   @last_decrypted_n = {id: "", n: "", dec_n: ""}
 
   def initialize(@use_polling = true)
   end
 
   def update_decrypt_function
-    @decrypt_function = fetch_decrypt_function
+    fetch_decrypt_function
   end
 
   private def fetch_decrypt_function(id = "CvFH_6DNRCY")
@@ -53,7 +53,21 @@ struct DecryptFunction
       decrypt_function << {operations[op_name], value}
     end
 
-    return decrypt_function
+    @decrypt_function = decrypt_function
+
+    function_name = player.match(/a\.get\("n"\)\)&&[\(]b=(?<nfunc>[a-zA-Z0-9]+)\(b\)/m).not_nil!["nfunc"]
+    function_body = player.match(/^#{Regex.escape(function_name)}=(?<body>function\(\w\)\{.*?"enhanced_except_[^\}]+\}[^\}]+\})/m).not_nil!["body"]
+    @decrypt_n_function = function_body
+  end
+
+  private def try_fetch_decrypt_function
+    if !@use_polling
+      now = Time.monotonic
+      if now - @decrypt_time > 60.seconds || @decrypt_function.size == 0 || !@decrypt_n_function
+        fetch_decrypt_function
+        @decrypt_time = Time.monotonic
+      end
+    end
   end
 
   def decrypt_signature(fmt : Hash(String, JSON::Any))
@@ -61,13 +75,8 @@ struct DecryptFunction
 
     sp = fmt["sp"].as_s
     sig = fmt["s"].as_s.split("")
-    if !@use_polling
-      now = Time.monotonic
-      if now - @decrypt_time > 60.seconds || @decrypt_function.size == 0
-        @decrypt_function = fetch_decrypt_function
-        @decrypt_time = Time.monotonic
-      end
-    end
+
+    try_fetch_decrypt_function
 
     @decrypt_function.each do |proc, value|
       sig = proc.call(sig, value)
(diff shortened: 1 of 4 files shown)


shoxie007:
@putara I'm cheering you on putara! Let us all know at Youtube-DL Issue #[29326](https://github.com/ytdl-org/youtube-dl/issues/29326) what your findings are and how you implemented the solution. 

In your opinion, is there is ANY way your solution could be implemented in Python?

unixfox:
Hello,

Thank you for your contribution.

In the PR https://github.com/iv-org/invidious/pull/2220 I'm working on fetching the videoplayback URLs from the internal YouTube API using an Android client.

I do think the given videoplayback URLs aren't throttled but I'm not sure that's why could you take a look at?
By either running Invidious patched with the PR #2220 or by fetching the videoplayback URLs manually using the technique described here: https://github.com/TeamNewPipe/NewPipeExtractor/issues/562

putara:
@unixfox Thanks, I will install #2220 locally and see if it can avoid throttling.

If the internal API works without any decryption, that will be great.
I personally am not a fan of polling a decryption function every minute. 😅

89z:
@unixfox doesnt change anything with throttling. As you can see from this cURL result, the returned URLs from InnerTube requests are still throttled:

~~~
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 12 60.6M   12 7727k    0     0   216k      0  0:04:47  0:00:35  0:04:12  312k
~~~

best workaround I found, is downloading the files in 10 MB chunks:

https://github.com/89z/mech/blob/v1.16.9/youtube/android.go#L52-L77

src/invidious/helpers/signatures.cr
```diff
@@ -70,4 +75,36 @@ struct DecryptFunction
...
+      playerjs = document.match(/src="(?<url>\/s\/player\/[^\/]+\/player_ias[^\/]+\/en_US\/base.js)"/).not_nil!["url"]
+      function_name = "(cache #2)"
+      function_body = @playerjs_decrypt_n_function[playerjs]?
+      if (!function_body)
+        player = YT_POOL.client &.get(playerjs).body
+        function_name = player.match(/a\.get\("n"\)\)&&\(b=(?<nfunc>[a-zA-Z0-9]+)\(b\)/m).not_nil!["nfunc"]
+        function_body = player.match(/^#{Regex.escape(function_name)}=(?<body>function\(\w\)\{.*?"enhanced_except_[^\}]+\}[^\}]+\})/m).not_nil!["body"]
+        @playerjs_decrypt_n_function[playerjs] = function_body
+      end
+      rt = Duktape::Runtime.new do |sbx|
```
SamantazFox, line 99: You should probably use a sandbox, instead of a runtime, as this javascript is what we can consider "untrusted"...

SamantazFox:
You should split the fetching and execute parts, btw. Add a Job to fetch the function every hour or every day, idk, store the string, and then run it when a video needs it.

src/invidious/helpers/signatures.cr
putara: Doesn't the runtime use a sandbox by default? (I'm not very familiar with crystal...)
https://github.com/jessedoyle/duktape.cr/blob/master/src/duktape/runtime.cr

SamantazFox: Ah, you are right. Their documentation suggests the opposite ^^

SamantazFox:
Marking this as "blocked" because this fix is not required anymore since #2220 has been merged.

@putara While we honnestly appreciate and thank you for your work, we don't want to add an extra dependency now, and will reconsider this PR if the problem occurs again.

putara:
@SamantazFox Yes, I totally agree with you. I will close this PR.

PR #2215, opened 2021-06-27
Bountiful improvements to frontend and channels

syeopite:
Due to the fact that #2059 brings in a ton of improvements and feature, I've went ahead and cleaned it up in order to do an initial merge. This should be heavily tested on an larger instance before merging. Please note that this is not an replacement for the upcoming UI rewrite

I'll do my best to summarize all of the changes this PR brings:

## Enhanced footer design: 

Desktop
footer

Mobile
mobilefooter

The description blurb above should be discussed.

## Enhanced filter UI
improved-filter

The buttons are more spread out and also gained the ability to be removed directly from the interface. I've also went ahead and moved all of these options out of the search bar and into the URL bar instead as search parameters.

## Better channel page

channel pages

about

featured channels

The default tab to `/channel/:ucid` is now the homepage. But of course, this can be configured in the preferences via the `view channel homepage by default` option.

## Tiny download refactoring
details download box

I've also went ahead and done a tiny part of the planned download refactoring.
 - Everything is now encapsulated within a details box
 - Both the DMCA and disabled downloading options are removed
 - Download is now an all or nothing feature like requested.

## Backend
~~The horrible mess that is the [extract_item(s)](https://github.com/iv-org/invidious/blob/master/src/invidious/helpers/helpers.cr#L401-L414) function has been [changed into extractors.py](https://github.com/syeopite/invidious/blob/ui-overhaul-patch/src/invidious/helpers/extractors.cr) making it far more maintainable. This also includes infrastructure for parsing YouTube categories.~~ 
Extracted and taken care of in #2228

~~All of the channel routes, save for one, has been removed from invidious.cr into it's own separate route file. This also comes with the addition of a helper method to access InnerTube's `navigation/resolve_url` API for the `/c/, /user/, /profile, etc` routes respectively.~~ 
Extracted and taken care of by #2227

~~The massive channels.cr file has also been debloated into several parts.~~ 
Extracted and merged into master by #2225

I've also extracted the channel information section from each template file and created a separate component for it. They've also been moved to their own sub-directory in the views folder. 

~~I've also went ahead and sanitized the majority of channel and author names within Invidious. However, I've likely missed some so this should be checked thoroughly.~~ Superseded by #2247

## Closes
Closes #483
Closes #2032
Closes #1969 
Closes #1923
Closes #529
Closes #740 
Closes #179
Supersedes #744
Some parts of #1563

PR #2225, opened 2021-06-28
Debloat channels.cr into multiple files

syeopite:
Cherry picked from #2215 with a few modifications:
  - channel folder is renamed to channels
  - parsing for channel home and featured channels are removed due to
    lack of infrastructure from other commits

(cherry picked from commit 44d18b8e147b47ad06a54cc6fd08423d9f39074d)


SamantazFox:
Can you please keep `produce_channel_playlists_url` around? The informations about the "playlists" protobuf object contained in this function is important!

syeopite:
Done

SamantazFox (changes requested):

src/invidious.cr
```diff
@@ -27,6 +27,7 @@ require "compress/zip"
 require "protodec/utils"
 require "./invidious/helpers/*"
 require "./invidious/*"
+require "./invidious/channels/**"
```
SamantazFox, line 30: As in `spec/helpers_spec.cr`, this should be a single star, to match all files in the `channels` directory, unless you plan to plan to add subfolders, in this case this is the spec file that needs to be fixed.

SamantazFox (approved):

SamantazFox:
@syeopite I just realized that `country` and `links` elements in the `AboutChannel` struct aren't part of the original source code (probably some leftover of your other changes). Could you just remove those?

https://github.com/syeopite/invidious/blob/5bcb689804b214229967c955e9e132a66c131186/src/invidious/channels/about.cr#L2-L22

https://github.com/syeopite/invidious/blob/5bcb689804b214229967c955e9e132a66c131186/src/invidious/channels/about.cr#L141-L144

syeopite:
Oops, somehow missed that during cherry-picking. Should be fixed now.

PR #2227, opened 2021-06-29
Extract channel routes

syeopite:
Cherry picked from #2215 with a few modifications. See commit messages.

### DIFF
--- a/src/invidious/routes/channels.cr
+++ b/src/invidious/routes/channels.cr
@@ -106,7 +106,7 @@ class Invidious::Routes::Channels < Invidious::Routes::BaseRoute
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
 
     begin
-      resolved_url = YoutubeAPI.resolve_url("https://youtube.com#{env.request.path}")
+      resolved_url = YoutubeAPI.resolve_url("https://youtube.com#{env.request.path}#{env.params.query.size > 0 ? "?#{env.params.query}" : ""}")
     rescue ex : InfoException
       raise InfoException.new("This channel does not exist.")
     end
@@ -118,10 +118,6 @@ class Invidious::Routes::Channels < Invidious::Routes::BaseRoute
       url = "/channel/#{ucid}/#{selected_tab}"
     else
       url = "/channel/#{ucid}"
-    end
-
-    if env.params.query.size > 0
-      url += "?#{env.params.query}"
     end
 
     env.redirect url
--- a/src/invidious/routes/channels.cr
+++ b/src/invidious/routes/channels.cr
@@ -105,7 +105,6 @@ class Invidious::Routes::Channels < Invidious::Routes::BaseRoute
   def brand_redirect(env)
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
 
-    # /profile endpoint uses the `user` parameter
     # /attribution_link endpoint needs both the `a` and `u` parameter
     # and in order to avoid detection from YouTube we should only send the required ones
     # without any of the additional url parameters that only Invidious uses.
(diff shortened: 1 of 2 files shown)


src/invidious/helpers/youtube_api.cr
```diff
@@ -89,16 +89,31 @@ def request_youtube_api_search(search_query : String, params : String, region =
   return _youtube_api_post_json("/youtubei/v1/search", data)
 end
 
+# Requests the youtube/v1/navigation/resolve_url endpoint with the required headers
+#
+# The resulting response is a JSON reply with the resolved url embeded within.
+# ```
+# request_youtube_api_resolve_url("https://youtube.com/c/google")                 # => {"endpoint": {"browseEndpoint": {"params": "EgC4AQA%3D", "browseId":"UCK8sQmJBp8GCxrOtXWBpyEA"}, ...}}
+# request_youtube_api_resolve_url("https://youtube.com/c/aaaaaaaaaaaaaaaaaaaaaa") # => InfoException
+# ```
```
SamantazFox, line 98: Please wrap your comments :)

src/invidious/helpers/youtube_api.cr
```diff
@@ -89,16 +89,31 @@ def request_youtube_api_search(search_query : String, params : String, region =
...
+def request_youtube_api_resolve_url(url)
+  data = {
+    "context" => make_youtube_api_context("US"),
+    "url"     => url,
```
SamantazFox, line 102: In order to avoid problems/typos, I think it would be preferable to hardcode the `https://www.youtube.com` url in here, and only append a path (`/c/user`, `/user/name`, ...)

src/invidious/routes/channels.cr
```diff
@@ -0,0 +1,149 @@
...
+
+    env.redirect "/channel/#{ucid}"
+  end
+
+  # Redirects brand url channels to a normal /channel/:ucid route
+  def brand_redirect(env)
+    locale = LOCALES[env.get("preferences").as(Preferences).locale]?
+
+    begin
+      resolved_url = request_youtube_api_resolve_url("https://youtube.com#{env.request.path}#{env.params.query.size > 0 ? "?#{env.params.query}" : ""}")
```
SamantazFox, line 109: you shouldn't put the user's URL parameters in the URL sent to youtube, as it will leak user information (preferences override).

src/invidious/routes/channels.cr
```diff
@@ -0,0 +1,149 @@
...
+    locale = LOCALES[env.get("preferences").as(Preferences).locale]?
+
+    user = env.get? "user"
+    if user
+      user = user.as(User)
+      subscriptions = user.subscriptions
+    end
+    subscriptions ||= [] of String
+
+    ucid = env.params.url["ucid"]
```
SamantazFox, line 136: You should add a verificatin after this that we have an UCID provided. if we didn't, we should resolve in here too (e.g: acessing `/channel/kevin/videos` won't go through `brand_redirect`, and we still won't have a valid UCID to work with).

syeopite: That shouldn't be an issue :) An invalid channel URL always [returns an error json response](https://bpa.st/NOKQ) which automatically raises an `InfoException` as seen here:

https://github.com/iv-org/invidious/blob/a88e5b3b84ae40d2c39f42d80afa79f1427e096f/src/invidious/helpers/youtube_api.cr#L126-L133

This is then catched and handled in the `brand_redirect` method.

src/invidious/helpers/youtube_api.cr
unixfox: `resolve_url` can resolve sub-domains of YouTube (`m.youtube.com`), I think for future usage it's better to not hardcode anything as this function could be used for other domains. One example is if a specific URL that we want only work under the `m.youtube.com`.

And we don't know all the domains that this endpoint support, so we can't predict the future utilization of this function in Invidious.

src/invidious/routes/channels.cr
SamantazFox: The problem is that `/channel/<name>/videos` is a valid URL, but we won't have the UCID for it.

src/invidious/helpers/youtube_api.cr
SamantazFox: mmmh, good point!

src/invidious/routes/channels.cr
syeopite: Ah oops looks like I completely misread your original message, sorry about that. I still fail to see how this is an problem though, `/channel/:name` isn't a valid endpoint even on YouTube so there's absolutely no reason to pass it into `brand_redirect`. Furthermore, the get_about_info function already verifies ucids and returns an InfoException when the channel doesn't exist:

SamantazFox: > I still fail to see how this is a problem though, `/channel/:name` isn't a valid endpoint even on YouTube so there's absolutely no reason to pass it into `brand_redirect`.

Oooh, right, for some reason I thought that it was a valid endpoint...

SamantazFox:
@syeopite can you please rebase this on `master` and solve conflicts?

SamantazFox (approved):

PR #2228, opened 2021-06-29
Overhaul extract_item(s) functions and add infrastructure for parsing YouTube categories

syeopite:
Extracted from #2215 with a few additional modifications


src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,320 @@
...
+      raw_items = content["items"].as_a
+    end
+
+    return raw_items
+  end
+end
+
+def extract_item(item : JSON::Any, author_fallback : String? = nil, author_id_fallback : String? = nil)
+  # Parses an item from Youtube's JSON response into a more usable structure.
+  # The end result can either be a SearchVideo, SearchPlaylist or SearchChannel.
```
SamantazFox, line 356: I guess this is the function's documentation? If yes, could you please move it *above* the function definition, so it's clearer :)

syeopite:
Perhaps I should wrap the returned results from `extract_items` into an `Any` object similar to `JSON::Any`.

SamantazFox:
> Perhaps I should wrap the returned results from `extract_items` into an `Any` object similar to `JSON::Any`.

In that case, you should probably remove the return type of that function and let type infering do the job, though it seems that their engine is having a hard time when there are many types involved. That's probably something for a future modification of the code ^^

src/invidious/helpers/helpers.cr
```diff
@@ -248,172 +248,40 @@ def html_to_content(description_html : String)
...
   end
+  return target.select(&.is_a?(SearchVideo)).map(&.as(SearchVideo))
 end
 
-def extract_items(initial_data : Hash(String, JSON::Any), author_fallback : String? = nil, author_id_fallback : String? = nil)
-  items = [] of SearchItem
+def extract_selected_tab(tabs)
+  # Extract the selected tab from the array of tabs Youtube returns
+  return selected_target = tabs.as_a.select(&.["tabRenderer"]?.try &.["selected"].as_bool)[0]["tabRenderer"]
+end
```
SamantazFox, line 267: imo, we should keep the extraction functions in the same file (i.e within `extractors.cr`)

src/invidious/helpers/serialized_yt_data.cr
```diff
@@ -0,0 +1,258 @@
+struct SearchVideo
```
SamantazFox, line 1: Adding more "miscellanous" files to the `helpers` folder seems a bad idea. What do you think about having a dedicated `data_struct` folder, where we will put the large and common data structures, one per file?

The idea would be to be able, in the end, to `require` a specific structure, without having to require the rest of the project every time, and probably improve build/testing time.

src/invidious/search.cr
```diff
@@ -463,5 +231,20 @@ def process_search_query(query, page, user, region)
     count, items = search(search_query, search_params, region).as(Tuple)
   end
 
-  {search_query, count, items, operators}
+  # Light processing to flatten search results out of Categories.
+  # They should ideally be supported in the future.
```
SamantazFox, line 235: What do you mean by that? I've seen that you made a similar chunk of code in `extract_videos()`, maybe you could make a function for that, or include it directly in your extractor code?

src/invidious/views/components/item.ecr
```diff
@@ -96,6 +96,7 @@
                 </div>
                 <% end %>
             </div>
+        <% when Category %>
```
SamantazFox, line 99: I iagine that this is "TODO"?

SamantazFox (changes requested):
TL;DR: Please provide some examples of data structures, add a separate function for `simpleText` / `runs/text` structures, use `#dig?()` and `#as_*?`

src/invidious/views/components/item.ecr
syeopite: It's more of a placeholder. The range of designs the YT categories has is far too varied for us to have it as a component.

src/invidious/search.cr
syeopite: categories

What you see here are YouTube categories. The current Invidious UI doesn't support them so we'd need to extract out the contents of the categories instead.

src/invidious/helpers/serialized_yt_data.cr
syeopite: Huh that's a good idea, Invidious structs would be far more organized that way. Though it really shouldn't affect build times too much. I don't understand the nitty gritty of the Crystal compiler but:

If `fileA` requires `fileC` and  `fileB` requires `fileA` then the contents of `fileC` is going to be available in the global namespace of `fileB` including whatever else it imports. In Invidious, since all of the imports are done in `invidious.cr` everything is already available from the get-go and any additional imports after that are redundant and likely removed by Crystal to my knowledge. 
 
I'll go ahead and make the change to see how it affects buildtimes though.

src/invidious/search.cr
SamantazFox: Oh, okay. For the second category ("people also watched"), I think we should discuss if we add that or not, as it is kind of an anti-feature (it's attention gathering content, not very relevant to the searched terms).

src/invidious/views/components/item.ecr
SamantazFox: Ah, okay

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,400 @@
...
+
+# Parses multiple items from Youtube's initial JSON response into a more usable structure.
+# The end result is an array of SearchItem.
+def extract_items(initial_data : Hash(String, JSON::Any), author_fallback : String? = nil,
+                  author_id_fallback : String? = nil) : Array(SearchItem)
+  items = [] of SearchItem
+
+  if unpackaged_data = initial_data["contents"]?.try &.as_h
+  elsif unpackaged_data = initial_data["response"]?.try &.as_h
+  elsif unpackaged_data = initial_data["onResponseReceivedActions"]?.try &.as_a.[0].as_h
```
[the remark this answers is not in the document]
syeopite: According to the documentation the `#as_*?` methods only checks if the underlying type is the requested one and internally it's pretty much just a call to `#is_a?`. 

https://github.com/crystal-lang/crystal/blob/3184e1927ddb88dcaef8ab21c8e3fb3e21944f84/src/json/any.cr#L245-L247

So the try block is needed in this case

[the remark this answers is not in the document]
SamantazFox: it calls `#as_h` if the field exists and the underlying type is a `Hash`, otherwise it returns `nil`.

The try block does the same (check if field exists or returns `nil`, and tries to return a `Hash`)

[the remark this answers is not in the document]
syeopite: > it calls #as_h if the field exists and the underlying type is a Hash, otherwise it returns nil.

No it doesn't. The `#as_*?` methods only checks if the underlying value is the requested type. It doesn't do anything else. 

```crystal
  # Checks that the underlying value is `Array`, and returns its value.
  # Returns `nil` otherwise.
  def as_a? : Array(Any)?
    as_a if @raw.is_a?(Array)
  end
...
  # Checks that the underlying value is `Hash`, and returns its value.
  # Returns `nil` otherwise.
  def as_h? : Hash(String, Any)?
    as_h if @raw.is_a?(Hash)
  end
```
https://github.com/crystal-lang/crystal/blob/3184e1927ddb88dcaef8ab21c8e3fb3e21944f84/src/json/any.cr#L231-L235


In this case, we have to validate that:
1. A specific key exist ( and then retrieve the value)
2. That it is a hash
 
Therefore, we'll have to do an expression in this syntax: 
`initial_data["response"]?.try &.as_h`
If we only did `as_h?` then when the underlying value isn't a "response", it'll cause a KeyError.
`initial_data["response"].as_h?`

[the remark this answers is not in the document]
SamantazFox: ok

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,400 @@
...
+    if item_contents = (item["videoRenderer"]? || item["gridVideoRenderer"]?)
+      return self.parse(item_contents, author_fallback)
+    end
+  end
+
+  private def parse(item_contents, author_fallback)
+    video_id = item_contents["videoId"].as_s
+    title = item_contents["title"].try { |t| t["simpleText"]?.try &.as_s || t["runs"]?.try &.as_a.map(&.["text"].as_s).join("") } || ""
+
+    author_info = item_contents["ownerText"]?.try &.["runs"]?.try &.as_a?.try &.[0]?
```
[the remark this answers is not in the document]
syeopite: Yeah it only uses `"runs"` for some reason. InnerTube is fairly inconsistent.

src/invidious/helpers/extractors.cr
```diff
@@ -49,7 +47,7 @@ private module Parsers
       # Instead, in its place is the amount of people currently watching. This behavior should be replicated
       # on Invidious once all features of livestreams are supported. On an unrelated note, defaulting to the current
       # time for publishing isn't a good idea.
```
SamantazFox, line 60: There is indeed a start timestamp for live videos:
image

It's available as:
```
"microformat": {
  "playerMicroformatRenderer": {
    ...
    "liveBroadcastDetails": {
      "isLiveNow": true,
      "startTimestamp": "2020-02-22T20:14:45+00:00"
    },
    "uploadDate": "2020-02-22"
  }
},
...
```

SamantazFox: for a stream that ended:
```
"playerMicroformatRenderer": {
	...
  "publishDate": "2021-07-26",
	"liveBroadcastDetails": {
	"isLiveNow": false,
		"startTimestamp": "2021-07-26T07:44:26+00:00",
		"endTimestamp": "2021-07-27T00:58:16+00:00"
	},
	"uploadDate": "2021-07-26"
}
```

SamantazFox: Ah, nevermind. This parser never sees data from the `player` or `next` endpoints.

syeopite: > Ah, nevermind. This parser never sees data from the `player` or `next` endpoints.

Not yet at least! I'll be migrating all of the extraction logic we have in future PR(s) to the parsers. 

But even once that is implemented, I don't think we should be querying the `player` endpoint for every `videoRenderer` that represents a live-stream or premiere.

syeopite:
Added some specs and migrated (some) YouTube and Invidious objects to a separate data_struct directory.

src/invidious/data_structs/invidious/video_preferences.cr
```diff
@@ -0,0 +1,29 @@
+module InvidiousStructs
+  # Struct containing all values for video preferences
+  struct VideoPreferences
```
SamantazFox, line 3: Can you please revert your changes on this struct? I'm already working on a preferences overhaul :)
Same for `Config` and `Preferences`.

src/invidious/data_structs/invidious/channel.cr
```diff
@@ -0,0 +1,126 @@
...
+  struct Channel
+    include DB::Serializable
+
+    property id : String
+    property author : String
+    property updated : Time
+    property deleted : Bool
+    # TODO I don't believe the subscripted attribute is actually used.
+    # so this can likely be removed.
+    property subscribed : Time?
```
syeopite, line 16: The subscribed property tracks when the channel was subscribed on PubSub, and to resubscribe if it has been more than 4 days.

SamantazFox:
Sorry in advance, but I'll again ask to split the thing x)

The many force-push makes it hard to review such a large PR. Could you keep the cleaning/moving of data structure for another time?

syeopite:
Done. PR should be much smaller now

SamantazFox:
> Done. PR should be much smaller now

Thanks! Yeah, almost 4 times smaller ^^

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,562 @@
...
+
+private ITEM_PARSERS = {
+  Parsers::VideoRendererParser,
+  Parsers::ChannelRendererParser,
+  Parsers::GridPlaylistRendererParser,
+  Parsers::PlaylistRendererParser,
+  Parsers::CategoryRendererParser,
+}
+
+record AuthorFallback, name : String? = nil, id : String? = nil
```
SamantazFox, line 19: wouldn't it be better to only accept `String`s, defaulting to `""`, so we don't have to do that sort of things afterwards? (taken from L53-54):
```crystal
author = author_fallback.name || ""
author_id = author_fallback.id || ""
```

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,562 @@
...
+      author_id = item_contents["channelId"]?.try &.as_s || author_fallback.id || ""
+
+      author_thumbnail = HelperExtractors.get_thumbnails(item_contents)
+      # When public subscriber count is disabled, the subscriberCountText isn't sent by InnerTube.
+      # Always simpleText
+      # TODO change default value to nil
+      subscriber_count = item_contents.dig?("subscriberCountText").try &.["simpleText"].try { \
+         |s| short_text_to_number(s.as_s.split(" ")[0]) } || 0
+
+      auto_generated = !item_contents["videoCountText"]? ? true : false
```
SamantazFox, line 145: I'm unsure about what result is expected, but this should be simplified.

Something like [`.nil?`](https://crystal-lang.org/api/1.1.1/Object.html#nil?:Bool-instance-method) is probablly better suited for that purpose (it's a conditional, so it returns a boolean directly, non need for `? :`):

```crystal
# True if key `"videoCountText"` doesn't exist
# False if key exists
item_contents["videoCountText"]?.nil?

# Inverted results
!(item_contents["videoCountText"]?.nil?)
```

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,562 @@
...
+  #
+  # YouTube sometimes sends the thumbnail as:
+  # {"thumbnails": [{"thumbnails": [{"url": "example.com"}, ...]}]}
+  def self.get_thumbnails_plural(container : JSON::Any) : String
+    return container.dig("thumbnails", 0, "thumbnails", 0, "url").as_s
+  end
+
+  # Retrieves the ID required for querying the InnerTube browse endpoint.
+  # Raises when it's unable to do so
+  def self.get_browse_endpoint(container)
```
SamantazFox, line 480: Wouldn't the name `get_browse_id` be more suited (it returns the ID, not the endpoint URL)

src/invidious/helpers/helpers.cr
```diff
@@ -248,168 +248,40 @@ def html_to_content(description_html : String)
...
-  elsif i = item["searchPyvRenderer"]? # Ad
   end
+  return target.select(&.is_a?(SearchVideo)).map(&.as(SearchVideo))
 end
 
-def extract_items(initial_data : Hash(String, JSON::Any), author_fallback : String? = nil, author_id_fallback : String? = nil)
-  items = [] of SearchItem
+def extract_selected_tab(tabs)
+  # Extract the selected tab from the array of tabs Youtube returns
+  return selected_target = tabs.as_a.select(&.["tabRenderer"]?.try &.["selected"].as_bool)[0]["tabRenderer"]
```
SamantazFox, line 266: This could return `["tabRenderer"]["contents"]`? The two places where it's used seem to have to do:
`extract_selected_tab(tabs)["contents"]`

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,562 @@
...
+        self.extract(target)
+      end
+    end
+
+    private def self.extract(target)
+      raw_items = [] of JSON::Any
+      content = extract_selected_tab(target["tabs"])["content"]
+
+      content["sectionListRenderer"]["contents"].as_a.each do |renderer_container|
+        renderer_container_contents = renderer_container["itemSectionRenderer"]["contents"].as_a[0]
```
SamantazFox, line 354: `.as_a` is not required, `JSON::Any` supports `[]` with an Int index

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,562 @@
...
+  # Continuation items (on YouTube) are items which are appended to the
+  # end of the page for continuous scrolling. As such, in many cases,
+  # the items are lacking information such as author or category title,
+  # since the original results has already rendered them on the top of the page.
+  #
+  # The way they are structured is too varied to be accurately written down here.
+  # However, they all eventually lead to an array of parsable items after traversing
+  # through the JSON structure.
+  module Continuation
+    def self.process(initial_data : Hash(String, JSON::Any))
```
SamantazFox, line 422: Now that I think about it, all youtube API functions should return a `JSON::Any` object, and not a `Hash`. It's a useless conversion that probably implies to copy many elements in memory.

Though, that will be something for another PR...

src/invidious/helpers/extractors.cr
```diff
@@ -0,0 +1,562 @@
...
+    results = extractor.process(unpackaged_data)
+    if !results.nil?
+      results.each do |item|
+        parsed_result = extract_item(item, author_fallback, author_id_fallback)
+
+        if !parsed_result.nil?
+          items << parsed_result
+        end
+      end
+      return items
```
SamantazFox, line 557: redundant `return` statement?

syeopite: It's just for short-circuiting the loop. If you want, I'll can go ahead and change that to a `break statement.`

src/invidious/helpers/helpers.cr
syeopite: Irrc, the `"tabRenderer"` contains some other useful information besides the `"contents"`. I'd rather not lose the flexibility of getting those other attributes.

src/invidious/helpers/extractors.cr
syeopite: Having that distinction looks better imo, but I've went ahead and pushed 43ea8fa70698ef94701fdf9da419300b9a6a0710 which should resolve that.

src/invidious/helpers/extractors.cr
SamantazFox: Oh, didn't see that! Yeah, if you can put a `break`, that would be clearer :)

src/invidious/helpers/extractors.cr
SamantazFox: Imho, it makes it harder to read. And I don't know how crystal handles that, but less unions is probably better for performance (generated code is less complex).

PR #2230, opened 2021-06-30
Remove workaround for kemalcr/kemal/issues/575

unixfox:
Removing the workaround for a bug described in https://github.com/kemalcr/kemal/issues/575 and fixed in https://github.com/kemalcr/kemal/pull/576.

The workaround was introduced by @omarroth in https://github.com/iv-org/invidious/commit/d30a972a#diff-06fc4c84acaa6461d7a42730dcb5d21997eb524bb17fcb14f9d82c32035fa4a2R918.

### DIFF
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -700,26 +700,6 @@ def proxy_file(response, env)
   end
 end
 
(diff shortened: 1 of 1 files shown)


SamantazFox (approved):
Yep, looks good to me :)

Issue #1416, opened 2020-10-16
"HTTP::Server::ClientError" and "Broken pipe (IO::Error)" spam the log


TheFrenchGhosty:
This does spam the log:

```
   from ../../../usr/lib/crystal/io/evented.cr:156:7 in 'unbuffered_write'
   from ../../../usr/lib/crystal/io/buffered.cr:217:5 in 'unbuffered_flush'
   from ../../../usr/lib/crystal/io/buffered.cr:219:5 in 'close'
   from ../../../usr/lib/crystal/http/server/request_processor.cr:62:11 in 'process'
   from ../../../usr/lib/crystal/http/server.cr:498:5 in '->'
   from ../../../usr/lib/crystal/primitives.cr:255:3 in 'run'
   from ???
 Unhandled exception in spawn: Error while flushing data to the client (HTTP::Server::ClientError)
   from ../../../usr/lib/crystal/http/server/response.cr:258:9 in 'unbuffered_flush'
   from ../../../usr/lib/crystal/io/buffered.cr:219:5 in 'close'
   from ../../../usr/lib/crystal/http/server/request_processor.cr:62:11 in 'process'
   from ../../../usr/lib/crystal/http/server.cr:498:5 in '->'
   from ../../../usr/lib/crystal/primitives.cr:255:3 in 'run'
   from ???
 Caused by: Error writing to socket: Broken pipe (IO::Error)
```

However, this doesn't seem to cause any actual problem.

SamantazFox:
@TheFrenchGhosty is is still a problem?

unixfox:
> @TheFrenchGhosty is is still a problem?

Yes it is, up to this day I still very frequently get this issue.

unixfox:
Related to https://github.com/crystal-lang/crystal/issues/9065

SamantazFox:
Closed by #2235

unixfox:
> Closed by #2235

I thought I wouldn't close this issue until we fix the original problem.

TheFrenchGhosty:
This is an upstream issue.

SamantazFox:
Ah, yeah: #9065

PR #2235, opened 2021-07-02
Mute unbuffered_flush IOError exception

unixfox:
Related #1416, it doesn't really fix the real error, but it mutes the `Error while flushing data to the client` exception until we come up with a fix for the error.

Like explained in #1416, this exception `Error while flushing data to the client` doesn't really do any actual harm to the communication between the client and the server. But this exception message, really continuously spam the logs and make debugging and finding other errors really difficult.

That's why I would rather mute it so that I can more easily find the real other errors in my logs and better test the pull requests.

This code override these lines: https://github.com/crystal-lang/crystal/blob/master/src/http/server/response.cr#L274-L279

### DIFF
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -700,6 +700,16 @@ def proxy_file(response, env)
   end
 end
 
+class HTTP::Server::Response
+  class Output
+    private def unbuffered_flush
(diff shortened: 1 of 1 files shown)


SamantazFox (approved):
Looks good to me! You should probably report the issue upstream, btw;

PR #2254, opened 2021-07-15
Add 2fa to Invidious in the form of TOTP

### LINKED ISSUE #418: Add support for 2FA (OTP, U2F)
Please add support for OTP (HOTP) and/or U2F. I am sure there are some librarys for this.

I personally prefer U2F as second factor more than OTP.

syeopite:
Closes  #418

The only problem with this PR currently is that users would have to manually enter the base32 secret since we don't have a way to construct QR codes without adding another dependency to Invidious. 

I've also went ahead and moved some of the account related views to their own folder.


locales/en-US.json
```diff
@@ -423,5 +423,17 @@
...
+    "setup-totp-instructions-enter-code": "Enter the following <strong>secret</strong> code:",
+    "setup-totp-instructions-validate-code": "Enter the 6 digit number on your screen. Be sure to do it under thirty seconds!",
+    "setup-totp-submit-button": "Setup TOTP",
+    "general-totp-empty-field": "The TOTP code is a required field",
+    "general-totp-invalid-code": "The TOTP code entered is invalid",
+    "general-totp-enter-code-field": "6 digit number",
+    "general-totp-enter-code-header": "Two-factor authentication",
+    "general-totp-verify-button": "Verify",
+    "remove-totp-header": "Remove two-factor authentication",
+    "remove-totp-confirm-message": "Are you sure you would like to remove two-factor-authentication?"
```
SamantazFox, line 438: You should have used underscores, imo

src/invidious.cr
```diff
@@ -345,6 +345,12 @@ Invidious::Routing.get "/preferences", Invidious::Routes::PreferencesRoute, :sho
 Invidious::Routing.post "/preferences", Invidious::Routes::PreferencesRoute, :update
 Invidious::Routing.get "/toggle_theme", Invidious::Routes::PreferencesRoute, :toggle_theme
 
+Invidious::Routing.get "/setup_2fa", Invidious::Routes::Accounts, :setup_2fa_page
+Invidious::Routing.post "/setup_2fa", Invidious::Routes::Accounts, :setup_2fa
+Invidious::Routing.get "/remove_2fa", Invidious::Routes::Accounts, :remove_2fa_page
+Invidious::Routing.post "/remove_2fa", Invidious::Routes::Accounts, :remove_2fa
+Invidious::Routing.post "/validate_2fa", Invidious::Routes::Accounts, :validate_2fa
```
SamantazFox, line 352: I think that it'd be a good idea to have everything under `/2fa/` (`/2fa/setup`, `/2fa/remove`, etc..).

xbdmHQ:
Any updates?

syeopite:
Rebased. I believe the PR is still being blocked by a lack of an automatic migration system. Though it should work if you manually run `./invidious --migrate` now that we have #2878

syeopite:
On another note, purely for readability and consistency reasons, I feel like most of the methods added here should be renamed and properly documented.

Issue #2231, opened 2021-07-01
[Discussion] Code quality upgrade


SamantazFox:
In order to improve code quality, I'm listing here some recommendations for crystal code. This list is subject to evolutions.

**Feel free to discuss/propose changes below!**

## JSON object trasversal

Avoid using lengthy `object["key"]?.try &.["subkey"]` statements and prefer `.dig?("key", "subkey", ...)` (Check out [the documentation for more infos](https://crystal-lang.org/api/1.0.0/JSON/Any.html#dig?(index_or_key:String%7CInt,*subkeys):JSON::Any?-instance-method)).

For instance, this code:
https://github.com/iv-org/invidious/blob/cf619f24a96820a8a54ec47426dc41164db76bc9/src/invidious/videos.cr#L862-L869
Becomes:
```
primary_results = yt_initial_data
  .dig?("contents", "twoColumnWatchNextResults", "results", "results", "contents")

sentiment_bar = primary_results
  .try &.as_a.select { |object| object["videoPrimaryInfoRenderer"]? }[0]?
  .dig?("videoPrimaryInfoRenderer", "sentimentBar", "sentimentBarRenderer", "tooltip")
  .try &.as_s
```

Perflyst:
Would be nice to lint / enforce code quality via CI later on

syeopite:
Not exactly code quality related but I feel like we should enforce and encourage more documentation in the Invidious code. And ideally, it should be [made compatible with Crystal's own documentation tool](https://crystal-lang.org/reference/syntax_and_semantics/documenting_code.html) to make it familiar to other crystal tools.

The same can be said for [the coding style but I'm pretty sure Invidious already follows most of these.](https://crystal-lang.org/reference/conventions/coding_style.html)

syeopite:
I'll suggest we start using ameba https://github.com/crystal-ameba/ameba

syeopite:
Results of running Ameba on Invidious as of  https://github.com/iv-org/invidious/commit/34508966027fce3f460d9670eeecef67b92565a0

`80 | [W] Lint/NotNil: Avoid using not_nil!`

`71 | [C] Style/QueryBoolMethods: Consider using 'property?' for boolean properties`

`54 | [C] Style/VariableNames: Var name should be underscore-cased: page_type, not pageType`

`41 | [W] Metrics/CyclomaticComplexity: Cyclomatic complexity too `

`25 | [C] Style/ParenthesesAroundCondition: Redundant parentheses`

[...]

PR #2255, opened 2021-07-15
switch to innertube API for about channels

### LINKED ISSUE #2291: [Bug]  isFamilyFriendly always returns false for channels
**Describe the bug**
Invidious api always returns false for "isFamilyFriendly" even when it should be true

**Steps to Reproduce**
1. Go to /api/v1/channels/UCbCmjCuTUZos6Inko4u57UQ?fields=author,isFamilyFriendly&pretty=1
2. Notice that "isFamilyIsFriendly" is false when it should be true

**Screenshots**
image

**Additional context**
https://github.com/iv-org/invidious/blob/master/src/invidious/channels/about.cr#L103
I think this line would need to be changed to (have not tested, guessing based on how [FreeTube ](https://github.com/FreeTubeApp/yt-channel-info/blob/master/app/youtube-grabber.js#L129)does it)
`initdata["metadata"]["channelMetadataRenderer"]["isFamilySafe"].as_bool`

I am not sure if this line also needs to be changed
https://github.com/iv-org/invidious/blob/master/src/invidious/channels/about.cr#L80

unixfox:
Fixes #2291

This uses the innertube API for fetching the `about` for the YouTube channels.

This code has been in testing for almost 1 week on https://yewtu.be


src/invidious/channels/about.cr
```diff
@@ -29,29 +29,14 @@ struct AboutRelatedChannel
 end
 
 def get_about_info(ucid, locale)
-  result = YT_POOL.client &.get("/channel/#{ucid}/about?gl=US&hl=en")
-  if result.status_code != 200
-    result = YT_POOL.client &.get("/user/#{ucid}/about?gl=US&hl=en")
+  begin
+    initdata = request_youtube_api_browse(browse_id: ucid, params: "")
```
SamantazFox, line 33: `params` should be equal to `EgVhYm91dA==` (protobuf object to get the "about" page), no?

src/invidious/channels/about.cr
```diff
@@ -29,29 +29,14 @@ struct AboutRelatedChannel
...
-  if about.xpath_node(%q(//div[contains(@class, "channel-empty-message")]))
-    raise InfoException.new("This channel does not exist.")
-  end
-
-  initdata = extract_initial_data(result.body)
-  if initdata.empty?
-    error_message = about.xpath_node(%q(//div[@class="yt-alert-content"])).try &.content.strip
-    error_message ||= translate(locale, "Could not get channel info.")
-    raise InfoException.new(error_message)
+  if initdata["alerts"]?.try &.[0]?.try &.["alertRenderer"]?.try &.["type"]? == "ERROR"
```
SamantazFox, line 38: can you please use the `dig?()` method instead of this `.try` chain? (See #2231)

src/invidious/channels/about.cr
unixfox: Great idea, this reduced the received data from 1660KB to 98KB

src/invidious/channels/about.cr
unixfox: Not sure how to use the dig function on an array though.

SamantazFox: You can even do: `initdata.dig?("alerts", 0, "alertRenderer", "type")

unixfox: Awesome this function is magic.

src/invidious/channels/about.cr
```diff
@@ -99,9 +84,9 @@ def get_about_info(ucid, locale)
     description = initdata["metadata"]["channelMetadataRenderer"]?.try &.["description"]?.try &.as_s? || ""
     description_html = HTML.escape(description).gsub("\n", "<br>")
 
-    paid = about.xpath_node(%q(//meta[@itemprop="paid"])).not_nil!["content"] == "True"
-    is_family_friendly = about.xpath_node(%q(//meta[@itemprop="isFamilyFriendly"])).not_nil!["content"] == "True"
-    allowed_regions = about.xpath_node(%q(//meta[@itemprop="regionsAllowed"])).not_nil!["content"].split(",")
+    paid = false
```
SamantazFox, line 87: How are we going to know if a channel is paid or not? Did youtube remove the info?

unixfox: Well, I tried my best to find a paid channel, and I found nothing and on the innertube API there is no info about it. I think paid channels got removed a long time ago.
This info still exist on the HTML `about` page of a channel though for backward compatibilities.

syeopite:
Instead of switching to the browse endpoint and querying the about page, wouldn't it be better to use an existing initdata collection and parse the channel attributes out of there? 

We should try to avoid querying YouTube twice every time we need some information about a channel.

src/invidious/channels/about.cr
syeopite: In that case why not just remove the paid attribute? YouTube has deprecated it and Invidious doesn't even use it anywhere besides in the serialization methods.

SamantazFox:
Can you please rebase on `master`?

unixfox:
> Can you please rebase on `master`?

Done

src/invidious/channels/about.cr
```diff
@@ -29,29 +28,14 @@ struct AboutRelatedChannel
 end
 
 def get_about_info(ucid, locale)
-  result = YT_POOL.client &.get("/channel/#{ucid}/about?gl=US&hl=en")
-  if result.status_code != 200
-    result = YT_POOL.client &.get("/user/#{ucid}/about?gl=US&hl=en")
+  begin
+    initdata = YoutubeAPI.browse(browse_id: ucid, params: "EgVhYm91dA==")
```
SamantazFox, line 32: Can you please add a comment that explains the base64 string?:
```
# "EgVhYm91dA==" is the base64-encoded protobuf object {"2:string":"about"}
```

SamantazFox (approved):

syeopite:
https://github.com/iv-org/invidious/pull/2255#issuecomment-881359929

I'll go ahead and take care of this in another PR then :)

SamantazFox (approved):

SamantazFox:
Uh, okay, for some reason Github told me files changed and requested again a review....

Issue #1952, opened 2021-03-29
[Bug] Playlist: Only plays the second video after a video finishes


austinhuang0131:
**Describe the bug**
Expect Invidious to play the following video after the current one finishes, but instead it keeps directing to the second video in the playlist.

**Steps to Reproduce**
1. This issue exists on both invidious.snopyta.org and invidious.tube.
2. Access *any* playlist.
3. Play one video in it until it finishes
4. Expect to load the following video, instead loads the second video of the playlist, regardless of which video was played in step 3.

**Logs**
Not applicable.

**Screenshots**
Not applicable.

**Additional context**
 - Browser (if applicable): Ungoogled Chromium
 - OS (if applicable): OSX

SamantazFox:
Nice catch! Can reproduce here on latest version (`2021.03.28-2c1cf42 @ master`)

TheFrenchGhosty:
I also noticed that issue.

sijanec:
> Nice catch! Can reproduce here on latest version (`2021.03.28-2c1cf42 @ master`)

Can reporoduce on `c4c2e01213 @ master`

Testsr:
Can reproduce on 2021.04.13-edb37c9 @ master.

L1pE:
In addition, having both "play next by default" and "autoplay next video" settings disabled, it still automatically loads the second video, when I'd be expect the current video to simply finish and stay on the same page.

TheFrenchGhosty:
A 20$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 20$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

unixfox:
@austinhuang0131 Could you test if you can still reproduce this issue on https://yewtu.be?

austinhuang0131:
@unixfox nope, seems fixed

unixfox:
Reverting the patch until the scrolling issue is fixed.

FranciscoPombal:
I hacked together a ViolentMonkey userscript that works around this bug.
It basically replaces `watch.js` with a version that has the following patch applied:

```diff
--- watch.js
+++ watch.js
@@ -147,20 +147,32 @@

     xhr.onreadystatechange = function () {
         if (xhr.readyState == 4) {
             if (xhr.status == 200) {
                 playlist.innerHTML = xhr.response.playlistHtml;

                 if (xhr.response.nextVideo) {
[...]

PR #2217, opened 2021-06-27
Use the new youtube api for comments

unixfox:
This PR is related to #1981.

This PR switch the endpoint used for fetching the comments from the old and soon deprecated `https://www.youtube.com/comment_service_ajax` to the InnerTube API with the endpoint `/next`.

Issue to fix:
```
invidious Invalid memory access (signal 11) at address 0xffff811a8158                                                                                                                        │
│ invidious [0x673488] print_backtrace at /usr/lib/crystal/core/exception/call_stack.cr:129:5                                                                                                  │
│ invidious [0x41ac70] __crystal_sigfault_handler at /usr/lib/crystal/core/signal.cr:347:3                                                                                                     │
│ invidious [0xf4e5f4] posix_spawn_file_actions_init at src/process/posix_spawn_file_actions_init.c:7:1                                                                                        │
│ invidious [0xcbd510] ???                                                                                                                                                                     │
│ invidious [0x47f334] -> at /invidious/lib/lsquic/src/lsquic/client.cr:210:11                                                                                                                 │
│ invidious [0x71df28] run at /usr/lib/crystal/core/primitives.cr:255:3                                                                                                                        │
│ invidious [0x41a184] -> at /usr/lib/crystal/core/fiber.cr:92:34                                                                                                                              │
│ invidious [0x0] ???
```

PR #2257, opened 2021-07-16
Playlist starts at the offset (fixes: #1952) 

diogorac:
Fixes: #1952 

I'm not familiar with the code base, but I really wanted a working playlist! 

I noticed that the ```nextVideo``` that the frontend receives was always the ID from the second video of the playlist. I looked at how the mixer is handled at  ```fetch_mix```, and saw that the list is shifted until it reaches the offset. 

It was not happening on ```def get_playlist_videos(db, playlist, offset, locale = nil, continuation = nil)``` even though we pass the offset. So I applied it. I don't know if that's the intended behavior.

Also made sure that the playlist is always filled with videos if possible. The first time a playlist is executed, it might not have an offset, that's why it takes the one from the ```continuation.video_count```

### DIFF
--- a/src/invidious/playlists.cr
+++ b/src/invidious/playlists.cr
@@ -437,17 +437,38 @@ def get_playlist_videos(db, playlist, offset, locale = nil, continuation = nil)
     db.query_all("SELECT * FROM playlist_videos WHERE plid = $1 ORDER BY array_position($2, index) LIMIT 100 OFFSET $3",
       playlist.id, playlist.index, offset, as: PlaylistVideo)
   else
-    if offset >= 100
-      # Normalize offset to match youtube's behavior (100 videos chunck per request)
-      offset = (offset / 100).to_i64 * 100_i64
+    videos = [] of PlaylistVideo
 
-      ctoken = produce_playlist_continuation(playlist.id, offset)
-      initial_data = request_youtube_api_browse(ctoken)
-    else
-      initial_data = request_youtube_api_browse("VL" + playlist.id, params: "")
+    until videos.size >= 50 || videos.size == playlist.video_count 
+      if offset >= 100
+        # Normalize offset to match youtube's behavior (100 videos chunck per request)
+        normalized_offset = (offset / 100).to_i64 * 100_i64
+        ctoken = produce_playlist_continuation(playlist.id, normalized_offset)
+        initial_data = request_youtube_api_browse(ctoken)
+      else
+        initial_data = request_youtube_api_browse("VL" + playlist.id, params: "")
+      end
+
+      videos += extract_playlist_videos(initial_data)
+
+      if continuation
+        until videos[0].id == continuation
+          videos.shift
+        end
+      elsif
+        until videos[0].index == offset
+          videos.shift
+        end
+      end
+
+      if offset == 0
+        offset = videos[0].index
+      end
+
+      offset += 50
     end
 
-    return extract_playlist_videos(initial_data)
+    return videos
   end
 end
 


syeopite (dismissed):

unixfox:

syeopite:
Oops, somehow missed this regression when testing it on my local instance a few hours back. It seems like all videos within the playlist are treated as the first within the selector widget; or in other words, we can't scroll back any further than the current video.

Here's a comparison:

With this PR
with this pr

Without this PR:
without this pr


Note the position of the two scrollbars.

diogorac:
I'm now loading some more videos before the offset, but not the whole playlist. It's similar to youtube's playlist behaviour.

syeopite (changes requested):

syeopite:
Also good job with the lookback! Along with #1952, you've fixed a massive issue with the playlist widget only showing the first 100 videos. 🎉

src/invidious/playlists.cr
```diff
@@ -437,17 +437,45 @@ def get_playlist_videos(db, playlist, offset, locale = nil, continuation = nil)
...
+    original_offset = offset
+    until videos.size >= 100 || videos.size == playlist.video_count || offset >= playlist.video_count
+      if offset >= 100
+        # Normalize offset to match youtube's behavior (100 videos chunck per request)
+        normalized_offset = (offset / 100).to_i64 * 100_i64
+        ctoken = produce_playlist_continuation(playlist.id, normalized_offset)
+        initial_data = request_youtube_api_browse(ctoken)
+      else
+        initial_data = request_youtube_api_browse("VL" + playlist.id, params: "")
+      end
```
SamantazFox, line 451: I'm wondering why you put that chunk of code in a loop. This retrieves a chunk of 100 videos, so calling it 100 times is extremely expensive computation and bandwidth-wise.

syeopite: Somehow I missed that during my review... 

Effectively that loop only goes through one iteration since at least one of it's conditions will become true no matter what:

If all 100 videos for the current paged are fetched and fulfilled? Loop ends

If all videos for the playlist are fetched? Loop ends.

Etc 

However, I think I do recall a few cases where some Invidious playlist pages doesn't show the full 100 videos while still having more pages to access.  

I'm assuming this is what @diogorac meant by:
> Also made sure that the playlist is always filled with videos if possible

So the loop is likely just an attempt to resolve that issue.

diogorac: Unless I'm missing something, it's only being called until we find the corresponding videoID (if we're missing an offset), have "enough" videos on the list and/or we reach the end of the play list. 
``` until videos.size >= 100 || videos.size == playlist.video_count || offset >= playlist.video_count```

I hadn't looked into the ```request_youtube_api_browse```, and just worked with what it gave me. But now that I did, I see no use for this normalized offset. It's just making the results to be limited to pages from 0-99/100-199/etc, where the requested offset video is not the first one. Which means, we have to shift towards it, and then request more videos to fill the playlist in front of us. Otherwise, if it's video number 99, we won't have any following videos.
Without the normalized offset, it will return a playlist exactly from the offset + 99 videos, which is more suited. And it would make the loop meaningless... However...

The main reason I added it, was because of URLs where there is no offset. eg:
https://yewtu.be/playlist?list=PLNYlAz8KKiVl6HigSczBEHDdZhb5GdrOB
Any of those links have no ```index```. Now one solution might be to add an index to the html. But it has happened to me before, when I click some similar youtube link, that gets redirected to my invidious instance and I face the same issue.
If there is no offset, we have to look for our video in each 100-page-chunk, and when we find it, we fill our playlist with more videos (again same 99th video example.)

File: src/invidious.cr [hunk 2]
```diff
@@ -66,7 +66,8 @@ SOFTWARE = {
   "branch"  => "#{CURRENT_BRANCH}",
 }
 
-YT_POOL = YoutubeConnectionPool.new(YT_URL, capacity: CONFIG.pool_size, timeout: 2.0, use_quic: CONFIG.use_quic)
+YT_POOL      = YoutubeConnectionPool.new(YT_URL, capacity: CONFIG.pool_size, timeout: 2.0, use_quic: CONFIG.use_quic)
+MAX_LOOKBACK = 20
```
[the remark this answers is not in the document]
diogorac: True, 20 might be too low. It's now configurable, but I put 50 as default.

Imo it shouldn't be too close to the total number of videos in a playlist, otherwise this will happen.
* max_lookback ->  90
* playlist size -> 1000
* Requesting playlist at video # 98 
* Looks back to 0
* Gets playlist 0 (...) 98 (...) 99

Or you can just increase the max number of returned videos. 100 lookback with 200 max_videos . Which in a big playlist means making 2 youtube requests per request. Up to you.

src/invidious/playlists.cr
syeopite: In that case wouldn't it be better to just request InnerTube's `youtubei/v1/next` endpoint? For instance, if you request with these post data — and everything else that InnerTube requires for a response: 
`videoId": "xpCj7oaWPnY", "playlistId": "PLsNB4peY6C6KhUZiP_YogoCH4gsMccbXN` 

The following JSON would then be returned:
```json
...
      "playlist": {
        "playlist": {
          ...
          "currentIndex": 13,
          "playlistId": "PLsNB4peY6C6KhUZiP_YogoCH4gsMccbXN",
          "totalVideos": 2371
        }
     }
```
Which contains the index of the current video within the requested playlist.

diogorac: Oh cool I didn't know about this API. That simplifies things.

 I added it and increased the lookback to 100 and max videos to 200.

src/invidious/playlists.cr
```diff
@@ -437,17 +437,23 @@ def get_playlist_videos(db, playlist, offset, locale = nil, continuation = nil)
     db.query_all("SELECT * FROM playlist_videos WHERE plid = $1 ORDER BY array_position($2, index) LIMIT 100 OFFSET $3",
       playlist.id, playlist.index, offset, as: PlaylistVideo)
   else
-    if offset >= 100
-      # Normalize offset to match youtube's behavior (100 videos chunck per request)
-      offset = (offset / 100).to_i64 * 100_i64
+    if continuation
+      initial_data = request_youtube_api_next(continuation, playlist.id)
+      offset = initial_data["contents"]["twoColumnWatchNextResults"]["playlist"]["playlist"]["currentIndex"].as_i
```
syeopite, line 442: Please switch to the [`#dig()` method of JSON::Any](https://crystal-lang.org/api/1.0.0/JSON/Any.html#dig(index_or_key:String%7CInt,*subkeys):JSON::Any-instance-method). See #2231

src/invidious/helpers/youtube_api.cr
```diff
@@ -89,6 +89,33 @@ def request_youtube_api_search(search_query : String, params : String, region =
...
+#
+#  - A playlist ID (parameters MUST be an empty string)
+#
+def request_youtube_api_next(continuation : String, playlistId : String)
+  # JSON Request data, required by the API
+  data = {
+    "context"    => make_youtube_api_context("US"),
+    "videoId"    => continuation,
+    "playlistId" => playlistId,
+  }
```
syeopite, line 114: The `continuation` parameter needs to be renamed to something along the lines of `video_id`. Please also update the documentation for it!

syeopite (changes requested):

src/invidious/helpers/youtube_api.cr
SamantazFox: This part will be overriden by #2217 and #2220 anyway (as those will be merged first).

diogorac: done!

src/invidious/playlists.cr
SamantazFox: > I hadn't looked into the `request_youtube_api_browse`, and just worked with what it gave me. But now that I did, I see no use for this normalized offset. It's just making the results to be limited to pages from 0-99/100-199/etc, where the requested offset video is not the first one.

This bit of code was made for playlist pages. You can still create a ctoken with the video ID you want, youtube will always return this video and the next 99 ones.

> Which means, we have to shift towards it, and then request more videos to fill the playlist in front of us. Otherwise, if it's video number 99, we won't have any following videos.

Not necessarily. If you're playing the 205th video in a playlist, you can always craft a ctoken with offset = 155, so the API will give you all the videos from 155 to 254, inclusive.

> The main reason I added it, was because of URLs where there is no offset. eg:
> https://yewtu.be/playlist?list=PLNYlAz8KKiVl6HigSczBEHDdZhb5GdrOB
> Any of those links have no `index`. Now one solution might be to add an index to the html. (which I did) But it has happened to me before, when I click some similar youtube link, that gets redirected to my invidious instance and I face the same issue.

we shouldn't need an `index=` parameter. We know what the current video is, we already have a list of videos, so we can do everything on the back-end side.

src/invidious.cr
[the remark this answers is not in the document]
SamantazFox: I think that it would be better to stay at 1 API call (so 100 videos), and try to get up to the 49 preceeding / 50 following videos (all following videos if `idx < 50` / all following videos if `idx > end-50`) on the backend, then add javascript to request more as the user scrolls.

src/invidious/playlists.cr
diogorac: >This bit of code was made for playlist pages. You can still create a ctoken with the video ID you want, youtube will always return this video and the next 99 ones.
>Not necessarily. If you're playing the 205th video in a playlist, you can always craft a ctoken with offset = 155, so the API will give you all the videos from 155 to 254, inclusive.

Yeah I understood that later, and that's why I changed it so it crafts the ctoken with the specific offset. 

>we shouldn't need an index= parameter. We know what the current video is, we already have a list of videos, so we can do everything on the back-end side.

And we don't need. But by having it, we makes things smoother on the backend. That way we don't have to search for the index with the```v1/next``` request.

But maybe I'm missing something?

src/invidious/helpers/youtube_api.cr
syeopite: > This part will be overriden by #2217 and #2220 anyway (as those will be merged first).

Should they be merged first? I personally think important bug fixes like this should get into master as soon as possible.

SamantazFox: The problem is that we currently have 3 PRs that implements `request_youtube_api_next`, and I'd like to do something in order to not have to deal with that later (i.e: I'm thinking about implementing `next` and `player` endpoints now and some code cleaning)

unixfox: Good idea to just add the endpoints in the code in a separate PR.

src/invidious/playlists.cr
SamantazFox: > And we don't need. But by having it, we make things smoother on the backend. That way we don't have to search for the index with the`v1/next` request.
> 
> But maybe I'm missing something?

Nope, you're right :) That index thing is a good idea, and it looks like youtube does it, too...

src/invidious/helpers/youtube_api.cr
SamantazFox: done

SamantazFox:
Waiting for 2217 to be reviewed and merged before doing any further review on this one!

syeopite:
Can you please rebase this on master?

diogorac:
Rebased, and updated! Tested locally, and seems to be working fine.

syeopite (approved):

SamantazFox:
In test on https://test.invidious.io/

config/config.example.yml
```diff
@@ -239,6 +239,13 @@ https_only: false
 ##
 #statistics_enabled: false
 
+##
+## Maximum videos to be loaded into the playlist before the video playing
+##
+## Accepted values: a positive integer
+## Default: 100
+##
+max_lookback: 100
```
SamantazFox, line 248: Imo, another config option is not necessary. The server should always load 100 videos, and, if required, we load more of them with XHR when the playlist widget is scrolled.

src/invidious/playlists.cr
```diff
@@ -437,17 +437,26 @@ def get_playlist_videos(db, playlist, offset, locale = nil, continuation = nil)
     db.query_all("SELECT * FROM playlist_videos WHERE plid = $1 ORDER BY array_position($2, index) LIMIT 100 OFFSET $3",
       playlist.id, playlist.index, offset, as: PlaylistVideo)
   else
-    if offset >= 100
-      # Normalize offset to match youtube's behavior (100 videos chunck per request)
-      offset = (offset / 100).to_i64 * 100_i64
+    if continuation
+      initial_data = YoutubeAPI.next({
+        "videoId"    => continuation,
```
SamantazFox, line 442: Please use `video_id` instead of `continuation` (I know that it was the original term used, but that should be fixed). Please also propagate accordingly, but keep the `&continuation=` API parameter as it is (we will do what's required to deprecate this).

SamantazFox (changes requested):

src/invidious/playlists.cr
diogorac: Gladly added it. `continuation` sounds weird to me

config/config.example.yml
diogorac: I've grown to like the 200 videos to be honest... But don't really have a strong opinion on this matter. I did it out of a suggestion from before. Maybe we can tackle this in another PR/Issue?

SamantazFox: Yeah, that was @syeopite, IIRC. @unixfox @TheFrenchGhosty what are your thoughts on this?

syeopite: Yeah it was originally at 20 which seems absurdly low to me. 

> The server should always load 100 videos, and, if required, we load more of them with XHR when the playlist widget is scrolled.

This is a great idea! The current behavior of looking back 100 videos is what's implemented on YouTube itself, but maybe we shouldn't restrict users of Invidious to the same constraint. 

However imho, I think loading 100 videos in both directions is enough for most use cases. If you need to get to an earlier video, then it's really just faster to go onto the actual playlist page itself.

SamantazFox: Yeah. Loading more than 100 videos will only be useful when we'll rework the playlist widget to reduce the size of the thumbnails and list indices, too.

diogorac:
Is there something else to be added?

unixfox:
> Is there something else to be added?

No I don't think so, I have been testing your PR on https://yewtu.be for almost 2 weeks now, and it seems to work great.

syeopite (changes requested):
Hi, we've just had a discussion about it on the Matrix room and a few changes need to be done. 

Firstly, as @SamantazFox said, please remove the loopback configuration and limit the hard-coded value to +- 50ish videos; more on that in the next paragraph. 

Secondly, could you please switch the playlist widget entirely to the InnerTube next endpoint? It returns all the videos we'll need to populate it. However, as I stated above, please instill an artificial limit of +- 50 videos. The current UI just isn't suited for navigating massive amounts of videos, it lacks positional markers, indices, and shows only a single video at a time.

unixfox:
Removing testing of this PR on https://yewtu.be until it is rebased against the master branch.

SamantazFox:
Merged from command line.

diogorac:
Hi, just send it here thanks!

33BFZMhRnVeMD12cY7m7BfprrGnwCtuLyP

Much appreciated

TheFrenchGhosty:
A 20$ bounty has been rewarded to Givewell.org (on behalf of @diogorac ) for this PR.

Transaction (ac4427fb30a700480b14a60725a2ca5e17a1aee9e6ebd32b33617821389a2c8f):

Amount sent: 0.00041820 BTC (~17.32€ or ~20$)

Mining fees (5 sat/byte): 0.0000071 BTC (~0.37€)

Issue #2256, opened 2021-07-15
Build fails with crystal 1.1.0 (2021-07-14)


tinglycraniumplacidly:
```
user@host:~/gits/invidious$ crystal build src/invidious.cr --release
Showing last frame. Use --error-trace for full trace.

In /usr/share/crystal/src/socket/addrinfo.cr:192:7

 192 | resolve(domain, service, family, Type::STREAM, Protocol::TCP) { |addrinfo| yield addrinfo }
       ^------
Error: no overload matches 'Socket::Addrinfo.resolve' with types Int32, Socket::Family, Bool, Socket::Type, Socket::Protocol

Overloads are:
 - Socket::Addrinfo.resolve(domain, service, family : Family | ::Nil = nil, type : Type = nil, protocol : Protocol = Protocol::IP, timeout = nil)
 - Socket::Addrinfo.resolve(domain, service, family : Family | ::Nil = nil, type : Type = nil, protocol : Protocol = Protocol::IP, timeout = nil, &block)
 ```

unixfox:
We don't support crystal 1.1.0 yet, you are advised to use crystal 1.0.0 until further notices.

SamantazFox:
Looks like an issue with crystal itself (The error points to the code at `/usr/share/crystal/src/socket/addrinfo.cr:192:7` and [we don't use the `resolve()` function](https://github.com/iv-org/invidious/search?q=resolve)).

What distro are you using? Are you using the upstream binaries or your package maintainer's?

tinglycraniumplacidly:
Hi @SamantazFox thanks for following up.

I'm using Ubuntu, and I followed the Manual setup as outlined here
https://crystal-lang.org/install/on_ubuntu/

I wasn't sure how to downgrade/get 1.0.0.

syeopite:
It might not be an upstream issue. I just quickly ran an error trace which pointed to these extension methods for a potential culprit
 
https://github.com/iv-org/invidious/blob/0d57a887ea59b6111d9a19b1ea1cea619ad9f129/src/invidious/helpers/helpers.cr#L763-L773

GauthierPLM:
Hello,

If you upgraded crystal to 1.1.0 and encounter issues, you can try to downgrade to 1.0.0 until a fix is done.
Unfortunately, the official Crystal repository does not contains the 1.0.0 version any more.

If you are lucky, the package could still be in `/var/cache/apt/archives/`.
[...]

PR #2263, opened 2021-07-19
fixes for crystal 1.1.0

tleydxdy:
fix for https://github.com/iv-org/invidious/issues/2256

### DIFF
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -761,7 +761,7 @@ class HTTP::Client
 end
 
 class TCPSocket
-  def initialize(host, port, dns_timeout = nil, connect_timeout = nil, family = Socket::Family::UNSPEC)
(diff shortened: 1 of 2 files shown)


unixfox:
Is this fix still working for crystal 1.0.0?

tleydxdy:
I don't have a 1.0.0 to test now, but seeing as the test passed, it should. There's not much that was changed, mostly to guide the compiler back to the old track. ofc testing on 1.0.0 is probably a good idea.

Perflyst:
Works fine with crystal 1.0.0

Perflyst (approved):

tleydxdy:
the json parsing part should be fixed by 1.1.1 now https://github.com/crystal-lang/crystal/pull/10972
However, it may not be a good idea to use Int64 anyway.

SamantazFox:
@tleydxdy Why using `Int64` is a bad idea? What's wrong about it?

syeopite:
Superseded by #2295

PR #2264, opened 2021-07-20
Add nightly crystal to CI matrix

caspiano:
This large n' lovely codebase serves as a great testbed for breakages in nightly builds of crystal, hence I've added nightly to the CI matrix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,32 +23,47 @@ on:
 
 jobs:
   build:
-    
+
     runs-on: ubuntu-latest
-    
+
+    name: "build - crystal: ${{ matrix.crystal }}, stable: ${{ matrix.stable }}"
+
+    continue-on-error: ${{ !matrix.stable }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        stable: [true]
+        crystal:
+          - 1.1.0
+          - 1.0.0
+        include:
+          - crystal: nightly
+            stable: false
+
     steps:
       - uses: actions/checkout@v2
-      
+
       - name: Install Crystal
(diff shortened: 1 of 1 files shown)


SamantazFox:

Can you please put 1.0.0 before 1.1.0, to so the version order is consistent (from older to newest)? ty!

TheFrenchGhosty:
Is CI able to restart itself when Nightly update?

I don't think so, so in my opinion it's pointless to have a build on nightly since it will only be the nightly of the day of the last commit.

I maybe don't get this PR though.

unixfox:
> Is CI able to restart itself when Nightly update?
> 
> I don't think so, so in my opinion it's pointless to have a build on nightly since it will only be the nightly of the day of the last commit.
> 
> (I maybe don't get the point of this PR though, but from what I understand you want CI to build on 1.0.0/1.1.0/Nightly)
> 

An easy fix would be to use cron in order to build invidious with nightly every day.

TheFrenchGhosty:
@unixfox It could work for master yes, but not for PRs sadly. If it's done this way PRs shouldn't be build on Nightly.

unixfox:
> @unixfox It could work for master yes, but not for PRs sadly. If it's done this way PRs shouldn't be build on Nightly.

We can have both. Ran on PRs and master, I think that's fine. I don't see any issues.
We would be testing the working code and the future improvements to the code.

syeopite:
>  If it's done this way PRs shouldn't be build on Nightly.

Adding a cron job isn't going to make it build on nightly, only master. Irrc I don't think there's even any method of configuring `and` rules on the workflow files but I digress.

Either way, @caspiano could you please go ahead and schedule the CI to build everyday?

syeopite (approved):

PR #2155, opened 2021-06-13
Multiple search fixes

SamantazFox:
Fixes #2140 

Other fixes:
* Empty search redirects to /search, not /
* Show the fullscreen search "home page" (from #1977) at /search
* Allow 'region=' parameter to be passed to /search
* Add "search" to fullscreen search page's `<title>`

PR #2269, opened 2021-07-22
Extract feed routes

syeopite:
This PR extracts all the feed routes from `invidious.cr` into `src/invidious/routes/feeds.cr`; the associated views has also been moved to their own sub-directory.

In addition, I've went ahead and switched the `/view_all_playlist` route into `/feed/playlists` for consistency with the other header tabs.


src/invidious/routes/feeds.cr
```diff
@@ -0,0 +1,434 @@
...
+
+    if !token
+      env.response.status_code = 403
+      return
+    end
+
+    user = PG_DB.query_one?("SELECT * FROM users WHERE token = $1", token.strip, as: User)
+    if !user
+      env.response.status_code = 403
+      return
```
SamantazFox, line 219: It should be `next`, not `return`, to allow the 403 error handler (and other things) to kick in, if required. `return` ends the execution chain abruptly, and that can cause issues (I had this problem with #2155)

src/invidious/routes/feeds.cr
```diff
@@ -0,0 +1,434 @@
...
+          video.to_xml(channel.auto_generated, params, xml)
+        end
+      end
+    end
+  end
+
+  def rss_private(env)
+    locale = LOCALES[env.get("preferences").as(Preferences).locale]?
+
+    env.response.content_type = "application/atom+xml"
```
SamantazFox, line 207: it could be a good idea to add the following line to Atom feeds handlers:
`env.response.header["Content-Type"] = "application/atom+xml"`

`env.response.content_type` doesn't override the `Content-Type` header, defined in `before_all` as HTML. Because of that, the Atom feeds are not detected as such by the browser.

SamantazFox (changes requested):

src/invidious/routes/feeds.cr
syeopite: Oops, most of those were automatically replaced. 

> It should be `next`, not `return`...

Well yes, except `next` can't exactly be used in this context.

I've just went ahead and added a `haltf` macro that functions similar to Kemal's own `halt` and used that instead.

SamantazFox (approved):

PR #2270, opened 2021-07-22
Fix raw parameter for videos in listen mode 

### LINKED ISSUE #1378: Provide Podcast RSS feeds with direct video links
The current RSS feeds are great, but not compatible with podcast apps.

Real podcast RSS feeds should be possible. The biggest problem seems to be a direct media link.

For video links we can only use those formats containing audio and video (360p and 720p). Invidious already has a mechanism to generate direct video links by appending the parameter `raw=1` on invidious video links. With the `quality` parameter we can control if 360p (`quality=medium`) or 720p (`quality=hd720`) is provided.

For audio-only links I thought the combination of parameters `raw=1` and `listen=true` could provide a direct link similar to the video links. Unfortunately, this does not work, as the `raw` parameter is designed for video links. I currently see no way to get a direct link to audio-only formats.

So we should be able to create podcast feeds similar to current RSS feeds and use the above method to insert valid media links. As we have to specify a quality beforehand we should provide different RSS feeds for each quality level. The UI could feature a drop-down menu for the different formats and qualities.

syeopite:
Closes #1378

### DIFF
--- a/src/invidious/routes/watch.cr
+++ b/src/invidious/routes/watch.cr
@@ -167,9 +167,11 @@ class Invidious::Routes::Watch < Invidious::Routes::BaseRoute
       if params.listen
         url = audio_streams[0]["url"].as_s
 
-        audio_streams.each do |fmt|
-          if fmt["bitrate"].as_i == params.quality.rchop("k").to_i
-            url = fmt["url"].as_s
+        if params.quality.ends_with? "k"
+          audio_streams.each do |fmt|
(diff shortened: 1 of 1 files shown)


SamantazFox:
@syeopite Have you deployed the change on your instance?

syeopite:
Yep. Its been tested and works perfectly. Any video url that's called with the `listen` and `raw` parameters would now send the raw audio file.

PR #2271, opened 2021-07-22
Extract API routes 

syeopite:


src/invidious/routes/API/v1/channels.cr
```diff
@@ -0,0 +1,267 @@
+class Invidious::Routes::V1Api < Invidious::Routes::BaseRoute
+  def home(env)
+    locale = LOCALES[env.get("preferences").as(Preferences).locale]?
```
SamantazFox, line 3: The locale shouldn't be taken into account in the API (it has to be stable for programmatic use)

syeopite: That was present before the extraction, should I remove it? I'd rather not do any noticable changes to the API with this PR — especially when it is a "breaking" change like this.

syeopite:
Any ideas on how to sort the remaining API routes? It's obvious that the video playback endpoints can all be moved to its own separate file. But the others? I'm not quite sure.

src/invidious/routes/API/v1/channels.cr
SamantazFox: Double check to be sure, but that shouldn't break anything. Moreover, it's already done in the `before_all` handler: https://github.com/iv-org/invidious/blob/master/src/invidious.cr#L287

src/invidious/routes/API/v1/routes.cr
```diff
@@ -0,0 +1,30 @@
...
+                    {"latest", "latest"},
+                    {"playlists", "playlists"},
+                    {"comments", "community"}, # Why is the route for the community API `comments`?,
+                    {"search", "channel_search"},
+                  } %}
+
+  Invidious::Routing.get "#{{{base_url}}}/channels/#{{{route[0]}}}/:ucid", Invidious::Routes::V1Api, :{{route[1]}}
+  Invidious::Routing.get "#{{{base_url}}}/channels/:ucid/#{{{route[0]}}}", Invidious::Routes::V1Api, :{{route[1]}}
+
+  {% end %}
```
SamantazFox, line 29: instead of a macro in here, you could do something like that:

``` crystal
get "#{{{base_url}}}/channels/:route/:ucid", Invidious::Routes::V1Api, :channel_routes_handler
get "#{{{base_url}}}/channels/:ucid/:route", Invidious::Routes::V1Api, :channel_routes_handler
```

``` crystal
def channel_routes_handler(env)
  switch env.params.url["route"]
  case "home"
    home(env)
  case "videos"
    videos(env)
  # More cases
  else
    # 404 error
  end
end  
```

src/invidious/routes/API/v1/channels.cr
syeopite: The problem comes when you take into account just how intertwined the backend and frontend of Invidious is.

syeopite: The problem comes when you take into account just how intertwined the backend and frontend of Invidious is. Take the `template_*` and `fetch_*` methods for comments:
 <details>

https://github.com/iv-org/invidious/blob/bc937a6434eaeb231c7621426fcf818fe07807fb/src/invidious/comments.cr#L186
https://github.com/iv-org/invidious/blob/bc937a6434eaeb231c7621426fcf818fe07807fb/src/invidious/comments.cr#L301-L302
https://github.com/iv-org/invidious/blob/bc937a6434eaeb231c7621426fcf818fe07807fb/src/invidious/comments.cr#L370-L384

</details>

They all require the locale to translate **key features** of the comments (timestamps, permalink, reply loading, etc). 

Examples like those are littered everywhere throughout the API logic; and without significant changes — that is also out of scope for this PR — none of them can be removed.

src/invidious/routes/API/v1/routes.cr
syeopite: Unfortunately due to an upstream bug with kemal this wouldn't work; any attempt to access `env.params.url[]` would just yield a 404. We'll have to use a workaround similar to this: https://github.com/iv-org/invidious/blob/3e5c353298df54463558d5f9f1ad0698fe145c74/src/invidious.cr#L1592 And change all of the channel API handlers to either implement the same or accept the ucid as a separate parameter.  

It really isn't that much work but I'd rather not use a feature that's currently broken upstream unless we absolutely have to.

src/invidious/routes/API/v1/channels.cr
SamantazFox: Ooooffff .... Ok, nevermind, lol

src/invidious/routes/API/v1/routes.cr
SamantazFox: Ohhh, yeah, just realized. I'd say that it isn't a bug: kemal doesn't (and can't possibly) know what we expect to be an UCID and what we expect to be a route.

Then, the solution is to do the following:
```crystal
get "#{{{base_url}}}/channels/:ucid/:route", Invidious::Routes::V1Api, :channel_routes_handler
```
```crystal
def channel_routes_handler(env)
  # Handle API calls made by confused programmers who inverted
  # the channel route and the UCID.
  if env.params.url["route"].matches?(/UC[a-zA-Z0-9_-]{22})/)
    # Inverted
    route = env.params.url["ucid"]
    ucid = env.params.url["route"]
  elseif env.params.url["ucid"].matches?(/UC[a-zA-Z0-9_-]{22})/)
    # Normal
    route = env.params.url["route"]
    ucid = env.params.url["ucid"]
  else
    # Invalid UCID!
    raise # SomeException
  end

  switch route
    # Different routes handlers
  end
end
```

syeopite: > Ohhh, yeah, just realized. I'd say that it isn't a bug: kemal doesn't...

That's one part of it, but it's not the bug I'm referring to. I'm talking about the errror that occurs when routes like this occurs: `/a/:a`, `/a/:a/:b`, that prevents us from accessing anything in `env.params.url` without causing a 404 — as in http status event not missing hash key.

In our case, the route `.../channels/:ucid/:route` proceeds `.../channels/:ucid` and thus this strange behavior occurs.

src/invidious/routes/API/v1/channels.cr
syeopite: Yep... In order to do any meaningful refactor of the API, we'll first need to do the backend split.

src/invidious/routes/API/v1/routes.cr
SamantazFox: but `/api/v1/channels/:ucid`, doesn't exist, no?

syeopite: It does. https://github.com/iv-org/invidious/blob/c6a3ac9675f814baf49d9f272f10231110a6910e/src/invidious/routes/API/v1/routes.cr#L15

src/invidious.cr
```diff
@@ -1831,373 +1833,6 @@ end
...
-
-      next response.to_json
-    end
-  end
-end
-
-get "/api/v1/insights/:id" do |env|
-  locale = LOCALES[env.get("preferences").as(Preferences).locale]?
-  next error_json(410, "YouTube has removed publicly available analytics.")
-end
```
SamantazFox, line 2125: You should'nt remove this endpoint. This is part of the v1 API, and imo we should keep it consistent, even if youtube removed the feature.

SamantazFox:
> Any ideas on how to sort the remaining API routes? It's obvious that the video playback endpoints can all be moved to its own separate file. But the others? I'm not quite sure.

What do you think of that?

* `routes/api_v1` (folder)
    - `video_playback.cr` (file)
        * storyboards
        * captions
        * annotations
        * comments
    - `feeds.cr` (file)
        * trending
        * popular
    - `channels.cr` (file)
        * home (a.k.a `channels/:ucid`)
        * videos
        * latest
        * playlists
        * community
    - `search.cr` (file)
        * search
        * search_suggestions
        * channel_search
    - `user.cr` (file)
        * Everything under `/auth` ?
    - `misc.cr` (file)
        * stats
        * mixes

src/invidious.cr
syeopite: Eh I disagree. I mean Invidious currently works off of a rolling release model and I don't see why the API has to be different  —  anything that is planned to be removed would obviously have a deprecation notice as well.

In the case of the insights endpoint, it has been deprecated for almost 2 years now. See https://github.com/iv-org/invidious/commit/aba2c5b9387a8968ac1e49fe181e7bc6ca4e3b1b. I just don't see the point of still keeping it within the code base.

syeopite:
Ended up using a slightly modified version of that plan.

src/invidious/routes/api/v1/videos.cr
```diff
@@ -0,0 +1,372 @@
...
+
+        cache_annotation(PG_DB, id, annotations)
+      end
+    else # "youtube"
+      response = YT_POOL.client &.get("/annotations_invideo?video_id=#{id}")
+
+      if response.status_code != 200
+        env.response.status_code = response.status_code
+        return
+      end
```
SamantazFox, line 289: ditto

src/invidious/routes/api/v1/videos.cr
```diff
@@ -0,0 +1,372 @@
...
+      rescue ex
+        comments = nil
+        reddit_thread = nil
+        content_html = ""
+      end
+
+      if !reddit_thread || !comments
+        env.response.status_code = 404
+        return
+      end
```
SamantazFox, line 354: ditto

SamantazFox (changes requested):

src/invidious/routes/api/v1/routes.cr
```diff
@@ -0,0 +1,67 @@
+# There is far too many API routes to define in invidious.cr
+# so we'll just do it here instead with a macro.
+macro define_v1_api_routes(base_url = "/api/v1")
```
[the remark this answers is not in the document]
syeopite: > You should put that macro under the Invidious::Routes::APIv1 namespace, so you can use selfinstead of the whole class name.

I can put it under the `Invidious::Routes::APIv1` namespace, but `self` doesn't exactly work with macros and neither can it be used to refer to types in that Object's namespace.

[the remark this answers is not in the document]
SamantazFox: > but `self` doesn't exactly work with macros

Oh, right, it's `{{@type.id}}`

SamantazFox (approved):

PR #2274, opened 2021-07-24
Change Invidious logo in README to be a svg.

syeopite:
Logo looks slightly off as I had to recreate it.


SamantazFox:
You could use the "Safari pinned tab" version. It's already in SVG and is porbably closer to the original raster image. It has a transparent background, though.

https://github.com/iv-org/invidious/blob/master/assets/safari-pinned-tab.svg

PR #2285, opened 2021-07-28
Localization overhaul

### LINKED ISSUE #571: ISO codes for languages

### LINKED ISSUE #1951: [Enhancement] Localizations / internationalization overhaul
As discussed in #1920, there are many issues regarding localizations (`l10n`, i.e different locales/translations) and internationalization (`i18n`, i.e making sure content can be translated).

This issue will be used to regroup those many different problems, and keep track of what have been fixed. Feel free to report and discuss l10n/i18n issues here.

---

# Open questions

## Which format for key strings (strings identifiers)?

This question comes from PR #1629, where new key strings were added ("year", "hour", "video", ...) but do not correspond to the original string ("This year", "Last hour", ...) like it was the case for all previous strings.

Two options are available:
- keep the current format (where the key string is the original english US string)
- switch to short identifiers (like "last_video", "error_invalid_playlist") to avoid confusion and make strings easier to maintain.

## How to handle singular/plural and other counting forms?

The current system is bad, as we don't want a regex in the translation files. We have also to take into account that in some languages, the form for **zero**, **one**, **two**, and **more** of something can be different. 

The Qt docs have a good article on the subject: https://doc.qt.io/archives/qq/qq19-plurals.html

## What to do with orphaned translation files?

The following locales still exist in the source tree, but are either very outdated or poorly translated. Plus, they're not even loaded by invidious: `bn_BD`, `eu`, `hu-HU`, `sk`, `sr_Cyrl`.

# Planned work

- [x] Solve discrepancies between locales, and add tests to ensure it won't happen again (Done in #1920)
- [ ] Use non-translated locales name instead of ISO codes in settings (See #1916)
- [ ] Find strings that are supposed to be translated but aren't in the locale files (https://github.com/iv-org/invidious/pull/1920#issuecomment-809055543)

syeopite:
Pretty much everything in here needs to be discussed.

Closes #1951 
Closes #571


syeopite:
There's also the question of how we'll distribute the compiled gettext files (.mo).

SamantazFox:
Looks like a good start to me :) the string IDs are coherent, and I really like the fact that you split the thing in multiple files. That will be easier to maintain!

The only thing I don't get id the file `locales/en_US/addons/languages.po`. What is this for? If you're not already aware of that, the long list of languages you removed is the list of available subtitle languages.

syeopite:
That file is for outsourcing the translations of different languages. See #571.

syeopite:
I'm placing this PR on hold for now. See https://github.com/iv-org/invidious/issues/1951#issuecomment-890459484

syeopite:
Everything needs to be rewritten for compatibility with Invidious in 2024

PR #2296, opened 2021-08-02
Fix "fatal role postgres doesn't exist"

SamantazFox:
Fix a frequent error with recent postgres docker images:
`FATAL:  role "postgres" does not exist`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       POSTGRES_PASSWORD: kemal
       POSTGRES_USER: kemal
     healthcheck:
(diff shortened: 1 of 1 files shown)


unixfox:
Healthcheck is working fine:
````
root@invidious:~/invidious# docker-compose ps
        Name                       Command                 State                Ports          
-----------------------------------------------------------------------------------------------
invidious_invidious_1   /invidious/invidious            Up             127.0.0.1:3000->3000/tcp
invidious_postgres_1    docker-entrypoint.sh postgres   Up (healthy)   5432/tcp     
````

PR #2298, opened 2021-08-02
Switch routing logic to use modules

### LINKED ISSUE #2273: [Enhancement] Change route classes to be modules

Currently, each of the route file uses a separate class as a namespace for storing different functions (routes). This is then initialized when building each route for kemal.

https://github.com/iv-org/invidious/blob/095c894548c2aaa884b8d2857dab01207c184345/src/invidious/routing.cr#L4

As of master, the route classes are initialized and allocated almost 40 times. 

As expected, the current method is rather inefficient. So, I propose switching to the crystal object designed for namespaces: modules. Each of the route can then be a class method within their module.

Use structs instead  — and also with class methods.

**Additional context**

syeopite:
Closes #2273

### DIFF
--- a/src/invidious/routes/embed.cr
+++ b/src/invidious/routes/embed.cr
@@ -1,5 +1,5 @@
-class Invidious::Routes::Embed < Invidious::Routes::BaseRoute
-  def redirect(env)
+module Invidious::Routes::Embed
+  def self.redirect(env)
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
 
     if plid = env.params.query["list"]?.try &.gsub(/[^a-zA-Z0-9_-]/, "")
@@ -23,7 +23,7 @@ class Invidious::Routes::Embed < Invidious::Routes::BaseRoute
     env.redirect url
   end
 
-  def show(env)
+  def self.show(env)
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
     id = env.params.url["id"]
 
--- a/src/invidious/routes/preferences.cr
+++ b/src/invidious/routes/preferences.cr
@@ -1,5 +1,5 @@
-class Invidious::Routes::PreferencesRoute < Invidious::Routes::BaseRoute
-  def show(env)
+module Invidious::Routes::PreferencesRoute
+  def self.show(env)
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
 
     referer = get_referer(env)
@@ -9,7 +9,7 @@ class Invidious::Routes::PreferencesRoute < Invidious::Routes::BaseRoute
     templated "preferences"
   end
 
-  def update(env)
+  def self.update(env)
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
     referer = get_referer(env)
 
@@ -219,7 +219,7 @@ class Invidious::Routes::PreferencesRoute < Invidious::Routes::BaseRoute
     env.redirect referer
   end
 
-  def toggle_theme(env)
+  def self.toggle_theme(env)
     locale = LOCALES[env.get("preferences").as(Preferences).locale]?
     referer = get_referer(env, unroll: false)
 
--- a/src/invidious/routing.cr
+++ b/src/invidious/routing.cr
@@ -1,15 +1,13 @@
 module Invidious::Routing
   macro get(path, controller, method = :handle)
     get {{ path }} do |env|
-      controller_instance = {{ controller }}.new
-      controller_instance.{{ method.id }}(env)
+      {{ controller }}.{{ method.id }}(env)
     end
   end
 
   macro post(path, controller, method = :handle)
     post {{ path }} do |env|
-      controller_instance = {{ controller }}.new
-      controller_instance.{{ method.id }}(env)
+      {{ controller }}.{{ method.id }}(env)
     end
   end
 end
--- a/src/invidious/routing.cr
+++ b/src/invidious/routing.cr
@@ -1,13 +1,11 @@
 module Invidious::Routing
-  macro get(path, controller, method = :handle)
-    get {{ path }} do |env|
-      {{ controller }}.{{ method.id }}(env)
-    end
-  end
+  {% for http_method in {"get", "post", "delete", "options", "patch", "put", "head"} %}
 
-  macro post(path, controller, method = :handle)
-    post {{ path }} do |env|
-      {{ controller }}.{{ method.id }}(env)
+    macro {{http_method.id}}(path, controller, method = :handle)
+      {{http_method.id}} \{{ path }} do |env|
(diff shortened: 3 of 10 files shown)


src/invidious/routes/embed.cr
```diff
@@ -1,5 +1,5 @@
-class Invidious::Routes::Embed < Invidious::Routes::BaseRoute
-  def redirect(env)
+module Invidious::Routes::Embed
+  def self.redirect(env)
```
SamantazFox, line 2: Why is `self.` required in modules? I'm not sure to understand that correctly.

src/invidious/routing.cr
```diff
@@ -1,15 +1,13 @@
...
-      controller_instance.{{ method.id }}(env)
+      {{ controller }}.{{ method.id }}(env)
     end
   end
 
   macro post(path, controller, method = :handle)
     post {{ path }} do |env|
-      controller_instance = {{ controller }}.new
-      controller_instance.{{ method.id }}(env)
+      {{ controller }}.{{ method.id }}(env)
```
SamantazFox, line 10: While you're at it, can you add macros for `delete` and `options`? Those will be required in the future ;)

src/invidious/routes/embed.cr
syeopite: Modules cannot be initialized so aren't able to use the defined instance methods of that module. Hence why we have to use class methods.

In PR #2277 you used an `extend self` keyword on that module.

```crystal
module YoutubeAPI
  extend self
```

And [an extend makes a type include methods defined in that module as class methods](https://crystal-lang.org/reference/syntax_and_semantics/modules.html).

SamantazFox: Oooh, yeah, right. So adding `self.` on all methods or `extend self` at the top is basically the same, right?

src/invidious/routes/preferences.cr
```diff
@@ -1,5 +1,5 @@
-class Invidious::Routes::PreferencesRoute < Invidious::Routes::BaseRoute
-  def show(env)
+module Invidious::Routes::PreferencesRoute
```
SamantazFox, line 1: To be consistant with the others, could you rename this one to `Invidious::Routes::Preferences`? `PreferencesRoute` is kinda redundant

src/invidious/routes/embed.cr
syeopite: Yep!

src/invidious/routes/preferences.cr
syeopite: That'll cause a name collision with the global Preferences constant.

SamantazFox: There is no such constant, only the `Preferences` class, but as those are in different namespaces, and will be rewritten soon, that won't be a problem!

syeopite: > There is no such constant, only the `Preferences struct`

Ah that's what I meant thanks!

> those are in different namespaces, and will be rewritten soon, that won't be a problem!

Different namespaces yes, but the `Preferences` struct is global and gets overwritten by the module instead in that scope. So, if I go ahead and rename the `PreferencesRoute` module to `Preferences` then Crystal would think I'm attempting to call the `#from_json` method of the module here:

https://github.com/iv-org/invidious/blob/048eef100d330a83100522e85e9425d2aa921057/src/invidious/routes/preferences.cr#L132-L163 

Instead of the the `Preferences` struct

SamantazFox: Ah, mmh, indeed. we'll take care of that once I've finished my user preferences rewrite PR!

SamantazFox (approved):

syeopite:
During testing, could someone benchmark the startup (and build) times of Invidious with and without this PR? I'm curious on how much of an impact this PR would have.

SamantazFox:
> During testing, could someone benchmark the startup (and build) times of Invidious with and without this PR? I'm curious on how much of an impact this PR would have.

<details>
<summary><b>Without your PR:</b></summary>

```
$ crystal build src/invidious.cr --time --stats --debug --error-trace
Parse:                             00:00:00.027782664 (   4.25MB)
Semantic (top level):              00:00:23.893573504 ( 185.28MB)
Semantic (new):                    00:00:00.014330577 ( 185.28MB)
Semantic (type declarations):      00:00:00.113333246 ( 185.28MB)
Semantic (abstract def check):     00:00:00.021591748 ( 185.28MB)
Semantic (ivars initializers):     00:00:01.331658047 ( 249.40MB)
Semantic (cvars initializers):     00:00:00.012517280 ( 249.40MB)
Semantic (main):                   00:00:10.539234510 ( 603.15MB)
Semantic (cleanup):                00:00:00.013054144 ( 603.15MB)
Semantic (recursive struct check): 00:00:00.004977480 ( 603.15MB)
Codegen (crystal):                 00:00:50.326443426 ( 993.15MB)
Codegen (bc+obj):                  00:02:27.294571605 ( 993.15MB)
Codegen (linking):                 00:00:05.001655404 ( 993.15MB)

Macro runs:
 - /usr/share/crystal/src/ecr/process.cr: 00:00:22.054894423

Codegen (bc+obj):
 - no previous .o files were reused
```
</details>

<details>
<summary><b>With your PR:</b></summary>

```
$ crystal build src/invidious.cr --time --stats --debug --error-trace
Parse:                             00:00:00.032818899 (   4.25MB)
Semantic (top level):              00:00:23.837683292 ( 185.28MB)
Semantic (new):                    00:00:00.013645024 ( 185.28MB)
Semantic (type declarations):      00:00:00.169417005 ( 185.28MB)
Semantic (abstract def check):     00:00:00.022178682 ( 185.28MB)
Semantic (ivars initializers):     00:00:01.278309002 ( 249.40MB)
Semantic (cvars initializers):     00:00:00.016092990 ( 249.40MB)
Semantic (main):                   00:00:10.337811049 ( 603.15MB)
Semantic (cleanup):                00:00:00.013407219 ( 603.15MB)
Semantic (recursive struct check): 00:00:00.005532737 ( 603.15MB)
Codegen (crystal):                 00:00:48.070305318 ( 993.15MB)
Codegen (bc+obj):                  00:02:28.038866572 ( 993.15MB)
Codegen (linking):                 00:00:04.948952902 ( 993.15MB)

Macro runs:
 - /usr/share/crystal/src/ecr/process.cr: 00:00:21.688229431

Codegen (bc+obj):
 - no previous .o files were reused
```
</details>

This was compiled on our test server, running Crystal 1.0.0. The difference is neglictible (I cleaned the cache between each run). One thing I noted, though, is that a recent commit (in the last two weeks or so) greatly improved the compile time of invidious (we were at 5-6 minutes for the `codegen (bc+obj)` step previously).

SamantazFox:
Also, can you please rebase on master? it's missing the `channels.cr` route!

Issue #2074, opened 2021-05-10
[Enhancement] Autodetect browser language


Elaborendum:
**Is your enhancement request related to a problem? Please describe.**

Just a minor (and maybe specific?) incovenience as it is tedious to have to change the language of every instance visited, since some don't work from time to time and to avoid overloading just one instance (plus, they change a lot, one instance might go down or be down for a few days). It becomes specially annoying if you clear cookies constantly.

Maybe it could also help discover that Invidious is actually localized in the user's language.

**Describe the solution you'd like**

Invidious should automatically detect the browser language and display itself in it if available.

**Describe alternatives you've considered**

Whitelist available instances from the cookie deletion and add the new ones as they come.

SamantazFox:
Good idea! We'll see what we can do :)

TheFrenchGhosty:
A 10$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 10$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

TheFrenchGhosty:
A 20$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 20$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

PR #2299, opened 2021-08-02
Determine best default locale using "Accept-Language" header

afrmtbl:
Fixes: #2074

This changes the default locale used for clients without saved preferences, and the default when a new user is created.

A few questions:

1. I'm not sure if these are the only places that need to be changed to take the header into account.
2. I wasn't sure where to put the header parsing code so it is in `utils.cr` for now (?)
3. I'm not very familiar with Crystal so any comments on code quality are welcome.

### DIFF
--- a/src/invidious/helpers/utils.cr
+++ b/src/invidious/helpers/utils.cr
@@ -476,37 +476,39 @@ struct LanguageEntry
   class InvalidLanguageEntry < Exception
   end
 
-  property tag, subtags, quality : Float64
+  property tag : String
+  property region : String?
+  property quality : Float64
 
-  def initialize(@tag : String, @subtags : Array(String), @quality : Float64)
+  def initialize(@tag : String, @region : String?, @quality : Float64)
     @quality = @quality.clamp(0.0, 1.0)
   end
 
   def to_s(io)
-    io << tag
-    unless subtags.empty?
-      io << '-' << @subtags.join '-'
+    io << @tag
+    if @region
+      io << '-' << @region
     end
   end
 
   def self.from_string(language_str : String) : self
-    range_and_quality = language_str.split ';'
+    locale_and_quality = language_str.split ';'
 
-    raise LanguageEntry::InvalidLanguageEntry.new if range_and_quality.empty?
+    raise LanguageEntry::InvalidLanguageEntry.new if locale_and_quality.empty?
 
-    if range_and_quality[1]?
-      quality = parse_quality(range_and_quality[1])
+    if quality_str = locale_and_quality[1]?
+      quality = parse_quality(quality_str)
     else
       quality = 1.00
     end
 
-    language_range = range_and_quality[0]
-    tags = language_range.split '-'
+    locale = locale_and_quality[0]
+    parts = locale.split '-'
 
-    language_tag = tags[0]
-    subtags = tags[1..]
+    tag = parts[0]
+    region = parts[1]?
 
-    return LanguageEntry.new(language_tag, subtags, quality)
+    return LanguageEntry.new(tag, region, quality)
   end
 
   private def self.parse_quality(quality_str : String) : Float64
--- a/src/invidious/helpers/utils.cr
+++ b/src/invidious/helpers/utils.cr
@@ -542,7 +542,7 @@ end
 
 # Alternate matching schemes at https://datatracker.ietf.org/doc/html/rfc4647#section-3
 # Performs a simple exact match search for now
-def first_language_match(available_languages : Set(String), preferred_languages : LanguageEntries) : String?
+def first_language_match(available_languages : Enumerable(String), preferred_languages : LanguageEntries) : String?
   first_lang = preferred_languages.find do |lang|
     available_languages.includes? lang.to_s
   end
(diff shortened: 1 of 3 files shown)


src/invidious/routes/login.cr
```diff
@@ -434,6 +434,16 @@ class Invidious::Routes::Login < Invidious::Routes::BaseRoute
 
         sid = Base64.urlsafe_encode(Random::Secure.random_bytes(32))
         user, sid = create_user(sid, email, password)
+
+        # Set the locale preference to the best matching one in the "Accept-Language" header
+        # Will be overwritten by the PREFS cookie further down if present
+        if language_header = env.request.headers["Accept-Language"]?
+          preferred_langs = parse_accept_language_header(language_header)
+          if first_match = first_language_match(LOCALES.keys.to_set, preferred_langs)
```
syeopite, line 442: > `LOCALES.keys.to_set`

Why convert it to a set? There shouldn't be any duplicates within the locales.

src/invidious/helpers/utils.cr
```diff
@@ -471,3 +471,81 @@ def fetch_random_instance
...
+
+  def to_s(io)
+    io << tag
+    unless subtags.empty?
+      io << '-' << @subtags.join '-'
+    end
+  end
+
+  def self.from_string(language_str : String) : self
+    range_and_quality = language_str.split ';'
```
syeopite, line 493: `range` here refers to a locale right? As in `fr-CH` or `en-GB`? Imo calling it `range` is rather confusing; it should be renamed.

src/invidious/helpers/utils.cr
```diff
@@ -471,3 +471,81 @@ def fetch_random_instance
...
+      quality = parse_quality(range_and_quality[1])
+    else
+      quality = 1.00
+    end
+
+    language_range = range_and_quality[0]
+    tags = language_range.split '-'
+
+    language_tag = tags[0]
+    subtags = tags[1..]
```
syeopite, line 507: Anything beyond the first sub-tag can likely be ignored; Invidious is unlikely to ever localize to anything that specific.

syeopite: Subtags can then be renamed to `territory`

syeopite (changes requested):

src/invidious/helpers/utils.cr
SamantazFox: I can't recall what the MDN docs say on that, but shouldn't be the logic in `pt-BR;fr-FR,0.5;en,0.1` to check if `pt-BR` is available, then `pt | pt-*`, then `fr-FR`, then `fr | fr-*`, and finally `en | en-*`?

syeopite: That's what it currently does.

afrmtbl: Yeah the reason I just did an exact match is because my Accept-Language header was `en-CA,en-US;q=0.7,en;q=0.3`.
So if it was checked as `en-CA -> en -> en-*` it might skip `en-US` -- which is the next best locale quality-wise -- and choose `en` first instead. I guess a drawback would be if the browser doesn't specify a more general locale at the end of the specific ones then Invidious will just return the default of English like before.

I know this is a very unlikely scenario with the locales Invidious actually supports so maybe its not a big deal but it just seemed like confusing behaviour. But it's also possible I'm just misunderstanding.

If there is different matching behaviour desired I wouldn't mind implementing it.

src/invidious/routes/login.cr
afrmtbl: Fixed!

src/invidious/helpers/utils.cr
afrmtbl: Agreed. I was just following the spec language but it's rather confusing.

src/invidious/helpers/utils.cr
afrmtbl: Subtags beyond the first are now ignored and I've changed the name to `region` as in the MDN page. Let me know if you would still prefer `territory` though.

Blacksmoke16:
You could probably leverage https://athenaframework.org/Negotiation/#Athena::Negotiation--language if you don't mind an extra dependency.

src/invidious/helpers/utils.cr
SamantazFox: Ah, it's a bit more complex than what I expected:
https://datatracker.ietf.org/doc/html/rfc4647#section-2.3

SamantazFox:
> You could probably leverage https://athenaframework.org/Negotiation/#Athena::Negotiation--language if you don't mind an extra dependency.

Thanks :)

@syeopite @unixfox @TheFrenchGhosty what do you think of that?

Blacksmoke16:
@TheFrenchGhosty It's essentially a shard that implements https://github.com/iv-org/invidious/pull/2299/files#diff-a0d9774df56c3ae1cd9598b0bbec5d9d6f0ca97e00242bac5fc9f0df69191135R475-R553.

I.e. like in the example, you give it the `accept-header` value and a list of priorities and it matches the best language based on those two things.

syeopite:
https://github.com/athena-framework/negotiation 

Looks pretty nice! I'd say we go for it. It's a bit slower compared with this PR but the matches are more flexible than it.

syeopite:
Superceded by #2324

PR #2324, opened 2021-08-12
Use `athena-negotiation` to detect language through Accept-Language header 

### LINKED ISSUE #2074: [Enhancement] Autodetect browser language

syeopite:
Closes #2074

Follow up to https://github.com/iv-org/invidious/pull/2299#issuecomment-896388300

Please give my bounty of #2074 to @afrmtbl when merging

### DIFF
--- a/src/invidious/routes/login.cr
+++ b/src/invidious/routes/login.cr
@@ -435,6 +435,12 @@ module Invidious::Routes::Login
         sid = Base64.urlsafe_encode(Random::Secure.random_bytes(32))
         user, sid = create_user(sid, email, password)
 
+        if language_header = env.request.headers["Accept-Language"]?
+          if language = ANG.language_negotiator.best(language_header, LOCALES.keys)
+            user.preferences.locale = language.header
+          end
+        end
+
         user_array = user.to_a
         user_array[4] = user_array[4].to_json # User preferences
 
--- a/src/invidious/routes/login.cr
+++ b/src/invidious/routes/login.cr
@@ -434,6 +434,7 @@ module Invidious::Routes::Login
 
         sid = Base64.urlsafe_encode(Random::Secure.random_bytes(32))
         user, sid = create_user(sid, email, password)
+
         user_array = user.to_a
         user_array[4] = user_array[4].to_json # User preferences
 
--- a/shard.lock
+++ b/shard.lock
@@ -1,5 +1,9 @@
 version: 2.0
 shards:
+  athena-negotiation:
+    git: https://github.com/athena-framework/negotiation.git
+    version: 0.1.1
+
   db:
     git: https://github.com/crystal-lang/crystal-db.git
     version: 0.10.1
--- a/shard.yml
+++ b/shard.yml
@@ -25,6 +25,9 @@ dependencies:
   lsquic:
     github: iv-org/lsquic.cr
     version: ~> 2.18.1-2
+  athena-negotiation:
+    github: athena-framework/negotiation
+    version: ~> 0.1.1
 
 crystal: ">= 1.0.0, < 2.0.0"
 
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -17,6 +17,7 @@
 require "digest/md5"
 require "file_utils"
 require "kemal"
+require "athena-negotiation"
 require "openssl/hmac"
 require "option_parser"
 require "pg"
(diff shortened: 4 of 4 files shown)


TheFrenchGhosty:
I don't really understand the difference between athena and just the "normal" guess of the browser language... Isn't the "normal guess" supposed to detect the browser language? What does this do "more"?

syeopite:
The implementation in #2299 does an exact match. So, if we have a header like this: 

`en,fr-CH;q=0.2,zh-Hans-CN;q=0.9,de;q=0.1`

And decide to use that against the languages supported by Invidious

We'll get `de` as that's the only locale that matches exactly. However, in the header both `fr-CH` and  `zh-Hans-CN` have a higher precedence than `de`. So what should actually happen is this:

- Attempt to match `zh-Hans-CN` (highest precedence) 
- Attempt to match `zh-CN`
- If that fails we'll attempt to match `zh-*`
- If that also fails we'll attempt to match `zh`.
- Decent down precedence list to find a match with the above steps.

This is the behavior seen in athena

TheFrenchGhosty:
Perfect. Let's go with this then.

src/invidious/routes/login.cr
```diff
@@ -434,6 +434,13 @@ module Invidious::Routes::Login
 
         sid = Base64.urlsafe_encode(Random::Secure.random_bytes(32))
         user, sid = create_user(sid, email, password)
+
+        if language_header = env.request.headers["Accept-Language"]?
+          if language = ANG.language_negotiator.best(language_header, LOCALES.keys)
+            user.preferences.locale = language.header
+          end
+        end
```
SamantazFox, line 442: Why are you parsing the language header in here?

syeopite: Most of the logic was directly taken from #2299 which parses the language header to update the user preferences here.
https://github.com/iv-org/invidious/blob/3a759bdafbb0077fda62cbb463b232f8675b3c28/src/invidious/routes/login.cr#L438-L445

But you're right, this is unnecessary. The `before_all` section should already update the user preferences

SamantazFox (approved):

afrmtbl:
@syeopite The reason I checked the header again on signup is for users that signup without any previously saved preferences.

The following code creates a new user:
https://github.com/iv-org/invidious/blob/7afa027b9576f012bb27d65c7fa1028bc3ffa250/src/invidious/routes/login.cr#L436
but uses the default user preferences:
https://github.com/iv-org/invidious/blob/75d04a92c8c42d7e334b42d23ae4588523760982/src/invidious/users.cr#L381

the preferences for the new user are then updated, but only if they have a saved `PREFS` cookie:
https://github.com/iv-org/invidious/blob/7afa027b9576f012bb27d65c7fa1028bc3ffa250/src/invidious/routes/login.cr#L462-L469

Without checking the header again during signup, the user will be created with the default locale of `en`.

syeopite:
Yep, I just realized that... I was actually writing a reply (discarded now) to my post above when you posted a comment here. 

Either way, I'll go ahead and add back the extra parsing there.

TheFrenchGhosty:
@afrmtbl to receive your reward either send your BTC address in a comment (preferred), or contact us directly, with a proof that this is indeed you.

afrmtbl:
> @afrmtbl to receive your reward either send your BTC address in a comment (preferred), or contact us directly, with a proof that this is indeed you.

The address is `bc1q2ymxdl6fuwvudwg2qgclnjdj8ftaqkdkyqdgh0`. thanks!

TheFrenchGhosty:
A 20$ bounty has been rewarded to @afrmtbl for this PR.

Transaction (48034f558cc7f2074ef1812f1320f50b8b48c4b7383a7680576b54e9866ef6a2):

Amount sent: 0.00041820 BTC (~17.05€ or ~20$)

Mining fees (5 sat/byte): 0.0000071 BTC (~0.29€)

Issue #2338, opened 2021-08-21
[Bug] API methods not working with username as :ucid


dbr:
**Describe the bug**

Querying channel using the username as the `:ucid` parameter returns a "Could not get channel info" error consistently across multiple instances

**Steps to Reproduce**
Access the `/api/v1/channels/videos/:ucid` API method:
```
https://invidious.snopyta.org/api/v1/channels/videos/BlenderFoundation
https://invidious.namazso.eu/api/v1/channels/videos/BlenderFoundation
```

Same is true for other similar API's with the `:ucid` parameter, including this one which is [~directly copied from this doc](https://github.com/iv-org/documentation/blob/master/API.md#get-apiv1channelsucid))
```
https://invidious.namazso.eu/api/v1/channels/BlenderFoundation
```

It's also true for any username I have tried

All works well if using the long random `UC...` ID string

**Logs**
N/a

**Screenshots**
N/a

**Additional context**
I believe this was working previously sometime last week

unixfox:
Due to the new usage of the internal youtube API we have no way of retrieving the videos of a channel based on his username.

We could expose another endpoint where you can get the ucid of a channel username then use this ucid on the videos API endpoint.

dbr:
Ahh, I see! A method to resolve username to UCID would work perfectly for me

Thanks for the quick responce!

syeopite:
I think we should be working towards a v2 API and deprecating the current one — though since should be done till after the Invidious backend split. There's just far too many issues with it.

unixfox:
> I think we should be working towards a v2 API and deprecating the current one — though we should wait for the backend split first. We can't possibly create a good and consistent v2 API with how integrated the backend and frontend of Invidious are

We still have to fix this breakage or introduce an alternative for working around the issue though.

syeopite:
I'll take care of this in a few hours once I wake up.

TheFrenchGhosty:
A 20$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 10$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

PR #2339, opened 2021-08-22
Add new v1 API endpoint for resolving channel urls

syeopite:
Implementation was done hastily and requires the user to use a full YouTube URL. It's essentially just a proxy towards the InnerTube `resolve_url` endpoint. If requested, I'll go ahead and restrict the input to only the special channel "brand" links.


unixfox:
Can you rebase? Is this PR supposed to fix https://github.com/iv-org/invidious/issues/2338 right?

Issue #2347, opened 2021-08-24
[Bug] Using shortlink redirects to channel instead of video


austinhuang0131:
**Describe the bug**
Generally speaking `/<video_id>` redirects `/watch?v=<video_id>`.

However, somehow, `/<video_id>` was fetched as a channel URL instead in this case.

This affects Privacy Redirect's ability to redirect `youtu.be` links

**Steps to Reproduce**
1. https://redirect.invidious.io/watch?v=dQw4w9WgXcQ redirects to the rickroll video, right?
2. https://redirect.invidious.io/dQw4w9WgXcQ redirects to a channel called "SubToMe" whose [custom channel URL](https://www.youtube.com/c/dQw4w9WgXcQ) is `dQw4w9WgXcQ`

syeopite:
Hmm this is strange. *Technically* Invidious's behavior is right. YouTube returns the first of those links as a `urlEndpoint` while the rest are given as a `watchEndpoint`.

As a result on YouTube the links traverse to:

* https://youtu.be/3EsuRXSv2U
* https://www.youtube.com/watch?v=YZPdI5Y2zCI&t=0s
* https://www.youtube.com/watch?v=o6vQEmMNT3E&t=0s
* https://www.youtube.com/watch?v=Xya6zsX3Hps&t=0s
* https://www.youtube.com/watch?v=gGv2rLBp01c&t=0s

And this is *technically* the behavior seen on Invidious. Just not the one you'd typically expect. Anyhow, #2365 should fix it.

syeopite:
https://github.com/iv-org/invidious/issues/2347#issuecomment-907391702

Closing as this has been fixed by #2365

PR #2365, opened 2021-08-31
Properly transform youtu.be links to be /watch routes in comments and descriptions

syeopite:
Follow up to https://github.com/iv-org/invidious/issues/2347#issuecomment-908884819

### DIFF
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -552,10 +552,9 @@ def content_to_comment_html(content)
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
         url = URI.parse(url)
 
-        if !url.host
-        elsif url.host == "youtu.be"
+        if url.host == "youtu.be"
           url = "/watch?v=#{url.request_target.lstrip('/')}"
-        elsif {"m.youtube.com", "www.youtube.com"}.includes? url
+        elsif !url.host || {"m.youtube.com", "www.youtube.com"}.includes? url
           if url.path == "/redirect"
             url = HTTP::Params.parse(url.query.not_nil!)["q"]
           else
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -552,7 +552,10 @@ def content_to_comment_html(content)
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
         url = URI.parse(url)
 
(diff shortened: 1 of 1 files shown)


src/invidious/comments.cr
```diff
@@ -552,7 +552,10 @@ def content_to_comment_html(content)
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
         url = URI.parse(url)
 
-        if !url.host || {"m.youtube.com", "www.youtube.com", "youtu.be"}.includes? url.host
+        if !url.host
+        elsif url.host == "youtu.be"
+          url = "/watch?v=#{url.request_target.lstrip('/')}"
+        elsif {"m.youtube.com", "www.youtube.com"}.includes? url
           if url.path == "/redirect"
```
syeopite, line 559: Afaik `youtu.be` doesn't actually have a `/redirect` path. It should be fine to remove handling for `youtu.be` here.

src/invidious/comments.cr
```diff
@@ -552,7 +552,10 @@ def content_to_comment_html(content)
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
         url = URI.parse(url)
 
-        if !url.host || {"m.youtube.com", "www.youtube.com", "youtu.be"}.includes? url.host
+        if !url.host
+        elsif url.host == "youtu.be"
+          url = "/watch?v=#{url.request_target.lstrip('/')}"
+        elsif {"m.youtube.com", "www.youtube.com"}.includes? url
```
SamantazFox, line 558:
```suggestion
        elsif {"m.youtube.com", "www.youtube.com"}.includes? url.host
```

syeopite: Ah thanks! Didn't realize I removed it.

src/invidious/comments.cr
```diff
@@ -552,7 +552,10 @@ def content_to_comment_html(content)
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
         url = URI.parse(url)
 
-        if !url.host || {"m.youtube.com", "www.youtube.com", "youtu.be"}.includes? url.host
+        if !url.host
```
SamantazFox, line 555: If I'm not mistaken, `if !url.host` should be handled the same way as if the domain is `*.youtube.com`, in the case where google uses relative links

syeopite: Oops, thanks!

SamantazFox (approved):

PR #2367, opened 2021-08-31
Skip container release for insignficant changes

syeopite:


.github/workflows/container-release.yml
```diff
@@ -4,9 +4,18 @@ on:
...
       - "master"
-  schedule:
-    - cron: 0 0 * * *
-
+    paths-ignore:
+      - "*.md"
+      - LICENCE
+      - TRANSLATION
+      - invidious.service
+      - .git*
```
unixfox, line 12: Isn't this including the .github folder too? What happen if we change something in it which is related to how the docker image is built, does it skip the compilation or not when the changes are pushed?

syeopite: No. The `paths-ignore` only includes files, hence all of the globbing in the directory paths. If you take a look at my previous workflow pull requests, the jobs ran during checks is always from the updated workflow. `.git*` *shouldn't* prevent the workflow files from being ran.

unixfox:
If you rebase this PR I can merge it.

PR #2368, opened 2021-08-31
Make building in release mode optional

syeopite:


unixfox:
I don't have the time right now but I think there is a better way than involving conditions in bash. Like doing that with environment variables.

syeopite:
Yep! CI time is reverted back to how it was before 891116f13bc3afc702c85535cbf252fab6bdfc7f. Feel free to merge it

Issue #2343, opened 2021-08-22
[Bug] Sometimes video speed goes to 9.999


ghost:
**Describe the bug**

Sometimes the video player glitches out and the speed shows as:

```
9.999
9999
3922
529e-
9x
```

**Workaround**
Clicking the speed button again resets it to `0.25x` and you can set it back to `1x`

**Steps to Reproduce**

Not really reproducable, but it happens randomly on videos.

**Screenshots**

![1](https://user-images.githubusercontent.com/66189242/130371664-300021f9-99bb-46e8-8db7-19f7cfbd0d80.jpg)
![21](https://user-images.githubusercontent.com/66189242/130371668-22a27086-4301-4ab1-873e-7c8f9ada2011.jpg)

**Additional context**

This happens on different instances (Currently I see this on `yewtu.be` `snoopyta`,  `invidio.xamh.de`)

- Browser: Snowhaze and Safari (All browsers are basically webkit)
- OS: iPadOS

Issue #2289, opened 2021-07-29
[Bug] Typo in new option in preferences


Elaborendum:
**Describe the bug**

There's a typo in the "Automatic Instance redirection" option in settings. It says "Automaticatic". Is it okay to report this as a bug?

**Steps to Reproduce**
1. Go to the Settings of any instance
2. Scroll down to "Miscellaneous preferences"
3. See typo

**Screenshots**

![imagen](https://user-images.githubusercontent.com/71190696/127439029-f1de8c5f-3284-47ed-8cfd-b25dac093fe2.png)

syeopite:
Yep definitely a bug. Thanks for reporting the issue!

SamantazFox:
Will be fixed next time a PR from weblate is merged.

Issue #2259, opened 2021-07-17
[Bug] `Show Annotations` switches off on Annotation Redirects


madprogramer:
**Describe the bug**

While annotations are shown, clicking on an annotation link will hide annotations on the redirected video. 
In other words, the URL parameter `iv_load_policy=1` (Show Annotations) disappears from redirects.

**Steps to Reproduce**

1. On any instance, go to a video and enable the `Show Annotations` policy. For example, let's take `Drumstick of Destiny` by  `WheezyWaiter`, on the ytprivate.com instance: https://ytprivate.com/watch?v=P9qb_nGTtr8&iv_load_policy=1 ![image](https://user-images.githubusercontent.com/3719664/126038402-ee3c5cda-96c0-4453-88b8-8191e02d1666.png)
2. Skip to around 3:00 in the video and click on the `< Hit Matt with the Stick` option.
3. You should be redirected to https://ytprivate.com/watch?annotation_id=annotation_644056&ei=HcY7XPbYNIGB3LUPqMWr6AM&feature=iv&src_vid=P9qb_nGTtr8&v=Yf98K18a-cw and annotations on `Hit Matt` should be hidden.
Ideally the redirect should have been to https://ytprivate.com/watch?annotation_id=annotation_644056&ei=HcY7XPbYNIGB3LUPqMWr6AM&feature=iv&src_vid=P9qb_nGTtr8&v=Yf98K18a-cw&iv_load_policy=1 with `iv_load_policy=1` passed to the new URL as well to preserve the annotation loading policy.

**Logs**

n/a

**Screenshots**

n/a

**Additional context**

n/a

**Possible Solution**
Intuitively, if the `annotation_id=` parameter is passed to the URL, it was accessed through another annotation. Ergo, it can be assumed that `iv_load_policy` should be set to 1 for all such redirects. In other words, a rewrite rule for adding in `iv_load_policy=1` for URLs which carry the `annotation_id` parameter should probably do the trick.

unixfox:
Can you replicate this issue on other Invidious instances (https://api.invidious.io/) because ytprivate is running a modified version of Invidious?

madprogramer:
can confirm it occurs on snopyta.org as well, feel free to try it elsewhere but it's a main-branch issue

SamantazFox:
Woohoo, yet another URL parameters handling problem xD

This is starting to be a major issue, lol

madprogramer:
Soon :)

[...]

Issue #2173, opened 2021-06-18
[Enhancement] Replace country-language codes to ISO language code in the translations


TheFrenchGhosty:
**Is your enhancement request related to a problem? Please describe.**

bn_BD.json 
en_US.json
hu-HU.json 
nb-NO.json 

All those (and more) could be replaced by non-localized version:

bn.json 
en.json
hu.json 
nb.json

**Describe the solution you'd like**

Simplify everything except for the regions that have actual differences, where the current names should be kept:  pt-BR.json / pt-PT.json / zh-CN.json / zh-TW.json 

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

SamantazFox:
You should have commented at #1951 :)

TheFrenchGhosty:
@SamantazFox Oh I wasn't aware of this issue! But yeah, it's indeed related.

SamantazFox:
I was thinking about that, but that's going to be super complex to do. I think that Weblate won't like it...

PR #2377, opened 2021-09-02
Preferences overhaul

SamantazFox:
This PR aims to fix a lot of issues and imrpove a lot of things related to user preferences.

## Fixes

- Redundant and incoherent URL parameters parsing code

- URL parameters being dropped in some cases:
  * TODO: #1219
  * TODO: #1610
  * TODO: #2130
  * TODO: #2259

- Inneficient backend handling of variables (i.e using enumerates instead of strings)
  * TODO #1992
  * Maybe? #2343

- Miscellanous fixes:
  * TODO: #1992
  * TODO: #2012
  * Maybe? #2044
  * TODO: #2289

## Improvements
 
- Allow a region (YouTube's `gl` parameter) to be passed to change results:
  * TODO: #1279
  * TODO: #2036

- Use full language names instead of ISO codes:
  * TODO: ##571
  * TODO: #2173

- Wording/explanation of the options (add tooltips to describe more precisely what options do)
  * TODO: #2118

- Missing preferences options:
  * TODO: #1450
  * TODO: #1870

- Other:
  * TODO: #606

## To be discussed

- #1290
- Changing other defaults


unixfox:
Related to https://github.com/iv-org/invidious/issues/2236, how are you going to do if 1080p is set as a default video quality but the client only support up to 720p with the video format that have both audio and video in it (like an iOS client)?

SamantazFox:
> Related to #2236, how are you going to do if 1080p is set as a default video quality but the client only support up to 720p with the video format that have both audio and video in it (like an iOS client)?

My idea is to sort videos by descending order from the preferred quality to the worst (and putting the others after) so in theory, the client should select the first supported format in the list.

E.g, with preferred quality = `1080p`, and a video with streams up to `2160p`:

```
1080p  # Theorically un-supported, so client skips it
720p   # Highest quality supported
480p
360p
240p
144p  # Worst available
2160p # Best available
```

src/invidious/user/preferences.cr
```diff
@@ -0,0 +1,757 @@
...
+    def from_s(input : String?) : Themes?
+      return if !input
+
+      case input.downcase
+      when "auto" ; Themes::Auto
+      when "light"; Themes::Light
+      when "dark" ; Themes::Dark
+        # Compatibility with old 'dark_mode' values
+      when "false"; Themes::Light
+      when "true" ; Themes::Dark
```
syeopite, line 550: Ditto

src/invidious/routes/misc.cr
```diff
@@ -5,24 +5,22 @@ class Invidious::Routes::Misc < Invidious::Routes::BaseRoute
...
 
     case preferences.default_home
-    when "Popular"
-      env.redirect "/feed/popular"
-    when "Trending"
-      env.redirect "/feed/trending"
-    when "Subscriptions"
+    when Settings::HomePages::Popular ; env.redirect "/feed/popular"
+    when Settings::HomePages::Trending; env.redirect "/feed/trending"
+    when Settings::UserHomePages::Subscriptions
```
syeopite, line 10: Please use the `then` keyword instead of `;`

`when Settings::HomePages::Popular then env.redirect "/feed/popular"`

src/invidious/config.cr
```diff
@@ -0,0 +1,217 @@
...
+      return false
+    end
+  end
+
+  def self.load
+    # Load config from file or YAML string env var
+    env_config_file = "INVIDIOUS_CONFIG_FILE"
+    env_config_yaml = "INVIDIOUS_CONFIG"
+
+    config_file = ENV.has_key?(env_config_file) ? ENV.fetch(env_config_file) : "config/config.yml"
```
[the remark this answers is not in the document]
SamantazFox: What about also restoring the `-c / --config` option?

src/invidious/routes/misc.cr
SamantazFox: Oh, didn't know about that, thanks! Was it recently added?

SamantazFox:
@github-actions Hey! It's draft, not stale!

PR #2379, opened 2021-09-03
Handle equirectangular projections for VR

syeopite:
Some VR videos on YouTube uses the equirectangular projection type. For example: https://www.youtube.com/watch?v=i6j5pPD2jnY

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -151,7 +151,12 @@ player.on('error', function (event) {
 // Enable VR video support
 if (video_data.vr && video_data.params.vr_mode) {
     player.crossOrigin("anonymous")
-    player.vr({projection: "EAC"});
+    switch (video_data.projection_type) {
+        case "EQUIRECTANGULAR":
+            player.vr({projection: "equirectangular"});
+        default: // Should only be "MESH". But We'll use this as a fallback.
+            player.vr({projection: "EAC"});
+    }
 }
 
 // Add markers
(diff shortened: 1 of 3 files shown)


TheFrenchGhosty:
Tested. It works perfectly and doesn't break the other type of VR videos.

However something I noticed: perspective are really distorted compared to the YouTube player is that just a feature/bug of the lib we are using or an implementation issue on our side?

See those screenshots from the video you linked (I aligned to bottom right side of the video to the same thing):

YouTube:

youtube

Invidious:

invidious

syeopite:
> feature/bug of the lib we are using or an implementation issue on our side?

It's a library issue unfortunately. https://github.com/videojs/videojs-vr/issues/85 

The VR handling in Invidious is really experimental at best; the quality is horrid compared with YouTube,  there is FOV issues, lack of support for 180°, etc. It's stable in that it *works* but I wouldn't really call it ideal.

assets/js/player.js
```diff
@@ -151,7 +151,12 @@ player.on('error', function (event) {
 // Enable VR video support
 if (video_data.vr && video_data.params.vr_mode) {
     player.crossOrigin("anonymous")
-    player.vr({projection: "EAC"});
+    switch (video_data.projection_type) {
+        case "EQUIRECTANGULAR":
+            player.vr({projection: "equirectangular"});
+        default: // Should only be "MESH". But We'll use this as a fallback.
```
syeopite, line 157:
```suggestion
        default: // Should only be "MESH" but we'll use this as a fallback.
```

PR #2380, opened 2021-09-03
Fix invidious showing yt redirect links

syeopite:
I haven't tested this yet but it should resolve the `youtube.com/redirect` issues.

### DIFF
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -481,11 +481,15 @@ def replace_links(html)
     url = URI.parse(anchor["href"])
 
     if {"www.youtube.com", "m.youtube.com", "youtu.be"}.includes?(url.host)
-      if url.path == "/redirect"
-        params = HTTP::Params.parse(url.query.not_nil!)
-        anchor["href"] = params["q"]?
+      if url.host == "youtu.be"
+        url = "/watch?v=#{url.request_target.lstrip('/')}"
       else
-        anchor["href"] = url.request_target
+        if url.path == "/redirect"
+          params = HTTP::Params.parse(url.query.not_nil!)
+          anchor["href"] = params["q"]?
+        else
+          anchor["href"] = url.request_target
+        end
       end
     elsif url.to_s == "#"
       begin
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
Tested.

It works on some videos, but not on others:

Works (description + comments) : https://redirect.invidious.io/watch?v=ghAbmH7To-w

Doesn't work (description): https://redirect.invidious.io/watch?v=pZVpggzQwQ0

src/invidious/comments.cr
```diff
@@ -481,11 +481,15 @@ def replace_links(html)
     url = URI.parse(anchor["href"])
 
     if {"www.youtube.com", "m.youtube.com", "youtu.be"}.includes?(url.host)
-      if url.path == "/redirect"
-        params = HTTP::Params.parse(url.query.not_nil!)
-        anchor["href"] = params["q"]?
+      if url.host == "youtu.be"
+        url = "/watch?v=#{url.request_target.lstrip('/')}"
```
SamantazFox, line 485: You probably don't want to use `request_target` here, as it returns the path and query concatenated.
E.g, for the URL: `youtube.be/abcdef12345?t=20s`, that will give `/abcdef12345?t=20s`, meaning that you'll get double query parts (`?watch=abcdef12345` and `?t=20s`)

src/invidious/comments.cr
```diff
@@ -554,7 +558,7 @@ def content_to_comment_html(content)
 
         if url.host == "youtu.be"
           url = "/watch?v=#{url.request_target.lstrip('/')}"
-        elsif !url.host || {"m.youtube.com", "www.youtube.com"}.includes? url
+        elsif !url.host || {"m.youtube.com", "www.youtube.com"}.includes? url.host
```
SamantazFox, line 561: I've though about that, and we should use `ends_with` instead. As it's mostly used in description and comments, it's not rare to see youtube URLs with the `www` or `m` subdomain missing.

unixfox:
Can we prioritize that because all the links in the description are broken?

TheFrenchGhosty:
Tested again using the previous tests video:

https://redirect.invidious.io/watch?v=ghAbmH7To-w works in comments - not in descriptions

https://redirect.invidious.io/watch?v=pZVpggzQwQ0 doesn't work in description (couldn't find a comment with a link)

TheFrenchGhosty:
Tested again after clearing the instance cache and it works perfectly, I was seeing outdated information.

PR #2390, opened 2021-09-10
Unregister captcha job

syeopite:
Invidious now fully uses InnerTube which lacks captchas*. There's no reason to have the captcha job constantly run in the background anymore. However, in case YouTube does add captchas into InnerTube, the source code of the job hasn't been removed. All this PR does is remove the job from the compiled binary.

\* To a certain degree — reports do indicate that there is some sort of block/rate-limiting but the tolerance is much higher than what YouTube typically has.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -153,10 +153,6 @@ if CONFIG.popular_enabled
   Invidious::Jobs.register Invidious::Jobs::PullPopularVideosJob.new(PG_DB)
 end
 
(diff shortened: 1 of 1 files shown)


unixfox:
Wouldn't it be better to disable it by default with compiler flags?

syeopite:
The compiler should automatically do that. If no references are made to the specific job then it just isn't included in the final binary.

PR #2397, opened 2021-09-12
Add dependency manager script for videojs stuff

syeopite:
This PR adds a script to auto-magically fetch player dependencies, with a design inspired by traditional package managers.


SamantazFox:
I was personally doing it in bash. Imho it's easier to use (only one more command at install/update) and doesn't require to recompile.

syeopite:
> ...and doesn't require to recompile.

No, the script doesn't require recompilation. You can manually run it if you'd like since the compile-time check is really just there for convenience; more often than not, if we're updating the player-dependencies, it's likely some part of Invidious' code would get updated as well. It's just faster (and easier) to not think about updating the VideoJS dependencies.

unixfox:
Can you rebase on master? I can try this PR if you would like.

unixfox:
> I was personally doing it in bash. Imho it's easier to use (only one more command at install/update) and doesn't require to recompile. It's also easier to maintain (much more people are able to do bash than crystal).

The issue with doing it in bash is that outside of docker the user would have to install yq for handling YAML files.

syeopite:
Rebased

unixfox:
It seems like Crystal doesn't support openssl 3 yet (https://www.openssl.org/blog/blog/2021/09/07/OpenSSL3.Final/) so that's why the docker image for ARM is failing because it uses openssl 3: https://pkgs.alpinelinux.org/package/edge/main/x86_64/openssl but not the official crystal docker image.

unixfox:
Just tested and dash preferred dash video quality is broken. I get the lowest video quality even though I forced to 1080p, I get the same issue with "best" quality.

syeopite:
Yeah, I noticed that too. Could you try and downgrade the video.js dependency to `7.10.2` again?

syeopite:
That seemed like the culprit. Should be fixed now

unixfox:
You can safely upgrade to video.js 7.11.0 because I just tested and dash quality selection still works fine until this version.

It's at video.js 7.11.1 that dash quality selection really stop working altogether.

syeopite:
So it's a Firefox issue? That explains how FreeTube is able to utilize a more recent version of video.js then.

SamantazFox:
> The issue with doing it in bash is that outside of docker the user would have to install yq for handling YAML files.

My bash script doesn't involve YAML at all. Bonus point, it can be run in parallel of the compilation.

> It seems like Crystal doesn't support openssl 3 yet (https://www.openssl.org/blog/blog/2021/09/07/OpenSSL3.Final/) so that's why the docker image for ARM is failing because it uses openssl 3: https://pkgs.alpinelinux.org/package/edge/main/x86_64/openssl but not the official crystal docker image.
> 
> We could fix this issue by using the sha1 shell command though.

I don't understand how that simple change fixes the problem. We're using `Digest` at other places, so invidious shouldn't compile altogether on Alpinea atm.

unixfox:
> > It seems like Crystal doesn't support openssl 3 yet (https://www.openssl.org/blog/blog/2021/09/07/OpenSSL3.Final/) so that's why the docker image for ARM is failing because it uses openssl 3: https://pkgs.alpinelinux.org/package/edge/main/x86_64/openssl but not the official crystal docker image.
> > We could fix this issue by using the sha1 shell command though.
> 
> I don't understand how that simple change fixes the problem. We're using `Digest` at other places, so invidious shouldn't compile altogether on Alpinea atm.

That's true and strange, we should investigate a bit more about that.

unixfox:
Need to test video.js 7.17.0

SamantazFox (changes requested):

SamantazFox:
I've tried this, and it works well. @unixfox and @TheFrenchGhosty: should we merge as-is?

TheFrenchGhosty:
@SamantazFox What about your review?

Otherwise, sure.

unixfox:
> I've tried this, and it works well. @unixfox and @TheFrenchGhosty: should we merge as-is?

Actually I just tried the script 2 days ago and it still worked great. It adds more compilation time for the first time so I'll have to add some docker cache for that.

unixfox (approved):

SamantazFox:
@TheFrenchGhosty We can still make changes later. For now those things don't have an impact on the script functionality.

@unixfox I'll probably use the makefile to make that less painful (i.e, run this script along with "shards install").

unixfox:
This script is also great for testing different video.js versions.

I actually tried 2 days to understand why the latest version (v7.17.0) didn't worked. I found something related to video.js using a codec that we haven't tested and thus resulting in a strange corrupted video error. But nothing more.

SamantazFox:
Here, merged.

Issue #2405, opened 2021-09-15
[Bug] non-JS player


TechnicalSuwako:
**Describe the bug**
If you disable JS, the player's right side will spill.

**Steps to Reproduce**
If you disable JS, you'll know.

**Screenshots**
with JS
![with JS](https://some.very.questionable.website/Screenshot_20210915_200248.png)

no JS not yet fix
![no JS not yet fix](https://some.very.questionable.website/Screenshot_20210915_200232.png)

no JS with fix
![no JS with fix](https://some.very.questionable.website/Screenshot_20210915_200545.png)

TechnicalSuwako:
You can see an example in my instance.　→　https://youtube.076.ne.jp/watch?v=ybpDr25fwns

syeopite:
Closed by #2405

Issue #2261, opened 2021-07-18
[Bug] Vertical black bars on Firefox


kurokawa90:
Hi-
1. There seems to be an issue on Firefox with vertical black bars on videos due to odd ratio (I guess?). This is true for all instances tried. However, it works perfectly on Brave the scaling is fine. Is it an issue of settings on Firefox?  
2. Is there a way to choose/customize the size of the player or it is now just one fixed size (could not find anything on wiki about the customization)?
Many thanks!

syeopite:
Duplicate of #1880. But your findings here are rather interesting though. Could you post them over at that thread? Thanks.

Issue #1880, opened 2021-03-10
[Enhancement] Video cover entire player area (ie. No black bars)


stranger-danger-zamu:
**Is your enhancement request related to a problem? Please describe.**
Videos have black bars instead of scaling to cover whatever space they need to fit into the browser window.

**Describe the solution you'd like**
Changing the `/assests/css/default.css` file to make videos cover the space and have no black bars.
Specifically the changes needed to achieve this are:
```css
diff --git a/assets/css/default.css b/assets/css/default.css
index 793295d..8c70960 100644
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -237,11 +237,11 @@ input[type="search"]::-webkit-search-cancel-button {
 
 @media only screen and (max-aspect-ratio: 16/9) {
   .player-dimensions.vjs-fluid {
-    padding-top: 46.86% !important;
+/*     padding-top: 46.86% !important; */
   }
 
   #player-container {
-    padding-bottom: 46.86% !important;
+/*     padding-bottom: 46.86% !important; */
   }
 }
 
@@ -453,7 +453,7 @@ span > select {
 }
 
 .player-dimensions.vjs-fluid {
-  padding-top: 82vh;
+/*   padding-top: 82vh; */
 }
 
 video.video-js {
@@ -463,8 +463,8 @@ video.video-js {
 
 #player-container {
   position: relative;
-  padding-bottom: 82vh;
-  height: 0;
+/*   padding-bottom: 82vh; */
+/*   height: 0; */
 }
 
 .pure-control-group label {
@@ -630,4 +630,4 @@ body.dark-theme {
 
 #filters {
   display: none;
-}
\ No newline at end of file
+}
```

**Describe alternatives you've considered**
Something like a user setting that allows the addition of custom user CSS that gets evaluated last. 

**Additional context**
I just commented out the CSS blocks necessary and haven't done thorough testing to see if it breaks any other features. I did however use Google Chrome's device toolbar to emulate all the devices and checked for any weirdness. From what I can tell it worked perfectly fine. 

Also didn't submit this as a PR since I wasn't sure if this was done for some reason.

TheFrenchGhosty:
Can you share some screenshot of the results? If it works fine you should definitely open a PR.

stranger-danger-zamu:
Black Bars:
![image](https://user-images.githubusercontent.com/70606846/110685140-f71a0380-81ab-11eb-9fdc-1f6f7a026507.png)

No Black Bars:
[...]

PR #2406, opened 2021-09-15
fix non-JS player #2405

TechnicalSuwako:
Issues #2405

### DIFF
--- a/assets/css/player.css
+++ b/assets/css/player.css
@@ -218,6 +218,10 @@ video.video-js {
 
 #player-container {
   position: relative;
+  padding-left: 0;
+  padding-right: 0;
+  margin-left: 1em;
+  margin-right: 1em;
   padding-bottom: 82vh;
   height: 0;
 }


unixfox:
Thank you for this fix. We will need to test it before merging it, especially on others browsers in order to understand if this didn't break how the default video player is displayed on chromium browsers.

SamantazFox:
I can tell you right now that this won't work. the problem lies higher in the HTML hierarchy. Basically, the problem is that the one of the elements containing the player forces an aspect ratio that is not the video's aspect ration. The JS player "solves" that by cropping the video so there are no black bars on the sides.

related #1880, #2261

TechnicalSuwako:
> Thank you for this fix. We will need to test it before merging it, especially on others browsers in order to understand if this didn't break how the default video player is displayed on chromium browsers.

quteborwserで確認しました。
友達はpalemoonかtorbraserで確認しました。

syeopite:
Thanks @TechnicalSuwako!

PR #2428, opened 2021-09-25
Fix (some) issues detected by Ameba

syeopite:
Fixes all issues detected by Ameba with the exception of six, but mainly these 4:

- Lint/UselessAssign - Most of the variables detected is used for ECR rendering
- Metrics/CyclomaticComplexity - Requires a far larger rewrite than I have time for 
- Lint/ShadowingOuterLocalVar - The detected issues are usually fine or intentional. 
- Style/RedundantReturn - Imo, an explicit return is easier to understand. 

Each of the commits fixes a separate issue.


src/invidious/comments.cr
```diff
@@ -73,9 +73,9 @@ def fetch_youtube_comments(id, cursor, format, locale, thin_mode, region, sort_b
   contents = nil
 
   if response["onResponseReceivedEndpoints"]?
-    onResponseReceivedEndpoints = response["onResponseReceivedEndpoints"]
+    on_response_received_endpoints = response["onResponseReceivedEndpoints"]
```
SamantazFox, line 76:
```suggestion
```

src/invidious/comments.cr
```diff
@@ -73,9 +73,9 @@ def fetch_youtube_comments(id, cursor, format, locale, thin_mode, region, sort_b
...
     header = nil
-    onResponseReceivedEndpoints.as_a.each do |item|
+    on_response_received_endpoints.as_a.each do |item|
```
SamantazFox, line 78:
```suggestion
    response["onResponseReceivedEndpoints"].as_a.each do |item|
```

src/invidious/channels/about.cr
```diff
@@ -149,7 +149,7 @@ def get_about_info(ucid, locale)
         end
       end
     end
-    tabs = tabs_json.reject { |node| node["tabRenderer"]?.nil? }.map { |node| node["tabRenderer"]["title"].as_s.downcase }
+    tabs = tabs_json.reject { |node| node["tabRenderer"]?.nil? }.map(&.["tabRenderer"].["title"].as_s.downcase)
```
SamantazFox, line 152: why `["tabRenderer"].["title"]`? Is there a reson for the `.` in here?

src/invidious/helpers/static_file_handler.cr
```diff
@@ -173,7 +173,7 @@ module Kemal
             return
           end
 
-          if @cached_files.sum { |element| element[1][:data].bytesize } + (size = File.size(file_path)) < CACHE_LIMIT
+          if @cached_files.sum(&.[1].[:data].bytesize) + (size = File.size(file_path)) < CACHE_LIMIT
```
SamantazFox, line 176: ditto

src/invidious/routes/channels.cr
```diff
@@ -28,7 +28,7 @@ module Invidious::Routes::Channels
         end
       end
       items = items.select(&.is_a?(SearchPlaylist)).map(&.as(SearchPlaylist))
-      items.each { |item| item.author = "" }
+      items.each(&.author=(""))
```
SamantazFox, line 31: Wouldn't `&.author = ""` (without the parenthesis) be simpler? Afaik, parenthesis aren't required and just add clutter.

src/invidious/users.cr
```diff
@@ -557,15 +557,15 @@ def get_subscription_feed(db, user, max_results = 40, page = 1)
...
+      videos.sort_by!(&.title)
     when "alphabetically - reverse"
-      videos.sort_by! { |video| video.title }.reverse!
+      videos.sort_by!(&.title).reverse!
     when "channel name"
-      videos.sort_by! { |video| video.author }
+      videos.sort_by!(&.author)
     when "channel name - reverse"
-      videos.sort_by! { |video| video.author }.reverse!
+      videos.sort_by!(&.author).reverse!
```
SamantazFox, line 564: Note so self: cherry pick that change onto #2377

src/invidious/helpers/proxy.cr
```diff
@@ -236,7 +236,7 @@ def get_spys_proxies(country_code = "US")
     proxies << {ip: ip, port: port, score: score}
   end
 
-  proxies = proxies.sort_by { |proxy| proxy[:score] }.reverse
+  proxies = proxies.sort_by { |proxy| proxy[:score] }.reverse!
```
SamantazFox, line 237: this should also use `sort_by!` instead of `sort_by`. And also the `(&.[:score])` form, no?

src/invidious/jobs/pull_popular_videos_job.cr
```diff
@@ -16,7 +16,7 @@ class Invidious::Jobs::PullPopularVideosJob < Invidious::Jobs::BaseJob
     loop do
       videos = db.query_all(QUERY, as: ChannelVideo)
         .sort_by(&.published)
```
SamantazFox, line 18: ditto `.sort_by!`

src/invidious/routes/api/manifest.cr
```diff
@@ -47,7 +47,7 @@ module Invidious::Routes::API::Manifest
     end
 
     audio_streams = video.audio_streams
-    video_streams = video.video_streams.sort_by { |stream| {stream["width"].as_i, stream["fps"].as_i} }.reverse
+    video_streams = video.video_streams.sort_by { |stream| {stream["width"].as_i, stream["fps"].as_i} }.reverse!
```
SamantazFox, line 50: ditto `.sort_by!`

src/invidious/users.cr
```diff
@@ -90,20 +88,18 @@ struct Preferences
...
+      if result.empty?
+        CONFIG.default_user_preferences.dark_mode
+      else
+        result
+      end
+    rescue ex
+      if value.read_bool
+        "dark"
+      else
+        "light"
```
SamantazFox, line 102: could you revert this one? it'll be less collisions with #2377 :)

src/invidious/routes/channels.cr
```diff
@@ -55,7 +55,7 @@ module Invidious::Routes::Channels
     end
 
     items, continuation = fetch_channel_playlists(channel.ucid, channel.author, continuation, sort_by)
-    items = items.select { |item| item.is_a?(SearchPlaylist) }.map { |item| item.as(SearchPlaylist) }
+    items = items.select(SearchPlaylist).map { |item| item.as(SearchPlaylist) }
```
SamantazFox, line 58:
```suggestion
    items = items.select(SearchPlaylist).map(&.as(SearchPlaylist))
```

SamantazFox: (I'm pretty sure that it's not even required to use `.as(type)` here)

src/invidious/routes/playlists.cr
```diff
@@ -243,7 +243,7 @@ module Invidious::Routes::Playlists
     if query
       begin
         search_query, count, items, operators = process_search_query(query, page, user, region: nil)
-        videos = items.select { |item| item.is_a? SearchVideo }.map { |item| item.as(SearchVideo) }
+        videos = items.select(SearchVideo).map { |item| item.as(SearchVideo) }
```
SamantazFox, line 246:
```suggestion
        videos = items.select(SearchVideo).map(&.as(SearchVideo))
```

src/invidious/helpers/signatures.cr
```diff
@@ -30,7 +30,7 @@ struct DecryptFunction
 
       case op_body
       when "{a.reverse()"
-        operations[op_name] = ->(a : Array(String), b : Int32) { a.reverse }
+        operations[op_name] = ->(a : Array(String), _b : Int32) { a.reverse }
```
SamantazFox, line 33: What is the point/goal of that change?

src/invidious/playlists.cr
```diff
@@ -51,7 +51,7 @@ struct PlaylistVideo
     if xml
       to_xml(auto_generated, xml)
     else
-      XML.build do |json|
+      XML.build do |xml| # Why was this `json`?
```
SamantazFox, line 54: Seems like it was a copy-pasta bug. Can you please remove the comment?

src/invidious/search.cr
```diff
@@ -62,7 +62,7 @@ struct SearchVideo
     if xml
       to_xml(HOST_URL, auto_generated, query_params, xml)
     else
-      XML.build do |json|
+      XML.build do |xml| # Again, why was this json?
```
SamantazFox, line 65: Could you please remove the comment?

src/invidious/views/watch.ecr
```diff
@@ -184,7 +184,7 @@ we're going to need to do it here in order to allow for translations.
                                 </option>
                             <% end %>
                             <% captions.each do |caption| %>
-                                <option value='{"id":"<%= video.id %>","label":"<%= caption.name %>","title":"<%= URI.encode_www_form(video.title) %>-<%= video.id %>.<%= caption.languageCode %>.vtt"}'>
+                                <option value='{"id":"<%= video.id %>","label":"<%= caption.name %>","title":"<%= URI.encode_www_form(video.title) %>-<%= video.id %>.<%= caption.language_code %>.vtt"}'>
```
SamantazFox, line 187: note to self: cherry pick change for #2377

SamantazFox (changes requested):

src/invidious/comments.cr
syeopite: There's a slight performance hit when you directly access the hash multiple times instead of saving to a variable first.

src/invidious/routes/api/manifest.cr
syeopite: The `sort_by` instead of `sort_by!` is intentional. We shouldn't modify the attributes on the structs directly.

src/invidious/helpers/signatures.cr
syeopite: To mark `b` as unused. Proc notation doesn't seem to support using `_`.

src/invidious/comments.cr
SamantazFox: Whaaat O.o Shouldn't that be optimized by the compiler?

src/invidious/routes/api/manifest.cr
SamantazFox: Ah, right.

SamantazFox (approved):

src/invidious/comments.cr
syeopite: Apparently not. Here's the benchmark I did eariler:

```
Access via variable  58.21M ( 17.18ns) (± 1.30%)  0.0B/op        fastest
      Direct access  31.18M ( 32.07ns) (± 1.22%)  0.0B/op   1.87× slower
```

<details>
<summary>
Source code
</summary>

```crystal
require "benchmark"

hash = {
  "hello" => 10,
  "yeet"  => 20,
  "word"  => 30,
  "omg"   => 40,
  "test"   => {
    "a" => 1,
    "b" => 2,
    "3" => 4,
  },
}

Benchmark.ips do |test|
  test.report("Access via variable") {
    if inner = hash["test"]?
      inner.as(Hash(String, Int32)).each do |k, v|
        {k, v}
      end
    end
  }

  test.report("Direct access") {
    if hash["test"]?
      hash["test"].as(Hash(String, Int32)).each do |k, v|
        {k, v}
      end
    end
  }
end
```

</details>

I mean the performance hit is negligible, but it is still present.

SamantazFox: I added a 3rd option to that benchmark, to be similar to whet we hav in the code, and it seems that this is due to the variable assignment *inside the if statement*:

```crystal
  test.report("Access via variable (2)") {
    if hash["test"]?
      inner = hash["test"]
      inner.as(Hash(String, Int32)).each do |k, v|
        {k, v}
      end
    end
  }
```

```
    Access via variable  19.55M ( 51.16ns) (± 7.05%)  0.0B/op        fastest
Access via variable (2)  10.41M ( 96.03ns) (± 7.09%)  0.0B/op   1.88× slower
          Direct access  10.33M ( 96.82ns) (± 4.87%)  0.0B/op   1.89× slower
```

syeopite: Ah right! I guess I applied an extra layer of optimization there with the if statement... Either way, that method is faster and I've applied it to this PR with commit 36ebdb3.

PR #2429, opened 2021-09-25
Expose config values for database connection pool

syeopite:
Modifying these might help alleviate Invidious's database problems.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -15,6 +15,29 @@ db:
   port: 5432
   dbname: invidious
 
+  ##
+  ## Granular pooling behavior controls
+  ## The following is entirely optional.
+  ##
+
+  ## Initial amount of connections in the pool 
+  #initial_pool_size:
+
+  ## Maximum amount of connections in the pool. 0 is unlimited.
+  #max_pool_size: 0
+
+  ## Amount of connections when idle
+  #max_idle_pool_size: 1 
+
+  ## Amount of seconds to wait if the connection pool is full, and a connection is unavailable. 
+  #checkout_timeout: 5 
+
+  ## Amount of retries when a connection is either lost or can't be established.
+  #retry_attempts: 1
+
+  ## Amount of seconds between each retry
+  #retry_delay: 1
+
 ##
 ## Database configuration using a single URI. This is an
 ## alternative to the 'db' parameter above. If both forms
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -33,7 +33,7 @@ db:
   #checkout_timeout: 5 
 
   ## Amount of retries when a connection is either lost or can't be established.
-  #retry_attempts: 1
+  #retry_attempts: 3
 
   ## Amount of seconds between each retry
   #retry_delay: 1
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -164,6 +164,29 @@ https_only: false
 #pool_size: 100
 
 ##
+## Granular pooling behavior controls for HTTP connections.
+## The following is entirely optional.
+##
+## Warning: This is NOT used to adjust the behavior of the database pool. Please look
+## Inside the DB YAML map for those controls.
+##
+
+## Initial amount of connections in the pool 
+#initial_http_pool_size: 0
+
+## Amount of connections when idle
+#max_idle_http_pool_size: 1
+
+## Amount of seconds to wait if the connection pool is full, and a connection is unavailable. 
+#http_pool_checkout_timeout: 5 
+
+## Amount of retries when a connection is either lost or can't be established.
+#http_pool_retry_attempts: 1
+
+## Amount of seconds between each retry
+#http_pool_retry_delay: 0.2
+
+##
 ## Enable/Disable the use of QUIC (HTTP/3) when connecting
 ## to the youtube API and websites ('youtube.com', 'ytimg.com').
 ## QUIC's main advantages are its lower latency and lower bandwidth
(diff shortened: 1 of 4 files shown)


config/config.example.yml
```diff
@@ -15,6 +15,29 @@ db:
...
+  #max_pool_size: 0
+
+  ## Amount of connections when idle
+  #max_idle_pool_size: 1 
+
+  ## Amount of seconds to wait if the connection pool is full, and a connection is unavailable. 
+  #checkout_timeout: 5 
+
+  ## Amount of retries when a connection is either lost or can't be established.
+  #retry_attempts: 1
```
unixfox, line 36: Should we bump this number to a higher value?

syeopite: Most of the reported issues are related to `checkout_timeout` and not `retry_attempts` so I don't think it'd help that much if we change that. But just in case, I've went ahead and increased it to three, which should hopefully make errors like `DB::PoolRetryAttemptsExceeded` (which is already quite rare) nonexistent.

SamantazFox:
IIRC it was discussed on Matrix: the problem is probably due to the backend youtube client using the `DB:Pool` module rather than the actual database pool configuration.

https://github.com/iv-org/invidious/blob/6d68fbc31d54e2561e00e88ab5006d4ed4da26d7/src/invidious/helpers/utils.cr#L53-L65

syeopite:
Yeah, I realized halfway through replying here https://github.com/iv-org/invidious/pull/2429#discussion_r716019125 that the checkout errors was due to the YouTube pool instead. Hopefully #2430 would help resolve that.

unixfox:
> IIRC it was discussed on Matrix: the problem is probably due to the backend youtube client using the `DB:Pool` module rather than the actual database pool configuration.
> 
> https://github.com/iv-org/invidious/blob/6d68fbc31d54e2561e00e88ab5006d4ed4da26d7/src/invidious/helpers/utils.cr#L53-L65

That's true. The usual issues are about the youtube connection pool not the DB pool, I think we should expose the same parameters in this PR to the yt pool.

SamantazFox:
> That's true. The usual issues are about the youtube connection pool not the DB pool, I think we should expose the same parameters in this PR to the yt pool.

Imo, we should start by doing something like an empty class that inherits from `DB:Pool` so we could better trace errors. E.g:
```crystal
class YTPool < DB::Pool
end
```

And as for the rest, the parameters should be entirely separate from the `db` config option.

config/config.example.yml
```diff
@@ -140,6 +163,29 @@ https_only: false
...
+
+## Amount of seconds to wait if the connection pool is full, and a connection is unavailable. 
+#http_pool_checkout_timeout: 5 
+
+## Amount of retries when a connection is either lost or can't be established.
+#http_pool_retry_attempts: 1
+
+## Amount of seconds between each retry
+#http_pool_retry_delay: 0.2
+
```
syeopite, line 188: Imo, we should move these to their own YAML dict.

src/invidious.cr
```diff
@@ -67,7 +67,17 @@ SOFTWARE = {
   "branch"  => "#{CURRENT_BRANCH}",
 }
 
-YT_POOL = YoutubeConnectionPool.new(YT_URL, capacity: CONFIG.pool_size, use_quic: CONFIG.use_quic)
+YT_POOL = YoutubeConnectionPool.new(YT_URL,
+
+  initial_pool_size: CONFIG.initial_http_pool_size,
+  max_pool_size: CONFIG.pool_size,
+  max_idle_pool_size: CONFIG.max_idle_http_pool_size,
+  checkout_timeout: CONFIG.http_pool_checkout_timeout,
```
syeopite, line 75: Instead of passing all of these by argument, we could just directly call `CONFIG` in the `YTPool` constructor.

config/config.example.yml
SamantazFox: yes, and that way you could re-use the same structure for the DB pool and the HTTP pool

SamantazFox:
imho, we should keep it simple: the pool parameters in the config should only be used for the HTTP pool. If users want to modify the pool config for the DB, they can [pass the pool parameters in the `database_url`](https://crystal-lang.org/reference/database/connection_pool.html#sample).

This will remove a lot of complexity from the config file.

syeopite:
I disagree. The `database_url` isn't exactly language agnostic, we can't just expect instance-maintainers to dig into Crystal's documentation in order to change pooling behavior, or learn how to construct it. 

Allowing both of these to be modified directly in the config file really doesn't add much complexity either. It's just a few more configurable values. Once we nest the pooling configs for the YouTube pool, it'll be perfectly fine.

SamantazFox:
> we can't just expect instance-maintainers to dig into Crystal's documentation in order to change pooling behavior, or learn how to construct it.

I'd expect them to do so if they want to change the pooling behavior.

It's advanced configuration, that is, afaik, specific to Crystal. An extremely small percentage of maintainers will change those options, and this is already possible using URL parameters. We shouldn't directly expose too many options to the rest (the vast majority), or they'll be lost.

> Allowing both of these to be modified directly in the config file really doesn't add much complexity either. It's just a few more configurable values. Once we nest the pooling configs for the YouTube pool, it'll be perfectly fine.

"Just a few" is 10 config options that 99% of the users will never change, which has to be documented, etc...
There are already 62 config options!

unixfox:
We still need this PR.

TheFrenchGhosty:
bump

Issue #2060, opened 2021-05-03
[Enhancement] Allow changing RefreshChannelsJob frequency


imiric:
**Is your enhancement request related to a problem? Please describe.**

`RefreshChannelsJob` currently has a hardcoded 1 minute refresh time:
https://github.com/iv-org/invidious/blob/8bc91ced4fb22437dd8198536b331d76f82dd351/src/invidious/jobs/refresh_channels_job.cr#L61-L62

While it's great that subscriptions are being updated now, refreshing my 348 subscriptions takes about 90s and uses up ~2MBps of downstream bandwidth.

**Describe the solution you'd like**

I'd like to avoid the constant refreshing from using up my bandwidth and don't need this job to run every minute. Ideally there would be an admin page in the web UI that shows the status of all jobs and allows running them manually and configuring their schedule.

**Describe alternatives you've considered**

Since that is a large feature maybe allow changing the schedule at startup via the CLI.
In the meantime, I'll change the hardcoded value and rebuild it, and maybe learn Crystal... :sweat_smile:

Jojonintendo:
That would explain the constant pinging I see on my PiHole too. As I don't even have subscriptions it's not really useful to me, and I'd love to be able to disable this, not necessarily via UI, config.yml would be fine too.

PR #2433, opened 2021-09-26
Decrease channel refresh frequency (1 min -> 1 h)

SamantazFox:
This is a temporary fix to reduce load on instances with many channels and avoid IP being flagged by Google.

### DIFF
--- a/src/invidious/jobs/refresh_channels_job.cr
+++ b/src/invidious/jobs/refresh_channels_job.cr
@@ -58,8 +58,9 @@ class Invidious::Jobs::RefreshChannelsJob < Invidious::Jobs::BaseJob
         end
       end
 
-      LOGGER.debug("RefreshChannelsJob: Done, sleeping for one minute")
-      sleep 1.minute
+      # TODO: make this configurable
+      LOGGER.debug("RefreshChannelsJob: Done, sleeping for one hour")
+      sleep 1.hour
       Fiber.yield
     end
   end


syeopite (changes requested):
Please also modify the backoff timer.

https://github.com/iv-org/invidious/blob/9ba3e1cdb45cc4eafc33c02e773aaee286aac551/src/invidious/jobs/refresh_channels_job.cr#L12

https://github.com/iv-org/invidious/blob/9ba3e1cdb45cc4eafc33c02e773aaee286aac551/src/invidious/jobs/refresh_channels_job.cr#L46-L51

unixfox:
Not sure if it's a good idea to set that to a very long duration. This could have a negative effect on having a subscription feed very slow to update.

I think we should just allow the user to change the frequency but let it to 1min by default.

We never seen anyone being blocked by google for refreshing a lot of channels at the same time per minute. What we seen is people that are already blocked by google because someone in their subnet of IP addresses is doing something malicious.

I don't want to penalise the majority when it's just the minority that is having issues.

syeopite:
> What should I change? The delay is already doubling on error

The problem is that it initially starts off at a single second. And although it still grows exponentially, we're going to be spamming YouTube's RSS endpoint far more often than we need to as we wait for the backoff timer to grow to a decent size.  

>  After thinking about it, we should probably reset that time if no error hapenned for a long time

Perhaps we should look into external shards for that? The job system in Invidious' needs an rework anyway and last time I checked, there are quite a few shards for managing tasks. We really shouldn't be reinventing the wheel with Invidious' job system as our needs grow more complex. 

> Imho, the default shouldn't be 1 minute...

Definitely.

SamantazFox:
> The problem is that it initially starts off at a single second. And although it still grows exponentially, we're going to be spamming YouTube's RSS endpoint far more often than we need to as we wait for the backoff timer to grow to a decent size.

so maybe two minutes?

> Perhaps we should look into external shards for that? The job system in Invidious' needs an rework anyway and last time I checked, there are quite a few shards for managing tasks. We really shouldn't be reinventing the wheel with Invidious' job system as our needs grow more complex.

I'll let you look onto that, as you have far more experience in available crystal shards than most if not all of us ^^

unixfox:
> > This could have a negative effect by having a subscription feed very slow to update.
> 
> This change won't affect the speed at which feeds are updated. it will only change the frequency (how often it's done).

If an error occurs during the refresh Invidious won't try to refresh further until the next time so if 3 errors happen consecutively then it will be 3 hours without any new videos in the subscriptions feeds. That's mainly why Omar decided to have a very short refresh.

> > I think we should just allow the user to change the frequency but let it to 1min by default.
> 
> Imho, the default shouldn't be 1 minute, because on a small instance with a slow connection, running that job so often means that the next iteration starts before the end of the previous one. See #2060.

I don't think that's how it works, it waits before starting another job. At least that's my experience from refreshing thousands of yt channels.

Allowing to customize the time is a good idea but we should let a short duration by default.

> > We never seen anyone being blocked by google for refreshing a lot of channels at the same time per minute. What we seen is people that are already blocked by google because someone in their subnet of IP addresses is doing something malicious.
> 
> The channel refreshing is using XML feeds:
> 
> https://github.com/iv-org/invidious/blob/45f8145e67baf598555b659c7ab924431b2f5d6d/src/invidious/channels/channels.cr#L180
> 
> and the frozen subscription has been an issue for quite some time (see #674).

I don't see any proof that it was caused by a block on the XML yt page, nor that invidious caused this blockage.

SamantazFox:
> If an error occurs during the refresh Invidious won't try to refresh further until the next time so if 3 errors happen consecutively then it will be 3 hours without any new videos in the subscriptions feeds. That's mainly why Omar decided to have a very short refresh.

Nah, there is the backoff timer for that. Basically, if one channel is not refreshed, it will wait `backoff` time before refreshing the next channel. So if 3 errors happen in a row, the backoff will be 8s.

> I don't think that's how it works, it waits before starting another job. At least that's my experience from refreshing thousands of yt channels.

Ah, you're right. Though, only 1 minute between each refresh is a bit too low (too quick).

> Allowing to customize the time is a good idea but we should let a short duration by default.

The default should be good for single-user instances, i.e refresh at a reasonnable rate that won't hammer a slow connection. Only large instances require to (and should) refresh often.

syeopite (changes requested):
Once the backoff timer is increased, I'd say this is good to go. ~ 2 minutes would probably be fine.

706f6f7079646f6f646f6f:
default should not change! or make it a setting...

> Not sure if it's a good idea to set that to a very long duration. This could have a negative effect by having a subscription feed very slow to update.
> 
> I think we should just allow the user to change the frequency but let it to 1min by default.
> 
> We never seen anyone being blocked by google for refreshing a lot of channels at the same time per minute. What we seen is people that are already blocked by google because someone in their subnet of IP addresses is doing something malicious.
> 
> I don't want to penalise the majority when it's just the minority that is having issues.

yes!

SamantazFox:
> Good to merge?

yes!

PR #2434, opened 2021-09-26
Extract data structs to separate file

syeopite:
Depends on #2228


SamantazFox:
While the example data in the spec testing the JSON extraction looks pretty, It'd be more efficient to download JSON extracts directly from youtube [1]. This is for a very simple reason that it will be too complex to maintain over time, and makes the overall spec less useful (while it will help ensure we don't break the extractor code, it won't help detecting youtube changes)

[1]: Similarly to what yt-dlp does; i.e a cron job to get fresh data like once a month. Though, of course, not for data related to premiered videos or lives, as those will "lose" this status after some time.

SamantazFox:
> However you're right. I should just remove the static ones; those were mainly added on a whim after the main `item_renderers` folder, which is meant to document extractor inputs per your suggestion on #2228. Though imho we shouldn't be putting them in Invidious directly. As I mentioned many times on the matrix room before, investing in a proper unofficial "official" InnerTube documentation would be a good idea.

https://github.com/iv-org/youtube-utils/ would be the perfect place to host those.

PR #2444, opened 2021-10-02
Only use the /redirect endpoint when automatically redirecting to another instance

syeopite:


unixfox:
Need testing?

SamantazFox (approved):

SamantazFox:
> In testing on yewtu.be

Nothing reported about it?

unixfox:
Nothing got reported about it.

PR #2451, opened 2021-10-03
Variable postgres username and database name

JuniorJPDJ:
Applied additional changes mentioned in #1678

PS. Could you guys please add `hacktoberfest-accepted` label to this PR?
I'd like to take part in [hacktoberfest](https://hacktoberfest.digitalocean.com/).


config/migrate-scripts/migrate-db-17cf077.sh
```diff
@@ -1,4 +1,7 @@
 #!/bin/sh
 
-psql invidious kemal -c "ALTER TABLE channels ADD COLUMN subscribed bool;"
-psql invidious kemal -c "UPDATE channels SET subscribed = false;"
+[ -z "$POSTGRES_USER" ] && POSTGRES_USER=kemal
```
syeopite, line 3: I'm not too familiar with scripting languages so correct me if I'm wrong here, but why force `POSTGRES_USER` to be kemal? If we're setting it via environmental variables now there shouldn't be any reason to force it as kemal.

JuniorJPDJ: It's not forced. It's fallback value.
When user is not set, it's getting set to kemal as fallback.
The same with db name.

PR #2456, opened 2021-10-04
Fix typo

jorgesumle:
It should say "automatic".


SamantazFox:
this breaks the locales. The key is wrong, but the tranlation is fine

PR #2460, opened 2021-10-05
hrefs: replace HTML.escape w/ URI.encode_www_form

mastihios:
Currently in some href strings URL parameter values are HTML-escaped when they should be URL-encoded instead.

An example of the issue is redirecting a playlist item to another instance. The current href string:

    /redirect?referer=watch?v=yiw6_JakZFc&list=PLFs4vir_WsTxontcYm5ctqp89cNBJKNrs

Because the "referer" value is not URL-encoded, the "list" parameter ends up being its own URL parameter alongside "referer". So the above href ends up redirecting to a URL without the "list" component:

    https://redirect.invidious.io/watch?v=yiw6_JakZFc

This patch replaces HTML.escape in such places with URI.encode_www_form, so for example the above href string ends up being:

    /redirect?referer=watch%3Fv%3Dyiw6_JakZFc%26list%3DPLFs4vir_WsTxontcYm5ctqp89cNBJKNrs

...which correctly redirects to:

    https://redirect.invidious.io/watch?v=yiw6_JakZFc&list=PLFs4vir_WsTxontcYm5ctqp89cNBJKNrs

### DIFF
--- a/src/invidious/views/playlists.ecr
+++ b/src/invidious/views/playlists.ecr
@@ -96,7 +96,7 @@
     <div class="pure-u-1 pure-u-md-4-5"></div>
     <div class="pure-u-1 pure-u-lg-1-5" style="text-align:right">
         <% if continuation %>
-            <a href="/channel/<%= ucid %>/playlists?continuation=<%= continuation %><% if sort_by != "last" %>&sort_by=<%= HTML.escape(sort_by) %><% end %>">
+            <a href="/channel/<%= ucid %>/playlists?continuation=<%= continuation %><% if sort_by != "last" %>&sort_by=<%= URI.encode_www_form(sort_by) %><% end %>">
                 <%= translate(locale, "Next page") %>
             </a>
         <% end %>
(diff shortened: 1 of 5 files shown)


mastihios:
Just to add another example of the current issue:

1. search for "abc=123&def=456"
2. click on the "Next Page" link at the bottom

You end up on `/search?q=abc=123&def=456&page=2`, which turns the query into "abc=123".

src/invidious/views/playlists.ecr
```diff
@@ -96,7 +96,7 @@
     <div class="pure-u-1 pure-u-md-4-5"></div>
     <div class="pure-u-1 pure-u-lg-1-5" style="text-align:right">
         <% if continuation %>
-            <a href="/channel/<%= ucid %>/playlists?continuation=<%= continuation %><% if sort_by != "last" %>&sort_by=<%= HTML.escape(sort_by) %><% end %>">
+            <a href="/channel/<%= ucid %>/playlists?continuation=<%= continuation %><% if sort_by != "last" %>&sort_by=<%= URI.encode_www_form(sort_by) %><% end %>">
```
syeopite, line 99: Why were we even encoding the sort_by value? It's just plain text. Definitely something for another PR...

syeopite (approved):
Looks good to me. Thanks @mastihios!

PR #2469, opened 2021-10-06
limit feeds and delete materialized views

unixfox:
Recreated from https://github.com/iv-org/invidious/pull/2058 and rebased.


999eagle:
fwiw I'm currently testing this PR on my Invidious instance (https://iv.catgirl.cloud) with the database index created manually. It's reducing the write load on my database a lot as some of the larger views are refreshed and thus rewritten very often. So once automatic migrations are done I hope this PR can be revisited!

Fijxu:
Applied this commit on my instance 12 hours ago. Works well and this is really really useful for selfhosted instances like mine. Invidious was literally killing my SSD writing a lot of GiB per hour, now it's just normal. 

image

Issue #1690, opened 2021-01-21
Expose version by default on api/v1/stats


Perflyst:
The version is public of the invidious instance is public in the footer anyways and for the instance list it would be nice if every instance shows its version

FireMasterK:
We have to think about the impact for private instances too, which are a lot more in number.

Perflyst:
Which impact?

SamantazFox:
You mean that `statistics_enabled` wouldn't affect the display of `"software":{...}` on `/api/v1/stats` ?

TheFrenchGhosty:
A 10$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 10$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

PR #2482, opened 2021-10-08
always show software

706f6f7079646f6f646f6f:
makes it show the software in the error

send the money from #1690 to bitcoin:1Archive1n2C579dMsAu3iC6tWzuQJz8dN

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -142,7 +142,7 @@ if CONFIG.decrypt_polling
 end
 
 if CONFIG.statistics_enabled
-  Invidious::Jobs.register Invidious::Jobs::StatisticsRefreshJob.new(PG_DB, SOFTWARE)
+  Invidious::Jobs.register Invidious::Jobs::StatisticsRefreshJob.new(PG_DB)
 end
 
 if (CONFIG.use_pubsub_feeds.is_a?(Bool) && CONFIG.use_pubsub_feeds.as(Bool)) || (CONFIG.use_pubsub_feeds.is_a?(Int32) && CONFIG.use_pubsub_feeds.as(Int32) > 0)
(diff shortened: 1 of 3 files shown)


706f6f7079646f6f646f6f:
like https://github.com/iv-org/invidious/blob/master/src/invidious/jobs/statistics_refresh_job.cr#L4 but only sofware

SamantazFox:
Hello! Sorry to say that, but In the current state, your PR can't be merged.

The reason why we put a bounty on that PR is that it requires quite a bit of work to ensure that the JSON structure stays the same whenever the stats are enabled or not. This also implies that this endpoint should never return an error, and instead indicate, by some boolean value or some status message, that the stats are disabled.

706f6f7079646f6f646f6f:
like this maybe?

706f6f7079646f6f646f6f:
@SamantazFox no good? why not? structure is always good now and status now too

syeopite:
Lack of an review doesn't mean the PR isn't "good". We're quite busy! Please have some patience until one of us gets to reviewing it :)

706f6f7079646f6f646f6f:
ok i wait

706f6f7079646f6f646f6f:
💩💩💩💩💩💩💩💩💩💩💩

706f6f7079646f6f646f6f:
BOUNTY but NO REVIEW

fucking RETACULOUS

so important YOU PAY MONEY - then NO MEREGE

@unixfox @TheFrenchGhosty @SamantazFox @syeopite @saltycrys @invidious-bot

TheFrenchGhosty:
@706f6f7079646f6f646f6f We don't owe you anything.

The team is busy so be patient.

PR #2490, opened 2021-10-11
Fix `Role "kemal" does not exist`

### LINKED ISSUE #2485: [Bug] Role "kemal" does not exist
**Describe the bug**

**Steps to Reproduce**
!-- Steps to reproduce the behavior:
1.  Follow the installation guide
2. When trying to run you get "postgres_1   | 2021-10-09 11:14:55.553 UTC [968] FATAL:  role "kemal" does not exist"

**Logs**

postgres_1   | 2021-10-09 11:14:55.553 UTC [968] FATAL:  role "kemal" does not exist

**Screenshots**

**Additional context**

SamantazFox:
Fixes #2485

For some reason, `pg_isready` required a DB name, and by default will try with a database that has the same name as the current user.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       POSTGRES_PASSWORD: kemal
       POSTGRES_USER: kemal
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U $$POSTGRES_USER"]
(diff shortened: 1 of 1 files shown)


unixfox:
Good PR, that's right it tries to use "kemal" as a database name.

PR #2492, opened 2021-10-11
Change <input value="..."> encoding to HTML.escape

mastihios:
In some places the value string of `<input>` elements is URL-encoded:

```
<input type="hidden" name="csrf_token" value="<%= URI.encode_www_form(csrf_token) %>">
```

Browsers already URL-encode these strings on submit, so the above results in double URL-encoding.

Invidious should just make sure that these values don't mess up HTML:

```
<input type="hidden" name="csrf_token" value="<%= HTML.escape(csrf_token) %>">
```

(This is the inverse of #2460 which concerned href strings.)

### DIFF
--- a/src/invidious/views/authorize_token.ecr
+++ b/src/invidious/views/authorize_token.ecr
@@ -72,7 +72,7 @@
                 <input type="hidden" name="expire" value="<%= expire %>">
             <% end %>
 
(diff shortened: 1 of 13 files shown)


SamantazFox:
bump

SamantazFox:
deployed on https://test.invidious.io
ping @unixfox @TheFrenchGhosty

PR #2510, opened 2021-10-14
Link to invidious.io/donate/ in the footer

TheFrenchGhosty:
Link to invidious.io/donate/ in the footer instead of having only *some* of the donation addresses listed.

The donate page will always be on /donate/, even if we change things, so it's 100% future proof.

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -426,7 +426,7 @@
     "next_steps_error_message": "After which you should try to: ",
     "next_steps_error_message_refresh": "Refresh",
     "next_steps_error_message_go_to_youtube": "Go to YouTube",
-    "footer_donate": "Donate: ",
+    "footer_donate_page": "Donate: ",
     "footer_documentation": "Documentation",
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
@syeopite Won't this be handled on Weblate itself? Wasn't this automated (or do I have to change the en_US translations)?

syeopite:
Weblate can't modify the actual translated strings. In this case, we want to change:

https://github.com/iv-org/invidious/blob/adc12addfa1a78ac27f0e61bb6a594e56ac5f035/locales/en-US.json#L429

to
```json
    "footer_donate": "Donate",
```

Though, it may be better to just create a new key and remove the old one:

```json
    "footer_donate_page": "Donate",
```

So our translators can propagate the change across the different languages Invidious supports.

SamantazFox:
> Yes, to my knowledge Weblate will automatically propagate the deletion/addition.

For now, it only propagates additions. I haven't set what is required for auto-deletion.

PR #2514, opened 2021-10-14
Propagate removal of footer_donate

syeopite:
Follow up to https://github.com/iv-org/invidious/pull/2510#issuecomment-943729037

(And this is why we should wait for the code owners' approval in their respective sections...)


TheFrenchGhosty:
Will requires a weblate reset after being merged.

SamantazFox:
I'll enable the auto-cleanup feature of weblate. That will be easier than doing PRs here.

PR #2527, opened 2021-10-21
Bump crystal version for ARM docker container to 1.2.0-r0

SamantazFox:
https://github.com/iv-org/invidious/runs/3955838103


unixfox:
Waiting for testing on my side, and we will be able to bump the crystal version on the x86 docker image too.

unixfox:
Can you upgrade to crystal 1.2.0 too here: https://github.com/iv-org/invidious/blob/master/docker/Dockerfile#L1?

Thank you

PR #2538, opened 2021-10-26
Retain video time position in video player

### LINKED ISSUE #2024: [Feature request] Retain player position.
On 1h+ videos, having to refresh or re-open the browser makes you lose the position.

Keep the last passed timestamp ±10s/20s in browser-local storage.

Keeping a timestamp journal.

***
Edit, to not send out a notification:  
`t=$(t)s` would be great! It enables faster sharing / linking as well — user can copy right away, and quickly remove the tail themselves.

Though, I imagine someone out there would be offended by their links being cluttered, an user switch, defaulting to on, might be a good idea.

bbielsa:
Closes #2024 

1. Add a checkbox toggle in the preferences page
2. Add `remember_position` field to `Preferences` and `VideoPreferences` structs
2. Modify player.js to save and load the current time stamp to an object in local storage to maintain the video time in a long term location

User preferences page checkbox for this new setting:
Screen Shot 2021-10-25 at 9 04 12 PM

Users video times are stored in a JSON object in local storage, the object has this structure:
```javascript
{
  "<video_id>": seconds,
  "jNQXAC9IVRw": 3
}
```
Screen Shot 2021-10-25 at 9 11 26 PM

The time stamp is only updated once per second while the video is playing, to reduce unnecessary IO.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -38,7 +38,7 @@ embed_url.searchParams.delete('v');
 short_url = location.origin + '/' + video_data.id + embed_url.search;
 embed_url = location.origin + '/embed/' + video_data.id + embed_url.search;
 
-var remember_position_key = "remember_position";
+var save_player_pos_key = "save_player_pos";
 
 var shareOptions = {
     socials: ['fbFeed', 'tw', 'reddit', 'email'],
@@ -201,7 +201,7 @@ if (video_data.premiere_timestamp && Math.round(new Date() / 1000) < video_data.
     player.getChild('bigPlayButton').hide();
 }
 
-if (video_data.params.remember_position) {
+if (video_data.params.save_player_pos) {
     const remeberedTime = get_video_time();
     let lastUpdated = 0;
 
@@ -384,12 +384,12 @@ function get_video_time() {
 function set_all_video_times(times) {
     const json = JSON.stringify(times);
 
-    localStorage.setItem(remember_position_key, json);
+    localStorage.setItem(save_player_pos_key, json);
 }
 
 function get_all_video_times() {
     try {
-        const raw = localStorage.getItem(remember_position_key);
+        const raw = localStorage.getItem(save_player_pos_key);
         const times = JSON.parse(raw);
 
         return times || {};
@@ -400,7 +400,7 @@ function get_all_video_times() {
 }
 
 function remove_all_video_times() {
-    localStorage.removeItem(remember_position_key);
+    localStorage.removeItem(save_player_pos_key);
 }
 
 function set_time_percent(percent) {
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -38,6 +38,8 @@ embed_url.searchParams.delete('v');
 short_url = location.origin + '/' + video_data.id + embed_url.search;
 embed_url = location.origin + '/embed/' + video_data.id + embed_url.search;
 
+var remember_position_key = "remember_position";
+
 var shareOptions = {
     socials: ['fbFeed', 'tw', 'reddit', 'email'],
 
@@ -199,6 +201,27 @@ if (video_data.premiere_timestamp && Math.round(new Date() / 1000) < video_data.
     player.getChild('bigPlayButton').hide();
 }
 
+if (video_data.params.remember_position) {
+    const remeberedTime = get_video_time();
+    let lastUpdated = 0;
+
+    set_seconds_after_start(remeberedTime);
+
+    player.on("timeupdate", e => {
+        const raw = player.currentTime();
+        const time = Math.floor(raw);
+
+        if(lastUpdated !== time) {
+            save_video_time(time);
+            lastUpdated = time;
+        }
+    });
+}
+else {
+    console.log("Removing data for remebered positions");
+    remove_all_video_times();
+}
+
 if (video_data.params.autoplay) {
     var bpb = player.getChild('bigPlayButton');
     bpb.hide();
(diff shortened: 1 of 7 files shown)


src/invidious/views/preferences.ecr
```diff
@@ -116,8 +116,15 @@
...
 
+            <div class="pure-control-group">
+                <label for="save_player_pos"><%= translate(locale, "preferences_save_player_pos_label") %></label>
+                <input name="save_player_pos" id="save_player_pos" type="checkbox" <% if preferences.save_player_pos %>checked<% end %>>
+            </div>
+
             <legend><%= translate(locale, "preferences_category_visual") %></legend>
 
+            <legend><%= translate(locale, "Visual preferences") %></legend>
+
```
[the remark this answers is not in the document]
bbielsa: Finally figured it out, fixed

bbielsa:
@unixfox Fixed, thanks

SamantazFox (approved):

unixfox:
Is the bug on Safari fixed?

SamantazFox:
> Is the bug on Safari fixed?

yes: https://github.com/iv-org/invidious/pull/2538/commits/7e007da7e5d3cae30b85f37f3b6155f87759085e

unixfox:
Great it's on by default again on https://yewtu.be

unixfox:
There is an issue related to the `t` parameter, a user reported me this issue:
> whenever i visit a video with the &t=10 appended to it, it doesn't go to 10 seconds in, it just goes 10 seconds further from where i was before i refreshed the page

bbielsa:
> > whenever i visit a video with the &t=10 appended to it, it doesn't go to 10 seconds in, it just goes 10 seconds further from where i was before i refreshed the page

I think the logical thing to do in this case is that we go with the value passed in through `t` and ignore the saved time stamp. Do you agree @unixfox?

unixfox:
> > > whenever i visit a video with the &t=10 appended to it, it doesn't go to 10 seconds in, it just goes 10 seconds further from where i was before i refreshed the page
> 
> I think the logical thing to do in this case is that we go with the value passed in through `t` and ignore the saved time stamp. Do you agree @unixfox?

Yes that's the correct way of handling that :).
And that's how the user that reported the bug thought too.

syeopite:
Potential alternative

> Stores volume, muted and playback rate to local storage to be restored in the next session.

https://github.com/mister-ben/videojs-persist

bbielsa:
@syeopite @unixfox Unfortunately this does not store the current video time, and I believe most of the information that videojs-persist stores is already stored in the user preferences: volume and default speed (although they are global settings and not video specific).

That being said this does look like a nice plugin and I am not against forking it into iv-org and adding the video time saving feature

bbielsa:
In the newest commit, the player does not use the saved playback position if the `t=` parameter is set in the URL. It will still save the playback position so if the user goes back at a later time **without** the `t=` parameter it will pick up where they left off.

Let me know if you had another method in mind for fixing this issue @unixfox

SamantazFox:
> In the newest commit, the player does not use the saved playback position if the `t=` parameter is set in the URL. It will still save the playback position so if the user goes back at a later time **without** the `t=` parameter it will pick up where they left off.
> 
> Let me know if you had another method in mind for fixing this issue @unixfox

What if the user has the `t=` paramter in the URL (because they started watching the video from a URL with that parameter) and then refreshes the page after a few minutes of watchtime? they're brought back to the timestamp from the `t=` parameter?

bbielsa:
>they're brought back to the timestamp from the t= parameter?

Yes, I don't think there's any other way to go about it, because if the `t=` is in the url then the user could be explicitly asking for the player to skip to that time stamp.

unixfox:
> > In the newest commit, the player does not use the saved playback position if the `t=` parameter is set in the URL. It will still save the playback position so if the user goes back at a later time **without** the `t=` parameter it will pick up where they left off.
> > Let me know if you had another method in mind for fixing this issue @unixfox
> 
> What if the user has the `t=` paramter in the URL (because they started watching the video from a URL with that parameter) and then refreshes the page after a few minutes of watchtime? they're brought back to the timestamp from the `t=` parameter?

That's correct and that's the current behavior of invidious so I don't think it's an issue.

assets/js/player.js
```diff
@@ -199,6 +201,32 @@ if (video_data.premiere_timestamp && Math.round(new Date() / 1000) < video_data.
     player.getChild('bigPlayButton').hide();
 }
 
+if (video_data.params.save_player_pos) {
+    const url = new URL(location);
+    const hasTimeParam = url.searchParams.has("t");
```
SamantazFox, line 206: I'm not sure why you made a separate variable for this?

assets/js/player.js
```diff
@@ -199,6 +201,32 @@ if (video_data.premiere_timestamp && Math.round(new Date() / 1000) < video_data.
...
+    }
+
+    const updateTime = () => {
+        const raw = player.currentTime();
+        const time = Math.floor(raw);
+
+        if(lastUpdated !== time) {
+            save_video_time(time);
+            lastUpdated = time;
+        }
```
SamantazFox, line 221: Btw, I recall that we discussed that somewhere, but can't find it: Could it be better to store the current time by increment of 5 seconds?

unixfox:
I rebased your commits against master.

SamantazFox:
@unixfox There was a merge conflict? Also, is more testing required?

unixfox:
> @unixfox There was a merge conflict? Also, is more testing required?

Yeah with the  locales/en-US.json file.

Everything is working fine on my side, I think it's safe to merge unless you want to wait for @bbielsa to reply to your review.

SamantazFox:
> Everything is working fine on my side, I think it's safe to merge unless you want to wait for @bbielsa to reply to your review.

They haven't replied for almost one month. I think we should merge now as the PR is perfectly functional, and we'll make adjustments later :)

unixfox (approved):

TheFrenchGhosty:
@bbielsa to receive your reward either send your BTC address in a comment (preferred), or contact us directly, with a proof that this is indeed you.

PR #2552, opened 2021-10-31
Kubernetes: Replace Helm chart with plain Kubernetes deployment

minus7:
I wanted to deploy the Helm chart but noticed that it was using an outdated chart repo as well as a deprecated Postgres image. It also had some features you'd need in a big deployment (pod autoscaler, but not providing a similar mechanism for the database), but was missing some things you'd want for personal deployments:

- ease of use: the chart did *not* run out of the box with the instructions provided in the readme
- reverse proxy/Ingress setup: not hard to do, but easier if you have a working starting point that you ideally don't need to modify
- automatic updates: indispensible for low-maintenance setups and against install-once-never-update

…so I changed it to a plain deployment and a pod for Postgres and dropped Helm. The above features are in there and the readme should provide all the necessary steps.

The database connection config is a bit cumbersome since Invidious does not support passing the database user/password/etc as separate environment variables.

I run this (on a k3s single node cluster) with a slightly different Ingress setup with Traefik's IngressRoute to do HTTP→HTTPS redirects which the Ingress in this PR does not do, but I didn't find a way to make it work without depending on Traefik.


kubernetes/README.md
```diff
@@ -1,41 +1,63 @@
-# Invidious Helm chart
+# Invidious Kubernetes Deployment
```
TheFrenchGhosty, line 1: This whole file should go in the documentation in my opinion.

I'll let @unixfox decide if that's the case (and review this PR).

unixfox:
Mmh dropping helm... I'm against it. I would prefer to explain how to generate plain old kubernetes yaml files from an helm chart.

Moreover, I've some plans to use k8s-at-home helm common chart in order to rewrite the invidious helm chart. Just haven't had the time to work on it.

minus7:
> Mmh dropping helm... I'm against it. I would prefer to explain how to generate plain old kubernetes yaml files from an helm chart.
> 
> Moreover, I've some plans to use k8s-at-home helm common chart in order to rewrite the invidious helm chart. Just haven't had the time to work on it.

Helm does make installing application a "one click" operation, which seems useful at first, but then you end up having to put the application's config into Helm in a needlessly abstracted form. Helm templates usually end up a big and complex mess proportional to the choices the user gets.
I prefer having a couple of manually executed steps, where you can easily understand the whole setup, and modify it by copying the YAML files. The deployment for Invidious is unlikely to change, so upgrading isn't an issue.

I fully expected this change to not be met with much enthusiasm. I mostly wrote it for myself and wanted to share it in case anyone wants to make use of it.
I could add the plain deployment in a separate directory and maybe try to fix the Helm chart, if you'd like.

SamantazFox:
> I could add the plain deployment in a separate directory and maybe try to fix the Helm chart, if you'd like.

Yeah, that could be a good idea, imo.

@unixfox what do you think about having plain k8s and helm alongside eachother?
Maybe also move everythin into a dedicated `deploy` folder, like so?
```
deploy/
-- kubernetes
-- helm
-- docker
-- (others, if we make packages?)
```

unixfox:
> > I could add the plain deployment in a separate directory and maybe try to fix the Helm chart, if you'd like.
> 
> Yeah, that could be a good idea, imo.
> 
> @unixfox what do you think about having plain k8s and helm alongside eachother? Maybe also move everythin into a dedicated `deploy` folder, like so?
> 
> ```
> deploy/
> -- kubernetes
> -- helm
> -- docker
> -- (others, if we make packages?)
> ```

Like I said in https://github.com/iv-org/invidious/pull/2552#issuecomment-956869575 I much prefer to only have one helm chart and then explain how to generate plain k8s files from the helm chart if someone doesn't want to use helm.

lenaxia:
> Mmh dropping helm... I'm against it. I would prefer to explain how to generate plain old kubernetes yaml files from an helm chart.
> 
> Moreover, I've some plans to use k8s-at-home helm common chart in order to rewrite the invidious helm chart. Just haven't had the time to work on it.

Do you still have plans to do this @unixfox? I'm working on deploying this, and can't say I'm a huge fan of the state of the current helm chart. Happy to do my deployment in k8s-at-home (now just bjw-s') and provide the sample there.

unixfox:
> > Mmh dropping helm... I'm against it. I would prefer to explain how to generate plain old kubernetes yaml files from an helm chart.
> > Moreover, I've some plans to use k8s-at-home helm common chart in order to rewrite the invidious helm chart. Just haven't had the time to work on it.
> 
> Do you still have plans to do this @unixfox? I'm working on deploying this, and can't say I'm a huge fan of the state of the current helm chart. Happy to do my deployment in k8s-at-home (now just bjw-s') and provide the sample there.

Sure feel free to drop a PR for using bjw-s as a base.

Issue #2319, opened 2021-08-10
[Bug] Unexpected char 'C' at line 1 - Youtube is now serving csv file instead of json file when exporting the subscriptions


dhuschde:
TEMPORARY SOLUTION: https://github.com/iv-org/invidious/issues/2319#issuecomment-900358678

-------

YouTube gives a csv file instead of a json file when exporting the subscriptions.

------

**Describe the bug**
I sadly can't import my YT Subs to Invie

**Steps to Reproduce**
Import your YT Subs to Invidious while having Google on German

**Logs**
Title: `Unexpected char 'K' at line 1, column 1 (JSON::ParseException)`
Date: `2021-08-10T16:16:31Z`
Route: `/data_control?referer=%2Ffeed%2Fpopular`
Version: `2021.07.27-5c76cdaa @ master`
<details>
<summary>Backtrace</summary>
<p>
   
```
Unexpected char 'K' at line 1, column 1 (JSON::ParseException)
  from /usr/share/crystal/src/json/lexer.cr:371:7 in 'raise'
  from /usr/share/crystal/src/json/lexer.cr:367:5 in 'unexpected_char'
  from /usr/share/crystal/src/json/lexer.cr:366:11 in 'unexpected_char'
  from /usr/share/crystal/src/slice.cr:229:5 in 'next_token'
  from /usr/share/crystal/src/json/parser.cr:13:12 in '->'
  from lib/kemal/src/kemal/route.cr:12:9 in '->'
  from /usr/share/crystal/src/primitives.cr:255:3 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:112:11 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
[...]

PR #2621, opened 2021-11-15
Implement importing Youtube subscriptions from CSV

0x24D:
Thought I would take a stab at #2319 as it seemed like a good first issue. This is my first time using Crystal so there may be a better way of doing this?

I don't have a YouTube account so I tested the CSV import using the export linked in [this comment](https://github.com/iv-org/documentation/pull/55#issuecomment-897172424). As I don't have an account I also haven't tested the added JSON check, on line 833, but given a correctly formatted file I believe it should still work.

Fixes: #2319


0x24D:
Ah, didn't see #2545 before. Closing.

PR #2626, opened 2021-11-17
podman

706f6f7079646f6f646f6f:


unixfox:
Closing as there is no need to do that if the docker hub registry is configured as a default registry like on Docker:
- https://halukkarakaya.medium.com/how-to-configure-default-search-registries-in-podman-ea930289692
- https://www.redhat.com/sysadmin/manage-container-registries

706f6f7079646f6f646f6f:
> Closing as there is no need to do that if the docker hub registry is configured as a default registry like on Docker:
> 
>     * https://halukkarakaya.medium.com/how-to-configure-default-search-registries-in-podman-ea930289692
> 
>     * https://www.redhat.com/sysadmin/manage-container-registries

not default behavior also not desirable

adding does not hurt at all, does not change anything for docker, only makes podman work

adding only makes people happy, not people mad

SamantazFox:
> Closing as there is no need to do that if the docker hub registry is configured as a default registry like on Docker:
> 
>     * https://halukkarakaya.medium.com/how-to-configure-default-search-registries-in-podman-ea930289692
>     * https://www.redhat.com/sysadmin/manage-container-registries

I don't see how this change would hurt. It just makes the repository to get thos images from explicit.

SamantazFox (approved):

unixfox:
While I really like podman, our main user base is using Docker and I don't really want to break the workflow of our main user base.

I prefer to document the fix for podman on the documentation website and thus be sure that we don't break anything for our Docker users.

And about the part where I explain in details my motivation:

Here are some example where this could break the existing workflow of some of our Docker users:
- Someone that changed the `registry-mirrors` parameter in Docker in order to use a caching server, like documented in https://docs.docker.com/registry/recipes/mirror/#configure-the-docker-daemon. Forcing to `docker.io` would bypass the caching server and break his workflow.
- Recently Docker hub is testing IPv6 and requires using an alternative default registry (registry.ipv6.docker.com) like explained in https://www.docker.com/blog/beta-ipv6-support-on-docker-hub-registry/. Forcing to docker.io would make the docker image fail to build on an IPv6 only machine as docker.io is only reachable from IPv4.

You may think in both cases the user just have to change the registry URL manually in order to fix the issue, sure but we are still breaking their workflow and that's bad.

And for podman not working by default, I don't fully agree with you because on CentOS 8 (and probably other RedHat based distributions) podman ask you by default which registry you want to use:
```
[root@podman-inv invidious]# podman build -t invidious -f docker/Dockerfile .
WARN[0000] missing "release" build argument. Try adding "--build-arg release=<VALUE>" to the command line 
[1/2] STEP 1/12: FROM crystallang/crystal:1.2.1-alpine AS builder
? Please select an image: 
    registry.fedoraproject.org/crystallang/crystal:1.2.1-alpine
    registry.access.redhat.com/crystallang/crystal:1.2.1-alpine
    registry.centos.org/crystallang/crystal:1.2.1-alpine
  ▸ docker.io/crystallang/crystal:1.2.1-alpine
```
Maybe it works differently on Debian or Ubuntu but at least on the Linux distribution by the same creator of podman it works like this.

In conclusion, I'm against merging this change and I prefer to add a specific documentation for podman where we explain how to setup a default registry.

706f6f7079646f6f646f6f:
so who to annoy?

all of podman people OR docker people with mirror and docker people with ipv6

probably more people with podman...

unixfox:
It's not about the quantity of people affected, but the fact that we officially supported Docker before podman.

I don't want to break the workflows that are using an officially supported software.

unixfox:
I just created a GitHub issue for adding podman in the documentation: https://github.com/iv-org/documentation/issues/166

PR #2627, opened 2021-11-17
[locales, ru] Editing numeral endings

AHOHNMYC:
There are three forms of numeral endings in Russian. I've fixed them using regex.

Some advices to Russian-speaking approvers:
```
([^.,0-9]|^)(\d*[02-9])?1([^.,0-9]|$)       // Именительный, ед.ч   x1 - волк,  кроме x11 (волков)
([^.,0-9]|^)(\d*[02-9])?[2-4]([^.,0-9]|$)   // Родительный, ед.ч    x2-4 - волка, кроме x12-x14 (волков)
([^.,0-9]|^)\d*([05-9]|1\d)([^.,0-9]|$)     // Родительный, мн.ч    x0, x5-9, x1X волков
// Где: x - возможная цифра, X - обязательная цифра
```


unixfox:
I'm not entirely sure (maybe some else from @iv-org/developers can confirm) but I think you are supposed to submit languages changes on weblate instead: https://hosted.weblate.org/engage/invidious/

AHOHNMYC:
I don't think Weblate can handle this case. Sad tradeoff extensibility vs handy translation managing.

SamantazFox:
Yeah, weblate can't handle that, because it's not a standard format. We plan to switch to GNU gettext in the future, which will allow us to properly support plural forms. See Issue #1951 and PR #2285.

Also, one thing I'm afraid of is weblate trying to get rid of those regexes (or add more) because they don't match the en-US.json (source) file.

Other than this, I'm not sure what you want to do. `([^.,0-9]|^)` and `([^.,0-9]|$)` makes sure that the regex only matches whole numbers, so those can be left out. So it leaves us with the following:
- `(\d*[0,2-9])?1`
- `(\d*[0,2-9])?[2-4]`
- `\d*([0,5-9]|1\d)`

I'm not sure why there's a comma in the list. Is it intentional (to match english thoushand separators) or is it a mistake? Also, that comma ommited, do I get the logic right?
* the first one matches every number ending with 1, except 11 (so 1, 21, 451, 1981)
* The seconds matches numbers ending with 2, 3 or 4, except 12, 13 and 14 (so 42, 982, 6502)
* The last one matches numbers who end in 5, 6, 7, 8, 9 (45, 666, 1337, 2008) or with 1x (10, 12, 15)

AHOHNMYC:
@SamantazFox, you got it exactly right. Comma was added by mistake. I've committed fix and edited PR original post. You have really sharp eye 👍 
Nice to see that great moving to GNU gettext. I've looked at this, and it handles plurals very well.
What can I do now with this PR?

syeopite:
Yes GNU Gettext is great! But one of it's major limitations is that it can't handle how floating point numbers can affect the plural form. 

Languages are wonderfully complex. We're lucky that Invidious doesn't really make use of floating point numbers or we'd need another format.

SamantazFox:
> Nice to see that great moving to GNU gettext. I've looked at this, and it handles plurals very well. What can I do now with this PR?

For now, not much. The current state of translations in invidious is not great. As I expected, I can't merge your PR as-is, otherwise weblate will try to remove the strings you provided. So, I'm working on a quick i18next-like implementation of plural forms. This format is supported by Weblate and doesn't require too much work to implement, so I'm currently doing it.

> P.S. Deeper googling showed me that this bicycle was already [built by Unicode](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#ru) and [exists in gettext](https://www.gnu.org/software/gettext/manual/html_node/msginit-Invocation.html) by default.

Yeah, that's part of the Unicode's Common Locale Data Repository (CLDR). @syeopite is working on implementing that in crystal.

SamantazFox:
WIP, see #2646

SamantazFox:
Now that #2646 have been merged: @AHOHNMYC do you mind making your changes through Weblate?
If you don't want to deal with weblate, I can explain how the new plurals work :)

AHOHNMYC:
@SamantazFox thank you for noticing!
Better late than never :) I've updated Russian plurals and translated some untouched strings on Weblate.
I think, you may close this PR as totally useless 😺

PR #2630, opened 2021-11-17
Escape video description in <meta> elements

### LINKED ISSUE #2629: [Bug] Video description is not escaped in <meta> elements
On `/watch` the video description is not escaped in the `<meta>` `og:description` and `twitter:description` elements in the html head, see screenshot.

description not escaped

A description with quotation marks or `>` in it might mess up the page's markup (although firefox is coping here).

Not really sure how this could be used maliciously, CSP thankfully blocks a lot here. This probably can't comment out the rest of the page's markup because there's currently a comment right after the `<title>` tag (the string `-->` would end the inserted comment). An inserted comment could stop the stylesheets and videojs being loaded and change the page title. Just what I could think of though.

<details>
<summary>In the CSP `style-src` allows inline styles and `img-src` allows data URLs so something like this is definitely possible.</summary>
Invidious video page but with a repeating image of a big rat as the background.
</details>

babababag:
Fixes #2629.

### DIFF
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -11,7 +11,7 @@
 <meta property="og:url" content="<%= HOST_URL %>/watch?v=<%= video.id %>">
 <meta property="og:title" content="<%= title %>">
 <meta property="og:image" content="/vi/<%= video.id %>/maxres.jpg">
-<meta property="og:description" content="<%= video.short_description %>">
+<meta property="og:description" content="<%= HTML.escape(video.short_description) %>">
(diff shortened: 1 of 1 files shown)


SamantazFox (approved):

SamantazFox:
Thanks for this PR :)
I'll wait for the CI to pass, and I'll merge this right after, given that it's a really minor change (it doesn't require testing)

SamantazFox:
Thanks for contributing to invidious @babababag :)
(That's probably one of our fastest merges, btw xD 17 minutes!)

PR #2576, opened 2021-11-10
Fix locales handling

SamantazFox:
Fixes #2572
Solves item "Use the locales full name instead of ISO codes" of #1951

Other than that, this changes how the "locale" is passed around:
- Previously, `translate()` would take a `Hash(String, JSON::Any)` as an argument, which implied that every function would set `locale = LOCALES["ISO_code"]`, or in essence, a copy of a complete locale dataset (keys + translation strings). This made the whole Hash being passed around, the entire object is copied to the stack on each function call, and probably impacting performance, among adding unnecessary complexity.
- Now, only the ISO code string is passed, and `translate()` will, by itself, look into the `LOCALES` constant (which is a `Hash(String, Hash(String, JSON::Any))`).

PR #2646, opened 2021-11-19
Better support of plurals in locales

SamantazFox:
Requires #2576


SamantazFox:
@matthewmcgarvey Could you please review this PR?

Note that for the `locales/*.json` part (except for `en-US.json`) the goal of the modifications is to avoid extra work for the translators. Otherwise, weblate will take care of cleaning the old translation keys and propose the new ones (though, it requires that I change weblate settings to `i18n` format first).

src/invidious/helpers/i18n.cr
```diff
@@ -107,6 +107,36 @@ def translate(locale : String?, key : String, text : String | Nil = nil) : Strin
...
+    translation = LOCALES[locale][plural_key].as_s
+  else
+    # Try #1: Fallback to singular in the same locale
+    singular_suffix = I18next::Plurals::RESOLVER.get_suffix(locale, 1)
+
+    if LOCALES[locale].has_key?(key + singular_suffix)
+      translation = LOCALES[locale][key + singular_suffix].as_s
+    else
+      # Try #2: Fallback to english (or return key we're already in english)
+      if locale == "en-US"
```
matthewmcgarvey, line 136: (minor nitpick) This should be an `elsif` and line 133 should be the `else`

matthewmcgarvey (approved):
It's difficult to understand all the code changes but I don't see any issues. Looks like a good change 👍

src/invidious/helpers/i18next.cr
```diff
@@ -209,4 +209,284 @@ module I18next::Plurals
...
+        return 0_u8
+      end
+    end
+
+    # Plural form of Slavic languages (E.g: Russian)
+    #
+    # Corresponds to i18next rule #4
+    # Rule: (n%10==1 && n%100!=11 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2)
+    #
+    def self.dual_slavic(count : Int) : UInt8
```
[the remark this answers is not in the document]
SamantazFox: All languages have different plural forms (japanese has no concept of plurals, english has singular and plural, arabic has 6 different forms, and so on), so for a given number, the proper variant has to be selected. Those variants IDs come from GNU Gettext, and are used in other projects, like i18n, which this PR tends to be compatible with.

The list of rules and languages is directly extracted from here:
https://github.com/i18next/i18next/blob/master/src/PluralResolver.js

So in essence, the system (entry point: `get_suffix`) takes a locale ISO code and a number. The ISO code is looked up for its associated "plural form" (an enumerate pointing to the plural rule to use) and is used, along with the number, to get the string suffix. Finally, the calley can append this suffix to the translation key, and get the right associated string.

E.g: `"ga"` (Irish), with number `4` should return the suffix `_2`

src/invidious/helpers/i18next.cr
```diff
@@ -105,67 +108,58 @@ module I18next::Plurals
...
       else
-        @@version = version
+        @version = version.to_u8
       end
+
+      self.init_rules
     end
 
-    def self.init_rules
+    def init_rules # : Hash(String, PluralForms)
```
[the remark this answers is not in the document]
SamantazFox: Yup. The compiler yellied at me for using both @forms and @@forms directly.
I need to clean the code, though.

matthewmcgarvey (approved):

SamantazFox:
Sorry for the multiple force pushes, I forgot to add the `_0` suffix on various languages without plurals (`fa`, `id`, `ja`, `ko`, `vi`, `zh-*`)

SamantazFox:
Rebased on master to fix conflicts.
Note for @matthewmcgarvey: I haven't migrated this spec file to `Spectator` because it doesn't seem to support variable text in `it` statements.

matthewmcgarvey:
@SamantazFox 🤔 Compiler error or something else? I think you'd use `sample`. Docs are https://gitlab.com/arctic-fox/spectator/-/wikis/Sample-Values

I believe it will have special output handling on failure so you might not need to do the special it name

SamantazFox:
> @SamantazFox 🤔 Compiler error or something else?

Compiler error:
```
$ crystal spec spec/i18next_plurals_spec.cr
Showing last frame. Use --error-trace for full trace.

There was a problem expanding macro '_spectator_metadata'

Code in macro 'it'

 4 | _spectator_metadata(__temp_36, :metadata,  )
     ^
Called macro defined in lib/spectator/src/spectator/dsl/metadata.cr:6:13

 6 | private macro _spectator_metadata(name, source, *tags, **metadata)

Which expanded to:

 > 1 | private def self.__temp_36
                        ^--------
Error: can't declare def dynamically
```

> I think you'd use `sample`. Docs are https://gitlab.com/arctic-fox/spectator/-/wikis/Sample-Values
> 
> I believe it will have special output handling on failure so you might not need to do the special it name

Ok, I'll try to make that work tomorrow!

matthewmcgarvey:
@SamantazFox would you be able to check out the master branch of spectator and see if the interpolation works using `sample`? https://github.com/icy-arctic-fox/spectator/issues/41

He's added support but he's going to wait for us to try it before releasing. If you don't have time that's fine.

SamantazFox:
@matthewmcgarvey So, I tried to use `sample` instead, and hit some problems:

`sample` doesn't support `Hash`es, so I wen with this:
```crystal
Spectator.describe
  describe "get_plural_form" do
    sample FORM_TESTS.keys do |locale|
      it "returns the right plural form for locale '#{locale}'" do
        expect(resolver.get_plural_form(locale)).to eq(FORM_TESTS[form])
      end
    end
  end
  
  # Snip
```

And that gives the following compiler error:
```
$ crystal spec spec/i18next_plurals_spec.cr              
Showing last frame. Use --error-trace for full trace.

There was a problem expanding macro 'it'

Code in spec/i18next_plurals_spec.cr:30:13

 30 | "pl"    => I18next::Plurals::PluralForms::Special_Polish_Kashubian,
                ^
Called macro defined in macro 'define_example'

 17 | macro it(what = nil, *tags, **metadata, &block)

Which expanded to:

 > 20 |             example.with_context(SpectatorTestContext::Group__temp_51::Group__temp_54::Group__temp_57) do
 > 21 |               
 > 22 |                 __temp_63
                        ^--------
Error: Undefined local variable or method 'resolver'
```

matthewmcgarvey:
@SamantazFox I might mess around with it later. I believe the output will be fine without string interpolation, though right?

SamantazFox:
@matthewmcgarvey I've made a mistake in my code, that's why I got this error.

matthewmcgarvey:
Is this failing because he needs to release the Spectator changes?

SamantazFox:
> Is this failing because he needs to release the Spectator changes?

Ohhhhh, you're right! Obviously the shard isn't updated in here ^^>

matthewmcgarvey:
@SamantazFox You might update the spectator issue that it wasn't an issue on the library's side

SamantazFox:
@matthewmcgarvey I already did :)

spec/i18next_plurals_spec.cr
```diff
@@ -0,0 +1,214 @@
+require "spectator"
+require "../src/invidious/helpers/i18next.cr"
+
+Spectator.configure do |config|
+  config.fail_blank
+  config.randomize
+end
```
matthewmcgarvey, line 7: Could you require spec_helper instead of this? Youd probably just need to add a require for the helper

SamantazFox: The problem of the spec helper is that it `require`s the whole invidious source code. I'd prefer to keep that module somewhat separated (I'd like this to be part of @syeopite's lens library in the future).

matthewmcgarvey (approved):

SamantazFox:
Note: this was deployed on the test instance, works fine.

Issue #2436, opened 2021-09-28
[Bug] Searching for videos on a channel by oldest upload loops the first page.


SkMyPP:
Go to any channel on any instance, sort the videos by oldest, go to the bottom of the page and click "Next page".

Instead of going to the next page, it will loop the first page of results no matter how many times you click "Next page" when sorting by oldest uploads.

SamantazFox:
Thanks for reporting this issue :)

TheFrenchGhosty:
A 20$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 10$ (in BTC) from the Invidious project.

More details: #1898

Issue #1599, opened 2020-12-22
[Bug] Playlists list continuation is broken


TheFrenchGhosty:
Clicking on "Next page" on the playlists list of a channel will trigger a bug, because the continuation is broken.

Example:

https://invidious.snopyta.org/channel/UCXuqSBlHAE6Xw-yeJA0Tunw/playlists

"Next page" continuation is ```?continuation=``` instead of an actual value:

https://invidious.snopyta.org/channel/UCXuqSBlHAE6Xw-yeJA0Tunw/playlists?continuation=

TheFrenchGhosty:
Updated example:

https://invidious.snopyta.org/channel/UC-lHJZR3Gqxm24_Vd_AJ5Yw/playlists

https://www.youtube.com/channel/UC-lHJZR3Gqxm24_Vd_AJ5Yw/playlists

greenfoo:
Is this bug still relevant?
I just checked with the latest version from `master` and the "Next" button works on PewDiePie's playlists without problems.
The API endpoint (`http://localhost:3000/api/v1/channels/playlists/UC-lHJZR3Gqxm24_Vd_AJ5Yw?continuation=...`) also seems to work ok.

TheFrenchGhosty:
It is still relevant.

There's not next button here: https://invidious.snopyta.org/channel/UC-lHJZR3Gqxm24_Vd_AJ5Yw/playlists

greenfoo:
Snopyta is not using the latest commit from `master`:
![snopyta](https://user-images.githubusercontent.com/42691086/113520817-0bc49e00-9596-11eb-8a49-68a4972f9b2c.png)

I just tried with the latest `master` and the `Next` button is there:
![next](https://user-images.githubusercontent.com/42691086/113520832-1aab5080-9596-11eb-93a0-cfc33450662d.png)

It looks like the commit where this issue was fixed is c0997f951c9af58a844d6098398ce0a39e43d052

TheFrenchGhosty:
mmh great then!

SamantazFox:
It works on the few channels I tried, however, there is no "previous page" button.

B0pol:
It is not broken anymore

SamantazFox:
@B0pol filters don't work. For instance, both of these return the same thing:

- https://invidious.snopyta.org/channel/UC-lHJZR3Gqxm24_Vd_AJ5Yw/playlists?sort_by=newest
[...]

PR #2653, opened 2021-11-20
Get all videos when sorted by oldest

0x24D:
I had originally implemented a partially working solution where the continuation was passed in the request headers, similiar to how this issue was fixed [here](https://git.gir.st/subscriptionfeed.git/commit/7ddb60edbec29e4d88bb4b2f8f427869acea8220). However after doing that I found [this comment](https://github.com/iv-org/invidious/issues/1599#issuecomment-814372849) which stated that using the page number rather than passing around the continuation was a design decision.

This pull request adds a new database table which keeps track of the continuation for each page on a channel. The new db store/lookup logic is only used where the current implementation doesn't work, so getting videos sorted by newest and popular still uses the existing continuation generation logic. This reduces the amount of entries that will be written to the database, and the amount of time looking up continuations in the database, as we only really need to store the continuations when sorting by oldest.

In normal operation (using the next page button) the continuation for the following page will be stored after the data for the current page is returned from the API call. However if a user manually goes to a page where we don't have the continuation (e.g. by modifying the page request header) we find the 'latest' continuation we have then iterate from there to get the continuation for the wanted page. During this iteration we store the returned continuation so that it can be easily looked up next time that page is requested.

Fixes: #2436

### DIFF
--- a/config/sql/channel_continuations.sql
+++ b/config/sql/channel_continuations.sql
@@ -5,8 +5,8 @@
 CREATE TABLE IF NOT EXISTS public.channel_continuations
 (
   id text NOT NULL,
-  page integer,
-  sort_by text,
+  page integer NOT NULL,
+  sort_by text NOT NULL,
   continuation text,
   CONSTRAINT channel_continuations_id_page_sort_by_key UNIQUE (id, page, sort_by)
 );
--- a/config/sql/channel_continuations.sql
+++ b/config/sql/channel_continuations.sql
@@ -1,0 +1,23 @@
+-- Table: public.channel_continuations
+
+-- DROP TABLE public.channel_continuations;
+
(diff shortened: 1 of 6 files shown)


SamantazFox:
Thanks for your contribution!

While your idea is good, the problem lies in how the continuation token is crafted. Youtube changed that some time ago and `produce_channel_videos_continuation` now produces bad tokens.

In addition to that, we won't be able to merge your PR soon because we don't have any migration tools.

broquemonsieur:
According to this IRC [message](https://matrix.to/#/!yAXokUArOJxHEACeib:pussthecat.org/$N1niZVoAyYUtAQz-UYYACAG__giCj6HqI1N4cXGMAhc?via=nitro.chat&via=matrix.org&via=tchncs.de), this PR should be closed

PR #2666, opened 2021-11-27
Move more routes to new pattern

matthewmcgarvey:
Related to #1121

I can see the end of this 👀 ... then we get to actually do the real work of figuring out what to do to migrate off of kemal.


SamantazFox:
Thanks for this PR :)
Regarding the "/data_control" route, #2545 must be merged first, so I'll mark yours as "blocked" in the meantime.

SamantazFox:
The blocking PR has been merged! You can now rebase this one :)

src/invidious/routes/preferences.cr
```diff
@@ -285,4 +285,191 @@ module Invidious::Routes::PreferencesRoute
...
+          elsif extension == "json" || type == "application/json"
+            subscriptions = JSON.parse(body)
+            user.subscriptions += subscriptions.as_a.compact_map do |entry|
+              entry["snippet"]["resourceId"]["channelId"].as_s
+            end
+          elsif extension == "csv" || type == "text/csv"
+            subscriptions = parse_subscription_export_csv(body)
+            user.subscriptions += subscriptions
+          else
+            haltf(env, status_code: 415,
```
matthewmcgarvey, line 406: `halt` -> `haltf`

src/invidious.cr
```diff
@@ -399,512 +402,18 @@ Invidious::Routing.get "/s_p/:id/:name", Invidious::Routes::Images, :s_p_image
 Invidious::Routing.get "/yts/img/:name", Invidious::Routes::Images, :yts_image
 Invidious::Routing.get "/vi/:id/:name", Invidious::Routes::Images, :thumbnails
 
+Invidious::Routing.get "/modify_notifications", Invidious::Routes::Notifications, :modify
+
+Invidious::Routing.post "/subscription_ajax", Invidious::Routes::Subscriptions, :toggle_subscription
+Invidious::Routing.get "/subscription_manager", Invidious::Routes::Subscriptions, :subscription_manager
```
SamantazFox, line 408: Shouldn't those routes also be excluded for api_only builds?

matthewmcgarvey: I'm not familiar with that, so... should these routes also be excluded for api_only builds?

unixfox: Yes

SamantazFox: After thinking a bit about that, yeah, they should be excluded.

SamantazFox (approved):

SamantazFox:
Deployed on https://test.invidious.io: `/data_control` doesn't work.
It's probably because the `next` statements shouldn't be replaced by `return` in this context (they're apparently made to jump to the next loop iteration).

SamantazFox (approved):

SamantazFox:
Works better now ^^

PR #2671, opened 2021-11-28
Remove dead code

matthewmcgarvey:
Some unused code I found while looking around the codebase.


src/invidious/comments.cr
```diff
@@ -60,8 +60,6 @@ def fetch_youtube_comments(id, cursor, format, locale, thin_mode, region, sort_b
   case cursor
   when nil, ""
     ctoken = produce_comment_continuation(id, cursor: "", sort_by: sort_by)
-    # when .starts_with? "Ug"
-    #   ctoken = produce_comment_reply_continuation(id, video.ucid, cursor)
```
matthewmcgarvey, line 64: This has been commented out for two years https://github.com/iv-org/invidious/commit/7b53b6bfeff1670c1c5d3dd299b7ed6162a5e3c0

src/invidious/trending.cr
```diff
@@ -20,13 +20,3 @@ def fetch_trending(trending_type, region, locale)
 
   return {trending, plid}
 end
-
-def extract_plid(url)
```
matthewmcgarvey, line 24: Last usage removed 5 months ago https://github.com/iv-org/invidious/commit/50267a6dd60d8bb95c477e33e4824f8b19bc6424

SamantazFox:
> The StaticFileHandler code was confusing. Looks like it was unused from the initial commit but the message using the term "Patch" makes me think the intent was to monkey patch instead of making a new class? Maybe there's a way that it's actually in use and I'm missing it. [6cd8845](https://github.com/iv-org/invidious/commit/6cd884555c3ea3a108fe20d5b7ccbba9bd49bb4c)

Yeah, that's a "monkey patch", indeed.
And I'm going to be honnest, that's a part of the codebase that I don't know much. I've been mainly focusing on rewriting the more immediate (and broken) parts of the code, i.e YT scraping and a few end-user related things. Except in some cases, I've personally preferred to not touch what was working.

SamantazFox:
So, I looked at the code, and the patch of the static file handler was to allow static files to be cached in memory, probably to improve performances by reducing disk I/O.

matthewmcgarvey:
@SamantazFox right but the handler is never actually added to the the Kemal configuration or used anywhere.

SamantazFox:
> @SamantazFox right but the handler is never actually added to the the Kemal configuration or used anywhere.

As you said, it's monkey patched. It overrides the following function: `Kemal::StaticFileHandler.call()`
https://github.com/kemalcr/kemal/blob/master/src/kemal/static_file_handler.cr

matthewmcgarvey:
My bad, I read Kemal, but in my mind it didn't process that this isn't kemal 🙈

SamantazFox (approved):

TheFrenchGhosty:
@matthewmcgarvey I know it's completely off topic: but glad to see you back :)

Issue #2612, opened 2021-11-14
[Bug] Unexpected char '�' at line 1, column 1 (JSON::ParseException)


randrej:
**Describe the bug**

Most, if not all videos can't be shown, the following text is displayed in place of the player

```
Title: `Unexpected char '' at line 1, column 1 (JSON::ParseException)`
Date: `2021-11-14T12:46:01Z`
Route: `/watch?v=YgVh_rSbP0w`
Version: `2021.11.13-00904ae @ master`
<details>
<summary>Backtrace</summary>
<p>
   
\```
Unexpected char '' at line 1, column 1 (JSON::ParseException)
  from /usr/share/crystal/src/json/lexer.cr:331:7 in 'raise'
  from /usr/share/crystal/src/json/lexer.cr:327:5 in 'unexpected_char'
  from /usr/share/crystal/src/json/lexer.cr:326:11 in 'unexpected_char'
  from /usr/share/crystal/src/slice.cr:213:5 in 'next_token'
  from /usr/share/crystal/src/json/parser.cr:13:12 in '_post_json'
  from src/invidious/yt_backend/youtube_api.cr:317:12 in 'extract_video_info:video_id'
  from src/invidious/videos.cr:970:10 in 'get_video:region'
  from src/invidious/routes/watch.cr:63:15 in 'handle'
  from lib/kemal/src/kemal/route.cr:13:9 in '->'
  from /usr/share/crystal/src/primitives.cr:266:3 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:167:15 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
[...]

PR #2623, opened 2021-11-16
Temporarily fix for #2612

SamantazFox:
Don't rely on the auto compression/decompression provided by the crystal stdlib.

PR #2677, opened 2021-11-28
Temporarily fix for #2612 (round 2)

SamantazFox:
Follow up to #2623 (and #2673)
Fixes https://github.com/iv-org/invidious/issues/2612
Ok, I got what was the problem.

Reads of interest:
* https://gitter.im/crystal-lang/crystal?at=61a3a3c9026dc63ca61e5489
* https://github.com/crystal-lang/crystal/issues/9901

### DIFF
--- a/src/invidious/yt_backend/youtube_api.cr
+++ b/src/invidious/yt_backend/youtube_api.cr
@@ -404,18 +404,9 @@ module YoutubeAPI
     url = "#{endpoint}?key=#{client_config.api_key}"
 
     headers = HTTP::Headers{
-      "Content-Type" => "application/json; charset=UTF-8",
+      "Content-Type"    => "application/json; charset=UTF-8",
+      "Accept-Encoding" => "gzip, deflate",
     }
-
-    # The normal HTTP client automatically applies accept-encoding: gzip,
-    # and decompresses. However, explicitly applying it will remove this functionality.
-    #
-    # https://github.com/crystal-lang/crystal/issues/11252#issuecomment-929594741
-    {% unless flag?(:disable_quic) %}
-      if CONFIG.use_quic
-        headers["Accept-Encoding"] = "gzip"
-      end
-    {% end %}
 
     # Logging
     LOGGER.debug("YoutubeAPI: Using endpoint: \"#{endpoint}\"")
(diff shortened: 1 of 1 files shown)


unixfox:
I've unfortunately no plans to test this PR as it broke too many times my instance. We can advertise the PR in https://github.com/iv-org/invidious/issues/2612 tho

SamantazFox:
> I've unfortunately no plans to test this PR as it broke too many times my instance.

yeah, understandable ^^

> We can advertise the PR in #2612 tho

Yup, good idea!

unixfox:
@SamantazFox I think it should be safe to merge now that we received a lot of positive feedback.

src/invidious/yt_backend/youtube_api.cr
```diff
@@ -404,38 +404,33 @@ module YoutubeAPI
...
     else
-      response = YT_POOL.client &.post(
-        url, headers: headers, body: data.to_json
-      )
+      # Using HTTP client
+      body = YT_POOL.client(client_config.proxy_region) do |client|
+        client.post(url, headers: headers, body: data.to_json) do |response|
+          self._decompress(response.body_io, response.headers["Content-Encoding"]?)
+        end
+      end
```
matthewmcgarvey, line 429: if you saved the response to a variable instead of using a block, it would match exactly the quic side of the condition. It seems the only difference is how the code handles the response body. The request piece could be pulled out of the if and de-duplicated.

SamantazFox: The problem is that the instance variable `body_io` isn't accessible outside of a block (I did what you say at the beginning, but It didn't work, so I reached to the `#crystal-lang` IRC channel for advice, and that's the response I got. Check the gitter.im link in the first post for the relevant conversation log)

unixfox:
Safe to merge?

SamantazFox:
@unixfox I want to ensure that there are no issues with it. I've seen a lot of "too many open files" errors during testing on the VPS, so I want to make sure it's not due to that fix.

SamantazFox:
In testing on https://test.invidious.io

matthewmcgarvey (approved):

TheFrenchGhosty:
> @unixfox I want to ensure that there are no issues with it. I've seen a lot of "too many open files" errors during testing on the VPS, so I want to make sure it's not due to that fix.

This hasn't been an issue in the last 24 hours, I'm assuming this PR is fine as-is.

Issue #2670, opened 2021-11-28
[Bug] Download widget replaces spaces in filename with `+`


babababag:
**Steps to reproduce**
1. Go to a video with spaces in the title e.g. https://inv.riverside.rocks/watch?v=UvpMGxHaw8U
2. Click Download
3. Filename has pluses instead of spaces

This happens because the filename is urlencoded before being sent in the Content-Disposition header ([here](/iv-org/invidious/blob/master/src/invidious/routes/video_playback.cr#L161) and [here](/iv-org/invidious/blob/master/src/invidious/routes/api/v1/videos.cr#L140)).
e.g. ```content-disposition: attachment; filename="lorem+ipsum"; filename*=UTF-8''lorem+ipsum```

Urlencoding is not compliant in the `filename` directive but apparently browsers ignore the RFC and decode it anyway. Urlencoding is ok in the `filename*` directive but only with percent symbols, so `+` means a literal plus not a space. https://httpwg.org/specs/rfc6266.html#disposition.parameter.filename

The fix is to replace `+` with a literal space. I would PR but idk crystal and it'd take 2 seconds for one of you guys to do it anyway.

SamantazFox:
Wait, why it closed this in here?!?

unixfox:
> Wait, why it closed this in here?!?

Oops I didn't think adding "fix" in my commit from another repo would close the issue: https://github.com/unixfox/invidious-custom/commit/385d07ce3c96a388de32190469329e8a629f8899

SamantazFox:
That is not supposed to happen D:

babababag:
@SamantazFox @unixfox

This is still happening, at least in Firefox.

It's still happening because this content-disposition header is still urlencoding strings with `+` instead of `%20`:
https://github.com/iv-org/invidious/blob/170e75499816af460d76d05ce4d440d37a01e0fb/src/invidious/routes/video_playback.cr#L161-L161

There are two directives allowed in the content-disposition header, `filename` and `filename*`. The `filename*` directive only supports percent-encoded characters and some common literal ascii characters, including the plus character, so `+` means a literal plus *(this is the issue)*. The `filename` directive only officially supports a subset of [these characters](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Code_page_layout), but browsers accept percent-encoded characters here too. In both cases replacing all the `+` characters with `%20` in this content-disposition header would fix the problem.

---

#### Merged pull request

[...]

PR #2684, opened 2021-11-29
Decode title from download widget

SamantazFox:
Fixes https://github.com/iv-org/invidious/issues/2670


unixfox:

PR #2695, opened 2021-12-03
Add override to mute IO::Error in HTTP::Server::Response::Output

SamantazFox:
Same as #2690, I just renamed the branch for clarity when it'll be merged.

### DIFF
--- a/src/invidious/helpers/crystal_class_overrides.cr
+++ b/src/invidious/helpers/crystal_class_overrides.cr
@@ -53,38 +53,23 @@ end
 #
 class HTTP::Server::Response
   class Output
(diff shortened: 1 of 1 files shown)


SamantazFox:
Ok, so using `previous_def` is not working. I'm going to go with another solution, courtesy of @Blacksmoke16, which involves `HTTP::Handler`s.

Blacksmoke16:
@SamantazFox I'm not sure I'd suggest that given it'll silence that error globally. I.e. _ANY_ `IO::Error` that happens _ANYWHERE_ in the application.

SamantazFox:
@Blacksmoke16 I was thinking about either catching the HTTP::Server::ClientError (which should also mute the IO::Error, if I understand exceptions correctly) or check the trace to know if we should raise or not.

PR #2744, opened 2021-12-26
fix: null ptr while retaining video time

tirz:
Fix

> Uncaught TypeError: localStorage is null

Some browser does not have their localStorage enabled. In that case, the Js thread will throw an exception while initializing the video.

Impact :
Since https://github.com/iv-org/invidious/pull/2538, we are not able to set the quality of video (480p, 720p, etc...). It is still viewable because the html tag `<video>` is loaded before, but the icon which allows us to set the quality on videojs will never be displayed.

Affected browsers :
 - Tor Browser
 - Firefox with `dom.storage` set to false in about:config
 - Chrome launched with `--disable-local-storage`

I also edited the file themes.js => same bug but does not have the same impact.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -59,6 +59,16 @@ videojs.Hls.xhr.beforeRequest = function(options) {
 
 var player = videojs('player', options);
 
+const storage = (() => {
+    try {
+        if (localStorage.length !== -1) {
+            return localStorage;
+        }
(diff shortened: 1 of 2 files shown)


SamantazFox (approved):
Looks good to me

unixfox:
in testing on https://yewtu.be

PR #2750, opened 2021-12-31
Removes annoying video switching | Fixes #2736

### LINKED ISSUE #2736: [Bug] 'Save the current video time' and 'Play next by default' settings conflict when trying to rewatch a video
**Describe the bug**

'Save the current video time' and 'Play next by default' settings conflict when trying to rewatch a video. Once finishing viewing a video play next by default will bring the user to the next video as expected. When the user goes back to the previous page the video jumps to the end and plays the next video causing a minor annoyance to the user.

**Steps to Reproduce**

1. Go to any updated Invidious instance
2. Go to your preferences and enable both 'Save the current video time' and 'Play next by default'
3. Navigate to a short video
4. When the next video plays navigate back to the previous video
5. See bug occur as you are navigated to the next video

**Logs**

N/A

**Screenshots**

N/A

**Additional context**

 - Workaround: User disables either setting or must catch the video quickly before the next video is played.

11Tuvork28:
Add check to avoid saving player position to the length of the given video by saving a minimum of 15 seconds at the end

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -215,7 +215,7 @@ if (video_data.params.save_player_pos) {
         const raw = player.currentTime();
         const time = Math.floor(raw);
 
-        if(lastUpdated !== time) {
+        if(lastUpdated !== time && raw <= video_data.length_seconds - 15) {
             save_video_time(time);
             lastUpdated = time;
         }


SamantazFox (approved):

SamantazFox:

11Tuvork28:
> Thanks for your contribution :)

The least I can do besides hosting an instance :)

PR #2762, opened 2022-01-05
Switch to using Spectator for specs

matthewmcgarvey:
https://gitlab.com/arctic-fox/spectator is one of my favorite Crystal libraries. It's attempting to add in a full-featured rspec-like way of writing specs. But why add it to invidious? One word... [mocks](https://gitlab.com/arctic-fox/spectator/-/wikis/Mocks-and-Doubles). I want to add testing to Invidious' parsing of XML and everything else but there are several places where database access and http requests are mixed in. Before refactoring the code to a better layout, I plan (and hope those who would like to add testing) to test the code as-is to the best of my ability using mocks for the external code.

The added dependency won't affect deploys. The built binary won't include the shard code because the library is not required by the app, only the specs.


SamantazFox:
Neat library! What's exactly the purpose of mocks? I'm not sure to understand that entirely.

matthewmcgarvey:
Sure, so say someone was crazy to want to test [`extract_video_info`](https://github.com/iv-org/invidious/blob/823121637146c4b5b13ab4f4326cb73ff16828d6/src/invidious/videos.cr#L827-L994) so that the function could be broken up or refactored. The problem is that there are api requests scattered in the function (`YoutubeAPI.player` is one) that make it difficult. The default specs library doesn't have a way to mock / replace that class with something that doesn't actually make the api requests but Spectator does.

You would do something like
```crystal
mock YoutubeAPI.class do
  stub player { Hash(String, JSON::Any).new }
end
```
and then when you call `YoutubeAPI.player` it will return the stubbed value instead of calling the real method.

Since making this PR I have found the Todo lists and this doesn't really align with anything I've seen there so feel free to turn it down.

SamantazFox:
> and then when you call `YoutubeAPI.player` it will return the stubbed value instead of calling the real method.

Oh, neat!

> Since making this PR I have found the Todo lists and this doesn't really align with anything I've seen there so feel free to turn it down.

The ToDo list is mostly a triage page for the different issues we have. I try to keep it up to date.
That's what we'll focus on in general, but proposing changes for things not in there is alsp welcome ^^

SamantazFox:
Could you please rebase?

SamantazFox (approved):

SamantazFox:
Just to double check: nothing else is required to do the switch, except that changing the `require` statement from `spec` (Crystal's library) to `spectator`. Then it can be run with `crystal spec` all the same, right?

matthewmcgarvey:
@SamantazFox The API is a bit different than the default Crystal spec library, though I think there is legacy support. But yea, `crystal spec` is the correct command

PR #2778, opened 2022-01-06
Create videos caching table as UNLOGGED

unixfox:
From the doc for `UNLOGGED` (https://www.postgresql.org/docs/10/sql-createtable.html):
> If specified, the table is created as an unlogged table. Data written to unlogged tables is not written to the write-ahead log (see Chapter 30), which makes them considerably faster than ordinary tables. However, they are not crash-safe: an unlogged table is automatically truncated after a crash or unclean shutdown. The contents of an unlogged table are also not replicated to standby servers. Any indexes created on an unlogged table are automatically unlogged as well.

For a table used for caching, it doesn't make sense to have a fail proof table, as this table is repopulated by invidious as soon as the data is removed and recommend this table to be truncated every week.

Moreover, having unlogged table improves the performance by a lot.

### DIFF
--- a/config/sql/videos.sql
+++ b/config/sql/videos.sql
@@ -2,7 +2,7 @@
 
 -- DROP TABLE public.videos;
 
-CREATE TABLE IF NOT EXISTS public.videos
+CREATE UNLOGGED TABLE IF NOT EXISTS public.videos
 (
   id text NOT NULL,
   info text,


SamantazFox:
Oooh, yep, could be a great idea! Though it requires migration first.

unixfox:
> Oooh, yep, could be a great idea! Though it requires migration first.

At least for the new invidious installations the unlogged parameter will be present. I don't think we should lock this PR until the migration framework is implemented.

SamantazFox:
> At least for the new invidious installations the unlogged parameter will be present. I don't think we should lock this PR until the migration framework is implemented.

The, can't that be `CREATE UNLOGGED TABLE`, rather than adding `ALTER TABLE`?

unixfox:
> > At least for the new invidious installations the unlogged parameter will be present. I don't think we should lock this PR until the migration framework is implemented.
> 
> The, can't that be `CREATE UNLOGGED TABLE`, rather than adding `ALTER TABLE`?

Yes that's what I tried but I always got syntax error.

SamantazFox:
`CREATE UNLOGGED TABLE IF NOT EXISTS public.videos (` ?
From: https://www.postgresql.org/docs/11//sql-createtable.html

unixfox:
> `CREATE UNLOGGED TABLE IF NOT EXISTS public.videos (` ? From: [postgresql.org/docs/11//sql-createtable.html](https://www.postgresql.org/docs/11//sql-createtable.html)

Good idea, I updated the code.

Issue #2707, opened 2021-12-08
[Feature request] (QUICK / Urgent!!) Help archiving dislikes


Myzel394:
**Is your feature request related to a problem? Please describe.**

Youtube is going to remove the public dislike counter. https://returnyoutubedislike.com is a project which aims to bring it back and archive it. [When you have the extension installed and watch a video which hasn't been archived yet, it will be stored](https://returnyoutubedislike.com/faq).

**Describe the solution you'd like**
Send all video ids which are opened in Invidious to their API to help archiving dislikes.

**Describe alternatives you've considered**
[archiveteam.org](http://tracker.archiveteam.org/youtube-dislikes/) also offers archiving, which could also be used.

**Additional context**
This feature needs to be addressed quickly as we don't have much time left (Youtube is going to remove the dislike button counter ultimately on December 13.) and server owners also need to update their instances once this feature is added.

Myzel394:
Did some light digging and it seems like [this api call is sending the votes](https://github.com/Anarios/return-youtube-dislike/blob/main/Extensions/chrome/return-youtube-dislike.background.js#L43)

unixfox:
> Send all video ids which are opened in Invidious to their API to help archiving dislikes.

While I really like the initiative, I don't think we are going to do that due to obvious privacy reasons. Every video that an Invidious instance would watch would be automatically sent to a remote server, that's not a good idea because the remote server could literally build a profile of the invidious instance based on the IP address.

TheFrenchGhosty:
I definitely agree with @unixfox here.

Also, just an FYI:  the whole backend of "returnyoutubedislike.com" is closed source, and the creators instantly close issues that ask for it, I wouldn't trust them.

The project itself is completely pointless... since it's basically currently just using the API, and when the access gets cut off (on december 13), the website will basically have the dislike of some thousands videos (at best).

I'm strongly in favor of just having people run archiveteam warrior though.

Myzel394:
Okay that makes sense - Maybe we could add some awareness to the README.md ?

unixfox:
Closing this issue as we don't want to implement this feature, sorry. But you can still discuss here.

Myzel394:
> that's not a good idea because the remote server could literally build a profile of the invidious instance based on the IP address.

However, I don't fully understand this. Afaik invidious instances aren't meant to be private (since there is literally an API which lists all available instances (which fulfill the requirements)) :O

unixfox:
[...]

PR #2781, opened 2022-01-07
Return Youtube Dislikes and Ratio

jesperbakhandskemager:
This pull requests adds back dislike count, and ratio, using the '[returnyoutubedislike.com](https://github.com/anarios/return-youtube-dislike/)' API, clientside so a server won't reach the daily max requests of 10.000 or the 100 per minute.

### DIFF
--- a/assets/js/watch.js
+++ b/assets/js/watch.js
@@ -7,6 +7,24 @@ String.prototype.supplant = function (o) {
     });
 }
 
+function set_real_dislike()
+{
+var xhr = new XMLHttpRequest();
+xhr.open('GET', 'https://returnyoutubedislikeapi.com/votes?videoId=' + video_data.id);
+xhr.onload = function() {
+    if (xhr.status === 200) {
+        var data = JSON.parse(xhr.responseText);
+	document.getElementById("dislikes").innerHTML = "<i class='icon ion-ios-thumbs-down'></i> " + data.dislikes.toLocaleString('en-US');
+	document.getElementById("rating").innerHTML = "Rating: " + Math.round(data.rating * 10) / 10 + " / 5";
+	
+    }
+    else {
+	document.getElementById("dislikes").innerHTML = "<i class='icon ion-ios-thumbs-down'></i> 0";
+    }
+};
+xhr.send();
+}
+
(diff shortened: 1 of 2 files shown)


jesperbakhandskemager:
> 

The reason it's clientside is becuase if a large instance uses the returnyoutubedislike.com api, there is a daily limit of 10.000 requests or a limit of 100 requests per minute, which a single user couldn't achieve

jesperbakhandskemager:
It could easily be added as an option in preferences, so that it's an opt-in feature, with a warning that it's sending the video id to an external server and requires javascript. That way 'hidden instances' won't get leaked either as [Unifox mentions](https://github.com/iv-org/invidious/issues/2707#issuecomment-989143553) 

Also it looks like they are planning on releasing the backend code Discord Screenshot

unixfox:
Until the backend is open sourced, I'm opposed to implement this feature in the invidious codebase.
We can have that as an userscript or make the original browser extension working with invidious that's fine for me.

TheFrenchGhosty:
Discussion about its implementation in NewPipe as a reference: https://github.com/TeamNewPipe/NewPipe/issues/7469

As you can see [here in my comment](https://github.com/TeamNewPipe/NewPipe/issues/7469#issuecomment-1003392155), there's a lot of issues with it.

makerio90:
maybe you could have it so you can load dislikes by pushing a button, that way you limit the amount of data sent to the server.

jesperbakhandskemager:
> maybe you could have it so you can load dislikes by pushing a button, that way you limit the amount of data sent to the server.

That would also be a good option

TheFrenchGhosty:
This feature won't be implemented unless major change is done to "Return YouTube Dislike" (as pointed by my message in the NewPipe issue: https://github.com/TeamNewPipe/NewPipe/issues/7469#issuecomment-1003392155  ) mainly open sourcing the backend, and having the data public. After seeing the author's answers in the issue, I'm not expecting this to happen.

"Return YouTube Dislike" is free to implement Invidious support (like SponsorBlock is doing), but we won't do it, and won't endorse it (unlike SponsorBlock, that is fully open source and has its data public).

Issue #483, opened 2019-04-12
[Meta] Enhance and expand channel overview


Perflyst:
This is a meta issue for features which are available on YouTube if you are on a channel but missing on Invidious

- [ ] Profile image
- [ ] Profile banner

- [ ] Community (e.g. https://www.youtube.com/user/BBC/community)
- [ ] Channels, Featured channels (e.g. https://www.youtube.com/user/BBC/channels)
- [ ] About (e.g. https://www.youtube.com/user/BBC/about)

omarroth:
Related: #152

elypter:
channel search is also something very useful

omarroth:
Channel search is supported using [`channel:UCID` syntax](https://invidio.us/search?q=channel%3AUC6nSFpj9HTCZ5t-N3Rm3-HA+mind+field). Although should be made more user-friendly with #179.

omarroth:
Added banner and thumbnail with 0d0d3edeae1e632afc302e240b5ef7481c21b4b0.

Featured channels and information provided on the "About" page is already provided in the API, so adding it to the channel page should be fairly straight-forward.

omarroth:
Added communities as `/api/v1/channels/comments/:ucid` with 196ee1aa8bb7e722fe32b77bc00c52b99d96b9ae. Similarly to the above, should be a mostly straight-forward addition for a "community" page now that support has been added to the API.

omarroth:
Added community page with bcd239ac2b438ec721ffe35868371a9c81573f0a.

elypter:
i get an error when opening this oage https://invidio.us/channel/UCLLibJTCy3sXjHLVaDimnpQ/community

Perflyst:
@elypter this bug got fixed with af592ea8c15eeb985edecd875c66b05486bf7f89

I have added "Subscriptions" to the features list in the first post

omarroth:
Added description with f71d5c429d4e3abc4e6a7eaf461fd46587316121.

Perflyst:
I do not think we should mark "About" as done. There is also Details, Links and Stats available

TheFrenchGhosty:
Still useful.

SamantazFox:
still relevant

PR #2787, opened 2022-01-08
Update to fetch related channels again

### LINKED ISSUE #2749: [Bug] Channels API relatedChannels=[]
**Describe the bug**
Linus Tech Tips has multiple channels, and at least a year ago the below API endpoint yielded real results for related channels.

**Steps to Reproduce**

1. Go to `/api/v1/channels/UCXuqSBlHAE6Xw-yeJA0Tunw?fields=relatedChannels`
2. The API yields an empty list `{"relatedChannels":[]}`

I'm not sure what all this list used to include, but for my purposes I was able to previously extract the `author` and `authorId`.

The below image is source from youtube, and are the channels that should show up in this list.

image

matthewmcgarvey:
Fixes #2749
Relates to #483

Fetching channel info seems to no longer return related channels. Instead, we have to fetch the data found in the **Channels** tab. Because this is an extra API request, I updated the code to only make it when related channels are actually used which is just one place.

The structs used in this included `DB::Serializable` but weren't actually saved to the database. I updated them to be `record`'s.

## Testing

Go to `http://0.0.0.0:3000/api/v1/channels/UCXuqSBlHAE6Xw-yeJA0Tunw?fields=relatedChannels` on master and see that it returns an empty list. Now go to it on this branch and it should return all of the channels found in this picture

<details>
<summary>LTT Featured Channels</summary>
<br>
Screen Shot 2022-01-08 at 12 27 38 PM
</details>

`http://0.0.0.0:3000/api/v1/channels/UC38UpcvhpPqY2QmttdvWemA?fields=relatedChannels` is an example of a channel that doesn't have any related channels and doesn't break

### DIFF
--- a/src/invidious/channels/about.cr
+++ b/src/invidious/channels/about.cr
@@ -1,33 +1,31 @@
 # TODO: Refactor into either SearchChannel or InvidiousChannel
-struct AboutChannel
-  include DB::Serializable
+record AboutChannel,
+  ucid : String,
+  author : String,
+  auto_generated : Bool,
+  author_url : String,
+  author_thumbnail : String,
+  banner : String?,
+  description_html : String,
+  total_views : Int64,
+  sub_count : Int32,
+  joined : Time,
+  is_family_friendly : Bool,
+  allowed_regions : Array(String),
+  related_channels : RelatedChannels?,
+  tabs : Array(String)
 
-  property ucid : String
-  property author : String
-  property auto_generated : Bool
-  property author_url : String
-  property author_thumbnail : String
-  property banner : String?
-  property description_html : String
-  property total_views : Int64
-  property sub_count : Int32
-  property joined : Time
-  property is_family_friendly : Bool
-  property allowed_regions : Array(String)
-  property related_channels : Array(AboutRelatedChannel)
-  property tabs : Array(String)
-end
+record RelatedChannels,
+  browse_id : String,
+  params : String?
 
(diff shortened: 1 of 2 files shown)


matthewmcgarvey:
This will definitely conflict with https://github.com/iv-org/invidious/pull/2215 but it's expected with mega PRs

src/invidious/channels/about.cr
```diff
@@ -86,37 +83,14 @@ def get_about_info(ucid, locale)
...
-          author:           related_title,
-          author_url:       related_author_url,
-          author_thumbnail: related_author_thumbnail,
-        })
+    tabs = initdata.dig("contents", "twoColumnBrowseResultsRenderer", "tabs").as_a
+    if tab = tabs.find { |tab| tab.dig?("tabRenderer", "title").try(&.as_s?) == "Channels" }
+      browse_id = tab.dig?("tabRenderer", "endpoint", "browseEndpoint", "browseId").try(&.as_s?)
+      params = tab.dig?("tabRenderer", "endpoint", "browseEndpoint", "params").try(&.as_s?)
+      if browse_id
+        related_channels = RelatedChannels.new(browse_id: browse_id, params: params)
```
SamantazFox, line 91: In regard of my other comment, this code block is not required :)

SamantazFox (changes requested):

SamantazFox (approved):

PR #2789, opened 2022-01-08
Add e-mail contact to the README

TheFrenchGhosty:
This will ideally be moved to the website at a later date.

Rendered version: https://github.com/iv-org/invidious/blob/TheFrenchGhosty-email/README.md#e-mail-us

(I know it doesn't look great but it should be temporary-ish)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@
   <img alt="Mastodon: @invidious@social.tchncs.de" src="https://img.shields.io/badge/Mastodon-%40invidious%40social.tchncs.de-darkgreen">
   </a>
   <br>
-  <a href="#contact">
+  <a href="#contact-the-team-directly">
   <img alt="Contact the team directly" src="https://img.shields.io/badge/Contact%20the%20team%20directly-darkgreen">
   </a>
 </div>
--- a/README.md
+++ b/README.md
@@ -48,13 +48,10 @@
   <a rel="me" href="https://social.tchncs.de/@invidious">
   <img alt="Mastodon: @invidious@social.tchncs.de" src="https://img.shields.io/badge/Mastodon-%40invidious%40social.tchncs.de-darkgreen">
   </a>
-  
-  <h5>E-mail us:</h5>
-  <ul>
-  <li>General Inquiries (forwarded to every team member): <code>contact [at] invidious [dot] io</code> (Note: before sending a bug report please check that it hasn't already be reported on GitHub - bug reports sent to this address will be copied to GitHub)</li>
-  <li>Security issues (forwarded to the two project owners, <a href="https://github.com/TheFrenchGhosty">@TheFrenchGhosty</a> and <a href="https://github.com/Perflyst">@Perflyst</a>): <code>security [at] invidious [dot] io</code></li>
-  </ul>
-    
+  <br>
+  <a href="#contact">
+  <img alt="Contact the team directly" src="https://img.shields.io/badge/Contact%20the%20team%20directly-darkgreen">
+  </a>
 </div>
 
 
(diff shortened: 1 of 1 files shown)


SamantazFox:
image

This is wayy to much text at the top of the Readme. Makes other infos harder to parse.
I'd recommend to make a dedicated section down the page.

Also, I think that "obfuscating" a bit the email addresse (like `contact +SPAMGUARD [at] invidious.io`, with a note saying to remove `+SPAMGUARD`) would be nice to have. The email stays valid and can be easily filtered out, while ensuring our inboxes aren't filled with junk.

Also, an associated PGP key for the security email would make sense.

TheFrenchGhosty:
@SamantazFox 

>This is wayy to much text at the top of the Readme. Makes other infos harder to parse.
>I'd recommend to make a dedicated section down the page.

I completely agree, the thing is that having it at the bottom of the README isn't ideal.

>I think that "obfuscating" a bit the email addresse (like contact +SPAMGUARD [at] invidious.io, with a note saying to remove +SPAMGUARD) would be nice to have.

Won't we still receive emails from a "+" address?

>Also, an associated PGP key for the security email would make sense.

It's planned, just not right now.

SamantazFox:
> I completely agree, the thing is that having it at the bottom of the README isn't ideal either, it make it so most of the way to contact us is at the top, while email is as the bottom (This is why I said "This will ideally be moved to the website at a later date." that way we could just link to it)

You can still use anchors (e.g `#features` to go to the "features" section) in the meantime.

> Won't we still receive emails from a "+" address?

Yup, but as the "from" address will have this special part, it's easy to write filters for it.

> It's planned, just not right now.

Ok, nice :)

TheFrenchGhosty:
@SamantazFox Alright, done

README.md
```diff
@@ -153,6 +150,19 @@ Weblate also allows you to log-in with major SSO providers like Github, Gitlab,
 - [HoloPlay](https://github.com/stephane-r/HoloPlay): Funny Android application connecting on Invidious API's with search, playlists and favorites.
 
 
+## Contact the team directly
+
+Every team member is available through GitHub or through the Matrix room (bridged to IRC), however, if you need/have too, you can contact the team directly via e-mail (remove `+SPAMGUARD` from the addresses):
+
+- General Inquiries (forwarded to every team member): `contact +SPAMGUARD [at] invidious [dot] io`
```
SamantazFox, line 157: "forwarded to all team members"

README.md
```diff
@@ -48,13 +48,10 @@
...
-  
-  <h5>E-mail us:</h5>
-  <ul>
-  <li>General Inquiries (forwarded to every team member): <code>contact [at] invidious [dot] io</code> (Note: before sending a bug report please check that it hasn't already be reported on GitHub - bug reports sent to this address will be copied to GitHub)</li>
-  <li>Security issues (forwarded to the two project owners, <a href="https://github.com/TheFrenchGhosty">@TheFrenchGhosty</a> and <a href="https://github.com/Perflyst">@Perflyst</a>): <code>security [at] invidious [dot] io</code></li>
-  </ul>
-    
+  <br>
+  <a href="#contact">
+  <img alt="Contact the team directly" src="https://img.shields.io/badge/Contact%20the%20team%20directly-darkgreen">
```
SamantazFox, line 53: Maybe just "email" would be enough?

SamantazFox (changes requested):

README.md
TheFrenchGhosty: Done

README.md
TheFrenchGhosty: Done

PR #2814, opened 2022-01-13
Fix search autofocus

### LINKED ISSUE #2791: [Bug] New search field auto focusing prevents alternate inputs on page load
**Describe the bug**
When an invidious page is loaded, with the recent update (afaict), the search text field is auto focused. This means any keys typed are sent to the search field even if the user isn't intending to make a search as their first action.

This has the effect of 'eating' inputs for example of the `space` key for page down (and `Shift+space` for page up), commonly used by some users like myself in various browsers. Or for playing a video on a video page using `space` (the prior default behavior).

Mightn't be a bug per se but behavior perhaps not considered as a result of a recent change, though the prevention of using `space` on video pages on initial load may be a bug since that seems unintentional. If we compare the behavior to say search engines (eg: duckduckgo.com, bing.com, etc) one can see the search field isn't auto focused so I'm not sure if this new behavior was meant to address something or not (such as a user request).

**Steps to Reproduce**
1. Open an instance to any page (I tested on both invidious.snopyta.org and invidious-us.kavin.rocks).
2. Observe the search field is auto focused.

**Screenshots**
Example

**Additional context**
Tested on both Vivaldi (Chromium) and Firefox.

SamantazFox:
Make the following changes:
- search box only autofocuses on the full screen search page (`/search`)
- search box can be easily focused with the `/` key (requires JS)

Fixes #2791

### DIFF
--- a/assets/js/handlers.js
+++ b/assets/js/handlers.js
@@ -142,4 +142,16 @@
         var csrf_token = target.parentNode.querySelector('input[name="csrf_token"]').value;
         xhr.send('csrf_token=' + csrf_token);
     }
+
+    // Handle keypresses
+    window.addEventListener('keydown', (event) => {
+        // Ignore modifier keys
+        if (event.ctrlKey || event.metaKey) { return; }
+
+        // Focus search bar on '/'
+        if (event.key == "/") {
+            document.getElementById('searchbox').focus();
+            event.preventDefault();
+        }
+    });
 })();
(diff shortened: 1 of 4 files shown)


SamantazFox:
Go to https://test.invidious.io to try it out.

SamantazFox:
@unixfox Nothing to report?

unixfox:
Seems fine

PR #2821, opened 2022-01-17
Handle invalid channel id in channel: search

### LINKED ISSUE #2812: [Bug] Crash when channel: is put in search bar
**Describe the bug**
Crash when `channel:` is put in search bar

**Steps to Reproduce**
1. Go to [https://invidious.kavin.rocks/channel/UC7YOGHUfC1Tb6E4pudI9STA](https://invidious.kavin.rocks/channel/UC7YOGHUfC1Tb6E4pudI9STA)
2. Type `channel:metal` in the search bar
3. Press Enter

**Logs**
```
Title: `Missing hash key: "header" (KeyError)`
Date: `2022-01-13T13:37:15Z`
Route: `/search?q=channel%3Ametal`
Version: `2022.01.11-e7394fe @ master`
<details>
<summary>Backtrace</summary>
<p>
   
Missing hash key: "header" (KeyError)
  from /usr/share/crystal/src/hash.cr:1034:9 in '[]'
  from src/invidious/search.cr:8:12 in 'channel_search'
  from src/invidious/search.cr:213:20 in 'search'
  from /usr/share/crystal/src/primitives.cr:266:3 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/invidious/helpers/static_file_handler.cr:167:15 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
  from /usr/share/crystal/src/http/server.cr:500:5 in '->'
  from /usr/share/crystal/src/primitives.cr:266:3 in 'run'
  from ???

</p>
</details>
```

**Screenshots**

**Additional context**
Browser: Firefox 95.0.2

matthewmcgarvey:
Fixes #2812

The original error was from the json parsing failing, but the error after fixing it was still not the helpful.

Screen Shot 2022-01-18 at 6 55 57 PM

### DIFF
--- a/src/invidious/routes/search.cr
+++ b/src/invidious/routes/search.cr
@@ -56,7 +56,7 @@ module Invidious::Routes::Search
       begin
         search_query, count, videos, operators = process_search_query(query, page, user, region: region)
       rescue ex : ChannelSearchException
-        return error_template(404, "Unable to find channel with id of '#{HTML.escape(ex.channel)}'. Are you sure that's an actual channel id? It will look like 'UC4QobU6STFB0P71PMvOGN5A'.")
+        return error_template(404, "Unable to find channel with id of '#{HTML.escape(ex.channel)}'. Are you sure that's an actual channel id? It should look like 'UC4QobU6STFB0P71PMvOGN5A'.")
       rescue ex
         return error_template(500, ex)
       end
--- a/src/invidious/routes/search.cr
+++ b/src/invidious/routes/search.cr
@@ -55,6 +55,8 @@ module Invidious::Routes::Search
 
       begin
         search_query, count, videos, operators = process_search_query(query, page, user, region: region)
+      rescue ex : ChannelSearchException
+        return error_template(404, "Unable to find channel with id of '#{ex.channel}'. Are you sure that's an actual channel id? It will look like 'UC4QobU6STFB0P71PMvOGN5A'.")
       rescue ex
         return error_template(500, ex)
       end
--- a/src/invidious/routes/search.cr
+++ b/src/invidious/routes/search.cr
@@ -56,7 +56,7 @@ module Invidious::Routes::Search
       begin
         search_query, count, videos, operators = process_search_query(query, page, user, region: region)
       rescue ex : ChannelSearchException
-        return error_template(404, "Unable to find channel with id of '#{ex.channel}'. Are you sure that's an actual channel id? It will look like 'UC4QobU6STFB0P71PMvOGN5A'.")
+        return error_template(404, "Unable to find channel with id of '#{HTML.escape(ex.channel)}'. Are you sure that's an actual channel id? It will look like 'UC4QobU6STFB0P71PMvOGN5A'.")
       rescue ex
         return error_template(500, ex)
       end
--- a/src/invidious/search.cr
+++ b/src/invidious/search.cr
@@ -1,6 +1,7 @@
 class ChannelSearchException < InfoException
-  def initialize(channel : String)
-    super "Unable to find channel with id of '#{channel}'. Are you sure that's an actual channel id?"
+  getter channel : String
+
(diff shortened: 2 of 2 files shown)


unixfox:
Maybe you could add an example too like:
> Unable to find channel with id of 'metal'. Are you sure that's an actual channel id?
> Example channel id: UC4QobU6STFB0P71PMvOGN5A for  jawed's channel.

src/invidious/search.cr
```diff
@@ -1,12 +1,18 @@
+class ChannelSearchException < InfoException
+  def initialize(channel : String)
+    super "Unable to find channel with id of '#{channel}'. Are you sure that's an actual channel id?"
```
SamantazFox, line 3: It would be preferrable to have a generic exception in here, then to properly catch it and use translation strings for the error in the route handler.

matthewmcgarvey: By generic, I'm thinking you mean to just keep the `channel` as an instance variable and then prepare this message in the route handler. But if you mean that I should use `InfoException`, I'd worry that it would be errors this PR was not meant to handle

SamantazFox: I was thinking about an exception without a message (i.e `class Foo < InfoException; end`) with the rescuing and error message done in the route handler.

PS: after looking at the code, the error template doesn't allow for a parametric string anyway.

matthewmcgarvey:
Updated the route to catch the error and provide the message. Giving an example in the message and the http status is now 404 instead of the normal 500.

Description is updated with picture of current error

src/invidious/routes/search.cr
```diff
@@ -55,6 +55,8 @@ module Invidious::Routes::Search
 
       begin
         search_query, count, videos, operators = process_search_query(query, page, user, region: region)
+      rescue ex : ChannelSearchException
+        return error_template(404, "Unable to find channel with id of '#{ex.channel}'. Are you sure that's an actual channel id? It will look like 'UC4QobU6STFB0P71PMvOGN5A'.")
```
SamantazFox, line 59: `ex.channel` should be HTML escaped to prevent XSS injections!

SamantazFox (approved):

src/invidious/routes/search.cr
```diff
@@ -55,6 +55,8 @@ module Invidious::Routes::Search
 
       begin
         search_query, count, videos, operators = process_search_query(query, page, user, region: region)
+      rescue ex : ChannelSearchException
+        return error_template(404, "Unable to find channel with id of '#{HTML.escape(ex.channel)}'. Are you sure that's an actual channel id? It will look like 'UC4QobU6STFB0P71PMvOGN5A'.")
```
SamantazFox, line 59:
```suggestion
        return error_template(404, "Unable to find channel with id of '#{HTML.escape(ex.channel)}'. Are you sure that's an actual channel id? It should look like 'UC4QobU6STFB0P71PMvOGN5A'.")
```

PR #2827, opened 2022-01-20
More code cleanup

SamantazFox:
Lots of code cleanup.

Mostly:
 - Fix warnings raised by Ameba
 - Remove some dead code
 - Remove unused function parameters & return variables
 - Use `.dig?(key)` rather than long `[key]?.try` chains


src/invidious/playlists.cr
```diff
@@ -379,12 +374,15 @@ def fetch_playlist(plid, locale)
     author_thumbnail = ""
     ucid = ""
   else
-    author_info = playlist_sidebar_renderer[1]["playlistSidebarSecondaryInfoRenderer"]?.try &.["videoOwner"]["videoOwnerRenderer"]?
+    author_info = playlist_sidebar_renderer[1].dig?(
```
matthewmcgarvey, line 377: `[1]` could be added to the dig

matthewmcgarvey (commented):
Just the one minor nitpick in the code.

Also, is there anything in `.ameba.yml` that can be removed now?

src/invidious/playlists.cr
SamantazFox: given that the `if` right above ensurtes that we have `playlist_sidebar_renderer.size >= 2`, I though that for style, it could be left as is (the `1` alone at the beginning of the line was looking sad ^^>)

SamantazFox:
> Also, is there anything in `.ameba.yml` that can be removed now?

Nope. I haven't fixed anything related to cyclomatic complexity. I'll try to split the largest functions when possible (e.g user imports or video proxying) in another PR.

matthewmcgarvey (approved):

src/invidious/helpers/static_file_handler.cr
```diff
@@ -175,9 +175,9 @@ module Kemal
 
           if @cached_files.sum(&.[1][:data].bytesize) + (size = File.size(file_path)) < CACHE_LIMIT
             data = Bytes.new(size)
-            File.open(file_path) do |file|
-              file.read(data)
-            end
+
+            File.open(file_path) { |f| f.read(data) }
```
SamantazFox, line 179:
```suggestion
            File.open(file_path, &.read(data))
```

PR #2833, opened 2022-01-22
Channel helpers cleanup

matthewmcgarvey:
Looking for the cause of the database bloat, I found a lot of the default argument to these channel helper methods to be unnecessary.

Commits should show my step-by-step process.

### DIFF
--- a/src/invidious/channels/channels.cr
+++ b/src/invidious/channels/channels.cr
@@ -152,12 +152,14 @@ def get_batch_channels(channels)
   return final
 end
 
-def get_channel(id)
+def get_channel(id) : InvidiousChannel
   channel = Invidious::Database::Channels.select(id)
-  return channel if channel
 
-  channel = fetch_channel(id, pull_all_videos: false)
-  Invidious::Database::Channels.insert(channel)
+  if channel.nil? || (Time.utc - channel.updated) > 2.days
+    channel = fetch_channel(id, pull_all_videos: false)
+    Invidious::Database::Channels.insert(channel, update_on_conflict: true)
+  end
+
   return channel
 end
 
--- a/src/invidious/channels/channels.cr
+++ b/src/invidious/channels/channels.cr
@@ -114,8 +114,9 @@ class ChannelRedirect < Exception
   end
 end
 
-def get_batch_channels(channels, refresh = false, pull_all_videos = true, max_threads = 10)
+def get_batch_channels(channels)
   finished_channel = Channel(String | Nil).new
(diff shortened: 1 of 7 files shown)


src/invidious/channels/channels.cr
```diff
@@ -151,23 +152,18 @@ def get_batch_channels(channels, refresh = false, pull_all_videos = true, max_th
...
-    Invidious::Database::Channels.insert(channel)
-  end
+def get_channel(id)
+  channel = Invidious::Database::Channels.select(id)
+  return channel if channel
 
-  return channel
+  channel = fetch_channel(id, pull_all_videos: false)
+  Invidious::Database::Channels.insert(channel)
+  channel
```
SamantazFox, line 161: could you please use the `return` statement in here, for clarity?

src/invidious/channels/channels.cr
```diff
@@ -151,23 +152,18 @@ def get_batch_channels(channels, refresh = false, pull_all_videos = true, max_th
   return final
 end
 
-def get_channel(id, refresh = true, pull_all_videos = true)
-  if channel = Invidious::Database::Channels.select(id)
-    if refresh && Time.utc - channel.updated > 10.minutes
-      channel = fetch_channel(id, pull_all_videos: pull_all_videos)
-      Invidious::Database::Channels.insert(channel, update_on_conflict: true)
-    end
```
SamantazFox, line 159: I think it could be a good idea to keep a time-based condition for channel refresh. Currently channel infos are stored once in the DB and never updated, which could make user wonder why some channels have outdated infos (e.g description). Something like twice a week, maybe?

SamantazFox (approved):
Other than the two suggestions, it looks good to me :)
(Though, you could squash some commits to reduce the backlog spam when we'll merge that)

src/invidious/channels/channels.cr
matthewmcgarvey: I didn't realize this when making this change, but we already have a job for this https://github.com/iv-org/invidious/blob/master/src/invidious/jobs/refresh_channels_job.cr

Looking through the code now, it looks like it refreshes every channel's data every 30 minutes. Could probably be tweaked 😬

SamantazFox: Note that the channel refreshing job can be disabled by config, making this function the only way to update the channel's data.

SamantazFox (approved):

SamantazFox:
Could you please rebase this on master (conflict because of #2666).

PR #2834, opened 2022-01-22
Refresh channels cleanup

matthewmcgarvey:
Separated out the job's purpose from implementation code. The only difference in the implementation is that now, the `lim_fibers` is set back to `max_fibers` if an error occurs but it was because the channel no longer exists. Before it would not update `lim_fibers` so it would be the same as it was before.

I think there are some potential issues with this job implementation (and it probably carries over to others) but I didn't attempt to make any changes to that.

- _EVERY_ channel is updated every time the job is run instead of just channels that we would identify as needing update
- The backoff is never reset for the life of the job so if this job fails the second day invidious runs and moves the backoff to 1 day, two years later it will still wait a full day if a request fails if the app hasn't been restarted
- The backoff applies to individual channel fibers which means that if there's a youtube problem, we tie up all of the max channel threads in sleeping rather than the main fiber for the job
- That's a LOT of implementation code! Is there an abstraction we can extract?
- We mark channels as deleted, but SELECT sql query does not account for it so we try to refetch deleted channels over and over
- Since we fetch the originally stored author, we could only update the author if it's changed

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -132,7 +132,7 @@ end
 # Start jobs
 
 if CONFIG.channel_threads > 0
-  Invidious::Jobs.register Invidious::Jobs::RefreshChannelsJob.new(PG_DB)
+  Invidious::Jobs.register Invidious::Jobs::RefreshChannelsJob.new
 end
 
 if CONFIG.feed_threads > 0
--- a/src/invidious/database/channels.cr
+++ b/src/invidious/database/channels.cr
@@ -76,6 +76,14 @@ module Invidious::Database::Channels
 
     return PG_DB.query_all(request, as: InvidiousChannel)
   end
+
+  def select_all : Array(InvidiousChannel)
+    request = <<-SQL
+      SELECT * FROM channels
+    SQL
+
+    return PG_DB.query_all(rqeuest, as: InvidiousChannel)
+  end
 end
 
 #
--- a/src/invidious/database/channels.cr
+++ b/src/invidious/database/channels.cr
@@ -82,7 +82,7 @@ module Invidious::Database::Channels
       SELECT * FROM channels
     SQL
 
-    return PG_DB.query_all(rqeuest, as: InvidiousChannel)
+    return PG_DB.query_all(request, as: InvidiousChannel)
   end
 end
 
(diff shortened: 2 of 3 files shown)


SamantazFox:
> Separated out the job's purpose from implementation code. The only difference in the implementation is that now, the `lim_fibers` is set back to `max_fibers` if an error occurs but it was because the channel no longer exists. Before it would not update `lim_fibers` so it would be the same as it was before.

Yup, I think that the entire refreshing system has to be re-though to better handle updates without stressing the DB. For instance, time configurations could be dynamic, based on the amount of users and the popularity of a channel, unless a static value is set from the config.

I.e: refreshing should be done only a couple of times a day for a single user, and on large instance, popular channels should be refreshed more often than less popular ones, while being overall refreshed more frequently than on a single-user instance.

That would have to be discussed, obviously.

> The backoff is never reset for the life of the job so if this job fails the second day invidious runs and moves the backoff to 1 day, two years later it will still wait a full day if a request fails if the app hasn't been restarted

Oh, right. Could that be easily fixed?

> The backoff applies to individual channel fibers which means that if there's a youtube problem, we tie up all of the max channel threads in sleeping rather than the main fiber for the job

Mmmh, yeah. That should be an easy fix, I guess.

> That's a LOT of implementation code! Is there an abstraction we can extract?

What do you mean?

> We mark channels as deleted, but SELECT sql query does not account for it so we try to refetch deleted channels over and over

On that part, I'm not sure why it's done that way. Maybe it's because flagged channels (for policy violation, DMCA claims, or whatever) can disappear for some time, and then be available back, when the case is solved between the owner and Youtube?

> Since we fetch the originally stored author, we could only update the author if it's changed

I'm not sure to understand what you mean? The author name is, with the channel status (deleted?), the only two infos that we're updating from Youtube. See the `InvidiousChannel` structure:
https://github.com/iv-org/invidious/blob/662cf2f6333c7a3c01997862f6f98d17aa103ebd/src/invidious/channels/channels.cr#L1-L9

matthewmcgarvey:
I think it's pretty clear now that this change is pretty pointless. I'm going to close it. There's more important things to do here

PR #2846, opened 2022-01-28
Ignore "/" key handling if search box is focused

SamantazFox:
Fixes a side effect of https://github.com/iv-org/invidious/pull/2814
See: https://github.com/iv-org/invidious/issues/2791#issuecomment-1018264144

### DIFF
--- a/assets/js/handlers.js
+++ b/assets/js/handlers.js
@@ -146,7 +146,17 @@
     // Handle keypresses
     window.addEventListener('keydown', (event) => {
         // Ignore modifier keys
-        if (event.ctrlKey || event.metaKey) { return; }
+        if (event.ctrlKey || event.metaKey) return;
+
+        // Ignore shortcuts if any text input is focused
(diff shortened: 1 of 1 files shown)


SamantazFox:
Deployed on https://test.invidious.io

assets/js/handlers.js
```diff
@@ -150,7 +150,10 @@
 
         // Focus search bar on '/'
         if (event.key == "/") {
-            document.getElementById('searchbox').focus();
+            let searchbox = document.getElementById('searchbox');
+            if (searchbox == document.activeElement) return;
```
matthewmcgarvey, line 154: Should we ignore it if any text box is focused?

SamantazFox: While there is currently only one text box on every page (the search box) you're entirely right! This will avoid problems in the future!

SamantazFox:
deployed on https://test.invidious.io

PR #2847, opened 2022-01-30
rename hls to vhs in the player

unixfox:
hls method is deprecated: https://github.com/videojs/http-streaming#how-to-use


SamantazFox:
For reference, here's the deprecation commit:
https://github.com/videojs/http-streaming/commit/22af0b29e56e8a25d55737b0148da0ab5e4d461b

PR #2856, opened 2022-02-03
Fix related videos

### LINKED ISSUE #2830: [Bug] "0 views" on all related videos
**Describe the bug**
On the "watch" page, all related videos (right column) display 0 views

**Steps to Reproduce**
1. Enable related videos column
2. Got to any video
3. See error

**Screenshots**
image

SamantazFox:
Closes #2830

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -885,16 +885,24 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
 
   # Top level elements
 
-  primary_results = player_response
-    .dig?("contents", "twoColumnWatchNextResults", "results", "results", "contents")
+  main_results = player_response.dig?("contents", "twoColumnWatchNextResults")
+
+  raise BrokenTubeException.new("twoColumnWatchNextResults") if !main_results
+
+  primary_results = main_results.dig?("results", "results", "contents")
+
+  raise BrokenTubeException.new("results") if !primary_results
 
   video_primary_renderer = primary_results
-    .try &.as_a.find(&.["videoPrimaryInfoRenderer"]?)
-      .try &.["videoPrimaryInfoRenderer"]
+    .as_a.find(&.["videoPrimaryInfoRenderer"]?)
+    .try &.["videoPrimaryInfoRenderer"]
 
   video_secondary_renderer = primary_results
-    .try &.as_a.find(&.["videoSecondaryInfoRenderer"]?)
-      .try &.["videoSecondaryInfoRenderer"]
+    .as_a.find(&.["videoSecondaryInfoRenderer"]?)
+    .try &.["videoSecondaryInfoRenderer"]
+
+  raise BrokenTubeException.new("videoPrimaryInfoRenderer") if !video_primary_renderer
+  raise BrokenTubeException.new("videoSecondaryInfoRenderer") if !video_secondary_renderer
 
   # Likes/dislikes
 
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -455,7 +455,7 @@ struct Video
 
                       qualities.each do |quality|
                         json.object do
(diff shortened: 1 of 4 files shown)


src/invidious/videos.cr
```diff
@@ -455,7 +455,7 @@ struct Video
 
                       qualities.each do |quality|
                         json.object do
-                          json.field "url", rv["author_thumbnail"]?.try &.gsub(/s\d+-/, "s#{quality}-")
+                          json.field "url", rv["author_thumbnail"].try &.gsub(/s\d+-/, "s#{quality}-")
```
matthewmcgarvey, line 458: Since you removed the `?` you don't need the `.try` anymore as it will never be nil

src/invidious/videos.cr
```diff
@@ -871,30 +898,61 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
...
+    .dig?("secondaryResults", "secondaryResults", "results")
+
+  raise BrokenTubeException.new("results") if !primary_results
+  raise BrokenTubeException.new("secondaryResults") if !secondary_results
 
   video_primary_renderer = primary_results
-    .try &.as_a.find(&.["videoPrimaryInfoRenderer"]?)
-      .try &.["videoPrimaryInfoRenderer"]
+    .as_a.find(&.["videoPrimaryInfoRenderer"]?)
+    .try &.["videoPrimaryInfoRenderer"]
```
matthewmcgarvey, line 916: This is fine, but looking at all this, I'm wondering if a json path library would be useful for us https://github.com/json-path/JsonPath (this is a java one)

I've only ever used it in testing, but the equivalent of this would be an jsonpath of `$[?(@.videoPrimaryInfoRenderer)].videoPrimaryInfoRenderer` and it would return an array and we'd try to take the first item from it. Upside is that we'd have cleaner JSON searching, downside is that jsonpath is kind of confusing

src/invidious/videos.cr
```diff
@@ -871,30 +898,61 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
...
+
+  # If nothing was found previously, fall back to end screen renderer
+  if related.empty?
+    # Container for "endScreenVideoRenderer" items
+    player_overlays = player_response.dig?(
+      "playerOverlays", "playerOverlayRenderer",
+      "endScreen", "watchNextEndScreenRenderer", "results"
+    )
+
+    secondary_results.try &.as_a.each do |element|
```
matthewmcgarvey, line 947:
```suggestion
    player_overlays.try &.as_a.each do |element|
```

matthewmcgarvey (changes requested):
Minor fixes, otherwise looks fine

src/invidious/videos.cr
SamantazFox: Ah, yep, good catch!

SamantazFox:
```suggestion
                          json.field "url", rv["author_thumbnail"].gsub(/s\d+-/, "s#{quality}-")
```

matthewmcgarvey (approved):

src/invidious/videos.cr
SamantazFox: I don't think so.

The syntax is indeed quite confusing, and would make maintenance harder, especially for new comers (We already don't have a lot of contributions, I'd prefer to not require them to get familiar with such syntax).

The basic crstal syntax is good enough and quite legible, as long as we respect some styling rules to avoid a nightmare of `.try` statements and cascading conditions.

SamantazFox:
Oh, I have forgotten something!
I should fall back to the "short" count if the other is not available.

image

SamantazFox:
Better! Deployed at https://test.invidious.io

matthewmcgarvey (approved):

Issue #2857, opened 2022-02-03
[Enhancement] Show Youtube channel verification


jonas-w:
> I found no other issue with this being asked.  

**Is your enhancement request related to a problem? Please describe.**

Maybe it's not wanted in Invidious. But I miss the verification check mark next to comments or on the channel page. But Maybe I'm the only one.

**Describe the solution you'd like**

Add the youtube channel verification check mark to the comments and next to the channel.

**Describe alternatives you've considered**

Adding own Invidious verification check marks (e.g. Verified FOSS Content Creator, ...)

**Additional context**

> If this isn't wanted in Invidious i can understand that. I just want to know the reasons why it doesn't exist till now.

unixfox:
This seems like a very simple feature to implement, would you be interested to contribute?

jonas-w:
Yes I would be interested. But I'm not familiar with the invidious codebase and crystal language. Is it complicated to get into it?

unixfox:
It's not that complicated. If you like programming, you will like programming in crystal.

Here is the documentation for installing Invidious locally: https://docs.invidious.io/Installation.md and how to contribute: https://github.com/iv-org/invidious#contribute

We can also try to help you to get into it, we have : 
- a matrix channel: https://matrix.to/#/#invidious:matrix.org
- an IRC channel: https://web.libera.chat/?channel=#invidious

jonas-w:
Could the docs be outdated. I'm currently going through the manual installation. I'm using Manjaro (sorry im lazy) and on the "Setup Postgresql" Part i get permission denied on e.g. `psql invidious kemal < /home/invidious/invidious/config/sql/channels.sql`. I Solved this by using the invidious user. 
And in the shards update and install part i needed to use root priviliges because it got permission denied on "/home/manjaro/bin/pkg-config". And also the compiling part is only working on root priviliges.

unixfox:
It's probably because we assume the user is running in root, which to be fair we shouldn't assume that.

If you think the documentation should be adapted, then you can submit some modifications on this file: https://github.com/iv-org/documentation/blob/master/Installation.md 😃.

jonas-w:
But in the docs you assume that the crystal commands get executed by the user invidious because of the `sudo -iu invidious`. Or did i oversee something?

unixfox:
Oops sorry, you are right. The one time I used the documentation for installing manually way ages ago and back then IIRC it assumed you used root, so that's why I thought it was still the case.

[...]

PR #2859, opened 2022-02-03
Youtube verification badge

jonas-w:
> #2857 Feature Added partly. Only on channel view currently.

# TODO: Verification Badge in:

- [x] Channel View
- [x] Comments
- [x] Search videos
- [x] Related videos
- [x] Trending videos / Default Tab / the rest works
- [x] Playlist videos
- [x] (Popular Page Videos)
- [x] (Channel Page Videos)


src/invidious/views/community.ecr
```diff
@@ -20,6 +20,9 @@
         <div class="channel-profile">
             <img src="/ggpht<%= URI.parse(channel.author_thumbnail).request_target %>">
             <span><%= author %></span>
+            <% if channel.verified %>
+                <i class="icon ion ion-md-checkmark-circle"></i>
+            <% end %>
```
SamantazFox, line 25: Note that the content can be inlined with the `if` statement like so:
```
<% if channel.verified %><i class="icon ion ion-md-checkmark-circle"></i><% end %>
```

This reduces the amount of characters sent to the client (the indentations in this file are not removed).

jonas-w: Okay thank you, as mentioned in the issue this is my first time using crystal language.

SamantazFox:
There is also here, in the video watch page:
https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/views/watch.ecr#L238-L245

the data is extracted here:
https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/videos.cr#L1036-L1045

And here is the corresponding JSON structure:
```json
"videoSecondaryInfoRenderer": {
  "owner": {
    "videoOwnerRenderer": {
      "badges": [
        {
          "metadataBadgeRenderer": {
            "icon": {
              "iconType": "CHECK_CIRCLE_THICK"
            },
[SNIP]
```

SamantazFox:
Here, in the related videos:
https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/views/watch.ecr#L315-L321

Data is parsed here:
https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/videos.cr#L810

Associated JSON structure:
```json
"compactVideoRenderer": {
  "ownerBadges": [
    {
      "metadataBadgeRenderer": {
        "icon": {
          "iconType": "CHECK_CIRCLE_THICK"
        },
        "style": "BADGE_STYLE_TYPE_VERIFIED",
        "tooltip": "Verified",
[SNIP]
```

unixfox:
By the way you can use `crystal tool format` for linting/formatting automatically the code :)

jonas-w:
Okay i've added comments and channel name in watch.ecr and i'm currently working on the related videos.
And in this line it tries to convert a bool to a string:
https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/videos.cr#L560

Could you please review my code and give me a tip to go around this  error?

SamantazFox:
> Okay i've added comments and channel name in watch.ecr and i'm currently working on the related videos. And in this line it tries to convert a bool to a string:
> 
> https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/videos.cr#L560
> 
> Could you please review my code and give me a tip to go around this error?

I'd recommend to search if the badge exists first (use `.find()` to search in array children) which will give you a `Bool` value, that you can convert to `String`:
```crystal
author_verified_badge = related["ownerBadges"]?.try do |badges_array|
  badges_array.as_a.find(&.dig("metadataBadgeRenderer", "tooltip").as_s.== "Verified")
end

author_verified = (author_verified_badge && !author_verified_badge.empty?)
```
        
to explain the code:
* `related["ownerBadges"]?` search the JSON element (in our case, a `"compactVideoRenderer"`) for a `ownerBadges` child: if child exists, returns that, otherwise return `nil`
* `.try do` only execute block if previous statement didn't return `nil`
* `.as_a` transform JSON entry to Array
* `.find()` something in this array, according to those criteras:
  - `&` (simply refers to the array elments)
  - `.dig("x","y")` search for a child named `x` and grand-child named `y`
  - `.as_s` convert JSON object to String
  - `.==` compares to another String

We end up with an array if the `"ownerBadges"` field exist, and `nil` if it doesn't.

We just have to check that the result is not `nil`, and that, in case it's an array, that array is not empty (otherwise, that would mean that `find` didn't find our criteras).

SamantazFox:
As to why it tries to convert to string: this is in part because it's stored in the database as a huge JSON string, and because that data structure wasn't really though to be expanded. So for now we'll keep strings, and in the future I'll try to refactor the whole video structures to make the handling of data more consistent across the different parts of the code.

Also note that when you'll use the structure (in the HTML), don't expect the field to always be present: as data is cached in the DB, older entries won't have that field. So use `rv["authorVerified"]? == "true"` when trying to get the value!

jonas-w:
Okay sorry for the inconvenience. I will implement this later!

jonas-w:
I'm currently struggling with https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/channels/channels.cr#L158

> I'm confused where this comes from and where i could get the verification status

### Sorry for asking so much questions!

SamantazFox:
> Okay sorry for the inconvenience. I will implement this later!

> Sorry for asking so much questions!

Don't be sorry ^^ It's normal to not be familiar with other people's code :)


> I'm currently struggling with
> 
> https://github.com/iv-org/invidious/blob/df599c0fc3226cdb5ce4cccfd26c02228976cac3/src/invidious/channels/channels.cr#L158
> 
> I'm confused where this comes from and where i could get the verification status

I'm not sure: what are your looking for, exactly?
The other places where youtube results are parsed?

Also, do you have suitable JSON API responses to see how data is structured? If you don't, don't hesitate to post screenshots of places in the interface that are missing the checkmark, so I can get those and point you to the right place :)

jonas-w:
For `components/item.ecr` to work correctly i also need to init `ChannelVideo` and in that method it fetches from an videos.xml and this confuses me a bit because very other part i implemented had to do with JSON.

SamantazFox:
@jonas-w mmmh. I'll recommend you to not bother with `ChannelVideo`. It's deeply rooted with the notification system [^1]. You can use a conditional like that to avoid `ChannelVideo` items in the `else` branch of `items.ecr`:
```crystal
if !item.is_a?(ChannelVideo) && item.author_verified
```

[^1]: the notifications need exact timestamps, which we can't get from the internal API (it only says "published 3 days ago"). So that's why we're extracting data from the RSS feed ^^

jonas-w:
Okay thank you! That item.is_a? function is what i've search everything for but found nothing :)

jonas-w:
Okay as far as i can see. This feature should be done. Maybe some refactoring. But the core functionality is there!

src/invidious/channels/about.cr
```diff
@@ -70,7 +71,9 @@ def get_about_info(ucid, locale) : AboutChannel
     # if banner.includes? "channels/c4/default_banner"
     #  banner = nil
     # end
+    author_verified_badges = initdata["header"]?.try &.["c4TabbedHeaderRenderer"]?.try &.["badges"]?
```
SamantazFox, line 74: Please use `.dig?()` rather than `.try` chains! See: https://github.com/iv-org/invidious/issues/2231#issue-935010191

SamantazFox, line 74: Please use `.dig?()` rather than `.try` chains! See: https://github.com/iv-org/invidious/issues/2231#issue-935010191

src/invidious/channels/about.cr
```diff
@@ -70,7 +71,9 @@ def get_about_info(ucid, locale) : AboutChannel
...
 
+    author_verified = (author_verified_badges && author_verified_badges.size > 0)
```
SamantazFox, line 76: I see that you copied this from the example I gave you in the discussions. However, you forgot to use `.as_a` on `author_verified_badges`, and you also forgot to check if the badges array has a "verified" badge child.

src/invidious/comments.cr
```diff
@@ -144,7 +144,8 @@ def fetch_youtube_comments(id, cursor, format, locale, thin_mode, region, sort_b
 
               content_html = node_comment["contentText"]?.try { |t| parse_content(t) } || ""
               author = node_comment["authorText"]?.try &.["simpleText"]? || ""
-
+              verified = (node_comment["authorCommentBadge"]? != nil)
+              json.field "verified", (verified || false)
```
SamantazFox, line 148: you can inline the `!= nil` statement with the `json.field` line. Also note that you don't need to fallback to `false` (using the (`||`) operator), as your verified wheck will always be boolean.

And please keep the existing blank lines between code chunks, to improve readibility :)

src/invidious/helpers/serialized_yt_data.cr
```diff
@@ -140,7 +142,7 @@ struct SearchPlaylist
       json.field "author", self.author
       json.field "authorId", self.ucid
       json.field "authorUrl", "/channel/#{self.ucid}"
-
+      json.field "authorVerified", self.author_verified
```
SamantazFox, line 145: ditto empty new lines between code chunks

src/invidious/videos.cr
```diff
@@ -570,6 +570,10 @@ struct Video
     info["authorThumbnail"]?.try &.as_s || ""
   end
 
+  def author_verified : Bool
+    info["authorVerified"].as_bool
```
SamantazFox, line 574: please use `.try &.as_bool || false`, as the video parsing should fail as nicely as possible.
We don't want the users to be sent to the "you found a bug" page for just a missing field ^^

src/invidious/videos.cr
```diff
@@ -1037,7 +1047,9 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
 
   author_info = video_secondary_renderer.try &.dig?("owner", "videoOwnerRenderer")
   author_thumbnail = author_info.try &.dig?("thumbnail", "thumbnails", 0, "url")
+  author_verified_badge = author_info.try &.["badges"]?
```
SamantazFox, line 1050: Same as above: you forgot to check if the badges element is an array, and if it has a "verified" badge child!

src/invidious/views/channel.ecr
```diff
@@ -20,7 +20,7 @@
     <div class="pure-u-2-3">
         <div class="channel-profile">
             <img src="/ggpht<%= URI.parse(channel.author_thumbnail).request_target %>">
-            <span><%= author %></span>
+            <span><%= author %></span><% if !channel.verified.nil? && channel.verified %> <i class="icon ion ion-md-checkmark-circle"></i><% end %>
```
SamantazFox, line 23: I'd recommend using an non-breakable space (`&nbsp;` in HTML) over a normal space, so it won't be inadvertently squished by the browser ;)

src/invidious/views/components/item.ecr
```diff
@@ -45,7 +45,7 @@
                 <p dir="auto"><%= HTML.escape(item.title) %></p>
             </a>
             <a href="/channel/<%= item.ucid %>">
-                <p dir="auto"><b><%= HTML.escape(item.author) %></b></p>
+                <p dir="auto"><b><%= HTML.escape(item.author) %><% if !item.is_a?(MixVideo) && !item.author_verified.nil? && item.author_verified %> <i class="icon ion ion-md-checkmark-circle"></i><% end %></b></p>
```
SamantazFox, line 48: In this part of the code, you're only dealing with a `MixVideo` (note the `when MixVideo` on line 35).
That means that you don't need to do any changes that line, as it seems that you want to exclude `MixVideo` kind of videos. They don't have an owner badge structure?

src/invidious/yt_backend/extractors.cr
```diff
@@ -129,6 +133,7 @@ private module Parsers
         live_now:           live_now,
         premium:            premium,
         premiere_timestamp: premiere_timestamp,
+        author_verified:    author_verified || false,
```
SamantazFox, line 136: you don't need the `|| false` fallback here either!

src/invidious/yt_backend/extractors.cr
```diff
@@ -102,7 +102,11 @@ private module Parsers
       premium = false
 
       premiere_timestamp = item_contents.dig?("upcomingEventData", "startTime").try { |t| Time.unix(t.as_s.to_i64) }
+      author_verified_badge = item_contents["ownerBadges"]?.try do |badges_array|
+        badges_array.as_a.find(&.dig("metadataBadgeRenderer", "tooltip").as_s.== "Verified")
+      end
 
+      author_verified = (author_verified_badge && author_verified_badge.size > 0)
```
SamantazFox, line 109: Given how often this code is repeated, you should make it a function! You can pass to that function the `JSON::Any` object that contains the `"ownerBadges"` field, i.e `item_contents` in this context and `related` in a previous file.

Also, please put said function under `HelperExtractors`, in that same file (`yt_backend/extractors.cr`). That's where the other similar utility functions are.

SamantazFox (changes requested):
Overall, you did well :) Just some small mistakes and a few places where code could be improved (either cleaned a bit, or de-duplicated).

src/invidious/yt_backend/extractors.cr
jonas-w: Sorry that i didnt  answer, i have currently many personal problems and i hope that i can get tomorrow back to coding!

src/invidious/comments.cr
```diff
@@ -328,7 +329,9 @@ def template_youtube_comments(comments, locale, thin_mode, is_replies = false)
       end
 
       author_name = HTML.escape(child["author"].as_s)
-
+      if child["verified"]?.try &.as_bool
+        author_name += " <i class=\"icon ion ion-md-checkmark-circle\"></i>"
```
ChunkyProgrammer, line 333: Not sure of the degree that Invidious wants to copy design elements from YT but I would like to note that YouTube would use 'checkmark-circle' for Verified comments by the uploader and would just use 'checkmark' for verified comments by other commentors

src/invidious/yt_backend/extractors.cr
SamantazFox: Yeah, sure, no problem ^^ Take care of yourself, that more important!

src/invidious/yt_backend/extractors.cr
jonas-w: Tried removing it and it seems to be required or at least at some point

src/invidious/yt_backend/extractors.cr
jonas-w: Okay sorry, this took a bit longer than "tomorrow". 

I currently refactored everything except this extractor method. 
I remember that every time i tried implementing this as a method it breaks and i needed like a few exception cases where i would have written another helper method but at the end it felt like it was more code than before. Probably because this is the first time writing in crystal lang.

src/invidious/comments.cr
jonas-w: @SamantazFox opinion? Do you want this in invidious?

SamantazFox: The more youtube-like features it has, the better ^^

SamantazFox:
Hello! Sorry for the long wait, I was quite busy IRL.

It seems that the verified badge is missing in playlists. The data is not available?
https://test.invidious.io/playlist?list=PL8mG-RkN2uTx9Er8AqFDnD6mFvje5fybs

Same question for the "next" video in playlist (above recommended):
image

Other than that, it looks good!

jonas-w:
> @jonas-w mmmh. I'll recommend you to not bother with `ChannelVideo`. It's deeply rooted with the notification system [1](#user-content-fn-1-a2d5bbe900ed98aa3ed1854f8c60bba5). You can use a conditional like that to avoid `ChannelVideo` items in the `else` branch of `items.ecr`:
> 
> ```crystal
> if !item.is_a?(ChannelVideo) && item.author_verified
> ```
> 
> ## Footnotes
> 
>     1. the notifications need exact timestamps, which we can't get from the internal API (it only says "published 3 days ago"). So that's why we're extracting data from the RSS feed ^^ [leftwards_arrow_with_hook](#user-content-fnref-1-a2d5bbe900ed98aa3ed1854f8c60bba5)

IIRC correctly Playlists were also in that xml format and you told me not to bother with that

SamantazFox:
> IIRC correctly Playlists were also in that xml format and you told me not to bother with that

Ah! I though that playlist items have their own class. I'll check that.

SamantazFox:
Mmmh, yep. `Playlist` and `PlaylistVideo` are two different classes, so if the verified badge is available in playlist data, you should be able to display it there as well!

jonas-w:
Oh okay, will have a look at it!

SamantazFox:
@jonas-w What's the status?

SamantazFox:
I'll merge as-is, as it's working fine. If anything needs to be added, we'll do that in another PR!

SamantazFox:
@jonas-w And thanks for your contribution to invidious :D

jonas-w:
Oh shit I'm so so sorry! My mind is so full with so much stuff...(┛◉Д◉)┛彡┻━┻ But I'm thankful you still merged it as-is.

PR #2874, opened 2022-02-08
Small fixes

### LINKED ISSUE #1234: US and UK english autoload subtitle not available in options
I have set the subtitles to show automatically if they are in english (but not auto generated ones)
this works for videos that name the language of subtitle "english" but not with those that are named "english (united stats)" and "english (united kingdom)"
like this video
https://invidious.snopyta.org/watch?v=54XLXg4fYsc

can you fix this so that english from us and uk are also automatically shown when simple english is selected in "Default captions:"

### LINKED ISSUE #1810: [Bug] Error message isn't HTML escaped
The error message isn't HTML escaped, leading to messages like https://github.com/iv-org/invidious/issues/1755.

> Route: `/feed/trending?type=News®ion=US`

should be

> Route: `/feed/trending?type=News&region=US`

<hr>

The error message is created here: https://github.com/iv-org/invidious/blob/b29e60a97aaef1d513208e177ecae8ffc671a3d9/src/invidious/helpers/errors.cr#L25-L44
See https://crystal-lang.org/api/0.35.1/HTML.html#escape(string:String):String-class-method

### LINKED ISSUE #2118: [Enhancement] Mention that 360° videos requires WebGL

Follow up to https://github.com/iv-org/invidious/pull/2010

360° videos only works with WebGL the only reason I know that was by asking @syeopite when it didn't work for me.

The preferences should be reworded from `Interactive 360 degree videos` to `Interactive 360° videos (requires WebGL)` (no reason not to use `°`)

Find a VR lib that doesn't requires WebGL

**Additional context**

N/A

### LINKED ISSUE #2839: [Bug] Pressing any numpad number skips to the end of the video
**Describe the bug**

Pressing a number on the numpad skips to the end of the video, unlike normal buttons

**Steps to Reproduce**

1. go to a video
2. start playing it
3. press any numpad number
4. video skips to end

### LINKED ISSUE #2873: [Bug] missing dependency in embed video: player.overlay is not a function (video-overlay.min.js)
**Describe the bug**
Hi a feature was added to add a button to link back to the instance from a embed indivious frame (see #1995)
However this feature doesn't seem to work anymore, I've looked on other instances to see if I was the only one having the issue but it doesn't seems like it.
If I'm going in on an embed link I get the following error under firefox 96: player.overlay is not a function.
I've tried copying the content of "videojs-overlay.min.js" inside player.js and it fixed the issue right away so it really seems like a missing dependency.

Thanks!

**Steps to Reproduce**
1- Go to any embed link (ex: https://yewtu.be/embed/0gct28UEKws)
2- Invidious should show up next to the seek bar but it's not
3- CTRL+SHIFT+P or F12 then console show the error mentioned above.

### LINKED ISSUE #2876: [Bug] Most issue pages broken since February 3rd
**Describe the bug**

Originally discussed on Matrix.

Most (if not all?) error page are broken  since a PR from February 3rd

**Steps to Reproduce**

Broken: https://yt.artemislena.eu/watch?v=plHrNpNd9zk (dated February 3rd)

Works: https://inv.cthd.icu/watch?v=plHrNpNd9zk (dated February 2nd)

example from https://github.com/iv-org/invidious/issues/2863

**Logs**

N/A

**Screenshots**

broken

**Additional context**

N/A

SamantazFox:
Closes #1234 (US/UK english auto-generated subtitles not available (and more, actually))
Closes #1810 (Error message isn't HTML escaped)
Closes #2118 (Mention that 360° videos requires WebGL)
Closes #2839 (pressing any numpad key jumps to end of video)
Closes #2873 (error on embed page "player.overlay is not a function")
Closes #2876 (empty error message on many pages)

Also adds a compile option to not fetch player dependencies (useful for `make verify`) and fixes a TypeError in `handlers.js` (mentioned in a comment I can't find).


Makefile
```diff
@@ -62,7 +62,8 @@ test:
 	crystal spec
 
 verify:
-	crystal build src/invidious.cr --no-codegen --progress --stats --error-trace
+	crystal build src/invidious.cr -Ddont_fetch_videojs \
```
matthewmcgarvey, line 65: I think `skip_videojs` makes sense

src/invidious.cr
```diff
@@ -114,16 +114,18 @@ LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
...
-#
-# Running the script by itself would show some colorful feedback while this doesn't.
-# Perhaps we should just move the script to runtime in order to get that feedback?
-
-{% puts "\nChecking player dependencies...\n" %}
-{% if flag?(:minified_player_dependencies) %}
-  {% puts run("../scripts/fetch-player-dependencies.cr", "--minified").stringify %}
-{% else %}
-  {% puts run("../scripts/fetch-player-dependencies.cr").stringify %}
+{% unless flag?(:dont_fetch_videojs) %}
```
matthewmcgarvey, line 117: 🙏 `if !flag?(...)`

unless is such a pain for me to read

assets/js/player.js
```diff
@@ -612,6 +612,9 @@ window.addEventListener('keydown', e => {
         case '7':
         case '8':
         case '9':
+            // Ignore numpad numbers
+            if (code > 57) break;
```
matthewmcgarvey, line 557: Instead of using the code, let's just parse `decoratedKey` to an integer and multiply by 10. Wouldn't that work for both number sets?

Makefile
SamantazFox: what about `skip_videojs_download`?

matthewmcgarvey: 👍

assets/js/player.js
SamantazFox: yeah, that should work. Though, I'm unsure about extending that feature to the numpad. I happen to often press numpad keys by mistake when reaching stuff next to my keyboard, and I didn't like the video jumping to the end...

matthewmcgarvey: I've felt that pain before. I'm fine with that, though one thing I looked into is what those numbers actually are but the things I tested with showed the same numbers for the numpad and the other number row. Maybe it's something to do with me using an external keyboard on a macbook pro and is different on windows/linux

matthewmcgarvey (approved):

SamantazFox:
this last item has been deployed and tested on the test instance

PR #2878, opened 2022-02-09
Add custom migration implementation

### LINKED ISSUE #2620: [Discussion] Database migrations framework
This subject is a bit related to #2498, as most database migration tools come with a web framework.

We **must** use a database migration engine before doing any changes to the DB (like #2469), in order to avoid issues that could be caused by a manual migration. It will also ease the initial DB deployment by not requiring the script file currently used for docker, that prevents the dockerfile/docker-compose files to be used standalone, without cloning.

This issue is a discussion around the different options available, their pros and their cons.

Personally, my vote will be either for micrate or a fork of mg to implement MySQL/PGSQL support.

## Candidtates

- [amberframework/micrate](https://github.com/amberframework/micrate)
  * **Pros**
    - Supports all major DB types: PGSQL, MySQL and SQLite
    - Is independant from the amber framework (no dependencies other than the DB drivers)
  * **Cons**
    - DB migrations folder path is hard-coded, so only one of the DB types can be used at once
    - DB migrations can only be in separate SQL files. They can't be embedded in the final crystal executable.
    - Unknown license (see https://github.com/amberframework/micrate/issues/81)

- [luckyframework/avram](https://github.com/luckyframework/avram)
  * **Pros**
    - Has its own "abstraction language". No need to write heavy SQL statements ourselves.
    - Integrates with the rest of the lucky framework
    - Active development
  * **Cons**
    - Heavily depends on the lucy framework (implies that we'll use it for the web part, too)
    - Only supports PGSQL (and this will probably never change)
    - Very complex, because of that "abstraction language"

- [hkalexling/mg](https://github.com/hkalexling/mg)
  * **Pros**
    - Extremely simple
    - SQL statements are embedded in Crystal classes with `up` and `down` methods
  * **Cons**
    - Only supports SQLite

- [vladfaust/migrate.cr](https://github.com/vladfaust/migrate.cr)
  * **Pros**
    - Migration folder can be specified at initialization (unlike micrate) 
  * **Cons**
    - Outdated (last commit from August 2019, shard.yml refers to an old version of crystal)
    - DB migrations can only be in separate SQL files. They can't be embedded in the final crystal executable.
    - Only supports PGSQL

matthewmcgarvey:
Fixes #2620

Instead of adding a migration library that is unmaintained, restrictive, or we'd have to tweak anyways, why not make our own? As you can see it really doesn't take that much code to get a minimal version working.

This implementation has things to point out:
- Migrations are versioned, that version is provided inside the class (filename is ignored but useful for file ordering, could be changed in the future)
- Migrations are run in the order of the versions, but ALL unapplied migrations are run. So if version 8 has run, but version 6 hasn't it will still be applied the next time migrations are run. This is useful if you have a migration you _always_ want to run last but add migrations after that one. It can be version 9999 but not force later migrations to be 10000.
- No rollback capability, maybe the concept is useful for local dev but it can be added later if we need it
- No CLI
- No ability to create or drop the database
- Each individual migration is run within a transaction so if any of the sql statements fail, the migration will not bel applied 

Feel free to push back on this, my feelings won't be hurt 😄


src/invidious/migrations/0007_create_playlists_table.cr
```diff
@@ -0,0 +1,47 @@
...
+            (
+                'Public',
+                'Unlisted',
+                'Private'
+            );
+        END IF;
+      END;
+      $$
+      LANGUAGE plpgsql;
+      SQL
```
matthewmcgarvey, line 25: This is the only sql that was changed to allow for it to be run if the privacy type already exists. Got the sql to do it from https://stackoverflow.com/questions/56647514/create-a-type-if-not-exist-postgresql

src/invidious/migration.cr
```diff
@@ -0,0 +1,38 @@
+abstract class Invidious::Migration
+  macro inherited
+    Invidious::Migrator.migrations << self
```
matthewmcgarvey, line 3: migrations are automatically added to this list when the file is required which is why there's no need to do anything with migrations after they're defined.

src/invidious/migration.cr
```diff
@@ -0,0 +1,38 @@
...
+  def migrate
+    # migrator already ignores completed migrations
+    # but this is an extra check to make sure a migration doesn't run twice
+    return if completed?
+
+    @db.transaction do |txn|
+      up(txn.connection)
+      track(txn.connection)
+      @completed = true
+    end
```
matthewmcgarvey, line 28: This is the core code that runs the migration

unixfox:
Thank you for your work! 

I'm definitively going to try it on https://github.com/iv-org/invidious/pull/2469/files#diff-64021f64c486d8a36eae29fce4c1b13e9c95f5cbb713ca7bcd26f8d38517eeb8

src/invidious/migrations/0001_create_videos_table.cr
```diff
@@ -0,0 +1,28 @@
+module Invidious::Migrations
+  class CreateVideosTable < Migration
+    version 1
+
+    def up(conn : DB::Connection)
+      conn.exec <<-SQL
+      CREATE UNLOGGED TABLE IF NOT EXISTS public.videos
```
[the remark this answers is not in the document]
matthewmcgarvey: I left this as it was and made a new migration altering the table

src/invidious/database/migrations/0008_create_playlists_table.cr
```diff
@@ -0,0 +1,48 @@
...
+        index int8[]
+      );
+      SQL
+
+      conn.exec <<-SQL
+      GRANT ALL ON public.playlists TO current_user;
+      SQL
+    end
+
+    private def privacy_type_exists?(conn : DB::Connection) : Bool
```
matthewmcgarvey, line 37: Instead of the weird pgsql query, I extracted out this check to look for the privacy type and only creating it if this method returns false.

src/invidious.cr
```diff
@@ -111,6 +112,8 @@ end
 OUTPUT = CONFIG.output.upcase == "STDOUT" ? STDOUT : File.open(CONFIG.output, mode: "a")
 LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
 
+# Run migrations
+Invidious::Database::Migrator.new(PG_DB).migrate
```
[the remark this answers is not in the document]
SamantazFox: We could do the following:
- Check for migrations
- If migrations have to be run, exit with an informational message (and a specific error code)
- If `--migrate-db` is passed, run migrations

That way, it's at the discretion of the instance owner to do as they want: if invidious exits with specified error code, they can hook up a wrapper script to stop all other instances, run a backup, and then migrate.

src/invidious/database/migrations/0008_create_playlists_table.cr
```diff
@@ -0,0 +1,48 @@
...
+      !conn.query_one?(<<-SQL, as: Int32).nil?
+      SELECT 1 AS one
+      FROM pg_type
+      INNER JOIN pg_namespace ON pg_namespace.oid = pg_type.typnamespace
+      WHERE pg_namespace.nspname = 'public'
+        AND pg_type.typname = 'privacy'
+      LIMIT 1;
+      SQL
```
SamantazFox, line 45: As a matter of style/readibility, a separate request variable (like in the other DB functions) would be nice!

src/invidious/database/migrator.cr
```diff
@@ -0,0 +1,41 @@
...
+      puts "Running migration: #{migration.class.name}"
+      migration.migrate
+    end
+  end
+
+  private def load_migrations : Array(Invidious::Database::Migration)
+    self.class.migrations.map(&.new(@db))
+  end
+
+  private def load_run_migrations : Array(Int64)
```
SamantazFox, line 28: maybe rename this to `load_versions`? Makes the purpose clearer imo.

SamantazFox (approved):
Looks good to me!

src/invidious.cr
```diff
@@ -101,6 +102,10 @@ Kemal.config.extra_options do |parser|
     puts SOFTWARE.to_pretty_json
     exit
   end
+  parser.on("--migrate", "Run any migrations") do
+    Invidious::Database::Migrator.new(PG_DB).migrate
+    exit
+  end
```
matthewmcgarvey, line 108: This was a good idea, only issue is that it really messes with my flow of using docker-compose

src/invidious.cr
```diff
@@ -111,6 +116,11 @@ end
 OUTPUT = CONFIG.output.upcase == "STDOUT" ? STDOUT : File.open(CONFIG.output, mode: "a")
 LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
 
+# Run migrations
+if Invidious::Database::Migrator.new(PG_DB).pending_migrations?
+  puts "There are pending migrations. Run `invidious --migrate` to apply the migrations."
+  exit 46
```
matthewmcgarvey, line 122: You said use a specific error code 😛. But for real... what error code should I return?

src/invidious.cr
SamantazFox: Oh, right. We should probably make a wrapper script for that?

Also, we probably shouldn't `exit` if migrations are run. Passing `--migrate` would always bump the DB no matter what, without breaking the execution process. And we could make that the default on Docker, so owners of small instances aren't too surprised.

src/invidious.cr
SamantazFox: I don't know xD I'll have to double check that, to ensure that we don't use something that either kemal or crystal already do.

src/invidious.cr
matthewmcgarvey: So I'm concerned that we're going to roll out a breaking change to everyone with this PR. Everyone will update as normal and then find out that they can't run the service and will have to update their command, and I'm not sure the best way to roll that out.

What I'm going to do is leave this as it is, and remove the pending migration check. That way we can merge this change, it won't break anyone, and people can test out the migrations process. Then we can make a follow-up PR with the implementation that requires the migrations.

src/invidious.cr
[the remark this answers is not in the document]
unixfox: Mmmh this definitively need to be tackled properly. I'm running 10 instances of Invidious at the same time, so I don't want this to become an issue haha.

unixfox:
@SamantazFox Please don't merge this PR without extensive testing. I would even consider asking some people to test before merging that to the public.
Don't merge before I've tested it too, I really don't want to break yewtu.be.

SamantazFox:
@unixfox Sure!

src/invidious.cr
```diff
@@ -116,11 +116,6 @@ end
 OUTPUT = CONFIG.output.upcase == "STDOUT" ? STDOUT : File.open(CONFIG.output, mode: "a")
 LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
 
-# Run migrations
-if Invidious::Database::Migrator.new(PG_DB).pending_migrations?
-  puts "There are pending migrations. Run `invidious --migrate` to apply the migrations."
-  exit 46
-end
```
SamantazFox: "This is so that we don't break deploys with this PR."

Well, when we'll make any DB changes (like #2469), the code won't correspond to the DB anymore, so invidious will still be broken until the migration is done.

SamantazFox: As an in-between, we could keep the check + message, but not exit (so it's easier to trace why invidious is broken).

matthewmcgarvey: I made this change because it seems like the rollout of requiring migrations to have run needs more thought which would could give more attention in a followup PR. By removing that, we can merge in the migration code without worrying about breaking users

matthewmcgarvey:
@SamantazFox Is there a blocker for this anymore now that it won't run migrations unless the CLI command is used? If this is merged we could talk further about how we want migrations to be done by end users

SamantazFox:
I don't see any. I was waiting on @unixfox approval, as per his previous message.

unixfox:
I'm not sure if I'll have time to try, you may merge anyway as it is not automatic yet.

TheFrenchGhosty:
@SamantazFox Does this mean we can finally merge PRs that implement database changes? (after they're changed to work, obviously)

unixfox:
> @SamantazFox Does this mean we can finally merge PRs that implement database changes? (after they're changed to work, obviously)

No because it's not automatic yet. Merging them will brake invidious.

src/invidious.cr
broquemonsieur: I am struggling to run these CLI arguments, when I type `sudo docker exec -it invidious-invidious-1 invidious -v` I get

> OCI runtime exec failed: exec failed: unable to start container process: exec: "invidious": executable file not found in $PATH: unknown

Issue #2875, opened 2022-02-08
[Bug] User unable to change its own password


Malakai-DB:
Hello,

Just FYI, before opening the issue, I did :
  * Read the FAQ!
  * Use the search function to check if there is already an issue open for your problem!

**Describe the bug**
I created a user on my own invidious instance and tried to change it's password from the webui without success.

**Steps to Reproduce**
1. Login with your user
2. Go to : http(s)://your.instance/change_password
3. Input your current password once and your new password twice
4. After hitting "Change password" you are sent back to homepage without effect (password is still the old one)

**Logs**
There is no error on the application logs, nor on the postgres database logs. Nothing special in the browser console either.
In the application logs we can see the following : 
```
2022-02-08 15:25:19 UTC [info] 200 GET /preferences?referer=%2Ffeed%2Fsubscriptions 1.83ms
2022-02-08 15:25:20 UTC [info] 200 GET /api/v1/auth/subscriptions?fields=authorId 1.68ms
2022-02-08 15:25:24 UTC [info] 200 GET /change_password?referer=%2Ffeed%2Fsubscriptions 879.17µs
2022-02-08 15:25:24 UTC [info] 200 GET /api/v1/auth/subscriptions?fields=authorId 3.47ms
2022-02-08 15:25:28 UTC [info] 200 POST /api/v1/auth/notifications?fields=videoId,title,author,authorId,publishedText,published,authorThumbnails,liveNow 178034.63ms
```
I activated debug logs, but I can still not see any error / warning that could explain why the password isn't changed.

**Screenshots**
Not applicable.

**Additional context**
Invidious is installed with docker compose (and postgres database also) behind an apache2 reverse proxy.
After setting up the database, I imported the sql scripts found [here](https://github.com/iv-org/invidious/tree/master/config/sql) to create the necessary tables and set the rights.
Everything else is working as expected (subscriptions, watching videos, creating playlists, etc).

I am on version 2022.02.02-0ca3337@master using the docker image from quay.io/invidious/invidious

unixfox:
Can you post your config.yml without the personal/confidential information?

SamantazFox:
I'd say it's probably a side effect of #2492

Malakai-DB:
> Can you post your config.yml without the personal/confidential information?

```
channel_threads: 1
feed_threads: 1
output: STDOUT
log_level: info
db:
  user: dbuser
  password: dbpassword
  host: postgres
  port: 5432
  dbname: mydb
[...]

PR #2890, opened 2022-02-10
DB: fix inverted arguments in User.update_password()

SamantazFox:
Closes https://github.com/iv-org/invidious/issues/2875


SamantazFox:
Oh, and it's been deployed on https://test.invidious.io
It seems to work fine.

PR #2895, opened 2022-02-12
Persist player volume/speed

### LINKED ISSUE #846: Remember current volume setting
Hello guys.

I noticed that when switching to another video, the volume is reset to whatever the account setting is. Maybe we could store the value somewhere else so that it persists at least for the current session?

Greetings from Germany
NH

MathiusD:
Closes #846 

The purpose of this PR is to propose the addition of the [videojs-persist](https://github.com/mister-ben/videojs-persist) module.

This one proposes the following things:
Stores volume, muted and playback rate to local storage to be restored in the next session.

It seems to me that it can be interesting for the users (in particular all that is associated with the management of volume) therefore I propose its addition via this PR

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -176,16 +176,54 @@ if (video_data.params.video_start > 0 || video_data.params.video_end > 0) {
     player.currentTime(video_data.params.video_start);
 }
 
-/*
-    If the video settings are default, we enable the management of the settings by
-    the videojs-persist module otherwise we apply the preferences.
-*/
-if (video_data.params.volume == 100 && video_data.params.speed == "1.0")
-    player.persist();
-else {
-    player.volume(video_data.params.volume / 100);
-    player.playbackRate(video_data.params.speed);
+player.volume(video_data.params.volume / 100);
+player.playbackRate(video_data.params.speed);
+
+/**
+ * Method for get content of Cookie
+ * @param {String} name Name of cookie
+ * @returns cookieValue
+ */
+function getCookieValue(name) {
+    var value = document.cookie.split(";").filter(item => {
+        return item.includes(name + "=");
+    });
+    return value != null && value.length >= 1 ? value[0].substring((name + "=").length, value[0].length) : null;
 }
+
+/**
+ * Method for update Prefs cookie (Or create if missing)
+ * @param {number} newVolume New Volume defined (Null if unchanged)
+ * @param {number} newSpeed New Speed defined (Null if unchanged)
+ */
+function updateCookie(newVolume, newSpeed) {
+    var volumeValue = newVolume != null ? newVolume : video_data.params.volume;
+    var speedValue = newSpeed != null ? newSpeed : video_data.params.speed;
+    var cookieValue = getCookieValue('PREFS');
+    if (cookieValue != null) {
+        var cookieJson = JSON.parse(decodeURIComponent(cookieValue));
+        cookieJson.volume = volumeValue;
+        cookieJson.speed = speedValue;
+        document.cookie = document.cookie.replace(getCookieValue('PREFS'), encodeURIComponent(JSON.stringify(cookieJson)));
+    } else {
+        var date = new Date();
+        //Set expiration in 2 year
+        date.setTime(date.getTime() + 63115200);
+        document.cookie = 'PREFS=' +
+            encodeURIComponent(JSON.stringify({ 'volume': volumeValue, 'speed': speedValue })) +
+            '; expires=' + date.toGMTString() + '; SameSite=Strict; path=/';
+    }
+    video_data.params.volume = volumeValue;
+    video_data.params.speed = speedValue;
+}
+
+player.on('ratechange', function () {
+    updateCookie(null, player.playbackRate());
+});
+
+player.on('volumechange', function () {
+    updateCookie(Math.ceil(player.volume() * 100), null);
+});
 
 player.on('waiting', function () {
     if (player.playbackRate() > 1 && player.liveTracker.isLive() && player.liveTracker.atLiveEdge()) {
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -210,10 +210,9 @@ function updateCookie(newVolume, newSpeed) {
     var date = new Date();
     // Set expiration in 2 year
     date.setTime(date.getTime() + 63115200);
-    var ipv4Regex = /(([0-1]?[0-9]{1,2}\.)|(2[0-4][0-9]\.)|(25[0-5]\.)){3}(([0-1]?[0-9]{1,2})|(2[0-4][0-9])|(25[0-5]))/;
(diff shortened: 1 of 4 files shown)


unixfox:
Mmh isn't this going to conflict with the already existing volume in the preferences?

SamantazFox:
Why not make a small JS snippet that changes the `PREFS` cookie's value?
That way, it won't collide with the existing preferences, and it works on devices without local storage.

MathiusD:
This is a good compromise indeed, to manage devices without localStorage.

I'll modify that and come back to annoy you

MathiusD:
To justify the modification of the `http_only` parameter within the modification of the preference cookie on the server side, this is necessary to allow its visibility and its edition on the client side (otherwise it remains hidden). I allowed myself this modification on the preference cookie because I believe that the leakage of preferences via an XSS flaw (as mentioned [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#security)) is relatively minimal if it occurs. Otherwise I simply could not proceed via the cookies and came back to the exploitation of the [videojs-persist](https://github.com/mister-ben/videojs-persist) module or to a rewriting of the module.

MathiusD:
Thinking about it, would it be interesting to allow or not allow propagation of player settings to preferences by adding a preference like `playerSettingsPropagation` or equivalent ?

(Yes, I'm asking because I think it might be useful, but I don't know if you think it is, so I'm asking before adding an option that might not be relevant)

SamantazFox:
@MathiusD I don't think that adding such an option is required. Propagation of player settings is a quite common feature, and [something expected by the users](https://github.com/iv-org/invidious/issues/846).

MathiusD:
@SamantazFox It is true, however, that I thought it would be potentially interesting to allow its non-propagation for users who would regularly find themselves in a situation such as having to adjust the sound only for certain videos whose sound is slightly different (and who would therefore not want to keep the ephemeral changes, especially if modification is associated with the player speed)

SamantazFox:
@MathiusD yeah. Imo, we should keep it simple for now, and add a config option later if that's really required (a.k.a requested by the users).

src/invidious/routes/preferences.cr
```diff
@@ -222,10 +222,10 @@ module Invidious::Routes::PreferencesRoute
 
       if CONFIG.domain
         env.response.cookies["PREFS"] = HTTP::Cookie.new(name: "PREFS", domain: "#{CONFIG.domain}", value: URI.encode_www_form(preferences.to_json), expires: Time.utc + 2.years,
-          secure: secure, http_only: true)
+          secure: secure, http_only: false)
```
SamantazFox, line 225: Could you please rebase against master, as I merged a PR that changed that code?

SamantazFox (approved):
Code looks good to me.

src/invidious/routes/preferences.cr
MathiusD: Rebased

SamantazFox:
I'll try to deploy that ASAP on the test instance

assets/js/player.js
```diff
@@ -179,6 +179,52 @@ if (video_data.params.video_start > 0 || video_data.params.video_end > 0) {
...
 
+/**
+ * Method for get content of Cookie
+ * @param {String} name Name of cookie
+ * @returns cookieValue
+ */
+function getCookieValue(name) {
+    var value = document.cookie.split(";").filter(item => {
+        return item.includes(name + "=");
+    });
```
mhmdanas, line 190:
```suggestion
    var value = document.cookie.split(";").filter(item => item.includes(name + "="));
```

assets/js/player.js
```diff
@@ -179,6 +179,52 @@ if (video_data.params.video_start > 0 || video_data.params.video_end > 0) {
...
+    var speedValue = newSpeed != null ? newSpeed : video_data.params.speed;
+    var cookieValue = getCookieValue('PREFS');
+    if (cookieValue != null) {
+        var cookieJson = JSON.parse(decodeURIComponent(cookieValue));
+        cookieJson.volume = volumeValue;
+        cookieJson.speed = speedValue;
+        document.cookie = document.cookie.replace(getCookieValue('PREFS'), encodeURIComponent(JSON.stringify(cookieJson)));
+    } else {
+        var date = new Date();
+        //Set expiration in 2 year
```
mhmdanas, line 210: Minor style nit:
```suggestion
        // Set expiration in 2 year
```

assets/js/player.js
MathiusD: Resolved with 1e3f4ed3983f3b101a259f8164511665f25f878e

SamantazFox:
So I've deployed this on test.invidious.io, and there is a small problem: it creates a second `PREFS` cookie, which causes weird issues with the preferences page:

image
image
(Nope, I don't read/speak arabic ^^)

MathiusD:
@SamantazFox Arf... Yes, this is due to the change of the http_only parameter which makes a difference between the 2 cookies.
However, it acts strangely in the sense that the new preference cookie does not contain any instructions on the language and this is interpreted as a preference on Arabic.

I'm going to test again locally to find a patch to this problem, I'll come back to annoy you when I have something that solves the problem

SamantazFox:
@MathiusD note that the missing `locale` entry doesn't select arabic per se, but instead causes the drop down to not have any `selected` elements, causing it to show the first item.

As for the `http_only` parameter, I just thought that I should have saved the preferences beforehand. Maybe that would have helped.

MathiusD:
@SamantazFox
> note that the missing `locale` entry doesn't select arabic per se, but instead causes the drop down to not have any `selected` elements, causing it to show the first item.

Thank you very much for the answer, indeed it gives me a track of resolution in case of creation of cookie via the script (Namely specify default values that would be injected as the `video_data` or by an equivalent means)

> As for the `http_only` parameter, I just thought that I should have saved the preferences beforehand. Maybe that would have helped.

Yes, however, in case of an instance update, it seems to me quite hypothetical that all users will think to re-save their cookies under the pretext that my modification breaks the preferences.

> Edit: nope, there is still an issue:

In this case it is even more important if it does not solve the problem

MathiusD:
@SamantazFox

Well, I still have to work on it because I don't have a repro of the bug, but I have a question though. Is it voluntary to set the cookie on .domain.example and not on domain.example?

I ask because in fact locally if I try to add the cookie for player opts while it is positioned exactly on the same domain [In my case localhost], the cookie is just unalterable until the previous cookie disapear. (By redefining it via the preferences for example)

MathiusD:
Normally, the error found on test.invidious.io is patched

SamantazFox:
> Is it voluntary to set the cookie on .domain.example and not on domain.example?

Yep, as you figured out, this is not something that we explicitely do.
In invidious, we only set the cookie's domain to the value of the `domain` config.

> except in case of ip and localhost.

Note that in the localhost and IP cases, the `domain` in config.yml MUST be empty. This makes the config variable `domain` equal to `nil` in Crystal, and once passed to the cookie function, this prevents the cookie from being associated to a domain name.

Also, rather than those lenghty checks and creating a new cookie, you can copy the existing cookie value, and only replace the required bits.

MathiusD:
> Yep, as you figured out, this is not something that we explicitely do.
> In invidious, we only set the cookie's domain to the value of the `domain` config.

Yes that's what I saw I was just surprised that it was targeting not `domain.example` but `.domain.example`

> Note that in the localhost and IP cases, the `domain` in config.yml MUST be empty. This makes the config variable `domain` equal to `nil` in Crystal, and once passed to the cookie function, this prevents the cookie from being associated to a domain name.
>
> Also, rather than those lenghty checks and creating a new cookie, you can copy the existing cookie value, and only replace the required bits.

Here I create a new cookie in case the user simply doesn't have a preference (or if it's not accessible but then the browser doesn't allow replacing it)

In the case of an update, due to the targeted domain, I'm forced to re-specify at least the domain to avoid the case that posed problems on `test.invidious.io` with 2 cookies (By default define/update a cookie via the target js `domain.example` and not `.domain.example` [In this case, even if we are in an update process, it creates a new cookie] [I had to manage the case of a local host or a precise ip from the js, that said this information could just as well be transmitted as video_data, rather than determined via regexes, it would even lighten the cost of the script execution])

SamantazFox:
For the domain part, it seems that not setting the domain in `src/invidious/user/cookies.cr` solves it. There is probably a bug in the crystal standard library.

MathiusD:
> Edit2: that seem to be a problem with FF. The cookie is set as expected by the server:

In this case, should we do a workaround for FF (Like what was done for Opera) ?

SamantazFox:
Oh! This is from the [RFC 2109](https://www.ietf.org/rfc/rfc2109.txt):

>    Domain=domain
>      Optional.  The Domain attribute specifies the domain for which the
>      cookie is valid.  An explicitly specified domain must always start
>      with a dot.

MathiusD:
Ah, well, I think we have the solution

SamantazFox:
Just lemme check if that RFC is not obselete...

SamantazFox:
What you can do is reduce the regex to a simple expression like:
```javascript
/^((\d+\.){3}\d+|[A-Fa-f0-9]*:[A-Fa-f0-9:]*:[A-Fa-f0-9:]+)$/
```

It's far from perfect, but it will catch all IP addresses.

SamantazFox:
@MathiusD Thanks for your contribution :D

PR #2906, opened 2022-02-16
Add possibility to don't clip url in video description

### LINKED ISSUE #2218: Don't clip URLs in video descriptions.
image from `/watch?v=ZFUU59BAILc`
image
Can't copy a whole section, the links are clipped, ending with '...'

Show all links in full. They may be clipped in HTML so they don't overflow (yet can still be copied).

MathiusD:
Closes #2218 

The purpose of this PR is to allow or not the clip of urls (as requested in [this issue](https://github.com/iv-org/invidious/issues/2218)).

As mentioned by @syeopite [here](https://github.com/iv-org/invidious/issues/2218#issuecomment-937724866), the goal was to make this modification configurable. Here it is configurable via the user preferences.

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -63,7 +63,7 @@
     "preferences_autoplay_label": "Autoplay: ",
     "preferences_continue_label": "Play next by default: ",
     "preferences_continue_autoplay_label": "Autoplay next video: ",
-    "preferences_clip_urls_label":"Clip url displayed: ",
+    "preferences_clip_urls_label": "Clip url displayed: ",
     "preferences_listen_label": "Listen by default: ",
     "preferences_local_label": "Proxy videos: ",
     "preferences_speed_label": "Default speed: ",
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -63,6 +63,7 @@
     "preferences_autoplay_label": "Autoplay: ",
     "preferences_continue_label": "Play next by default: ",
     "preferences_continue_autoplay_label": "Autoplay next video: ",
+    "preferences_clip_urls_label":"Clip url displayed: ",
     "preferences_listen_label": "Listen by default: ",
     "preferences_local_label": "Proxy videos: ",
     "preferences_speed_label": "Default speed: ",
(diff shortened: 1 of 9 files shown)


unixfox:
IIRC the translation should be handled in weblate

MathiusD:
> IIRC the translation should be handled in weblate

Indeed, I had put 2 translation keys because it seemed more appropriate, I can remove them if necessary

SamantazFox:
I agree with matthew here.

There are already so many config options, and adding more for such simple details is simply going to be confusing for the average user. Imo, the description should stay "clean" (legible) for the majority of users, which is best achieved by keeping the URLs short.

I don't think copying entire descriptions is something one does often, and if that's needed, Imho a userscript is better suited at this purpose (i.e a JS snippet that would add a button to copy the description and some logic to clean out the links).

As an in-between solution, we could try to increase the size of the displayed text, while still clipping URLs that are too long.

MathiusD:
In this sense, this pr is no longer necessary (unless you choose to lengthen the text of the links).
And it should be indicated in the original output that the user should turn to a user script

locales/en-US.json
```diff
@@ -63,6 +63,7 @@
     "preferences_autoplay_label": "Autoplay: ",
     "preferences_continue_label": "Play next by default: ",
     "preferences_continue_autoplay_label": "Autoplay next video: ",
+    "preferences_clip_urls_label":"Clip url displayed: ",
```
mhmdanas, line 66:
```suggestion
    "preferences_clip_urls_label": "Clip url displayed: ",
```

locales/fr.json
```diff
@@ -402,6 +402,7 @@
     "adminprefs_modified_source_code_url_label": "URL du dépôt du code source modifié",
     "footer_documentation": "Documentation",
     "footer_original_source_code": "Code source original",
+    "preferences_clip_urls_label":"Tronquer les liens affichés : ",
```
mhmdanas, line 405:
```suggestion
    "preferences_clip_urls_label": "Tronquer les liens affichés : ",
```

MathiusD: Resolved

locales/en-US.json
MathiusD: Resolved

PR #2910, opened 2022-02-18
Expose version by default on api/v1/stats

### LINKED ISSUE #1690: Expose version by default on api/v1/stats

MathiusD:
Close #1690

The purpose of this PR is to display version information within the `/api/v1/stats` route even when statistics are disabled.

(Note : I don't want to be awarded the bounty associated to the issue this PR is fixing.)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,7 +34,6 @@ services:
           dbname: invidious
         full_refresh: false
         https_only: false
-        statistics_enabled: false
         domain:
     healthcheck:
       test: wget -nv --tries=1 --spider http://127.0.0.1:3000/api/v1/comments/jNQXAC9IVRw || exit 1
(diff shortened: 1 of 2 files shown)


unixfox (approved):

docker-compose.yml
```diff
@@ -34,6 +34,7 @@ services:
           dbname: invidious
         full_refresh: false
         https_only: false
+        statistics_enabled: false
```
SamantazFox, line 37: This is not required, as the default value is `false` :)

SamantazFox (approved):

docker-compose.yml
MathiusD: I had put it because I had tested by altering the parameter, but indeed there is no interest to leave it at its default value within the docker-compose

SamantazFox:
@MathiusD Thanks for your contribution to invidious :D

PR #2912, opened 2022-02-19
Add currentTimestamp in shared url by videojs-share

### LINKED ISSUE #586: Feature request: Share at current timestamp link
From https://github.com/omarroth/invidious/issues/564#issuecomment-499152471

MathiusD:
Close #586

The purpose of this PR is to add the current timestamp within the shared link via the [videojs-share](https://www.npmjs.com/package/videojs-share) interface

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -51,7 +51,7 @@ var player = videojs('player', options);
 
 /**
  * Function for add time argument to url
- * @param {String} url 
+ * @param {String} url
  * @returns urlWithTimeArg
  */
 function addCurrentTimeToURL(url) {
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -60,9 +60,8 @@ function addCurrentTimeToURL(url) {
     var currentTime = Math.ceil(player.currentTime());
     if (currentTime > 0)
         urlUsed.searchParams.set('t', currentTime);
-    else
-        if (urlUsed.searchParams.has('t'))
-            urlUsed.searchParams.delete('t');
+    else if (urlUsed.searchParams.has('t'))
+        urlUsed.searchParams.delete('t');
     return urlUsed;
 }
 
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -35,20 +35,10 @@ if (player_data.aspect_ratio) {
 
 var embed_url = new URL(location);
 embed_url.searchParams.delete('v');
-short_url = location.origin + '/' + video_data.id + embed_url.search;
+var short_url = location.origin + '/' + video_data.id + embed_url.search;
 embed_url = location.origin + '/embed/' + video_data.id + embed_url.search;
 
 var save_player_pos_key = "save_player_pos";
-
-var shareOptions = {
-    socials: ['fbFeed', 'tw', 'reddit', 'email'],
-
-    url: short_url,
-    title: player_data.title,
-    description: player_data.description,
-    image: player_data.thumbnail,
-    embedCode: "<iframe id='ivplayer' width='640' height='360' src='" + embed_url + "' style='border:none;'></iframe>"
-}
 
 videojs.Vhs.xhr.beforeRequest = function(options) {
     if (options.uri.indexOf('videoplayback') === -1 && options.uri.indexOf('local=true') === -1) {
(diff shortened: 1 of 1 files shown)


assets/js/player.js
```diff
@@ -59,6 +49,38 @@ videojs.Vhs.xhr.beforeRequest = function(options) {
...
 var player = videojs('player', options);
 
+/**
+ * Function for add time argument to url
+ * @param {String} url 
+ * @returns urlWithTimeArg
+ */
+function addCurrentTimeToURL(url) {
+    var urlUsed = new URL(url);
+    urlUsed.searchParams.delete('start');
```
SamantazFox, line 59: Can those be inlined like so?
`var urlUsed = new URL(url).searchParams.delete('start');`

assets/js/player.js
```diff
@@ -59,6 +49,38 @@ videojs.Vhs.xhr.beforeRequest = function(options) {
...
+    var currentTime = Math.ceil(player.currentTime());
+    if (currentTime > 0)
+        urlUsed.searchParams.set('t', currentTime);
+    else
+        if (urlUsed.searchParams.has('t'))
```
SamantazFox, line 64: Is this check necessary? You don't seel to check if the `start` parameter exists just above.

SamantazFox (commented):
Looks good to me

assets/js/player.js
```diff
@@ -59,6 +49,38 @@ videojs.Vhs.xhr.beforeRequest = function(options) {
...
+
+    get url() {
+        return addCurrentTimeToURL(short_url);
+    },
+    title: player_data.title,
+    description: player_data.description,
+    image: player_data.thumbnail,
+    get embedCode() {
+        return "<iframe id='ivplayer' width='640' height='360' src='" +
+            addCurrentTimeToURL(embed_url) + "' style='border:none;'></iframe>";
```
SamantazFox, line 80: There are extra spaces at the end of the line, btw!

assets/js/player.js
MathiusD: Yes, it wouldn't change much in the end

assets/js/player.js
MathiusD: Yes, here it is necessary because if for example we have a `t` argument defined, it will always be present while the player is at the beginning of the video. In this case it must be removed (in addition to the start argument which is systematically removed to avoid having 2 arguments specifying the starting point [Example : `start=1m2&t=30`]) otherwise we may find ourselves in a situation where we give t=30 when we want to share from the beginning.

assets/js/player.js
MathiusD: I'm cleaning this up!

assets/js/player.js
MathiusD: If indeed, I can not reduce it otherwise the variable `urlUsed` receives the result `void` of the method `.delete('start')`

assets/js/player.js
SamantazFox: Could you just inline the `else if` then? (just a matter of style, and trying to reduce the amoun of useless indentations, as the JS is not minified).

MathiusD: Of course! (By the way, since it is mentioned, is there any plan to minify the JS ? [Because I confess I haven't seen any issue in this sense and that it has an interest in reducing the weight of the page])

assets/js/player.js
```diff
@@ -59,6 +49,37 @@ videojs.Vhs.xhr.beforeRequest = function(options) {
 
 var player = videojs('player', options);
 
+/**
+ * Function for add time argument to url
+ * @param {String} url
```
SamantazFox, line 54: there is also an extra space in here!

MathiusD: Crap, I'll take it right off

assets/js/player.js
SamantazFox: We won't minify the JS in order to keep it legible by an end user. The only JS that can be minified are the VideoJS dependencies (thought it's not the case by default).

While invidious is proxying video data (= a lot of bytes), I'm in favor of trying to minimize the amounf of data sent to the user at other places, in order to reduce the stress on large instances (if we can save a kilobyte on HTML and JS, that can be a lot when an instance gets a million requests a day).

SamantazFox:

assets/js/player.js
MathiusD: Okay, I didn't know this was voluntary. In this case I understand even more the desire to reduce the weight of assets (Scripts, etc)

PR #2915, opened 2022-02-21
Allow configuration of the time between 2 RefreshChannelsJob

### LINKED ISSUE #2060: [Enhancement] Allow changing RefreshChannelsJob frequency

MathiusD:
Close #2060

The purpose of this PR is to allow the configuration of the time between 2 RefreshChannelsJob.

This PR proposes this configuration by adding the `channel_refresh_time` parameter in the configuration file (Cf usage in `config/config.example.yml`)

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -315,7 +315,7 @@ https_only: false
 channel_threads: 1
 
 ##
-## Time between channel_refresh
+## Time between two jobs for crawling videos from channels
 ##
 ## Accepted values: a valid time interval (hours:min:seconds)
 ## Default: 00:30:00
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -315,6 +315,14 @@ https_only: false
 channel_threads: 1
 
 ##
+## Time between channel_refresh
+##
+## Accepted values: a valid time interval (hours:min:seconds)
+## Default: 00:30:00
+##
+channel_refresh_time: 00:30:00
+
+##
 ## Forcefully dump and re-download the entire list of uploaded
 ## videos when crawling channel (during subscriptions update).
 ##
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -315,12 +315,13 @@ https_only: false
 channel_threads: 1
 
 ##
-## Time between two jobs for crawling videos from channels
+## Time interval between two executions of the job that crawls
+## channel videos (subscriptions update).
 ##
-## Accepted values: a valid time interval (like 1h30m or 90min)
+## Accepted values: a valid time interval (like 1h30m or 90m)
 ## Default: 30m
 ##
-channel_refresh_interval: 30min
+#channel_refresh_interval: 30m
 
 ##
 ## Forcefully dump and re-download the entire list of uploaded
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -317,10 +317,10 @@ channel_threads: 1
 ##
 ## Time between two jobs for crawling videos from channels
 ##
-## Accepted values: a valid time interval (hours:min:seconds)
-## Default: 00:30:00
+## Accepted values: a valid time interval (like 1h30m or 90min)
+## Default: 30m
 ##
-channel_refresh_interval: 00:30:00
+channel_refresh_interval: 30min
 
 ##
 ## Forcefully dump and re-download the entire list of uploaded
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -320,7 +320,7 @@ channel_threads: 1
 ## Accepted values: a valid time interval (hours:min:seconds)
 ## Default: 00:30:00
 ##
-channel_refresh_time: 00:30:00
+channel_refresh_interval: 00:30:00
 
 ##
 ## Forcefully dump and re-download the entire list of uploaded
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       INVIDIOUS_CONFIG: |
         log_level: Info
         channel_threads: 1
-        channel_refresh_interval: 00:30:00
+        channel_refresh_interval: 30m
         check_tables: true
         feed_threads: 1
         db:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,9 @@ services:
     environment:
       # Adapted from ./config/config.yml
       INVIDIOUS_CONFIG: |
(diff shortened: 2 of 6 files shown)


src/invidious/helpers/utils.cr
```diff
@@ -18,23 +18,39 @@ def elapsed_text(elapsed)
   "#{(millis * 1000).round(2)}µs"
 end
 
-def decode_length_seconds(string)
-  length_seconds = string.gsub(/[^0-9:]/, "")
-  return 0_i32 if length_seconds.empty?
+module TimeSpanConverter
```
SamantazFox, line 21: Please put that module in `config.cr`, with the other converters.

src/invidious/config.cr
```diff
@@ -56,11 +56,13 @@ end
...
   include YAML::Serializable
 
-  property channel_threads : Int32 = 1           # Number of threads to use for crawling videos from channels (for updating subscriptions)
-  property feed_threads : Int32 = 1              # Number of threads to use for updating feeds
-  property output : String = "STDOUT"            # Log file path or STDOUT
-  property log_level : LogLevel = LogLevel::Info # Default log level, valid YAML values are ints and strings, see src/invidious/helpers/logger.cr
-  property db : DBConfig? = nil                  # Database configuration with separate parameters (username, hostname, etc)
+  property channel_threads : Int32 = 1 # Number of threads to use for crawling videos from channels (for updating subscriptions)
+  @[YAML::Field(converter: TimeSpanConverter)]
+  property channel_refresh_time : Time::Span = 30.minutes # Time between two jobs for crawling videos from channels
```
SamantazFox, line 61: While you're at it, could you please put the comments above their respective variable, rather than inline? (note that it has to be before the `@[]` converter statement).

SamantazFox (changes requested):

src/invidious/helpers/utils.cr
MathiusD: I'm moving this right now

src/invidious/config.cr
MathiusD: No problem, I was following the way it was done at that place

config/config.example.yml
```diff
@@ -314,6 +314,14 @@ https_only: false
 ##
 channel_threads: 1
 
+##
+## Time between two jobs for crawling videos from channels
+##
+## Accepted values: a valid time interval (hours:min:seconds)
+## Default: 00:30:00
+##
+channel_refresh_time: 00:30:00
```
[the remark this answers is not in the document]
MathiusD: > imo, using a similar system to the timestamp in the URL (`t` parameter) would be preferrable (so `1h30m` would be converted to `90m` or 5400 seconds).

Yes, that's basically what I was going to do. But I stopped at this encoding which seemed to me correct (Even if less intuitive than a `XhYm`)

> Also, the default unit should be minutes. Refreshing channels is quite long, so anything under a minute is basically the same thing as running the refresh job continuously.

Admittedly, I had not taken this point into consideration

> And finally, could you please rename this variable to `channel_refresh_interval`, as it's more accurate (less ambiguous about the variable's purpose)?

Yes, thanks for the variable name, it is more accurate

src/invidious/helpers/utils.cr
MathiusD: By the way the other conversion modules are in preferences, I have to put it in `user/preferences.cr` in this case even if it is not used for the config or I put it in `config.cr`, where there is currently no conversion module ?

SamantazFox: Oh, right, I haven't move those yet! Yeah, put them in `config.cr`.

MathiusD: Ok !

config/config.example.yml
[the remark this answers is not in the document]
MathiusD: > imo, using a similar system to the timestamp in the URL (t parameter) would be preferrable (so 1h30m would be converted to 90m or 5400 seconds).
> 
> Also, the default unit should be minutes. Refreshing channels is quite long, so anything under a minute is basically the same thing as running the refresh job continuously.

It's done, so `1h2m` -> `62m`, `90m` -> `90m`, `12` -> `12m`

docker-compose.yml
```diff
@@ -23,7 +23,9 @@ services:
     environment:
       # Adapted from ./config/config.yml
       INVIDIOUS_CONFIG: |
```
TheFrenchGhosty, line 25: Please don't change those, this will conflict with https://github.com/iv-org/invidious/pull/2917

MathiusD: Okay, I'll clean it up right away

config/config.example.yml
```diff
@@ -314,6 +314,14 @@ https_only: false
 ##
 channel_threads: 1
 
+##
+## Time between two jobs for crawling videos from channels
```
SamantazFox, line 318:
```suggestion
## Time interval between two executions of the job that crawls
## channel videos (subscriptions update).
```

config/config.example.yml
```diff
@@ -314,6 +314,14 @@ https_only: false
...
+##
+## Accepted values: a valid time interval (like 1h30m or 90min)
+## Default: 30m
+##
+channel_refresh_interval: 30min
```
SamantazFox, line 323:
```suggestion
#channel_refresh_interval: 30m
```

SamantazFox: The variable should be commented out because the variable is optional (not as important as `db`, e.g). Also, changing `min` for `m` to stay consistent with the default value in the comment above.

SamantazFox (approved):
Code is good. Some minor changes needed in the exampke config file.

PR #2917, opened 2022-02-22
Enhance the development compose file

TheFrenchGhosty:
Follow up to https://github.com/iv-org/documentation/pull/160#issuecomment-1039778281

Related to https://github.com/iv-org/documentation/pull/200 (wait for both to be ready - so that we can merge them at the same time)

Massively enhance the development compose:

- Make the production and development compose use the same (sane and clean) bases
- Change the restart policy for `unless-stopped` since `always` is bad practice
- Remove the network (that's completely useless)
- Remove the resources limiters (that's not really useful, and users should only use it if they need to)
- Remove the `willfarrell/autoheal` container: we can't vouch for it's usefulness, and its security (it has access to the docker socket, which is a massive security risk)
- Bump the postgres version to 14 since it's the latest and 10 [EOL in 8 months](https://www.postgresql.org/support/versioning/) ~~and move to an Alpine-based Postgres image~~
- Layout change (service then DB)
- ~~Enforce a container_name~~ **Edit:** https://github.com/iv-org/documentation/pull/200#discussion_r811676660

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,7 @@ services:
       - invidious-db
 
   invidious-db:
-    image: docker.io/library/postgres:14-alpine
+    image: docker.io/library/postgres:14
     restart: unless-stopped
     volumes:
       - postgresdata:/var/lib/postgresql/data
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,7 +34,7 @@ services:
       - invidious-postgres
 
   invidious-postgres:
-    image: postgres:14
+    image: postgres:14-alpine
     restart: unless-stopped
     volumes:
       - postgresdata:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,18 +1,8 @@
-version: '3'
+# This docker-compose file is made for development purpose and build an image from source, if you want to use Invidious in production, see the documentation.
```
SamantazFox, line 1:
```suggestion
# Warning: This docker-compose file is made for development purposes.
# Using it will build an image from the locally cloned repository.
#
# If you want to use Invidious in production, see the [docker-compose.yml file provided
# in the installation documentation: https://docs.invidious.io/Installation.md
```

SamantazFox (approved):
Looks good to me. That first comment could be improved, though.

SamantazFox:
@unixfox what do you think about this?

TheFrenchGhosty:
@SamantazFox the discussion is still going on in the pull for the production compose, we'll merge both PR at the same time.

PR #2922, opened 2022-02-22
Download widget fix

### LINKED ISSUE #2670: [Bug] Download widget replaces spaces in filename with `+`

SamantazFox:
That should close #2670 for good.

### DIFF
--- a/src/invidious/frontend/watch_page.cr
+++ b/src/invidious/frontend/watch_page.cr
@@ -26,11 +26,15 @@ module Invidious::Frontend::WatchPage
     return String.build(4000) do |str|
       str << "<form"
       str << " class=\"pure-form pure-form-stacked\""
-      str << " action='/latest_version'"
-      str << " method='get'"
+      str << " action='/download'"
+      str << " method='post'"
       str << " rel='noopener'"
       str << " target='_blank'>"
       str << '\n'
+
+      # Hidden inputs for video id and title
+      str << "<input type='hidden' name='id' value='" << video.id << "'/>\n"
+      str << "<input type='hidden' name='title' value='" << HTML.escape(video.title) << "'/>\n"
 
       str << "\t<div class=\"pure-control-group\">\n"
 
@@ -48,8 +52,7 @@ module Invidious::Frontend::WatchPage
 
         height = itag_to_metadata?(option["itag"]).try &.["height"]?
 
-        title = URI.encode_www_form("#{video.title}-#{video.id}.#{mimetype.split("/")[1]}")
-        value = {"id": video.id, "itag": option["itag"], "title": title}.to_json
+        value = {"itag": option["itag"], "ext": mimetype.split("/")[1]}.to_json
 
         str << "\t\t\t<option value='" << value << "'>"
         str << (height || "~240") << "p - " << mimetype
@@ -61,8 +64,7 @@ module Invidious::Frontend::WatchPage
       video_assets.video_streams.each do |option|
         mimetype = option["mimeType"].as_s.split(";")[0]
 
-        title = URI.encode_www_form("#{video.title}-#{video.id}.#{mimetype.split("/")[1]}")
-        value = {"id": video.id, "itag": option["itag"], "title": title}.to_json
+        value = {"itag": option["itag"], "ext": mimetype.split("/")[1]}.to_json
 
         str << "\t\t\t<option value='" << value << "'>"
         str << option["qualityLabel"] << " - " << mimetype << " @ " << option["fps"] << "fps - video only"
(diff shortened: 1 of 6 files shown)


SamantazFox:
image
Works as expected

SamantazFox:
@matthewmcgarvey could you please review this?

src/invidious/routes/video_playback.cr
```diff
@@ -278,8 +272,10 @@ module Invidious::Routes::VideoPlayback
       haltf env, status_code: 404
     end
 
-    url = URI.parse(url).request_target.not_nil! if local
-    url = "#{url}&title=#{title}" if title
+    if local
+      url = URI.parse(url).request_target.not_nil!
+      url += "&title=#{URI.encode_www_form(title, space_to_plus: false)}" if title
```
matthewmcgarvey, line 277: Does encoding it twice mess anything up? It's also encoded here https://github.com/iv-org/invidious/pull/2922/files#diff-a86ced8aa99e3403588538decc008851d3d33dfadbb00d392d5ec8b4696f7df4R309-R312

src/invidious/routes/video_playback.cr
```diff
@@ -242,31 +242,25 @@ module Invidious::Routes::VideoPlayback
...
-      title = URI.decode_www_form(download_widget["title"].as_s)
-
-      if label = download_widget["label"]?
-        return env.redirect "/api/v1/captions/#{id}?label=#{label}&title=#{title}"
-      else
-        itag = download_widget["itag"].as_s.to_i
-        local = "true"
-      end
+    # Sanity checks
+    if id.nil? || id.size != 11 || !id.matches?(/^[\w-]+$/)
```
matthewmcgarvey, line 249:
```suggestion
    if id.nil? || !id.matches?(/^[\w-]{11}$/)
```
I'm not sure if that's the actual way to do it, but I'm the reviewer not the coder, I don't have to get it right 😄

src/invidious/frontend/watch_page.cr
```diff
@@ -0,0 +1,108 @@
...
+    def initialize(
+      @full_videos,
+      @video_streams,
+      @audio_streams,
+      @captions
+    )
+    end
+  end
+
+  def download_widget(locale : String, video : Video, video_assets : VideoAssets) : String
```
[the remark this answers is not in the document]
SamantazFox: Oohh!! It looks cool! I'll have to look into it :)

src/invidious/routes/video_playback.cr
SamantazFox: Yup, that's valid! However, checking the size first is done on purpose: if the string's length is not 11, we don't need to run the regex engine against the string. That saves a small amount of time.

I don't know if it has an important impact, though, given how high level Crystal is. It'd probably be faster to do a loop on the string's bytes and count & check at the same time.

(Yeah, my C habits may have a hard time, but I can't stop thinking that small optimizations like those could still make quite a difference at a larger scale).

src/invidious/routes/video_playback.cr
```diff
@@ -242,31 +242,25 @@ module Invidious::Routes::VideoPlayback
...
 
-    local ||= env.params.query["local"]?
-    local ||= "false"
-    local = local == "true"
+    title = env.params.query["title"]?
 
-    if !id || !itag
-      haltf env, status_code: 400, response: "TESTING"
+    if title && CONFIG.disabled?("downloads")
+      return error_template(403, "Administrator has disabled this endpoint.")
```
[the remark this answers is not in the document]
SamantazFox: Yes it does, because `latest_version` is primarily used to fetch the latest URL for video playback. In > 90% of the cases, no title is passed.

src/invidious/routes/video_playback.cr
SamantazFox: Yup, you're right. I fixed that.

unixfox:
It's possible that this PR broke something, related: https://github.com/iv-org/invidious/issues/2943

Issue #2672, opened 2021-11-28
[Enhancement] Routing layer refactor


matthewmcgarvey:
For a few months now, an effort has been underway to refactor Invidious' routing layer and generally addressing the huge file that was `src/invidious.cr` (you can see some of the connected PRs through this issue https://github.com/iv-org/invidious/issues/1121). I thought I'd make this issue to have one open to make the effort more "official" and I might come and go with the work as life gets more busy or I start feeling burnt out again so I want to have my thoughts written down so others can pick up where I left off if they so choose or make fun of how bad of an idea it was. Whichever makes more sense.

## Need

Before this work began, [`src/invidious.cr`](https://github.com/iv-org/invidious/blob/1978c3d3bd9b3221b4dd47a4ad779708cef984f5/src/invidious.cr) held all the endpoint code and was close to 6000 lines of code.

Here's some of the main pain points
- Large files
- Global routes
- Global methods
- No testing around endpoint code

## Considerations

My very first PR setting up the initial patterns of routing and the jobs https://github.com/iv-org/invidious/pull/1399

The goal is to extract routes into classes/modules that can be more easily read in isolation. By doing this, we also pave the way to potential opportunities to switch to a different framework that would be a better fit for this codebase.

## Goals

- [x] Extract jobs from main file and into new pattern
- [ ] Extract routes from main file and into new pattern
- [ ] Organize global helper methods into better places
- [ ] Migrate to official new routing (not my made up way)
- [ ] Testing???

## Routing Options

There's a bunch of options as for how we might land this effort:

### What we have is good enough

This option would be that somehow, everyone is satisfied with my temporary solution https://github.com/iv-org/invidious/blob/82f3eda82b9e97a94f98e6f325b428d985d9852d/src/invidious/routing.cr#L1-L11

It wouldn't be the worst thing in the world. The only thing I would want, probably is to get the routing moved to a different file because they're starting to take up a lot of space as well. It does bother me that this is not an "official" way of writing kemal apps though.

### We switch to a different framework

This might be the "best" solution as far as the outcome might be nicer than any of the other options, but I think it would take the most work and bring the most risk.

There are definitely plenty of other options:
- https://github.com/grip-framework/grip
- https://athenaframework.org/
- https://luckyframework.org/
- https://amberframework.org/
- https://spider-gazelle.net/#/

Some tradeoffs to consider when evaluating these frameworks would be:
- How "different" of a setup do they impose that would require us to change more or less to fit into
[...]

Issue #2498, opened 2021-10-12
[Meta] Invidious and Kemal | Should Invidious switch web frameworks?


syeopite:
This is a meta issue for discussing the future of Invidious' web framework. 

Kemal has been with Invidious since it's very inception. From the first commit omarroth made, to the latest the community has done. It's been a constant throughout Invidious' history. However, this is where problems arise: Kemal is a microframework. It's far too simple. It doesn't even have support for a more structured routing.  

We had to essentially implement our own full stack framework on top of Kemal: 

- Authentication
- Database manipulation
- Jobs
- File handling
- Structured routing
  
and so on...

It's all things we had to create *in house* on top of Kemal. And at one point, we've got to ask ourselves, is it really worth it? Our own additions on top of Kemal increases the complexity of Invidious tenfold and makes everything almost impossible to maintain; most of our time is already spent catching up to YouTube, do we really want more code to deal with? Invidious and it's Kemal based infrastructure already experiences crashes, memory leaks, and so many missing features like migrations and automatic database creation.

If Invidious wants to improve, it shouldn't stick with Kemal. We need to switch to a proper full stack framework. 

Of course, this would also requiring rewriting a *massive* portion of Invidious, not to mention the need to learn a new framework. Invidious could potentially get a slight performance hit or even increased build times too. It's all factors we have to take into account. Though, in the long run, I think switching frameworks could benefit Invidious greatly. 

All in all, we really have to discuss if sticking with such a simple framework is a good idea.

Rjevski:
Maybe an unpopular opinion considering all of the maintainers seem to be proficient in Crystal and (I am assuming) enjoy working with the language, but I would suggest considering switching languages entirely.

Crystal is IMO a significant barrier to entry when it comes to new contributors as it's a pretty obscure language and has little commercial usage (I'm being generous here - I don't actually know any company that uses it) meaning learning it is unlikely to be directly translatable to income, thus the only practical reason most people would learn the language is to contribute to Invidious which is a significant thing to ask for - I'd rather have the project direct all volunteer effort into keeping up with YouTube rather than wasting it on technicalities such as an obscure language/tooling they won't even be able to capitalize on commercially.

[...]

PR #2925, opened 2022-02-23
Move last remaining routes in main file to new pattern

matthewmcgarvey:
Related #2672
Related #2498

With this change, all routes will have been moved to the new setup.


src/invidious.cr
```diff
@@ -154,8 +154,8 @@ if CONFIG.popular_enabled
   Invidious::Jobs.register Invidious::Jobs::PullPopularVideosJob.new(PG_DB)
 end
 
-connection_channel = Channel({Bool, Channel(PQ::Notification)}).new(32)
-Invidious::Jobs.register Invidious::Jobs::NotificationJob.new(connection_channel, CONFIG.database_url)
+CONNECTION_CHANNEL = Channel({Bool, Channel(PQ::Notification)}).new(32)
```
matthewmcgarvey, line 157: Made a constant so that it could be referenced in the route file

src/invidious.cr
```diff
@@ -399,92 +402,19 @@ Invidious::Routing.get "/s_p/:id/:name", Invidious::Routes::Images, :s_p_image
...
 Invidious::Routing.get "/vi/:id/:name", Invidious::Routes::Images, :thumbnails
 
+Invidious::Routing.get "/channel/:ucid/live", Invidious::Routes::Live, :check
+Invidious::Routing.get "/user/:user/live", Invidious::Routes::Live, :check
+Invidious::Routing.get "/c/:user/live", Invidious::Routes::Live, :check
+
 # API routes (macro)
 define_v1_api_routes()
+Invidious::Routing.get "/api/v1/auth/notifications", Invidious::Routes::API::V1::Authenticated, :notifications_get
+Invidious::Routing.post "/api/v1/auth/notifications", Invidious::Routes::API::V1::Authenticated, :notifications_post
```
matthewmcgarvey, line 412: Terrible names on these but not what it could be instead

src/invidious/routes/api/v1/authenticated.cr
```diff
@@ -397,4 +397,22 @@ module Invidious::Routes::API::V1::Authenticated
...
+  end
+
+  def self.notifications_post(env)
+    env.response.content_type = "text/event-stream"
+
+    topics = env.params.body["topics"]?.try &.split(",").uniq.first(1000)
+    topics ||= [] of String
+
+    create_notification_stream(env, topics, CONNECTION_CHANNEL)
+  end
```
SamantazFox, line 417: I think that you could merge those two in a single function, as the only difference is where the parameters are fetched from (`env.params.query` vs `env.params.body`).

src/invidious/routes/captcha.cr
```diff
@@ -0,0 +1,8 @@
+module Invidious::Routes::Captcha
```
SamantazFox, line 1: could you please put that under the `Login` route, as it is related to Google log-in?

src/invidious/routes/live.cr
```diff
@@ -0,0 +1,34 @@
+module Invidious::Routes::Live
```
SamantazFox, line 1: Same here: this route is related to channels, so it should go under `Invidious::Routes::Channels`

SamantazFox (changes requested):

SamantazFox (approved):

PR #2930, opened 2022-02-23
Add video_context buttons in suggested video on watch view

### LINKED ISSUE #1928: [Enhancement] Add audio mode button to related video suggestions

I've noticed that the related video suggestions lacks the "audio mode" button displayed elsewhere

Add the button. I personally don't care about this but seeing as we've got those buttons everywhere else we may want to put it on the related video suggestions as well.

MathiusD:
Closes #1928

The purpose of this PR is to propose the addition of video playing options (like audio only) on the suggested videos within the route `/watch`

### DIFF
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -320,7 +320,12 @@ we're going to need to do it here in order to allow for translations.
                                         <% end %>
                                     </div>
 
-                                    <div class="pure-u-10-24" style="text-align:right">
+                                    <div class="pure-u-4-24">
+                                        <% endpoint_params = "?v=#{rv["id"]}" %>
+                                        <%= rendered "components/video-context-buttons" %>
+                                    </div>
+
+                                    <div class="pure-u-6-24" style="text-align:right">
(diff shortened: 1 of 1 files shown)


SamantazFox:
image

I think that this icons shouldn't be in here. They don't integrate well with the rest, and end up way too small, which impacts functionality (note that they're super tiny, even on a large HD display).

MathiusD:
In this case, I can propose to place them to the right of the title and to increase their size (Like example attached for example)
image

SamantazFox:
That won't be usable on mobile. The risk of miss-click is too high.

MathiusD:
> That won't be usable on mobile. The risk of miss-click is too high.

That's not wrong. But otherwise, apart from placing them closer to the number of views (or replacing it) I don't know where to place them (because for me, placing them close to the channel would also cause miss-click issues on mobile)

SamantazFox:
Imo, that's something we should simply not implement.
It's not possible to properly integrate those at the moment.

MathiusD:
That's not wrong, in that case I might as well close the PR and indicate it on the associated issue

PR #2936, opened 2022-02-25
Increase size of links displayed in video description

### LINKED ISSUE #2218: Don't clip URLs in video descriptions.

MathiusD:
Closes #2218

As mentioned [here](https://github.com/iv-org/invidious/issues/2218#issuecomment-1050319538), the goal is to increase the length of links displayed to 50 characters (This follows the discussion in this PR [#2906] and then in the original issue [#2218])

### DIFF
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -587,7 +587,7 @@ def content_to_comment_html(content)
           end
         end
 
-        text = %(<a href="#{url}">#{text}</a>)
+        text = %(<a href="#{url}">#{reduce_uri(url)}</a>)
       elsif watch_endpoint = run["navigationEndpoint"]["watchEndpoint"]?
         length_seconds = watch_endpoint["startTimeSeconds"]?
         video_id = watch_endpoint["videoId"].as_s
@@ -595,10 +595,10 @@ def content_to_comment_html(content)
         if length_seconds && length_seconds.as_i > 0
           text = %(<a href="javascript:void(0)" data-onclick="jump_to_time" data-jump-time="#{length_seconds}">#{text}</a>)
         else
-          text = %(<a href="/watch?v=#{video_id}">#{text}</a>)
+          text = %(<a href="/watch?v=#{video_id}">#{reduce_uri("/watch?v=#{video_id}")}</a>)
         end
       elsif url = run.dig?("navigationEndpoint", "commandMetadata", "webCommandMetadata", "url").try &.as_s
-        text = %(<a href="#{url}">#{text}</a>)
+        text = %(<a href="#{url}">#{reduce_uri(url)}</a>)
       end
     end
 
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -573,10 +573,11 @@ def content_to_comment_html(content)
 
     if run["navigationEndpoint"]?
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
+        base_url = URI.parse(url)
         url = URI.parse(url)
 
         if url.host == "youtu.be"
-          url = "/watch?v=#{url.request_target.lstrip('/')}"
+          url = "youtube.com/watch?v=#{url.request_target.lstrip('/')}"
         elsif url.host.nil? || url.host.not_nil!.ends_with?("youtube.com")
           if url.path == "/redirect"
             # Sometimes, links can be corrupted (why?) so make sure to fallback
@@ -587,7 +588,12 @@ def content_to_comment_html(content)
           end
         end
 
-        text = %(<a href="#{url}">#{reduce_uri(url)}</a>)
+        if base_url.host.not_nil!.ends_with?("youtube.com") && base_url.path != "/redirect"
+          displayed_url = "youtube.com#{base_url.request_target}"
+        else
+          displayed_url = url
+        end
+        text = %(<a href="#{url}">#{reduce_uri(displayed_url)}</a>)
       elsif watch_endpoint = run["navigationEndpoint"]["watchEndpoint"]?
         length_seconds = watch_endpoint["startTimeSeconds"]?
         video_id = watch_endpoint["videoId"].as_s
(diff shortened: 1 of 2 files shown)


src/invidious/comments.cr
```diff
@@ -587,18 +587,18 @@ def content_to_comment_html(content)
...
+        text = %(<a href="#{url}">#{reduce_uri(url)}</a>)
       elsif watch_endpoint = run["navigationEndpoint"]["watchEndpoint"]?
         length_seconds = watch_endpoint["startTimeSeconds"]?
         video_id = watch_endpoint["videoId"].as_s
 
         if length_seconds && length_seconds.as_i > 0
           text = %(<a href="javascript:void(0)" data-onclick="jump_to_time" data-jump-time="#{length_seconds}">#{text}</a>)
         else
-          text = %(<a href="/watch?v=#{video_id}">#{text}</a>)
+          text = %(<a href="/watch?v=#{video_id}">#{reduce_uri("/watch?v=#{video_id}")}</a>)
```
SamantazFox, line 598: In this case, you probably want to recreate the original youtube URL (`www.youtube.com/watch?v=<id>&t=<time>`), otherwise something like `/watch?v=<id>&t=<time>` won't make sense to the user.

src/invidious/helpers/utils.cr
```diff
@@ -365,3 +365,14 @@ def fetch_random_instance
 
   return filtered_instance_list.sample(1)[0]
 end
+
+def reduce_uri(uri : URI | String, max_length : Int32? = 50, suffix : String? = "...") : String
```
SamantazFox, line 369: No need of making `max_length` and `suffix` nillable. Assigning a default value is enough to make the parameters optional. And that can let you avoid those useless `.nil?` checks ^^

Also, please use the horizontal ellipsis unicode character `…` (U+2026) instead of three dots.

SamantazFox (changes requested):

src/invidious/comments.cr
MathiusD: In this case, to be consistent with the link target it would be better to display something like `<currentHost>/watch?v=<id>`

src/invidious/helpers/utils.cr
MathiusD: > No need of making `max_length` and `suffix` nillable. Assigning a default value is enough to make the parameters optional. And that can let you avoid those useless `.nil?` checks ^^

Here the possibility to put `nil` is an artifact of my local version with the configuration [That I had mentioned [here](https://github.com/iv-org/invidious/issues/2218#issuecomment-1050322509)] (Where the default value was taken from the config), I remove that immediately and yes the check `nil?` is obsolete now

> Also, please use the horizontal ellipsis unicode character … (U+2026) instead of three dots.

Indeed, I do it right now

src/invidious/comments.cr
SamantazFox: I'd prefer if we don't chznge the current behavior, which is displaying the youtube URL (like in the original description), but redirecting to the watch page.

src/invidious/comments.cr
```diff
@@ -587,18 +587,18 @@ def content_to_comment_html(content)
...
         end
 
-        text = %(<a href="#{url}">#{text}</a>)
+        text = %(<a href="#{url}">#{reduce_uri(url)}</a>)
       elsif watch_endpoint = run["navigationEndpoint"]["watchEndpoint"]?
         length_seconds = watch_endpoint["startTimeSeconds"]?
         video_id = watch_endpoint["videoId"].as_s
 
         if length_seconds && length_seconds.as_i > 0
           text = %(<a href="javascript:void(0)" data-onclick="jump_to_time" data-jump-time="#{length_seconds}">#{text}</a>)
```
SamantazFox, line 596: You forgot that one ^^

MathiusD: For the link to a section of the video you think it's interesting to reduce it too?

SamantazFox: * `www.youtube.com` = 15 chars
* `/watch?v=` = 9
* a video id is 11 chars
* `&t=...` = from 4 to 12 chars (`10h25m32s` = 9 chars)

So the total is 47 characters. So it won't be needed :)

src/invidious/comments.cr
```diff
@@ -573,10 +573,11 @@ def content_to_comment_html(content)
 
     if run["navigationEndpoint"]?
       if url = run["navigationEndpoint"]["urlEndpoint"]?.try &.["url"].as_s
+        base_url = URI.parse(url)
         url = URI.parse(url)
 
         if url.host == "youtu.be"
-          url = "/watch?v=#{url.request_target.lstrip('/')}"
+          url = "youtube.com/watch?v=#{url.request_target.lstrip('/')}"
```
SamantazFox, line 580: note that by adding `youtube.com` here youtube URLs will redirect to youtube rather than invidious.

MathiusD: Indeed I should have done as with `displayed_url` before, I correct this immediately

src/invidious/comments.cr
```diff
@@ -574,31 +574,35 @@ def content_to_comment_html(content)
...
+        text = %(<a href="#{url}">#{reduce_uri(displayed_url)}</a>)
       elsif watch_endpoint = run["navigationEndpoint"]["watchEndpoint"]?
         length_seconds = watch_endpoint["startTimeSeconds"]?
         video_id = watch_endpoint["videoId"].as_s
 
         if length_seconds && length_seconds.as_i > 0
           text = %(<a href="javascript:void(0)" data-onclick="jump_to_time" data-jump-time="#{length_seconds}">#{text}</a>)
         else
-          text = %(<a href="/watch?v=#{video_id}">#{text}</a>)
+          text = %(<a href="/watch?v=#{video_id}">#{reduce_uri("youtube.com/watch?v=#{video_id}")}</a>)
```
SamantazFox, line 602: In this case, you don't need to call `reduce_uri` :)
See: https://github.com/iv-org/invidious/pull/2936/files#r815028109

SamantazFox (approved):
Looks good to me!

SamantazFox:
I'll test that as soon as I can :)

src/invidious/helpers/utils.cr
```diff
@@ -365,3 +365,11 @@ def fetch_random_instance
 
   return filtered_instance_list.sample(1)[0]
 end
+
+def reduce_uri(uri : URI | String, max_length : Int32 = 50, suffix : String = "…") : String
+  str = uri.to_s.sub(/https?:\/\//, "")
```
SamantazFox, line 370:
```suggestion
  str = uri.to_s.sub(/^https?:\/\//, "")
```

SamantazFox: That's minor, but I realized that it's needed for e.g web archive links, which have another `https://` in the URL path.

MathiusD: > That's minor, but I realized that it's needed for e.g web archive links, which have another `https://` in the URL path.

It's true I had forgotten that too, I'll change it right away

PR #2944, opened 2022-03-01
Fix deprecated helm chart dependency

jmhbnz:
When attempting to deploy invidious via helm today I ran the initial `helm dep build` and encountered `Error: no repository definition for https://kubernetes-charts.storage.googleapis.com/. Please add the missing repos via 'helm repo add'`

This is because that repository is now deprecated, the new mirror is https://charts.helm.sh/stable/.  Refer: https://github.com/helm/charts/issues/24432 

This pull request fixes the mirror so others won't be caught out by the same issue.

### DIFF
--- a/kubernetes/values.yaml
+++ b/kubernetes/values.yaml
@@ -14,7 +14,7 @@ autoscaling:
   targetCPUUtilizationPercentage: 50
 
 service:
-  type: clusterIP
+  type: ClusterIP
   port: 3000
   #loadBalancerIP:
 
@@ -34,12 +34,17 @@ securityContext:
 
 # See https://github.com/helm/charts/tree/master/stable/postgresql
 postgresql:
-  postgresqlUsername: kemal
-  postgresqlPassword: kemal
-  postgresqlDatabase: invidious
-  initdbUsername: kemal
-  initdbPassword: kemal
-  initdbScriptsConfigMap: invidious-postgresql-init
+  image:
+    registry: quay.io
+  auth:
+    username: kemal
+    password: kemal
+    database: invidious
+  primary:
+    initdb:
+      username: kemal
+      password: kemal
+      scriptsConfigMap: invidious-postgresql-init
 
 # Adapted from ../config/config.yml
 config:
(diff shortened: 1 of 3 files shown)


unixfox:
IMO we should switch to https://github.com/bitnami/charts/tree/master/bitnami/postgresql instead of keeping relying on helm charts stable.

jmhbnz:
> IMO we should switch to https://github.com/bitnami/charts/tree/master/bitnami/postgresql instead of keeping relying on helm charts stable.

Sure thing, I can update and test the deployment with that version of postgres chart tomorrow and will update this pull request if all working ok 👍🏻

kubernetes/values.yaml
```diff
@@ -34,12 +34,17 @@ securityContext:
...
 # See https://github.com/helm/charts/tree/master/stable/postgresql
 postgresql:
-  postgresqlUsername: kemal
-  postgresqlPassword: kemal
-  postgresqlDatabase: invidious
-  initdbUsername: kemal
-  initdbPassword: kemal
-  initdbScriptsConfigMap: invidious-postgresql-init
+  image:
+    registry: quay.io
```
unixfox, line 38: Why quay instead of docker hub?

kubernetes/values.yaml
```diff
@@ -34,12 +34,17 @@ securityContext:
 
 # See https://github.com/helm/charts/tree/master/stable/postgresql
```
unixfox, line 35: I think this comment need to be updated too.

unixfox (changes requested):

kubernetes/values.yaml
jmhbnz: Reviewing the installation instructions the project seemed to have a preference that quay.io was preferred: https://docs.invidious.io/Installation.md 

> The Invidious docker image is only [available on Quay](https://quay.io/repository/invidious/invidious) because, unlike Docker Hub, [Quay is Free and Open Source Software](https://github.com/quay/quay/blob/master/LICENSE). This is reflected in the docker-compose.yml file used in this walk-through.

The helm chart was already pulling invidious from quay.io, so I have kept that approach for pulling this dependency :)

kubernetes/values.yaml
jmhbnz: Great spotting, fixed thanks!

unixfox (approved):

unixfox:
merged, thank you :)

PR #2544, opened 2021-10-27
Overhaul search filters

syeopite:
Cherry-picked from #2215 along with changes to resolve #2464

- Decluttered the search bar by moving filters to URL parameters
- Extracted filters box into a separate component
- Slightly improved design
- Add ability to remove filters by clicking on applied ones
- Rename duration filters (See #2464)
- Add `4 - 20 minutes` as a duration filter (See #2464)

PR #2956, opened 2022-03-07
Overhaul search filters

SamantazFox:
Replaces #2544


SamantazFox:
Note to self: change translation keys in other languages

SamantazFox:
Poke @matthewmcgarvey: do you mind reviewing that?

src/invidious/search/filters.cr
```diff
@@ -228,6 +228,25 @@ module Invidious::Search
...
+      raw_params["date"] = [@date.to_s.underscore] if !@date.none?
+      raw_params["type"] = [@type.to_s.underscore] if !@type.all?
+      raw_params["sort"] = [@sort.to_s.underscore] if !@sort.relevance?
+
+      if !@duration.none?
+        raw_params["duration"] = [@duration.to_s.underscore]
+      end
+
+      if !@features.none?
+        raw_params["features"] = [Filters.format_features(@features)]
```
matthewmcgarvey, line 249: Should this split on the comma? It's going to look like `["live,4k,3d"]`

matthewmcgarvey (approved):
Looks good. Little difficult to look at everything in great detail but I'm confident this will work 👍

SamantazFox:
> Little difficult to look at everything in great detail but I'm confident this will work 👍

If that helps, each commit is made so that it can be reviewed separately (but still in order). Checking out at any commit should compile without an error message and the CI should always pass.

src/invidious/search/filters.cr
```diff
@@ -192,6 +192,42 @@ module Invidious::Search
       return {filters, channel, cleaned_query, subscriptions}
     end
 
+    def self.from_iv_params(params : HTTP::Params) : Filters
+      # Temporary variables
+      filters = Filters.new
+
+      if type = params["type"]?
+        filters.type = Type.parse?(type) || Type::All
+        params.delete("type")
```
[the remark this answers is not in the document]
SamantazFox: My idea is to use this as a sanitization function, by calling it from kemal's middleware (a dedicated handler, as it's done for session management). By removing the parameters, that will ensure that the non-sanitized inputs won't be used somewhere else. The goal would to end up an empty `HTTP::Params` object.

src/invidious/search/filters.cr
SamantazFox: No, it shoudln't split on the comma, and yup, that's expected :)

Previously, the features filter was passed in the query, as so:
```some search query features:live,4k,3d"```

Imo it's better to keep it that way, even with URL parameters, so the URL can be easily edited by hand, without needing the lenghy `&features=live&features=4k&features=3d....`.

SamantazFox:
I've fixed a few things since Matthew's review:
- Broken previous/next URLs on search page (forgot surrounding `<%= %>`)
- Broken `Query.process` (`when` should contain all conditions on a single line, separated by commas)
- Enhanced accessibility on filters (by using `<fieldset>` and `<legend>` HTML elements)

SamantazFox:
Stuff that remains "TODO": properly handle incompatible search filters (e.g `duration=short` + `type=channel`). Maybe some JS to make that behavior more explicit on the client side could be nice.

matthewmcgarvey (approved):

SamantazFox:
Deployed on test.invidious.io.
Bogus search query (to see new error message): https://test.invidious.io/search?q=LTT&type=channel&duration=short

matthewmcgarvey (approved):

SamantazFox:
Sorry for the multiple force-pushes :x

PR #2974, opened 2022-03-15
Request minified JSON from innertube

SamantazFox:
This will probably save a bit of bandwidth on large instances.
Thank to TiA4f8R fro telling us about it!


SamantazFox:
I've tested using [`yt-api-helper.sh`](https://github.com/iv-org/youtube-utils/blob/master/scripts/yt-api-helper.sh) and it worked fine, so I don't see why it wouldn't work on invidious.

unixfox (approved):

unixfox:
I'm for merging it then.

unixfox:
Merging as it works without any issue on my instance.

PR #2990, opened 2022-03-27
Safari audio double duration fix for iOS 15

138138138:
The previous method breaks Always Loop feature on iOS 15.
The previous player.currentTime(player.duration() + 1) sometimes breaks the entire player.
Now it jumps to (end - 1) seconds when the time goes between over half and (end - 2) seconds.
With Always Loop on, player will jump to the beginning after 1 second.


SamantazFox:
Hello! Thanks for your contribution!
Can you confirm that the fix still works on older iOS releases? I don't have a device to test this out.

138138138:
Tested working on iOS 15.4 and iOS 12.5.5 (both are iPhone).

SamantazFox (approved):

Issue #2850, opened 2022-02-01
[Enhancement] Detect rolling cipher/video encryption and force proxy or do not load the video page if proxy disabled.


unixfox:
Detect if rolling cipher/encryption is involved then force proxy in order to improve the user experience. An user don't specially know that he has to enable proxy in the settings for the music videos to work.

Moreover, if the proxy is disabled in the settings do not load the video page and throw an error. Do not throw an error for a local invidious instance.

unixfox:
Not really an issue anymore since we now want to switch to DASH by default.

PR #3011, opened 2022-04-08
Bump videojs to 7.12.1

### LINKED ISSUE #2065: [Bug] DASH video stalls after two 502 errors
**Describe the bug**

DASH video will stop playing and require either a quality change or pause and unpause to continue playing if the segment it tries to request returns a 502 error more than once.
**Steps to Reproduce**

I suggest switching to another player such as https://github.com/sampotts/plyr

### LINKED ISSUE #2385: [Enhancement] Fallback to proxying the requests if the user is blocking third party requests or google servers.
While discussing on the techlore matrix server I discovered that some people are blocking third party requests. By default invidious request through google servers for the video data so this is an issue for them.

Automatically detect if the user is blocking the requests and switch to proxying the requests through invidious servers.
This will probably only work with javascript, so with videojs.

**Additional context**
The person that I discussed with blocked the requests with ublock origin.

### LINKED ISSUE #2848: [Enhancement] Trying to use the latest video.js version
Here are my findings at trying to make the latest video.js version working.

What I discovered:

- videojs-http-source-selector is completely broken, but that's because we are using the old method for initiating the video.js player. I submitted a fix here: https://github.com/iv-org/invidious/pull/2847

- when selecting another quality from the quality selector, video.js will sometimes try to use an unsupported codec, either we try to understand what's going on or we remove the codec from the available dash playlist:

### LINKED ISSUE #2941: [Bug] No timestamp when hovering the progress bar
The video player does not show the timestamp for the point on the trackbar (when am I jumping to?)

Please add a timestamp to the trackbar hover thumbnail in the video player.

### LINKED ISSUE #2942: [Bug] Rotating pauses video
**Describe the bug**

When rotating from landscape to vertical on iOS, the video leaves fullscreen and pauses
**Steps to Reproduce**

**Screenshots**

https://user-images.githubusercontent.com/89056889/155855110-dc6e7454-c399-4d73-8ddd-0022ac3d87c0.MOV

**Additional context**

unixfox:
Related to #2848

Help with the issue in #2850

Fixes #2941
Fixes #2385
Fixes #2065
Fixes #2942

- This PR bumps video.js to the latest version 7.12.1. The versions after that seems to have some issues with the [videojs-http-source-selector](https://github.com/jfujita/videojs-http-source-selector) plugin which is by the way unmaintained, we should look into removing it with another way to offer a quality selection.
Webm is not served anymore through the manifest dash API because webm with dash is unfortunately not supported:
  * https://github.com/videojs/video.js/issues/7267
  * https://github.com/videojs/http-streaming/blob/main/docs/supported-features.md#container-formats

- I've added a small piece of code into the player javascript code so that it automatically reload the player on errors when using DASH, documentation: https://github.com/videojs/http-streaming/blob/main/docs/reload-source-on-error.md

- Also related to #2850, videojs **and the default video player of the browser (no JS)** will try to proxy the video/audio if it encounters an error. Change in silvermine videojs-quality-selector library is in https://github.com/iv-org/videojs-quality-selector/pull/1

- Finally, videojs will now send the range parameter as a query string in order to avoid any [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request). This is very useful for proxying the video data through a server under another domain. [That's what Piped is already doing](https://github.com/TeamPiped/Piped/blob/master/src/components/VideoPlayer.vue#L320).

---

### Nice to have
- [ ] Have a better retry in case of errors, don't try indefinitely every 10 seconds. Stop after X retries then continue later.


unixfox:
Ideally, I would like to still offer by default webm in the manifest API but not offering webm to video.js when it requests the dash manifest.

unixfox:
Bugs remaining:
- When 1080p is forced in the preferences then changing the quality from the quality icon to 720p or another quality then the lowest quality will be chosen.
- When auto is set and then changing quality from the quality icon to 720p or another quality then the player will throw an errror.

unixfox:
Okay, videojs 7.12.1 doesn't seem to break videojs-http-source-selector. There is no repeatable `video append of` errors when switching to 1080p in the quality selector when auto is set in the preferences.

videojs-dependencies.yml
```diff
@@ -1,9 +1,7 @@
-# Due to an firefox issue, we're stuck on 7.11.0. If you're hosting a private instance
-# and you're using a chromium based browser, feel free to bump this to the latest version
-# in order to get support for higher resolutions on more videos.
+# Due to a 'video append of' error see #3011, we're stuck on 7.12.1.
```
SamantazFox, line 1:
```suggestion
# Due to a 'video append of' error, we're stuck on 7.12.1 (see #3011).
```

assets/js/player.js
```diff
@@ -49,6 +49,10 @@ videojs.Vhs.xhr.beforeRequest = function(options) {
 
 var player = videojs('player', options);
 
+player.reloadSourceOnError({
+    errorInterval: 10
+});
```
SamantazFox, line 54: What's that for, exactly? Won't it interfere with `player.on('error', function (event)` (line 147)?

unixfox: Yes it does and in fact I plan to remove the player.error something.

But I'm still trying some things, with `retryOnError` option too. Hence this PR is still "blocked"

unixfox:
I've added more code! One major part is that videojs will now try to proxy the video if it encounters an error at the start of the playback of the video.
This should greatly reduce the amount of "the media could not be loaded" errors.

I've updated the original description for reflecting this change.

assets/js/player.js
unixfox: I've made it so it is only loaded when using dash because it's a dash only feature: https://github.com/videojs/http-streaming/blob/main/docs/reload-source-on-error.md

unixfox:
In testing on https://yewtu.be since 3 days ago.

unixfox:
Seems to work fine on my end. And the "nice to have" list is probably good for another pull request once I've more free time again to work on that.

assets/js/player.js
```diff
@@ -41,6 +41,12 @@ embed_url = location.origin + '/embed/' + video_data.id + embed_url.search;
 var save_player_pos_key = "save_player_pos";
 
 videojs.Vhs.xhr.beforeRequest = function(options) {
+    if (options.headers) {
+        if (options.headers.Range) {
+            options.uri = options.uri + '&range=' + options.headers.Range.split("=")[1];
+            delete options.headers.Range;
```
SamantazFox, line 47: Do note that removing the `Range` header needs to be handled on the server side.
@unixfox it works for you because you don't use Invidious' proxy!

unixfox: > Do note that removing the `Range` header needs to be handled on the server side. @unixfox it works for you because you don't use Invidious' proxy!

It's mostly handled on the youtube server side but I see that there are no code related to the `Range` parameter inside invidious but probably not mandatory to work because it worked fine on a vanilla invidious installation.

unixfox: Part of the code removed.

IronMaltese:
#2942 seems to be fixed aswell :)

assets/js/player.js
SamantazFox: > It's mostly handled on the youtube server side but I see that there are some code related to the `Range` parameter inside invidious

Tbh, the proxying part should be reworked, but last time I've tried to touch that part of the code, everything broke (see #2900).

unixfox: That's true. The range header being transformed into a query string is really only useful for me because I run a big instance with multiple external proxy servers. Now that I think again about it, I don't think it makes sense to modify that for everyone.

TheFrenchGhosty (changes requested):
If possible, please replace `assets/js/silvermine-videojs-quality-selector.min.js` with a non-minified version of it.

unixfox:
> If possible, please replace `assets/js/silvermine-videojs-quality-selector.min.js` with a non-minified version of it.

Why? It's the point of https://github.com/iv-org/videojs-quality-selector

And it's always been like this, and it's much faster for the browser to load a minified file. If we want to provide the source code then we can do it through a source map: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/SourceMap

TheFrenchGhosty:
Minified code doesn't save much while making the code unreadable. We have already moved some minified libraries to their non-minified for this reason.

FOSS shouldn't use any minified code, it almost obfuscate code just to save a really minimal amount of bandwidth.

unixfox:
Minified js do allow saving resources and makes page loading faster... if that wasn't the case "minifying" wouldn't exist.

But anyway, there is no point in discussing this when source maps do exist in order to solve the "source code" issue. It's a system that automatically show the source code file instead in the devtools.

You are reminding me that I should try to minify everything at compile time in invidious.

TheFrenchGhosty:
>You are reminding me that I should try to minify everything at compile time in invidious.

I'm not fundamentally against minify-ing at compile time, but I'm against shipping anything minified in the source code.

unixfox:
> >You are reminding me that I should try to minify everything at compile time in invidious.
> 
> I'm not fundamentally against minify-ing at compile time, but I'm against shipping anything minified in the source code.

Again source maps are solving the issue... Please read the docs.

TheFrenchGhosty:
>Please read the docs.

Okay, if I understand it correctly, it should be fine.

TheFrenchGhosty (approved):

assets/js/player.js
```diff
@@ -49,6 +49,42 @@ videojs.Vhs.xhr.beforeRequest = function(options) {
...
+    if (video_data.params.quality !== 'dash') {
+        if (!player.currentSrc().includes("local=true") && !video_data.local_disabled) {
+            var currentSources = player.currentSources();
+            for (var i = 0; i < currentSources.length; i++) {
+                currentSources[i]["src"] += "&local=true"
+            }
+            player.src(currentSources)
+        }
+        else if (player.error().code === 2 || player.error().code === 4) {
+            setTimeout(function (event) {
```
SamantazFox, line 62: Currently, this error handler retries fetching the video indefinitely. We should probably put a timeout/counter.

unixfox: This is exactly what I described in the "nice to have" list.

I plan to do another PR for that.

unixfox:
@SamantazFox How do we move forward with this PR? Anything else needed to test before merging?

SamantazFox:
@unixfox Sorry for the late reply: you've deployed it on your instance, right? If yes, I'm confident that we can merge it without further testing, as I haven't seen any issues related to that.

doubledge-spec:
here's a source selector fork being maintained by freetube
https://github.com/FreeTubeApp/videojs-http-source-selector

PR #3013, opened 2022-04-08
Rescue DB errors in get_video()

SamantazFox:
this should prevent the vast majority of issues related to DB problems.

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -1094,6 +1094,10 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
   end
 
   return video
+rescue DB::Error
+  # Avoid common `DB::PoolRetryAttemptsExceeded` error and friends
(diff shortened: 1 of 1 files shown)


unixfox:
I can see this being an "infinite" loop if the DB is not reachable for a long time and thus consuming all the CPU.

SamantazFox:
@unixfox How so? The `get_video()` function first tries to get the video the usual way (from the DB, check if it needs to be updated, fetch if required, and update entry) but if it encounters a DB error, it simply fetches the video info directly from innertube, and the DB is not touched again.

unixfox:
Oops sorry, I reviewed the code late at night and I forgot that this function bypass postgresql.

SamantazFox:
@unixfox no problem ^^ So it's good to merge, then?

unixfox (approved):

PR #3016, opened 2022-04-09
Updates the URL of the var url_faq

oliveiraleo:
The old URL was returning a HTTP 404 status code. My commit fixes that as it points the URL to the new file location.


SamantazFox:
@oliveiraleo Ah, neat! thanks for fixing :)

PR #3020, opened 2022-04-10
Improve user account UI

SamantazFox:


SamantazFox:
bump

assets/css/user.css
```diff
@@ -0,0 +1,132 @@
...
+	padding: 1em;
+}
+
+/* Background color accent using transparency */
+/* TODO: handle themes better, ffs */
+.light-theme .captcha { background-color: #0002; }
+.dark-theme  .captcha { background-color: #fff2; }
+
+@media (prefers-color-scheme: light) { .no-theme .captcha { background-color: #0002; } }
+@media (prefers-color-scheme: dark)  { .no-theme .captcha { background-color: #fff2; } }
```
anyone00, line 63: Are these colors missing 2 digit of the hex code?

anyone00: I'm not sure how theming is handled in invidious but this seems redundent

assets/css/user.css
```diff
@@ -0,0 +1,132 @@
...
+    text-align: center;
+}
+
+.login-container .username-pass,
+.register-container .username-pass {
+	width: max-content;
+	margin: auto;
+	padding: 0 2em;
+}
+
```
anyone00, line 50: from the dark/light mode code below you might add

```
.dark-theme  .login-container .username-pass::placeholder { background-color: #ffffff; }
@media (prefers-color-scheme: dark)  { .login-container .username-pass::placeholder { background-color: #ffffff; } }
```

anyone00: or nvm I think you deleted the placeholder text

B0pol:
It overflows on mobile

assets/css/user.css
SamantazFox: > Are these colors missing 2 digit of the hex code?

No, it's the short RGBA notation. It's the same as doing `#FFFFFF22` (RRGGBBAA).

> I'm not sure how theming is handled in invidious but this seems redundent

Short answer: badly, lol
Long answer: This due to 1) not using JS for the "auto" theme (we only use CSS for light/dark mode detection) and 2) because we're not using `--var: <properties>;` and `<property>: var(--var);` in order to have IE11 compatibility.

SamantazFox:
> It overflows on mobile

Ah, crap :/ Thanks for checking!

PR #3023, opened 2022-04-13
Filters visual improvement

AHOHNMYC:
Two small changes:
* Pointer cursor on filter panel toggler
* [Real minus sign](https://en.wikipedia.org/wiki/Plus_and_minus_signs) instead of [hyphen-minus](https://en.wikipedia.org/wiki/Hyphen-minus). Minus sign has same width as plus sign. Hyphen-minus is shorter, which leads to twitching text on panel toggle

GitHub shows no difference between ‒ and − because of monospace font 😸 (and in this sentence too, because of small font size, huh)

### DIFF
--- a/assets/css/search.css
+++ b/assets/css/search.css
@@ -5,6 +5,7 @@ summary {
 	font-size: 1.17em;
 	font-weight: bold;
 	margin: 0 auto 10px auto;
+	cursor: pointer;
 }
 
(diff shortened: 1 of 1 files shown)


SamantazFox:
Nice! I wasn't even aware that there was a minus sign, so I spent too much time trying to find a hyphen with matching length XD Thanks for the contribution :)

Issue #3024, opened 2022-04-14
[Bug] can't connect to postgresql with a unix socket


unixfox:
**Describe the bug**
4 days ago invidious stopped working with postgresql through a unix socket. It used to work.
I'm using postgresql 10 on docker.
Need to investigate

SamantazFox:
I'm keeping this issue open, until we have found the source of the problem.

witchent:
Is this supposed to be fixed or not? in the config example it is still listed and I just pulled my hair out trying to make it work, but it seems invidious defaults to localhost if nothing is specified after the @ sign, even if the url ends with?host=..

witchent:
Maybe updating db would work. I found this merge request:
https://github.com/will/crystal-pg/pull/241

witchent:
Found a working solution:
``database_url: postgres://user:password@socket-path/dbname``
where user, password and socket-path are percent encoded (so / -> %2F, = -> %3D etc)

Caian:
This is quite an old thread, but I found out that I had to pass the full path of the socket, not only the socket directory like in many other libraries:  `postgres://invidious:password@/invidious?host=/postgres/.s.PGSQL.5432`

PR #2928, opened 2022-02-23
Bump dependencies

SamantazFox:
* Update `sqlite3` from v0.18.0 to v0.19.0 ([changelog](https://github.com/crystal-lang/crystal-sqlite3/releases/tag/v0.19.0))
* Update `pg` from v0.24.0 to v0.26.0 ([[changelog](https://github.com/will/crystal-pg/blob/v0.26.0/CHANGELOG))
* Update `db` from v0.10.1 to v0.11.0 ([changelog](https://github.com/crystal-lang/crystal-db/releases/tag/v0.11.0), update required by `sqlite3` and `pg`)
* Update `spectator` from v0.10.4 to v0.10.5 ([changelog](https://gitlab.com/arctic-fox/spectator/-/blob/v0.10.5/CHANGELOG.md))

PR #3026, opened 2022-04-14
Revert "Bump dependencies"

SamantazFox:
Reverts iv-org/invidious#2928

It seems to be causing multiple DB connectivity issues

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -1,9 +1,5 @@
 version: 2.0
 shards:
-  ameba:
-    git: https://github.com/crystal-ameba/ameba.git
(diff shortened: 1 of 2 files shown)


unixfox:
Fixes https://github.com/iv-org/invidious/issues/3024

TheFrenchGhosty:
It would be better, in my opinion, to find what's the actual problem... but I guess we can revert for now.

SamantazFox:
@TheFrenchGhosty yes, that would be ideal, but at the moment, I don't have the time to investigate.

PR #3031, opened 2022-04-16
Fix filter checkbox

AHOHNMYC:
Due to different prefixes in id (`filter-features` in `input` and `filter-feature` in `label`) click on `label` didn't affect corresponding checkbox.


SamantazFox:
Oh, I though I fixed it... Seems like I forgot xD
Thanks for the PR :)

Issue #672, opened 2019-08-06
Footer font color is hard to read when using Dark Mode 


TheFrenchGhosty:
In my opinion the color of the font used in the footer is a bit too dark when using dark mode, I don't personally have a problem to read it but I guess it might be better to have a lighter color for people with vision problems.

White would be too "visible" so I guess using a color a bit darker than the one used for the "normal" text (#a0a0a0) would be the way to go.

![footer](https://user-images.githubusercontent.com/47571719/62502877-366abc00-b7e1-11e9-81f2-902997eb4860.png)

TheFrenchGhosty:
Still an issue.

saltycrys:
The color can be adjusted in `default.css`: https://github.com/iv-org/invidious/blob/e036d89a86cb7ab2d65ec2a9c3ea654c9774e70e/assets/css/default.css#L283-L287

TheFrenchGhosty:
bump

le-jun:
Let’s use this opportunity to tackle the interface accessibility question at the same time. According to W3C, the contrast between text and background should be at least 4.5:1 (AA level) and 7:1 (AAA level). Only modifying the text colour (no text size nor background change), then we could use:

* #adadad (7:1)
* #8a8a8a (4.55:1)

Ratio measured using [this tool](https://juicystudio.com/services/luminositycontrastratio.php#specify)

ahangarha:
This issue is already resolved. I see good contrast in dark mode.

TheFrenchGhosty:
@ahangarha No it isn't resolved.

ahangarha:
Based on my test on yewtu.be and invidious.snopyta.org, the contrast between text and background color in the footer and dark mode is 4.99 AA. It is a good contrast.

PR #3036, opened 2022-04-17
Increased footer text contrast to 7:1 ratio

le-jun:
Increased footer text contrast to 7:1 ratio for accessibility as stated in WCAG21. Being footer supplementary content, the 4.5:1 could have been used as well if prefered.

https://github.com/iv-org/invidious/issues/672#issuecomment-1100810591

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -302,7 +302,7 @@ p.video-data   { margin: 0; font-weight: bold; font-size: 80%; }
  */
 
 footer {
-  color: #919191;
+  color: #adadad;
   margin-top: auto;
   padding: 1.5em 0;
   text-align: center;


unixfox:
Thanks for your contribution.

Can you post a screenshot of before and after?

SamantazFox:
Before/after:
image
image

**Note:** You need to also change the `color` property under `footer a`, so it changes the links too:
image

But that makes text hard to read on white theme:
image

Imho, we should probably wait on #2215 (I'm working on merging that one).

le-jun:
> But that makes text hard to read on white theme:

Ha, you’re right. Haven’t thought about checking how the light/dark theme works. Anyway I agree with you on waiting for https://github.com/iv-org/invidious/pull/2215 if that’s still active.

PR #2934, opened 2022-02-24
Bump kemal dependency to v1.1.2

SamantazFox:
Supersedes #2921
Closes: https://github.com/iv-org/invidious/issues/2783

Solve broken playback on crystal 1.3.*
See: https://github.com/kemalcr/kemal/pull/628

PR #3041, opened 2022-04-17
bump to crystal 1.4.0

unixfox:
Invidious should be now stable with the latest version of crystal thanks to https://github.com/iv-org/invidious/pull/2934

### DIFF
--- a/docker/Dockerfile.arm64
+++ b/docker/Dockerfile.arm64
@@ -1,5 +1,5 @@
-FROM alpine:3.15 AS builder
-RUN apk add --no-cache 'crystal=1.2.2-r0' shards sqlite-static yaml-static yaml-dev libxml2-dev zlib-static openssl-libs-static openssl-dev musl-dev
+FROM alpine:edge AS builder
+RUN apk add --no-cache 'crystal=1.4.0-r0' shards sqlite-static yaml-static yaml-dev libxml2-dev zlib-static openssl-libs-static openssl-dev musl-dev
 
 ARG release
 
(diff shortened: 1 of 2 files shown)


unixfox:
Soon in testing on yewtu.be: https://github.com/unixfox/invidious-custom/actions/runs/2180947035

docker/Dockerfile.arm64
```diff
@@ -1,5 +1,5 @@
-FROM alpine:3.15 AS builder
-RUN apk add --no-cache 'crystal=1.2.2-r0' shards sqlite-static yaml-static yaml-dev libxml2-dev zlib-static openssl-libs-static openssl-dev musl-dev
+FROM alpine:edge AS builder
```
SamantazFox, line 1: could we not jump to edge, to avoid accidentally breaking images if Crystal 1.5 releases and invidious doesn't work with it?

unixfox: You can't have crystal 1.4.0 without using edge.

unixfox: And if crystal 1.5 comes out, the docker image will just fail to build. That's the point of specifying a version in `apk add`.

TheFrenchGhosty: We can *technically* just install Crystal from a tarball, or just COPY from an existing image.

Otherwise, why not just use: https://hub.docker.com/r/crystallang/crystal and use `1.4.0-alpine`

unixfox: > We can _technically_ just install Crystal from a tarball, or just COPY from an existing image.
> 
> Otherwise, why not just use: [hub.docker.com/r/crystallang/crystal](https://hub.docker.com/r/crystallang/crystal) and use `1.4.0-alpine`

This is the ARM dockerfile, not the intel x86.

TheFrenchGhosty: >This is the ARM dockerfile, not the intel x86 one.

Yeah I just noticed, might bad. We should ask them to make an arm image.

TheFrenchGhosty (approved):

SamantazFox (approved):

Issue #3043, opened 2022-04-18
[Bug] Regression - 0:00 time-code replaced with the video URL


TheFrenchGhosty:
**Describe the bug**

Surely related to https://github.com/iv-org/invidious/issues/3038

Since a recent PR (maybe https://github.com/iv-org/invidious/pull/2936), all time-code of "0:00" are replaced with the URL of the video on Invidious (shown with in text under youtube.com, ie: [youtube.com](https://invidious.io)) (in both the description AND the comments)

**Steps to Reproduce**

Find any video with time-codes in the description, 0:00 will be replaced by the video URL

**Logs**

N/A

**Screenshots**

Invidious:

![invidious](https://user-images.githubusercontent.com/47571719/163836679-0782d874-820f-490c-815e-cd5ca047e0dc.png)

YouTube:

![youtube](https://user-images.githubusercontent.com/47571719/163836702-a47a5eb9-9029-4212-9142-f003a4465918.png)

**Additional context**

Example used in the screenshots: https://redirect.invidious.io/watch?v=TEygSwHWhfA

Note: there was a bug before (I'm not sure it was ever reported) that made the page reload when clicking on 0:00

MathiusD:
Ah, maybe it's related to my PR, I'm looking at this right now

MathiusD:
Damn I forgot to add the 'Fix #XXX' in my PR, logically it should resolve current issue

PR #3044, opened 2022-04-19
Fix regression related of timestamp 0:00

MathiusD:
Purpose of this PR is submit fix for regression related of timestamp 0:00 described here : #3043


IronMaltese:
A very complicated pull request indeed, this needs a lot of further review and testing🦭

SamantazFox:
Mmmh, I'll have to test that! I'm not sure if that will break links to other videos in comments.

SamantazFox (approved):

PR #3065, opened 2022-04-30
Fix iOS 3 buttons separated lines

138138138:
On iOS 15,
before:
before
after:
after


SamantazFox:

PR #3092, opened 2022-05-09
Remove dislikes icon

unixfox:
Change only on the frontend side for the moment.

### DIFF
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -173,7 +173,7 @@ we're going to need to do it here in order to allow for translations.
 
             <p id="views"><i class="icon ion-ios-eye"></i> <%= number_with_separator(video.views) %></p>
             <p id="likes"><i class="icon ion-ios-thumbs-up"></i> <%= number_with_separator(video.likes) %></p>
-            <p id="dislikes"><i class="icon ion-ios-thumbs-down"></i> <%= number_with_separator(video.dislikes) %></p>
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
The plan was to only hide it from the frontend so that addons could "restore" the frontend code "easily".

unixfox:
Addons can inject code into the page too.

I don't see the point of keeping it in the code. Moreover accessibility wise it's not great to still have the text and icon in the code but visually hiden.

TheFrenchGhosty:
It was @SamantazFox plan, so I'll let her answer.

>Moreover accessibility wise it's not great to still have the text and icon in the code but visually hiden.

It was planed to be commented out in the HTML.

unixfox:
> > 
> 
> 
> 
> > Addons can inject HTML code into the page too.
> > 
> > I don't see the point of keeping it in the HTML page.
> > 
> > Moreover accessibility wise it's not great to still have the text and icon in the html code but visually hidden.
> 
> My idea was to only keep `<p id="dislikes"></p>`, so that add-ons can select the HTML element based using its id and fill the content.

Do I do that then?

TheFrenchGhosty:
>Do I do that then?

Yes

unixfox:
Done, ready to review.

unixfox:
I merged the change because we all agreed on keeping to the solution in https://github.com/iv-org/invidious/pull/3092#issuecomment-1122790432

PR #3112, opened 2022-05-24
Change README.md badge from 'Mastodon' to 'Fediverse'

Froggo8311:
Mastodon is one of multiple softwares that compose the Fediverse ("Federeated Universe"). Some of the most popular softwares include Misskey, Plemora, PeerTube, and Pixelfed, among others. 

As each instance (server) integrates using ActivityPub, any one of these softwares can be used to follow users or reply to posts(/toots/notes) on any instance.

Most people seem to not realize that Mastodon is different from the "umbrella term" Fediverse. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@
   </a>
   <br>
   <a rel="me" href="https://social.tchncs.de/@invidious">
-  <img alt="Mastodon: @invidious@social.tchncs.de" src="https://img.shields.io/badge/Mastodon-%40invidious%40social.tchncs.de-darkgreen">
+  <img alt="Mastodon: @invidious@social.tchncs.de" src="https://img.shields.io/badge/Fediverse-%40invidious%40social.tchncs.de-darkgreen">
   </a>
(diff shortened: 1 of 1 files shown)


IronMaltese:
But the link directs to their mastodon, isn't it better that they're being more specific?

Froggo8311:
@IronMaltese Although the invidious account is on a mastodon instance, interactions are still able to be conducted from the other parts of the Fediverse. This just displays the more broad interaction between the account and others. Perhaps maybe it could be rewritten to state as `Fediverse (Mastodon)` in the readme, but that could be unnecessary.

TheFrenchGhosty:
>Most people seem to not realize that Mastodon is different from the "umbrella term" Fediverse. :)

I did realize that it was different, I just used the name of the software we're using.

>Although the invidious account is on a mastodon instance, interactions are still able to be conducted from the other parts of the Fediverse.

That's indeed a good point.

Thanks for the PR.

PR #3128, opened 2022-05-29
Add Yattee to README

arekf:
Hi, that'd be great if [Yattee](https://github.com/yattee/yattee) could be featured here, it utilizes Invidious APIs (accounts, playlists, browsing and search)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ Weblate also allows you to log-in with major SSO providers like Github, Gitlab, 
 - [WatchTube](https://github.com/WatchTubeTeam/WatchTube): Powerful YouTube client for Apple Watch.
 - [Yattee](https://github.com/yattee/yattee): Alternative YouTube frontend for iPhone, iPad, Mac and Apple TV.
 
+
 ## Liability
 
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty (approved):

TheFrenchGhosty:
Good job.

unixfox:
What's more difficult to implement in the comments and subtitles API of invidious than in the Piped API?

I was just wondering that when looking at the `Availability` table: https://github.com/yattee/yattee#availability

arekf:
> What's more difficult to implement in the comments and subtitles API of invidious than in the Piped API?
> 
> I was just wondering that when looking at the `Availability` table: https://github.com/yattee/yattee#availability

There was no particular technical reason for the comments, it's just that when I implemented this module I was having issues with getting comments from my private Invidious instance and then I never had a chance to work on it again after (and there's a lot of other stuff pending to do, keeping app like this for three different platforms is quite challenging for somebody that's not really an Apple developer 😅)

For the subtitles it is available for Piped because it gives HLS stream where it is included and it required no further implementation, system player can pick it up automatically.

PR #3140, opened 2022-06-05
CSS. Small IE11 fixes

AHOHNMYC:
Four small changes. Three on the screenshot, fourth is placeholder default color (taken from Chrome).

### DIFF
--- a/assets/css/search.css
+++ b/assets/css/search.css
@@ -68,7 +68,7 @@ fieldset, legend {
 .filter-options label { margin: 0 10px; }
 
 
-#filters-apply { text-align: end; }
+#filters-apply { text-align: right; }
 
 /* Error message */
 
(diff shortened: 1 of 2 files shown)


assets/css/search.css
```diff
@@ -68,7 +68,7 @@ fieldset, legend {
 .filter-options label { margin: 0 10px; }
 
 
-#filters-apply { text-align: end; }
+#filters-apply { text-align: right; }
```
SamantazFox, line 71: Please keep `text-align: end;`, as it is required for Right-to-Left languages. I guess that you changed this because IE11 doesn't support `end`? if that's the case, you can do that:

```css
#filters-apply {
	text-align: right; /* IE11 only */
	text-align: end;   /* Override for compatible browsers */
}

AHOHNMYC: @SamantazFox FYI, Invidious in Hebrew and Arabic (most obvious RTL languages) don't change elements order :P

SamantazFox: not *yet* ^^ But it's something I need to fix!

unixfox (approved):

PR #3157, opened 2022-06-22
Fix captions

138138138:
Captions should automatically show according to preferences.


138138138:
Some debug messages are shown as captions on DASH mode. I need to modify the code to make both non-DASH and DASH mode show captions correctly.

SamantazFox (approved):
Looks good to me! I'll try to deploy that on the test instance ASAP.

SamantazFox:
Thanks again for your contributions :)

PR #3158, opened 2022-06-22
Player MobileUi fast forward/backward rate

138138138:
The fast forward/backward seconds will be adjusted according to playback rate (same as YouTube app behavior).
5 seconds is used when the playback rate is 1x. Previously it was 10 seconds. I believe most of the users watch videos at 2x, so the change will not be obvious.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -145,7 +145,7 @@ function isMobile() {
 }
 
 if (isMobile()) {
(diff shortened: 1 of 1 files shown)


SamantazFox (approved):

SamantazFox:
Thank you for contributing to invidious :)

Issue #3074, opened 2022-05-01
[Bug] DASH playback results in very low Audio bitrate regardless of Video quality selected


LennyPenny:
**Describe the bug**
When using the DASH playback mode instead of the normal direct stream the Audio bitrate is very low (20kbit/s in my testing vs the normal 192kbit/s). I'm not sure how the dash playback internals of invidious work, but I guess audio quality is selected separately from video quality and google changed the default or something. 

**Steps to Reproduce**
(random video and instance for illustration purposes, should work with any recent invidious)
1. Watch [this video](https://yewtu.be/watch?v=urevinis_PU) with DASH playback off (can be verified if you only see "medium"/"hd720" quality options in the player)
2. Download the video from network tab and save somewhere
3. Go to the preferences page and change `Preferred video quality: ` to "DASH (adaptive quality)" - the preferred DASH quality doesn't matter, can set it to 1080p for consistency
4. Watch the same video and verify dash is on (1080p, 720p, 420p, etc options available)
5. For DASH playback audio and video streams are separate, so you will need to hunt the network tab a bit for the audio only stream (it's still in an mp4 container)
6. Download that audio stream and save it somewhere
7. Use the `ffprobe` command line utility from `ffmpeg` to compare the audio bitrates (or any tool of your choosing that does the same)

**Screenshots**
![image](https://user-images.githubusercontent.com/4027243/166148106-2550a155-d71c-41e9-a93b-0ebdbe7b780d.png)

Here I downloaded the mp4 of a video from the direct googlevideo stream (left) and the DASH playback audio stream from invidious. Note that left is a combined video/audio stream and right is audio only, so be sure to look at the audio stream on the right. 

**Additional context**
I tried this on firefox and chrome and macos and linux and the issue seems to be consistent.

SamantazFox:
I'm not getting the same results. Something I have noticed is that the reported audio stream bitrate is broken: In the audio files I downloaded it shows as 0 kpbs and 1 kbps respectively (see below).
 
Note: All streams were downloaded from the widget on invidious.

---

For the MP4 audio+video stream (a.k.a "HD720"):
```
$ ffprobe -hide_banner urevinis_PU-Concerning\ the\ Future\ of\ LGR.mp4
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from 'urevinis_PU-Concerning the Future of LGR.mp4':
  Metadata:
    major_brand     : mp42
    minor_version   : 0
    compatible_brands: isommp42
[...]

PR #3161, opened 2022-06-22
Select audio quality according to video quality

138138138:
Currently 64k m4a is selected as default. This sounds very bad to users with enough bandwidth.
The code assigns medium, small, dash 240p, dash 144p, dash worst to 64k m4a. Others will be 128k m4a.

### DIFF
--- a/src/invidious/views/components/player.ecr
+++ b/src/invidious/views/components/player.ecr
@@ -14,7 +14,10 @@
                 bitrate = fmt["bitrate"]
                 mimetype = HTML.escape(fmt["mimeType"].as_s)
 
-                selected = i == 0 ? true : false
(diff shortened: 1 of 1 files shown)


unixfox:
This might fix https://github.com/iv-org/invidious/issues/3074. Could you confirm?

julianfairfax:
Does audio quality decrease with video quality on YouTube? Seems odd to me but maybe I never noticed. If this does fix #3074 though it'd be amazing cause I'd finally be able to use Invidious without having to turn dash on when videos don't have non-dash, thus decreasing the audio quality, and then turning it off again to get proper audio quality, multiple times a day.

138138138:
This patch only affects audio only mode. I will look at fixing DASH soon.

138138138:
Some work still need to be done. I will release a fix for both audio mode and DASH later.

Issue #2513, opened 2021-10-14
[Bug] Difficult to understand qualitiy selection in listen mode


offa:
**Describe the bug**
The quality selection menu in listen mode shows some k-values, but it's not obvious what they mean.

**Steps to Reproduce**

1. Choose an Invidious instance and select a video
2. Switch to listen mode _(Headphones icon)_
3. Open quality selector menu _(gear wheel)_
4. Some thousand-k values are shown 

**Logs**

_– none –_

**Screenshots**

![iv](https://user-images.githubusercontent.com/8887756/137382800-8a6fb803-e19d-4eda-a206-2983dcc76264.png)

**Additional context**

- Browser: Firefox
- OS: Linux
- Proxy videos enabled
- Tested on several instances

unixfox:
> Thanks for notifying this! I'll add that to the TODO list of #2377, whose aim is to make the different user preferences (quality selectors, checkboxes, labels) much clearer and friendly.

The qualities offered are buggy tho, it shouldn't have 100000k and so on. Just normal audio qualities.

TheFrenchGhosty:
A 10$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 10$ (in BTC) from the Invidious project.

More details: #1898

offa:
Issue still exists.

offa:
Bug still exists.

dzinesco:
🐕 **Dog the Bounty Hunter** has claimed this bounty!

PR: https://github.com/iv-org/invidious/pull/5554

PR #3162, opened 2022-06-22
Fix low audio bitrate in DASH mode and audio only mode

138138138:
Fix https://github.com/iv-org/invidious/issues/3074
Also default highest quality m4a on audio only mode

### DIFF
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -61,7 +61,18 @@ module Invidious::Routes::API::Manifest
             next if mime_streams.empty?
 
             xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true) do
+              # ignore the 64k m4a stream, only consider the 128k m4a stream
+              best_m4a_stream = mime_streams[0]
+              best_m4a_stream_bitrate = 0
               mime_streams.each do |fmt|
+                bandwidth = fmt["bitrate"].as_i
+                if (bandwidth > best_m4a_stream_bitrate)
+                  best_m4a_stream_bitrate = bandwidth
+                  best_m4a_stream = fmt
+                end
+              end
+      
+              [best_m4a_stream].each do |fmt|
                 # OTF streams aren't supported yet (See https://github.com/TeamNewPipe/NewPipe/issues/2415)
                 next if !(fmt.has_key?("indexRange") && fmt.has_key?("initRange"))
 
--- a/src/invidious/views/components/player.ecr
+++ b/src/invidious/views/components/player.ecr
@@ -7,14 +7,25 @@
         <source src="<%= URI.parse(hlsvp).request_target %><% if params.local %>?local=true<% end %>" type="application/x-mpegURL" label="livestream">
     <% else %>
         <% if params.listen %>
-            <% audio_streams.each_with_index do |fmt, i|
+            <% # ignore the 64k m4a stream, only consider the 128k m4a stream
+               best_m4a_stream_index = 0
+               best_m4a_stream_bitrate = 0
+               audio_streams.each_with_index do |fmt, i|
+                 bandwidth = fmt["bitrate"].as_i
+                 if (fmt["mimeType"].as_s.starts_with?("audio/mp4") && bandwidth > best_m4a_stream_bitrate)
+                   best_m4a_stream_bitrate = bandwidth
+                   best_m4a_stream_index = i
+                 end
+               end
+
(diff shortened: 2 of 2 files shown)


138138138:
The code can be successfully built in docker, but the Github checks failed.
This may be some Crystal syntax error. I cannot figure out the issues. Please help me to check the code.

SamantazFox:
The problem is that it removes entirely lower quality audio... That's not ideal for users with limited bandwidth.
I'd prefer if you could fix the audio stream selection on the client side (JS).

138138138:
When playing DASH videos, an audio button is shown to change the quality of the stream.
Screenshot_20220625_195326

Now everything defaults to the best quality m4a.
Future requests on selecting audio quality based on video quality, or having a separate audio quality preference, or enabling webm, are welcome to be discussed.
However, I think this patch needs to be merged now, since many users are suffering from low audio quality in DASH mode.

unixfox:

julianfairfax:
Thank you so much for fixing this!

LennyPenny:
could you publish a docker image of this somehwere?

unixfox:
> could you publish a docker image of this somehwere?

There is my docker image: https://hub.docker.com/r/unixfox/invidious-custom with all the patches: https://github.com/unixfox/invidious-custom/tree/master/patches.
There is a breaking change in the patches, especially the patch from this pull request: https://github.com/iv-org/invidious/pull/2469

SamantazFox:
While you're at it, is it possible to fix the audio bitrate diplayed? See issue #2513

unixfox:
> While you're at it, is it possible to fix the audio bitrate diplayed? See issue #2513

I think this should be tackled into another pull request, let's merge this one which fixes a wide annoying bug.

LennyPenny (approved):

src/invidious/routes/api/manifest.cr
```diff
@@ -60,8 +60,8 @@ module Invidious::Routes::API::Manifest
             mime_streams = audio_streams.select { |stream| stream["mimeType"].as_s.starts_with? mime_type }
             next if mime_streams.empty?
 
-            xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true) do
-              mime_streams.each do |fmt|
+            mime_streams.each do |fmt|
+              xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true, label: fmt["bitrate"].to_s + "k") do
                 # OTF streams aren't supported yet (See https://github.com/TeamNewPipe/NewPipe/issues/2415)
                 next if !(fmt.has_key?("indexRange") && fmt.has_key?("initRange"))
```
SamantazFox, line 66: That part need to be moved right after `mime_streams.each do |fmt|` so that OTF streams don't get an entry.

SamantazFox (changes requested):
In addition two the two changes below, I'm really wondering about having two `AdaptationSet` for the same audio. Doesn't that break the standard?

138138138:
The above two changes are implemented. 
 video.js 
Regarding to AdaptationSet, theoretically audios with same content should be put into one AdaptationSet. However, most players cannot handle auto quality switching, and cannot provide a quality selector. By separating into two AdaptationSet, video.js will treat them as different languages audio tracks, which shows an audio track selector.
youtube.com also defaults to highest quality audio and does not change. Lower qualities are for YouTube Music.

src/invidious/routes/api/manifest.cr
```diff
@@ -62,16 +62,18 @@ module Invidious::Routes::API::Manifest
             mime_streams = audio_streams.select { |stream| stream["mimeType"].as_s.starts_with? mime_type }
             next if mime_streams.empty?
 
-            xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true) do
-              mime_streams.each do |fmt|
-                # OTF streams aren't supported yet (See https://github.com/TeamNewPipe/NewPipe/issues/2415)
-                next if !(fmt.has_key?("indexRange") && fmt.has_key?("initRange"))
+            mime_streams.each do |fmt|
+              # OTF streams aren't supported yet (See https://github.com/TeamNewPipe/NewPipe/issues/2415)
+              next if !(fmt.has_key?("indexRange") && fmt.has_key?("initRange"))
```
SamantazFox, line 68:
```suggestion

              # Different representations of the same audio should be groupped into one AdaptationSet.
              # However, most players don't support auto quality switching, so we have to trick them
              # into providing a quality selector.
```

SamantazFox:

Issue #705, opened 2019-08-18
LDAP support


Perflyst:
Add support for auth against LDAP

omarroth:
This seems quite out-of-scope to me. Is there a reason for preferring LDAP in particular? Would you mind providing an example use-case?

Perflyst:
People don't have to create new accounts on each service. One account, one login on all services.
As an admin you have the users central. Other apps like nextcloud, tinytinyrss or mastodon support LDAP too

PoGo606:
Same here, I try to let my users use the same account for all the provided services. LDAP is a great help for that ; at least for the authentication part : same login/pass accross all services.

HumanG33k:
Agree too, several organisations want to migrate to their self-hosted services (nextcloud, email, website) and ldap is a key tech for that and with tools like fusiondirectory it’s "easy" now.

PipeItToDevNull:
My use case is that I am a single user, but I host a lot of stuff and want to have one source for my authentication and credentials.

Perflyst:
still relevant

joseph-cheung:
still relevant!

joseph-cheung:
still relevant.
PLZ support LDAP.

unixfox:
Please stop spamming this issue, if you really need LDAP support then you are free to submit a pull request with the code with the LDAP support.

joseph-cheung:
> Please stop spamming this issue, if you really need LDAP support then you are free to submit a pull request with the code containing the LDAP support.

I apologize, but you are also abusing power. 
If you think this is a spam, it is my fault to communicate with you bureaucracy.

unixfox:
> > Please stop spamming this issue, if you really need LDAP support then you are free to submit a pull request with the code containing the LDAP support.
> 
> I apologize, but you are also abusing power.
> If you think this is a spam, it is my fault to communicate with you bureaucracy.

Apologizes accepted, but I'm just moderating this repository because we want to keep the issues clean from useless comments.

The developer that will implement your feature want to read useful comments, he doesn't want to spend extra time filtering the good comments from the bad comments. The time of open source developers is precious because they aren't getting paid, they are just volunteers.

Please keep in mind that we don't own you anything, we are spending our time on things that matter to us, so maybe no one is interested in LDAP and that's why it takes a long time to have this feature.

No need to reply to comment, you want just add a thumbs up if you agree.

yagarea:
Is there any progress on this issue ?

[...]

PR #3164, opened 2022-06-25
Implemented OAuth Authentication

choelzl:
Added Option for forced authentication back-end (overriding the query parameter)
Added Options to configure OAuth client (API endpoints, client_id, client_secret,...)

Related Issue: #705

### DIFF
--- a/src/invidious/helpers/oauth.cr
+++ b/src/invidious/helpers/oauth.cr
@@ -1,0 +1,55 @@
+require "oauth2"
+
+module Invidious::OAuthHelper
+  extend self
+
+  def get_provider(key)
+    if provider = CONFIG.oauth[key]?
+      provider
+    else
+      raise Exception.new("Invalid OAuth Endpoint: " + key)
+    end
+  end
+
+  def get(key)
+    if HOST_URL == ""
+      raise Exception.new("Missing domain and port configuration")
+    end
+    provider = get_provider(key)
+    redirect_uri = "#{HOST_URL}/login/oauth/#{key}"
+    OAuth2::Client.new(
+      provider.host,
+      provider.client_id,
+      provider.client_secret,
+      authorize_uri: provider.auth_uri,
+      token_uri: provider.token_uri,
+      redirect_uri: redirect_uri
+    )
+  end
+
+  def get_uri_host_pair(host, uri)
+    if (uri.starts_with?("https://"))
+      res = uri.gsub(/https*\:\/\//, "").split('/', 2)
+      [res[0], "/" + res[1]]
+    else
+      [host, uri]
+    end
+  end
+
+  def get_info(key, token)
+    provider = self.get_provider(key)
+    uri_host_pair = self.get_uri_host_pair(provider.host, provider.info_uri)
+    LOGGER.info(uri_host_pair[0] + " " + uri_host_pair[1])
+    client = HTTP::Client.new(uri_host_pair[0], tls: true)
+    token.authenticate(client)
+    response = client.get uri_host_pair[1]
+    client.close
+    LOGGER.info(response.body)
+    response.body
+  end
+
+  def info_field(key, token)
+    info = JSON.parse(self.get_info(key, token))
+    info[self.get_provider(key).field].as_s?
+  end
+end
--- a/src/invidious/helpers/oauth.cr
+++ b/src/invidious/helpers/oauth.cr
@@ -11,7 +11,7 @@ module Invidious::OAuthHelper
     end
   end
 
-  def get(key)
+  def make_client(key)
     if HOST_URL == ""
       raise Exception.new("Missing domain and port configuration")
     end
@@ -27,24 +27,22 @@ module Invidious::OAuthHelper
     )
   end
 
-  def get_uri_host_pair(host, uri)
-    if (uri.starts_with?("https://"))
-      res = uri.gsub(/https*\:\/\//, "").split('/', 2)
-      [res[0], "/" + res[1]]
+  def get_uri_host_pair(host, url)
+    if (url.starts_with?(/https*\:\/\//))
+      uri = URI.parse url
+      [uri.host || host, uri.path || "/"]
     else
-      [host, uri]
+      [host, url]
     end
   end
 
   def get_info(key, token)
     provider = self.get_provider(key)
     uri_host_pair = self.get_uri_host_pair(provider.host, provider.info_uri)
-    LOGGER.info(uri_host_pair[0] + " " + uri_host_pair[1])
     client = HTTP::Client.new(uri_host_pair[0], tls: true)
     token.authenticate(client)
     response = client.get uri_host_pair[1]
     client.close
-    LOGGER.info(response.body)
     response.body
   end
 
--- a/src/invidious/routes/login.cr
+++ b/src/invidious/routes/login.cr
@@ -235,4 +235,49 @@ module Invidious::Routes::Login
 
     env.redirect referer
   end
+
+  def self.user_flow_existing(env, email)
+    sid = Base64.urlsafe_encode(Random::Secure.random_bytes(32))
+    Invidious::Database::SessionIDs.insert(sid, email)
+    env.response.cookies["SID"] = Invidious::User::Cookies.sid(CONFIG.domain, sid)
+
+    # Since this user has already registered, we don't want to overwrite their preferences
+    if env.request.cookies["PREFS"]?
+      cookie = env.request.cookies["PREFS"]
+      cookie.expires = Time.utc(1990, 1, 1)
+      env.response.cookies << cookie
+    end
+  end
+
+  def self.user_flow_new(env, email, password)
+    sid = Base64.urlsafe_encode(Random::Secure.random_bytes(32))
+    if password
+      user, sid = create_user(sid, email, password)
+    else
+      user, sid = create_user(sid, email)
+    end
+
+    if language_header = env.request.headers["Accept-Language"]?
+      if language = ANG.language_negotiator.best(language_header, LOCALES.keys)
+        user.preferences.locale = language.header
+      end
+    end
+
+    Invidious::Database::Users.insert(user)
+    Invidious::Database::SessionIDs.insert(sid, email)
+
+    view_name = "subscriptions_#{sha256(user.email)}"
+    PG_DB.exec("CREATE MATERIALIZED VIEW #{view_name} AS #{MATERIALIZED_VIEW_SQL.call(user.email)}")
+
+    env.response.cookies["SID"] = Invidious::User::Cookies.sid(CONFIG.domain, sid)
+
+    if env.request.cookies["PREFS"]?
+      user.preferences = env.get("preferences").as(Preferences)
+      Invidious::Database::Users.update_preferences(user)
+
+      cookie = env.request.cookies["PREFS"]
+      cookie.expires = Time.utc(1990, 1, 1)
+      env.response.cookies << cookie
+    end
+  end
 end
--- a/src/invidious/routes/login.cr
+++ b/src/invidious/routes/login.cr
@@ -3,10 +3,9 @@
 module Invidious::Routes::Login
   def self.login_page(env)
     locale = env.get("preferences").as(Preferences).locale
+    referer = get_referer(env, "/feed/subscriptions")
 
     user = env.get? "user"
-
-    referer = get_referer(env, "/feed/subscriptions")
 
     return env.redirect referer if user
 
@@ -19,17 +18,13 @@ module Invidious::Routes::Login
     captcha = nil
 
     account_type = env.params.query["type"]?
-    account_type ||= "invidious"
+    account_type ||= ""
 
-    if CONFIG.auth_type.find(&.== account_type).nil?
-      if CONFIG.auth_type.size == 0
-        account_type = "invidious"
-      else
-        account_type = CONFIG.auth_type[0]
-      end
+    if CONFIG.auth_type.size == 0
+      return error_template(401, "No authentication backend enabled.")
+    elsif CONFIG.auth_type.find(&.== account_type).nil? && CONFIG.auth_type.size == 1
+      account_type = CONFIG.auth_type[0]
     end
-
-    oauth = CONFIG.auth_type.find(&.== "oauth") && (CONFIG.oauth.size > 0)
 
     captcha_type = env.params.query["captcha"]?
     captcha_type ||= "image"
@@ -39,36 +34,36 @@ module Invidious::Routes::Login
 
   def self.login_oauth(env)
     locale = env.get("preferences").as(Preferences).locale
-
     referer = get_referer(env, "/feed/subscriptions")
 
     authorization_code = env.params.query["code"]?
     provider_k = env.params.url["provider"]
-    if authorization_code
-      begin
-        token = OAuthHelper.get(provider_k).get_access_token_using_authorization_code(authorization_code)
-        email = OAuthHelper.info_field(provider_k, token)
 
-        if email
-          user = Invidious::Database::Users.select(email: email)
-          if user
-            user_flow_existing(env, email)
-          else
-            user_flow_new(env, email, nil)
-          end
-        end
-      rescue ex
-        return error_template(500, "Internal Error" + (ex.message || ""))
-      end
-    else
+    if authorization_code.nil?
       return error_template(403, "Missing Authorization Code")
+    end
+    begin
+      token = OAuthHelper.make_client(provider_k).get_access_token_using_authorization_code(authorization_code)
+
+      if email = OAuthHelper.info_field(provider_k, token)
+        if user = Invidious::Database::Users.select(email: email)
+          if CONFIG.auth_enforce_source && user.password != ("oauth:" + provider_k)
+            return error_template(401, "Wrong provider")
+          else
+            user_flow_existing(env, email)
+          end
+        else
+          user_flow_new(env, email, nil, "oauth:" + provider_k)
+        end
+      end
+    rescue ex
+      return error_template(500, "Internal Error")
     end
     env.redirect referer
   end
 
   def self.login(env)
     locale = env.get("preferences").as(Preferences).locale
-
     referer = get_referer(env, "/feed/subscriptions")
 
     if !CONFIG.login_enabled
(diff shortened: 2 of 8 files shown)


SamantazFox:
Hello! Thanks for the PR :)

I have some questions:
 - I'm not sure why you chose to implement OAuth alone. Why not support [openID connect](https://github.com/straight-shoota/openid_connect) instead?
 - I don't understand the comment "Force Authentication Backend" in the config file. There is no config option underneath. Is it meant for `auth_type`?
 - Have you plans to implement an interface (dedicated webpage) for OAuth/OpenID connection?

choelzl:
* The implementation works with OpenID backends. The main reason I used the OAuth library is that there is not [standard](https://crystal-lang.org/api/master/) or [certified](https://openid.net/developers/certified/) OpenID library for crystal yet.
* Yes what i meant is the `auth_type` parameter. This way only one back-end can be used if desired (forcing users to use the OAuth for example).
* What do you mean with `interface` / `dedicated webpage` ? A way to link existing accounts to OAuth ? If users rely on the same email for Invidious and OAuth, there is no need for it. Otherwise it would mean  either changing account emails, duplicating account or allowing multiple emails. I have no plans to add this for now. The main idea is to provide a authentication source similar to the google option.

SamantazFox:
> The implementation works with OpenID backends.

Ah ok, that's good to know! I asked because I didn't openID connect being mentionned ^^

> The main reason I used the OAuth library is that there is not [standard](https://crystal-lang.org/api/master/) or [certified](https://openid.net/developers/certified/) OpenID library for crystal yet.

While it's not certified by the openID connect organization (probably because Crystal is too small or the author never bothered with the paperwork), there is the [openid_connect shard](https://github.com/straight-shoota/openid_connect) that was created by [straight-shoota](https://github.com/straight-shoota), one of the core developers of Crystal.

>  What do you mean with `interface` / `dedicated webpage` ? A way to link existing accounts to OAuth ?

You [added a simple input field](https://github.com/iv-org/invidious/pull/3164/files#diff-5822d9bc4d51e77a2d71c83f550e3aa58b51ef504022794b4b3d3b5417a3c809R46-R51) similar to the "log in with google" one, but the only way to reach it is to edit the URL. There is no link pointing to that authentication method.

In addition, the OAuth/openID connect server is defined server-wide. IMO, that would be nicer if there was a list of identity providers that the user can choose from (, and that the server config would add other providers.

choelzl:
- I could use that library. Due to the low usage/stars and not being included in the default library makes me just a bit more insecure about using it. Is it maintained and mature ?
- I can add a link to the page login page (if oauth is configured).

The way I initially designed it was based on my desired use case, as a replacement authentication back-end with a single endpoint. I will make the two changes to support more potential use cases.

SamantazFox:
I need to read a bit more to understand OAuth before adding anything here!

In the meantime, can you please fix the CI? thanks :)

weblate3:
こんにちは
githubの初心者です。
質問ですがURLなどはどこから見ることができますか？
教えてください。お願いします。

choelzl:
@SamantazFox CI-fix and requested changes where made.

src/invidious/helpers/oauth.cr
```diff
@@ -0,0 +1,47 @@
+def oauth_get(idx)
```
SamantazFox, line 1: Coud you put everything in this file under the `Invidious::OAuthHelper` module, and remove the `oauth_` prefix from variables? (don't forget to `extend self` on the module ^^)

src/invidious/routes/login.cr
```diff
@@ -42,10 +115,20 @@ module Invidious::Routes::Login
     # https://stackoverflow.com/a/574698
     email = env.params.body["email"]?.try &.downcase.byte_slice(0, 254)
     password = env.params.body["password"]?
+    oauth = CONFIG.auth_type.find { |i| i == "oauth" } && (CONFIG.oauth.size > 0)
```
SamantazFox, line 118: Prefer `.includes?` to that heavy `.find` syntax :)

src/invidious/routes/login.cr
```diff
@@ -42,10 +115,20 @@ module Invidious::Routes::Login
...
+    oauth_providers = CONFIG.oauth
 
     account_type = env.params.query["type"]?
     account_type ||= "invidious"
 
+    if CONFIG.auth_type.size == 0
+      return error_template(401, "No authentication backend enabled.")
+    end
+
+    if CONFIG.auth_type.find { |i| i == account_type } == nil
```
SamantazFox, line 128: Similarly, prefer `x.nil?` over `x == nil`

SamantazFox (changes requested):
Sorry for the time it took me to review!

SamantazFox:
Thanks for making the requested changes!
I'm sorry in advance if it takes quite some time to merge. I need to educate myself a lot on authentication and authorization, as well as how to deploy the associated services in order to test the features!

SamantazFox:
Sorry, I haven't much worked on invidious in December/early January.

I've deployed your PR to the test instance at https://test.invidious.io, where I've set Github as a provider:
```yaml
auth_type: ["invidious", "oauth"]
oauth:
  github:
    host: github.com
    auth_uri: /login/oauth/authorize
    token_uri: /login/oauth/access_token
    info_uri: /login/oauth/userinfo
    client_id: redacted
    client_secret: redacted
```

I can't log in with Oauth, and I'm getting the following error:

Title: `Invalid OAuth Endpoint: github (Exception)`
Date: `2023-01-15T22:47:55Z`
Route: `/login?referer=%2Ffeed%2Fsubscriptions&type=oauth`
Version: `2023.01.10-05258d56 @ master`

<details>
<summary>Backtrace</summary>
<p>
   
```
Invalid OAuth Endpoint: github (Exception)
  from invidious.git/src/invidious/helpers/oauth.cr:13:9 in 'get'
  from invidious.git/src/invidious/routes/login.cr:394:20 in 'login'
  from invidious.git/src/invidious/routing.cr:59:20 in '->'
  from invidious.git/lib/kemal/src/kemal/route.cr:12:26 in '->'
  from invidious.git/src/invidious/helpers/handlers.cr:30:37 in 'process_request'
  from invidious.git/lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:212:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:94:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:145:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:70:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/ext/kemal_static_file_handler.cr:112:11 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/lib/kemal/src/kemal/exception_handler.cr:8:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/logger.cr:17:35 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/lib/kemal/src/kemal/init_handler.cr:12:7 in 'call'
  from /usr/share/crystal/src/http/server/request_processor.cr:51:11 in 'process'
  from /usr/share/crystal/src/http/server.cr:515:5 in 'handle_client'
  from /usr/share/crystal/src/http/server.cr:468:13 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:98:34 in '->'
  from ???
```
</p>
</details>

src/invidious/helpers/oauth.cr
```diff
@@ -0,0 +1,41 @@
...
+    if CONFIG.oauth[key]?
+      raise Exception.new("Invalid OAuth Endpoint: " + key)
+    end
+    v = CONFIG.oauth[key]
+    redirect_uri = HOST_URL + "/login/oauth/" + key
+
+    OAuth2::Client.new(v.host,
+      v.client_id, v.client_secret,
+      authorize_uri: v.auth_uri, token_uri: v.token_uri,
+      redirect_uri: redirect_uri)
```
SamantazFox, line 19:
```suggestion
    if provider = CONFIG.oauth[key]?
      redirect_uri = "#{HOST_URL}/login/oauth/#{key}"

      OAuth2::Client.new(
        provider.host,
        provider.client_id,
        provider.client_secret,
        authorize_uri: provider.auth_uri,
        token_uri: provider.token_uri,
        redirect_uri: redirect_uri
      )
    else
      raise Exception.new("Invalid OAuth Endpoint: " + key)
    end
```

SamantazFox: The check on `CONFIG.oauth[key]?` is inverted, so when the provider was set in the config, it'd always raise an error. I changed that in the suggestion above, with some small style changes.

I've also used string interpolation (`#{}`) for `redirect_uri`. This is a minor performance change. For reference, see https://crystal-lang.org/reference/1.7/guides/performance.html#use-string-interpolation-instead-of-concatenation

SamantazFox (changes requested):

SamantazFox:
After doing the change mentioned earlier, I am properly redirected to github where I can login.
But then I get "Invalid Authorization Code".

SamantazFox:
bump ?

choelzl:
have had a relatively busy period, will work on fixing and testing correctness in the coming weeks

SamantazFox:
> have had a relatively busy period, will work on fixing and testing correctness in the coming weeks

Sure, no problem ^^ take care!

choelzl:
@SamantazFox I updated the code with your suggested changes, with a minor refactor and a rebase onto master. 

There is just one minor issue where through the oauth procedure the "referer" gets lost. This could be avoided by storing it in a cookie at the start of the login flow.

SamantazFox:
I tried to login using OAuth with the new changes, but the authentication keeps failing.
For the record, I'm using Github as an OAuth provider. 

#### Here are the problems I encountered during testing:

First thing, I was getting HTML from github, so I had to add an `accept` header in `OAuthHelper.info`:
```diff
     if provider = CONFIG.oauth[key]?
       client = HTTP::Client.new(provider.host, tls: true)
       token.authenticate(client)
-      response = client.get provider.info_uri
+      headers = HTTP::Headers{"Accept" => "application/json"}
+      response = client.get(provider.info_uri, headers: headers)
       client.close
       response.body
```

After doing that, I keep getting an `Invalid Authorization Code` error. In the following code, I'd recommend to handle the different parts of the authentication flow separately, rather than having a "catch all" block, that is giving a misleading message to the user if anything fails:
https://github.com/iv-org/invidious/pull/3164/files#diff-34166896357fd68545cc771bc54d6e966f4498c8e3ef7c3e59f6fca5e0e8a71dR51-R65

And in the end, I'm always getting the same JSON response from Github: `{"error":"Not Found"}`
I haven't spent enough time on this to pinpoint the root cause of the problem.

mrbreil:
Maybe LD_AUTH_PROXY_USERNAME_HEADER Automatic Login is a simpler solution.  Everybody who Need oauth use a oauthproxy.

choelzl:
Sorry for the inactivity, had some things to sort. I am back working on this PR. 
I investigated the issues with Google:
- They use different hosts for authorization and token exchange. 
- The crystal oauth2 implementation does not support this. 

I will be adding support for a different host for the authorize step to fix this issue

choelzl:
@SamantazFox I made a google account and tested it, Here you can find the config I used for google 
image

Let me know if you have any comments. Also I am not sure how to best factorize/handle the error handling without making it long (with many repeating begin/rescue)

choelzl:
Github also has the issue of multiple domains in the oauth flow. The config with my changes would look as follows:
image

Github additionally has another problem  ( https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28 )
If the email of the user is not public the  app does not have access to it. 

I see two ways to handle this case:
- Fail the authentication
- Fallback to username/userid/login/...

choelzl:
Finally, for completeness, here is the config for authentik, an open source and self-hosted SSO/OpenID provider:
image

SamantazFox:
Nice, thanks for the update :)

> Sorry for the inactivity, had some things to sort.

Sure, no problem! Life happens, that's normal ^^

> Let me know if you have any comments. Also I am not sure how to best factorize/handle the error handling without making it long (with many repeating begin/rescue)

You could make it so that `auth_uri`, `info_uri` and `token_uri` all accept, well, URIs (domain + path). That's less configs entries, and less confusing too.

> Github additionally has another problem ( https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28 )
> If the email of the user is not public the app does not have access to it.

What unique identifier can we fallback to?

---

Also, you might take a look at https://github.com/crystal-lang/crystal/pull/13321
This was recently fixed in Crystal's stdlib, and might be of interest!

SamantazFox:
Also, can you please rebase your changes?
I've removed the broken "google login" code since the last time ^^

choelzl:
@SamantazFox  I rebased and made so that the different uri's support full paths. Below you can see the config 
image

I added an optional "field" value to the oauth configto modify the used field for a provider (by default "email" but can be set to "profile" for github to use usernames, as seen in the config above)

src/invidious/users.cr
```diff
@@ -3,6 +3,24 @@ require "crypto/bcrypt/password"
...
+    email:             email,
+    preferences:       Preferences.new(CONFIG.default_user_preferences.to_tuple),
+    password:          nil,
+    token:             token,
+    watched:           [] of String,
+    feed_needs_update: true,
+  })
+
+  return user, sid
+end
```
SamantazFox, line 22: The problem I see with an empty password is that you can'tto differentiate between two users with the same name but using two different authentication methods/providers (that could result in an account hijack).

I'd suggest using the password field to store the account type (like `oauth:github.com`), or to add a dedicated database field (and add a migration, which is a much cleaner solution).

SamantazFox (changes requested):

src/invidious/users.cr
choelzl: I see 3 scenarios: 
- User can use multiple services to get into his account (its a feature)
- User wants to hijack any account via oauth as it doesn't check/require password (its bad?). The administrator/host should select oauth providers that do not conflict usernames or where email address ownership is verified.
- User wants to hijack an oauth account via an internal account (safe), the password is nil and he cannot log into it.

The solutions are the follows
- Adding a field to the db: adds complexity and requires migration, might break the first scenario.
- Adding it to the password 
- Adding suffix to usernames to indicate auth_type  ("test@github", "test@internal",...), might also require migrations

I'd favor either the password or the username approach

choelzl:
I added the `auth_enforce_source` option, that is enabled by default. It will prevent users from using other providers (internal, other oauth,...) to login. 
It can be disabled by the host to allow users to access their account via multiple providers if so desired

SamantazFox:
I found this super interesting resource, btw:
https://portswigger.net/web-security/oauth

choelzl:
@SamantazFox can this PR be merged or are more changes needed ?

mentos1386:
@SamantazFox did you maybe managed to look at this PR?

imgurbot12:
@SamantazFox ping. 

would really love to see this feature implemented. best of luck!

imgurbot12:
i get why stale-bots exist, but they really do encourage worthless comments like this one when trying to ensure issues get resolved.

imgurbot12:
@syeopite can we flag this issue with the `exempt-stale` tag please? PR is just waiting for review at this point.

dragonchaser:
Is this still considered to be merged into invidious?

mrbreil:
Is this feature now available?

gregismotion:
> Is this feature now available?

Would love that but it doesn't seem merged.

krim404:
As this PR has been open for 2 years and has undergone multiple reviews, it should be considered either accepted or closed at this point and not be stalled further.

krim404:
@SamantazFox any chance for a review and hopefully merge of the PR including the requested changes done in 2023?

chazwarp923:
Any movement here? @SamantazFox

krim404:
as far as i can see the requested changes have been implemented, maybe its just ignored because the review system did not update correctly?
 
@SamantazFox @syeopite could you verify this please manually?

ikolsteren:
Is there a guide to get this configured?
Looking at the config examples and do not see the example to get it setup

krim404:
you need to compile it yourself. did this for arm64 and amd64 and works well - using it daily.

ikolsteren:
> you need to compile it yourself. did this for arm64 and amd64 and works well - using it daily.

Now I need to figure out how to do - I am running as the docker image and not seeing instructions on a compile "make" build - like the repo etc

ikolsteren:
> > you need to compile it yourself. did this for arm64 and amd64 and works well - using it daily.
> 
> I built a new server and cloned the master with the make and nothing yet Is there a particular release I need to clone to get the options to work or maybe there is an issue in my config - not seeing anything

If I pull in the request then the compile fails

Ejmathewp:
I would also love for this to be merged. Everything is behind Authentik for me and avoiding double login would be great

ikolsteren:
Any help on how to compile?

unixfox:
> Any help on how to compile?  

Please don't go offtopic. Ask on matrix or irc: https://github.com/iv-org/invidious?tab=readme-ov-file#chat-with-us

grapemix:
OIDC/SSO solution is needed. Plz continue on your great work.

unixfox:
Hello,
After discussing with the team.
We are not going to implement oauth in Invidious.
We lack of real free time in the team to maintain the source code and manage the community. Moreover, the contributor samantaz is not actively contributing to Invidious anymore, and she was one of the leading maintainer regarding this part of the code.
This is why we prefer to focus our time on keeping Invidious working with the constant YouTube changes.
You can however create a fork and apply the code from https://github.com/iv-org/invidious/pull/3164 to run in your Invidious instance. Though you will probably need to keep the code updated with each new Invidious code.

PR #3198, opened 2022-07-13
Fix a dead link to Docker install documentation

PrivateGER:
This just changes a link in the docker-compose file, the current one points to a 404.


SamantazFox:
Thanks for catching it!

PR #3199, opened 2022-07-13
Include `_helpers.js` in embedded view

AHOHNMYC:
`_helpers.js` already included in `template.ecr`, but this is not enough for /embed/ links.

image


SamantazFox:
Ah, lol, I've made that change too, but forgot to make a PR ^^
https://github.com/SamantazFox/invidious/tree/fix-missing-helpers-embed

PR #3244, opened 2022-08-06
Disable decrypt_polling by default + add comment

unixfox:
After some testing the decrypt polling doesn't work anymore, for example encrypted age-restricted videos are not decrypted anymore.

Disabling this function until we fix it and add back the previous comment that got removed in https://github.com/iv-org/invidious/commit/644ba469451f2348219f99684e44e42a276bfd46

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -352,10 +352,13 @@ feed_threads: 1
 ## Note: This part of the code generate a small amount of data every minute.
 ## This may not be desired if you have bandwidth limits set by your ISP.
 ##
-## Accepted values: true, false
-## Default: true
+## Note 2: This part of the code is currently broken, so changing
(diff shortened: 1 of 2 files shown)


TheFrenchGhosty (approved):

config/config.example.yml
```diff
@@ -349,13 +349,17 @@ feed_threads: 1
 ## Enable/Disable the polling job that keeps the decryption
 ## function (for "secured" videos) up to date.
 ##
-## Note: This part of the code generate a small amount of data every minute.
+## Note: This part of the code is currently broken, so changing
+## this setting has no impact.
```
SamantazFox, line 353: Could you please add this as a "Note 2"? That way it will be less diff both now and when the function will be fixed :)

PR #3262, opened 2022-08-12
Add/Change Javascript license notice

CalculationPaper:
Add missing js file and clarify the MIT license variant

### DIFF
--- a/src/invidious/views/licenses.ecr
+++ b/src/invidious/views/licenses.ecr
@@ -25,7 +25,7 @@
 
         <tr>
             <td>
-                <a href="/js/helpers.js?v=<%= ASSET_COMMIT %>">helpers.js</a>
+                <a href="/js/handlers.js?v=<%= ASSET_COMMIT %>">handlers.js</a>
             </td>
 
             <td>
@@ -33,7 +33,7 @@
             </td>
 
             <td>
-                <a href="/js/helpers.js?v=<%= ASSET_COMMIT %>"><%= translate(locale, "source") %></a>
+                <a href="/js/handlers.js?v=<%= ASSET_COMMIT %>"><%= translate(locale, "source") %></a>
             </td>
         </tr>
 
--- a/src/invidious/views/licenses.ecr
+++ b/src/invidious/views/licenses.ecr
@@ -25,6 +25,20 @@
 
         <tr>
             <td>
+                <a href="/js/helpers.js?v=<%= ASSET_COMMIT %>">helpers.js</a>
+            </td>
+
+            <td>
+                <a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL-3.0</a>
+            </td>
+
+            <td>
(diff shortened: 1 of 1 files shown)


src/invidious/views/licenses.ecr
```diff
@@ -23,6 +23,20 @@
...
+
+            <td>
+                <a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL-3.0</a>
+            </td>
+
+            <td>
+                <a href="/js/helpers.js?v=<%= ASSET_COMMIT %>"><%= translate(locale, "source") %></a>
+            </td>
+        </tr>
+
```
SamantazFox, line 39: There is no such file as `helpers.js` ^^
The license for [_helpers.js](https://github.com/iv-org/invidious/blob/master/assets/js/_helpers.js) is already in right above :)

src/invidious/views/licenses.ecr
```diff
@@ -169,7 +183,7 @@
             </td>
 
             <td>
-                <a href="https://choosealicense.com/licenses/mit/">MIT</a>
+                <a href="https://choosealicense.com/licenses/mit/">Expat</a>
```
SamantazFox, line 186: I'm not sure why you'd put `Expat` in here? What does that mean?

src/invidious/views/licenses.ecr
```diff
@@ -253,7 +267,7 @@
             </td>
 
             <td>
-                <a href="https://choosealicense.com/licenses/mit">MIT</a>
+                <a href="https://choosealicense.com/licenses/mit">Expat</a>
```
SamantazFox, line 270: https://github.com/videojs/videojs-vr/issues/264

src/invidious/views/licenses.ecr
CalculationPaper: Oh, my fault, sorry. I actually meant handlers.js I guess. Looks nearly the same and doesn't sound too different. Can use it for rhymes. :D

CalculationPaper:
I forgot too mention that changing "MIT" to "Expat" or "X11" makes it complies to the LibreJS browser-extension. According to the FSF there are multiple licenses the Massachusetts Institute of Technology (MIT) used/uses, so calling it "MIT license" could be misleading. A prominent example for "MIT licenses" are the "Expat" and "X11" variants and there is a notable difference between them. The "X11 License" have an advertising clause.

SamantazFox:
I feel like it's some "flame war" thing. I've always seen the name "MIT license" being used.

But in all cases, we should use what mentioned projects use. For instance, it should be "MIT" for `videojs-vtt-thumbnails`, `videojs-markers` and `videojs-http-source-selector`.

CalculationPaper:
You are free to call it how you want. However, LibreJS's machine readable format seems to require the distinction between those two different MIT Licenses.

TheFrenchGhosty (approved):
I don't really know the details, but if it make Invidious compatible/compliant with LibreJS, I'm for it.

Issue #3257, opened 2022-08-08
[Enhancement] Throw an error if the video ID result is not the same as the one requested


unixfox:
From https://github.com/iv-org/invidious/issues/3230

We should probably add an exception if we find that the videoID returned is not the correct one, like yt-dlp did: https://github.com/yt-dlp/yt-dlp/pull/4593/files

This may allow debugging future issues quicker.

PR #3278, opened 2022-08-23
throw error if the videoID returned is different

unixfox:
Closes https://github.com/iv-org/invidious/issues/3257

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -909,6 +909,10 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
         "reason" => JSON::Any.new(reason),
       }
     end
+  elsif video_id != player_response.dig("videoDetails", "videoId")
+    # YouTube may return a different video player response than expected.
+    # See: https://github.com/TeamNewPipe/NewPipe/issues/8713
+    raise VideoNotAvailableException.new("The video returned by YouTube isn't the requested one.")
   else
     reason = nil
   end
--- a/src/invidious/exceptions.cr
+++ b/src/invidious/exceptions.cr
@@ -30,3 +30,6 @@ end
 # Exception threw when an element is not found.
 class NotFoundException < InfoException
 end
+
+class VideoNotAvailableException < Exception
+end


src/invidious/videos.cr
```diff
@@ -909,6 +909,10 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
         "reason" => JSON::Any.new(reason),
       }
     end
+  elsif video_id != player_response.dig("videoDetails", "videoId")
+    # YouTube may return a different video player response than expected.
+    # See: https://github.com/TeamNewPipe/NewPipe/issues/8713
+    raise BrokenTubeException.new("Different video returned by YouTube compared to the one requested. Can't playback the video.")
```
SamantazFox, line 915:
```suggestion
    raise BrokenTubeException.new("YouTube returned the wrong video")
```

SamantazFox (commented):
[Don't mind my previous (now deleted) comment]

We should probably put something in info["error"] instead of rising an exception. I'm pretty sure that rising an exception for the wrong video ID will break sonner or later.

Also, if we raise something, We should raise an `InfoException` (no trace) or just an `Exception`. A `BrokenTubeException` is used to give the name of a JSON object that we can't find.

EDIT: We can also create a dedicated `VideoNotAvailableException`

unixfox:
> We should probably put something in info["error"] instead of rising an exception. I'm pretty sure that rising an exception for the wrong video ID will break sonner or later.

Then the user wouldn't have direct easy access to reporting the issue. Returning an info[error] means only a single message is displayed.

> Also, if we raise something, We should raise an `InfoException` (no trace) or just an `Exception`. A `BrokenTubeException` is used to give the name of a JSON object that we can't find.
> 
> EDIT: We can also create a dedicated `VideoNotAvailableException`

Ok for a  `VideoNotAvailableException`.

SamantazFox:
> Ok for a `VideoNotAvailableException`.

Go for it, then!

unixfox:
By throwing an `InfoException` I don't get the full error template, that's what I wanted to get...

unixfox:
Ok I got it, please review again.

src/invidious/videos.cr
```diff
@@ -909,6 +909,10 @@ def extract_video_info(video_id : String, proxy_region : String? = nil, context_
         "reason" => JSON::Any.new(reason),
       }
     end
+  elsif video_id != player_response.dig("videoDetails", "videoId")
+    # YouTube may return a different video player response than expected.
+    # See: https://github.com/TeamNewPipe/NewPipe/issues/8713
+    raise VideoNotAvailableException.new("Different video returned by YouTube compared to the one requested. Can't playback the video.")
```
SamantazFox, line 915: I'd reword the first sentence as follows:
"The video returned by Youtube isn't the requested one"

We can also remove the second sentence, as it will be obvious that the video can't be played (the user is on the error page).

unixfox: Reworded

SamantazFox (approved):

PR #3287, opened 2022-08-29
Customize scrollbar in Mozilla Firefox

zer0-x:
This uses the CSS's [`scrollbar-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars) to change the colors of the scroll bar. For now it's only supported in Firefox.

For Chromium [`::-webkit-scrollbar`](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar) is supported, but it's not a standard yet, and there may also be a large incompatibilities between implementations...

**Before:**

**After:**

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -370,6 +370,10 @@ span > select {
 }
 
 @media (prefers-color-scheme: light) {
+  html {
+    scrollbar-color: #303030 #f0f0f0;
+  }
+
   .no-theme a:hover,
   .no-theme a:active,
   .no-theme summary:hover  {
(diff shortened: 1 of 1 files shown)


unixfox:
What's the result on the white theme?

zer0-x:
I've uploaded new images 👍

unixfox:
Please squash your commits into one.

SamantazFox:
I'm not super convinced about the utility of that PR.

IMHO, we shouldn't change UI elements that belong to the browser, especially when the option to do that is barely supported.

PR #3294, opened 2022-08-31
Add table cleaning job

### LINKED ISSUE #2835: [Bug] Channel refresh job could be creating database bloat
**NOTE** This is only a theory as of right now.

## Theory

The `RefreshChannelsJob` calls `fetch_channel` for every channel in the database every 30 minutes. [link](https://github.com/iv-org/invidious/blob/8b0ed8b110752327b98565e4ecc4faf437c0bb3b/src/invidious/jobs/refresh_channels_job.cr#L33)

The `fetch_channel` code parses all videos from the channel's feed and tries to insert them into the database. [link](https://github.com/iv-org/invidious/blob/8b0ed8b110752327b98565e4ecc4faf437c0bb3b/src/invidious/channels/channels.cr#L237)

The database insertion uses `ON CONFLICT` to update the row if the id already exists. [link](https://github.com/iv-org/invidious/blob/8b0ed8b110752327b98565e4ecc4faf437c0bb3b/src/invidious/database/channels.cr#L92-L109)

All of this put together means that it's likely that almost every single channel video is updated every 30 minutes. That creates a dead tuple in postgres which means that the database table increase by `SELECT COUNT(*) FROM channel_videos` rows every 30 minutes. This does not mean that the detectable number of rows grows by that much, but that a large amount of rows are left for postgres to clean up at some point in time.

## Why is it just now a problem?

My theory doesn't make sense unless either (1) it has always bloated the database and no one has really noticed or (2) the job always failed before and no one noticed.

SamantazFox:
These few code additions do the same as [what we recommend in the documentation](https://docs.invidious.io/Database-Information-and-Maintenance.md), but without the admin needing to add a cron job. The only cron that remains to the admin is the DB vacuuming (Interval is dependent on the instance's size)

Note: if said admin followed the "1 hour restart" recommendation, the job will only run once on every instance start. As many other jobs, they're here for small instances that don't have scheduled restarts.

Closes #2835

@FireMasterK This will probably help with the recent problems your instance has been facing.

### DIFF
--- a/src/invidious/database/nonces.cr
+++ b/src/invidious/database/nonces.cr
@@ -4,7 +4,7 @@ module Invidious::Database::Nonces
   extend self
 
   # -------------------
-  #  Insert
+  #  Insert / Delete
   # -------------------
 
   def insert(nonce : String, expire : Time)
@@ -15,6 +15,15 @@ module Invidious::Database::Nonces
     SQL
 
     PG_DB.exec(request, nonce, expire)
+  end
+
+  def delete_expired
+    request = <<-SQL
+      DELETE FROM nonces *
+      WHERE expire < now()
+    SQL
+
+    PG_DB.exec(request)
   end
 
   # -------------------
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -304,10 +304,8 @@ https_only: false
 ## Number of threads to use when crawling channel videos (during
 ## subscriptions update).
 ##
-## Notes:
-##  - Setting this to 0 will disable the channel videos crawl job.
-##  - This setting is overridden if "-c THREADS" or
-##    "--channel-threads=THREADS" are passed on the command line.
+## Notes: This setting is overridden if either "-c THREADS" or
+## "--channel-threads=THREADS" is passed on the command line.
 ##
 ## Accepted values: a positive integer
 ## Default: 1
@@ -335,10 +333,8 @@ full_refresh: false
 ##
 ## Number of threads to use when updating RSS feeds.
 ##
-## Notes:
-##  - Setting this to 0 will disable the channel videos crawl job.
-##  - This setting is overridden if "-f THREADS" or
-##    "--feed-threads=THREADS" are passed on the command line.
+## Notes: This setting is overridden if either "-f THREADS" or
+## "--feed-threads=THREADS" is passed on the command line.
 ##
 ## Accepted values: a positive integer
 ## Default: 1
@@ -359,6 +355,39 @@ feed_threads: 1
 ## Default: false
 ##
 #decrypt_polling: false
+
+
+jobs:
+
+  ## Options for the database cleaning job
+  clear_expired_items:
+
+    ## Enable/Disable job
+    ##
+    ## Accepted values: true, false
+    ## Default: true
+    ##
+    enable: true
+
+  ## Options for the channels updater job
+  refresh_channels:
+
+    ## Enable/Disable job
+    ##
+    ## Accepted values: true, false
+    ## Default: true
+    ##
+    enable: true
+
+  ## Options for the RSS feeds updater job
+  refresh_feeds:
+
+    ## Enable/Disable job
+    ##
+    ## Accepted values: true, false
+    ## Default: true
+    ##
+    enable: true
 
 
 # -----------------------------
@@ -859,7 +888,7 @@ default_user_preferences:
   ## Default: false
   ##
   #automatic_instance_redirect: false
-  
+
   ##
   ## Show the entire video description by default (when set to 'false',
   ## only the first few lines of the description are shown and a
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -172,6 +172,8 @@ end
 CONNECTION_CHANNEL = Channel({Bool, Channel(PQ::Notification)}).new(32)
 Invidious::Jobs.register Invidious::Jobs::NotificationJob.new(CONNECTION_CHANNEL, CONFIG.database_url)
 
+Invidious::Jobs.register Invidious::Jobs::ClearExpiredItemsJob.new
+
 Invidious::Jobs.start_all
 
 def popular_videos
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -78,6 +78,10 @@ class Config
   property decrypt_polling : Bool = false
   # Used for crawling channels: threads should check all videos uploaded by a channel
   property full_refresh : Bool = false
+
+  # Jobs config structure. See jobs.cr and jobs/base_job.cr
+  property jobs = Invidious::Jobs::JobsConfig.new
+
   # Used to tell Invidious it is behind a proxy, so links to resources should be https://
   property https_only : Bool?
   # HMAC signing key for CSRF tokens and verifying pubsub subscriptions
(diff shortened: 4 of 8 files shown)


unixfox:
Delete doesn't fully reclaim the disk space unless vaccum is triggered. Even after vaccum not all the disk space is reclaimed.

That's why for the table videos it's better to truncate it because it will instantly reclaim the disk space.

Also you should allow the ability to turn on or off these jobs because if one is running invidious multiple times this could lead to some funky behavior if 10 invidious processes are deleting the same thing.

SamantazFox:
> Delete doesn't fully reclaim the disk space unless vaccum is triggered. Even after vaccum not all the disk space is reclaimed.
> 
> That's why for the table videos it's better to truncate it because it will instantly reclaim the disk space.

The problem of `TRUNCATE` is that it entirely wipes the table, which, if done too often, makes the caching useless. My idea was to have a regular `DELETE` running (that only removes expired entries) so the autovacuuming worker would be trigerred more often, but would have less work to do (there won't be a million rows to reclaim every day).

> Also you should allow the ability to turn on or off these jobs because if one is running invidious multiple times this could lead to some funky behavior if 10 invidious processes are deleting the same thing.

Afaik, a simple `DELETE` can be run by multiple instances in parallel. If it has already been run recently (in the last few minutes), there won't be much to delete anyway.

I considered to add such an option, but I'm not sure if that's really required; None of the commands locks the DB, and they're not run very often (even with 20 instances running, they'd be ran 20 times per hour).

This will indeed require some large scale testing.

unixfox:
> > Delete doesn't fully reclaim the disk space unless vaccum is triggered. Even after vaccum not all the disk space is reclaimed.
> > That's why for the table videos it's better to truncate it because it will instantly reclaim the disk space.
> 
> The problem of `TRUNCATE` is that it entirely wipes the table, which, if done too often, makes the caching useless. My idea was to have a regular `DELETE` running (that only removes expired entries) so the autovacuuming worker would be trigerred more often, but would have less work to do (there won't be a million rows to reclaim every day).

The issue on big instances is that autovacuum hammer the CPU like crazy, so doing a TRUNCATE avoid any CPU hammering by the autovacuum as the disk space is instantly reclaimed. I do agree with the cache purpose, but IMO the cache table is mildly useful because the cache table grows like crazy on big instances (6.7GB in 9 hours for me), and the video streams are often replaced with new one thus the rows cache are often updated.

The current cache system needs to be revisited.

I currently have a TRUNCATE every 12 hours, I can replace it with your query in order to see if that won't grow the DB too much.

> > Also you should allow the ability to turn on or off these jobs because if one is running invidious multiple times this could lead to some funky behavior if 10 invidious processes are deleting the same thing.
> 
> Afaik, a simple `DELETE` can be run by multiple instances in parallel. If it has already been run recently (in the last few minutes), there won't be much to delete anyway.
> I considered to add such an option, but I'm not sure if that's really required; None of the commands locks the DB, and they're not run very often (even with 20 instances running, they'd be ran 20 times per hour).

People may prefer to return these queries outside of invidious, so I think an option is a good idea.

src/invidious/database/nonces.cr
```diff
@@ -17,6 +17,15 @@ module Invidious::Database::Nonces
     PG_DB.exec(request, nonce, expire)
   end
 
+  def delete_expired
+    request = <<-SQL
+      DELETE FROM nonces *
```
unixfox, line 22: No idea why but nonces table is empty for me.

unixfox:
On top of that, `channel_videos` needs to be cleaned up too.

image

No idea how

src/invidious/database/channels.cr
```diff
@@ -28,6 +28,17 @@ module Invidious::Database::Channels
     PG_DB.exec(request, args: channel_array)
   end
 
+  def delete_not_subscribed
+    request = <<-SQL
+      DELETE FROM channels *
```
unixfox, line 33: This query takes AGES to finish on a big instance. After 300 seconds it hasn't finished completing. I do have 36232 users though.

SamantazFox:
> The issue on big instances is that autovacuum hammer the CPU like crazy, so doing a TRUNCATE avoid any CPU hammering by the autovacuum as the disk space is instantly reclaimed. I do agree with the cache purpose, but IMO the cache table is mildly useful because the cache table grows like crazy on big instances (6.7GB in 9 hours for me), and the video streams are often replaced with new one thus the rows cache are often updated.

Damn D:

> The current cache system needs to be revisited.

Yup, I'm definitely looking closely at solutions like Redis.

> I currently have a TRUNCATE every 12 hours, I can replace it with your query in order to see if that won't grow the DB too much.

I'm wondering if a more regular deletion would have the same "hammering" impact...

If `TRUNCATE`-ing is the only way, then a cron job is the only viable way atm (until we can make invidious run for multiple hours without a hiccup).

> People may prefer to return these queries outside of invidious, so I think an option is a good idea.

Okay. I'll add that.

src/invidious/database/channels.cr
SamantazFox: Maybe it's because it has never been done before? It took quite some time on the test instance too when I ran it...

unixfox (changes requested):
I'm okay with adding the following functions:
- Invidious::Database::Videos.delete_expired: has been tested on my instance for 15 days and seems to work fine. The database size stay at around 5GB while having a hundred of new videos added to the table per seconds.
- Invidious::Database::Nonces.delete_expired: The table is empty on my side, but why not.

Not okay with the following ones:
- Invidious::Database::Channels.delete_not_subscribed: It adds too much load on the database as explained in https://github.com/iv-org/invidious/pull/3294/files#r961975096, or we only execute the query on instances with like < 100 users

SamantazFox:
> Invidious::Database::Videos.delete_expired: has been tested on my instance for 15 days and seems to work fine. The database size stay at around 5GB while having a hundred of new videos added to the table per seconds.

Great! Did you notice any CPU/RAM usage?

> Invidious::Database::Nonces.delete_expired: The table is empty on my side, but why not.

I double checked: it only fills up if captcha are enabled on login/sign-up. I don't think yours has it enabled?

> Invidious::Database::Channels.delete_not_subscribed: It adds too much load on the database as explained in https://github.com/iv-org/invidious/pull/3294/files#r961975096, or we only execute the query on instances with like < 100 users

I'll entirely remove the code. The channels cache is also used for RSS iircs, and tbh, we don't keep much data anyway (the `channel_videos` is much bigger).

unixfox:
> > Invidious::Database::Videos.delete_expired: has been tested on my instance for 15 days and seems to work fine. The database size stay at around 5GB while having a hundred of new videos added to the table per seconds.
> 
> Great! Did you notice any CPU/RAM usage?

Difficult to say because my servers are often overloaded during the night in EU time (something I plan to fix) but I haven't seen any spike in CPU usage from postgres autovacuum which is nice. So I think it's ok.

> > Invidious::Database::Nonces.delete_expired: The table is empty on my side, but why not.
> 
> I double checked: it only fills up if captcha are enabled on login/sign-up. I don't think yours has it enabled?

Indeed I deactivated the captcha for logins a while ago.

unixfox (changes requested):
Add the ability to disable the job in the config.

SamantazFox:
> Before merging the PR I still would like the ability to disable this job in the config.

Sure, will do that tomorrow!

unixfox (approved):

SamantazFox:
I should have fixed the CI now.

PR #3325, opened 2022-09-22
Correct `peertubeify` URL in `README.md`

Benjamin-Loison:


unixfox:
Works in both cases but I guess having the correct URL is better.

Issue #3346, opened 2022-10-15
Twitter user linked in meta tags has been suspended


thecashewtrader:
**Describe the bug**
The twitter user `@omarroth1`, that is linked in the meta tag `twitter:site`, has been suspended. I'm willing to fix/change the user to a different one, but I'm not sure what to set it to.

Source file: https://github.com/iv-org/invidious/blob/master/src/invidious/views/watch.ecr#L22

PR #3347, opened 2022-10-15
Add meta tags to channels

thecashewtrader:
A note: this commit uses the same suspended username mentioned in #3346.

### DIFF
--- a/src/invidious/views/channel.ecr
+++ b/src/invidious/views/channel.ecr
@@ -10,7 +10,6 @@
 <meta property="og:image" content="/ggpht<%= channel_profile_pic %>">
 <meta property="og:description" content="<%= channel.description %>">
 <meta name="twitter:card" content="summary">
-<meta name="twitter:site" content="@omarroth1">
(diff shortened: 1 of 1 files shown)


SamantazFox:
Is `twitter:site` really needed? As you mentioned, the username used here doesn't exist anymore, so we don't have anything to replace it.

thecashewtrader:
According to [this document](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup), it is required, though it can be argued it's not consequential.
Anyway, https://twitter.com/iv-org is free for now, so could we take it and just put that here?

TheFrenchGhosty:
Discussed internally: I have no desire for us to ever get a Twitter account, if possible, removing all the lines related to "our" account is the way to go.

thecashewtrader:
Aight, fair enough.

SamantazFox (approved):

SamantazFox:
Thanks for your contribution to invidious!

Issue #1438, opened 2020-10-24
Invidious has a memory leak


TheFrenchGhosty:
This has already been reported multiples time, but like #1345 this is an issue made to regroup other issues where this problem was reported.

This is also the issue where we will keep track of the progress in fixing it.

---

Invidious has a massive memory leak.

As shown in this screenshot...

![image](https://user-images.githubusercontent.com/47571719/97091599-86bf7b00-162c-11eb-806b-01de7dd06a75.png)

...the memory usage of invidious can increase to 4GB in 15 minutes, starting from less than 500MB, the dips in memory usage are restarts or crashes, another issue related to crashes is available here: [ISSUE].

Previous issue where this was reported: #1415 #621 #721 

Previous discussion about it: #1051

SuperSandro2000:
A memory leak is probably not true. There are probably multiple ones.

GitWaifu:
i get memory leaks precisely on 12AM every night and if i change time zone in settings it reoccurs on 12AM for new time zone even if it is actually a bright day outside - looks like intentional malicious code is injected in about half of the instances

Perflyst:
> looks like intentional malicious code is injected in about half of the instances

Why does it look like that? Never heard of this "12AM memory leak" before. Do you run backups at 12AM? Do you run some other cronjobs at this time?
Does it also happen on a new clean server?

TheFrenchGhosty:
@GitWaifu 

>looks like intentional malicious code is injected in about half of the instances

So, according to you, omarroth intentionally added code to trigger a memory leak that end up crashing the instances, because he absolutely don't want you to host it?

Seriously? Don't use the software if you don't like it, but stop saying bugs are caused by malicious intent.

PS: read that: https://en.wikipedia.org/wiki/Correlation_does_not_imply_causation

HumanG33k:
there is a way to run a valgrind or something similar ?

Malicious intent even if  it's probably not the first cause should not be exclude. That kind of product make shadow to google and as they already did with useragent for Firefox or internet explorer. Why they not try by all ways to shutdown this kind of project. Also not in this ecosystem but in javascript one several lib were and will compromise it's the side effect of dependancies managers (should not be used if you want my point of view).   

The instance with trouble may use an installer (i did) and third party can in this case be the root cause to.

SuperSandro2000:
[...]

PR #3376, opened 2022-11-01
Dockerfile: Pass '-Ddisable_quic' to 'crystal build'

SamantazFox:
This is a follow-up to #3367 and a desperate attempt to fix the memory leak issues invidious is having.

### DIFF
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -52,7 +52,7 @@ jobs:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
 
-      - name: Build and push Docker AMD64 image for Push Event
+      - name: Build and push Docker AMD64 image without QUIC for Push Event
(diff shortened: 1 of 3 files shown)


unixfox:
Would prefer to have a special docker image tag for that instead.

SamantazFox:
> Would prefer to have a special docker image tag for that instead.

I think that many of the memory leak issues we're having at the moment are related to QUIC.
Until the dependency is updated and stabilized, I'd prefer that we don't ship it in the release containers.

In addition to that, the version of BoringSSL we're shipping is very outdated (and can't be updated without updating lsquic too). As it collides with the system's own OpenSSL lib, that's not good for security (though it has almost no impact in the context of a container behind a reverse proxy).

unixfox:
I'll update your code in order to have the default docker image without quic and an additionnal docker image with a special tag with quic included.

SamantazFox:
Forgot to mention that this is related to #1438

PR #3385, opened 2022-11-09
readme: Update matrix room ID

SamantazFox:


TheFrenchGhosty:
Why? matrix.org is standard.

JokerGermany:
As long as the Main Matrix alias is #invidious:matrix.org, this shouldn't be changed in my opinion.

BTW: Why matrix.pussthecat.org and not pussthecat.org?

SamantazFox:
> Why? matrix.org is standard.

Because the room isn't hosted on matrix.org anymore.

> BTW: Why matrix.pussthecat.org and not pussthecat.org?

Because that's the (sub)domain of PussTheCat's Matrix server ^^

TheFrenchGhosty:
@SamantazFox Matrix rooms aren't "hosted" anywhere, it's in the federation, any address pointed to it will work. We use matrix.org because it's the standard, we could use any address setup by any server that joined it.

matrix.pussthecat.org is the homeserver address, yes, but delegation is setup so even if we really want to use the address from my homeserver, it will be #invidious:pussthecat.org

SamantazFox:
yeah, I realized that

PR #3426, opened 2022-11-22
Fix values for dark_mode in config.example.yml

PrivacyDevel:


SamantazFox (changes requested):
Thanks for fixing this, I didn't notice that this was ignoring booleans.
It also takes `"light"` or `"auto"`

PrivacyDevel:
Yup, fixed it.

SamantazFox (approved):
Thank you :)

PR #3450, opened 2022-11-29
FIX: Invalid Int32: (ArgumentError) in ChannelRenderer

11Tuvork28:
It's only a temporary fix until the YT devs get their things together but till then it works fine.
It's live on my [instance](https://invidio.xamh.de/), works just fine.
I understand if this gets blocked because it's only a matter of time till it breaks again :(


SamantazFox:
https://github.com/iv-org/invidious/pull/3418 didn't solve the int32 error on your side?

11Tuvork28:
what the hell I 100% merged it. No it did not that's weird. Well I'm sorry for the PR then I didn't realize....

PR #3458, opened 2022-12-01
Docker: Added tini for proper signal forwarding

Craeckie:
Currently, the TERM signal from docker is not handled, so the container is killed after the timeout. Adding [tini](https://github.com/krallin/tini) as the entrypoint solves this.


SamantazFox:
> Can you also add it for https://github.com/iv-org/invidious/blob/master/docker/Dockerfile.arm64?

^ bump ?

SamantazFox:

Issue #2535, opened 2021-10-25
[Bug] When user-prefers-dark loads dark theme, changing to light theme requires to passes of the change theme function


vituperative:
It appears that when the browser loads the dark theme via the user-prefers-dark css directive, invidious itself still registers the theme as light, and a double change is required to successfully change to the light theme.

Steps to reproduce: 

1) ensure browser is automatically displaying dark theme by default (user-prefers-dark)
2) attempt to change theme to light
3) notice that on first attempt, with js disabled, the theme remains dark on save.
4) after the second attempt, the theme will change to light.

PR #3521, opened 2022-12-29
handle auto theme correctly with the manual toggle

tleydxdy:
If the user used the manual toggle, they will not be able to get back to auto since it will force set to light theme. This should fix that.

### DIFF
--- a/assets/js/themes.js
+++ b/assets/js/themes.js
@@ -22,9 +22,11 @@ function setTheme(theme) {
     if (theme === THEME_DARK) {
         toggle_theme.children[0].className = 'icon ion-ios-sunny';
         document.body.className = 'dark-theme';
-    } else {
+    } else if (theme === THEME_LIGHT) {
         toggle_theme.children[0].className = 'icon ion-ios-moon';
         document.body.className = 'light-theme';
+    } else {
+        document.body.className = 'no-theme';
     }
 }
 


tleydxdy:
found a problem, if auto theme is dark, the first click on the toggle will not do anything :(

SamantazFox:
> found a problem, if auto theme is dark, the first click on the toggle will not do anything :(

Yeah, you'll need to click twice on the theme button to switch. See #2535

SamantazFox:
I just realize that this change has no impact!

As the comment says, `setTheme` only takes either `THEME_LIGHT` or `THEME_DARK`.
This is passed from the event listener callback function above (on line 12).

So if `THEME_DARK` is not passed, it can only be `THEME_LIGHT` anyway!

tleydxdy:
it is also called from the function below which takes the value from the page, which can be "" or in my case it was "false" for some reason

SamantazFox:
Ah, crap, it can get a `true` or `false` value too :/
I think this is the backend that needs to be fixed first:
https://github.com/iv-org/invidious/blob/master/src/invidious/user/converters.cr

tleydxdy:
is the [converters.cr](https://github.com/iv-org/invidious/blob/master/src/invidious/user/converters.cr) part just stale code? I don't think env.params.query["dark_mode"] can be true/false anymore, maybe it used to

SamantazFox:
@AHOHNMYC can you help here? I'm not sure how/if the comment above that function needs to be updated.

SamantazFox:
@tleydxdy Thank you for noticing and providing a fix :)
@AHOHNMYC Thanks for your feedback!

PR #3554, opened 2023-01-13
Make DASH absolute urls when local

iBicha:
This PR makes the `BaseURL` in DASH manifest containing absolute urls instead of relative, which matches non-proxied DASH

### DIFF
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -42,7 +42,7 @@ module Invidious::Routes::API::Manifest
 
     if local
       adaptive_fmts.each do |fmt|
-        fmt["url"] = JSON::Any.new(URI.parse(fmt["url"].as_s).request_target)
+        fmt["url"] = JSON::Any.new("#{HOST_URL}#{URI.parse(fmt["url"].as_s).request_target}")
       end
     end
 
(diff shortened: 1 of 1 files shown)


unixfox:
This will break dash on the installations where the user forgot to change the `domain:` parameter and expose their invidious instance on another domain than `localhost` or if the user access invidious externally from the port 3000.

iBicha:
> This will break dash on the installations where the user forgot to change the `domain:` parameter and access their invidious instance on another domain than `localhost` or if the user access invidious externally from on the port 3000.

If domain is missing, wouldn't `HOST_URL` be empty anyway?

SamantazFox:
@iBicha yes, if `domain` is not set, `HOST_URL` will be empty!

iBicha:
> @iBicha yes, if `domain` is not set, `HOST_URL` will be empty!
> 
> Edit: why is this required? Is there a related issue?

This is only relevant to 3rd party apps. This allows the 3rd party to redirect/proxy the dash url without issues.
This is not important to Invidious frontend, but it does not affect it in anyway (but makes it consistent)

SamantazFox:
You'll probably want to change this line too!
https://github.com/iv-org/invidious/blob/1fb0a495925a60b2d4839b98440e220b7f95d10e/src/invidious/routes/api/manifest.cr#L32

iBicha:
> You'll probably want to change this line too!
> 
> https://github.com/iv-org/invidious/blob/1fb0a495925a60b2d4839b98440e220b7f95d10e/src/invidious/routes/api/manifest.cr#L32

Good catch!

SamantazFox (approved):

SamantazFox:

PR #3556, opened 2023-01-14
Login redirect to referer on logged-in user

### LINKED ISSUE #3553: [Enhancement] /authorize_token should redirect to account creation
When user opens `/authorize_token` and not logged in, it redirects to referrer, which is home page by default.
This is confusing, especially to new users coming from 3rd party apps.

`/authorize_token`, if not logged in, should redirect to account login/creation. Once logged-in/account created, the redirect should go back to authorizing the token page.
This means the referrer (in this case, the callback url) should keep following page by page until the user finally authorizes the token

Documentation (hey, make sure to login on invidious before clicking this login button) but that's not great UX

**Additional context**
For 3rd party app, [Playlet](https://github.com/iBicha/playlet)

iBicha:
Closes #3553

### What
This PR allows a 3rd party app ([Playlet](https://github.com/iBicha/playlet)) to send user to Invidious where they will login/signup, authorize token, and go back to the app.

### Changes
 For this idea to work, we needed 3 things:
- `/authorize_token` should redirect to login if not logged-in already
- Let the login page redirect to referer, and not always "/feed/subscriptions"
- Do not strip chars `,` and `:` from `referer`)

### Flow
- `/authorize_token`'s referer is not interesting because it is a reference to the 3rd party app. Because of that, it should redirect to `/login`, and refer back to the original `/authorize_token` query once logged-in
- The login page should redirect to `referer` if provided, which allows us to return back to the `/authorize_token` after a login or a signup
- Do not strip chars `,` and `:` from `referer`, because these chars are used in the scope (e.g. `GET:feed,GET:playlists,POST:tokens/unregister`)

I have tried this with already logged in user, login to existing user, and creating a new user. The redirection goes back to `/authorize_token`, and 3rd party apps are able to acquire tokens.


iBicha:
Solves for #3553

unixfox:
> Solves for #3553 

Please add a "Closes #3553" so that the PR automatically close the issue.

iBicha:
@SamantazFox as code owner, do you have any thoughts on this?

SamantazFox:
> @SamantazFox as code owner, do you have any thoughts on this?

Looks good to me!

> * `/authorize_token` should redirect to login if not logged-in already
> * Let the login page redirect to referer, and not always "/feed/subscriptions"

That's a good idea. To be fair, all authenticated endpoints should redirect to the login page.

SamantazFox (approved):

SamantazFox:

PR #3238, opened 2022-08-03
Cleanup videos

SamantazFox:
Follow up to #3214
Related: #441

In essence, parse most of the video data before storing it.

PR #3563, opened 2023-01-15
Don't continue when LOGIN_REQUIRED and no videoDetails element

### LINKED ISSUE #3402: [Bug] Error when viewing private video - Missing JSON element: "videoDetails" (BrokenTubeException)
**Describe the bug**

**Steps to Reproduce**

**Logs**

**Screenshots**

**Additional context**

unixfox:
~~I've no idea why playability `LOGIN_REQUIRED` is allowed to continue the parsing of the video, but this should not be allowed.~~

~~A video behind a login will not give more information than the playability status, hence there is no section for `videoDetails` in the JSON payload.~~

-----

Don't continue when the playability is `LOGIN_REQUIRED` and the element `videoDetails` at the root of the JSON payload is not found. When the element `videoDetails` is not present, this usually means that trying other clients won't do anything.

An age-restricted video has the element `videoDetails` but not a private video in order to keep the video description hidden, for example.

Closes #3402

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -66,8 +66,10 @@ def extract_video_info(video_id : String, proxy_region : String? = nil)
     reason ||= subreason.try &.[]("runs").as_a.map(&.[]("text")).join("")
     reason ||= player_response.dig("playabilityStatus", "reason").as_s
 
-    # Stop here if video is not a scheduled livestream
-    if !{"LIVE_STREAM_OFFLINE", "LOGIN_REQUIRED"}.any?(playability_status)
+    # Stop here if video is not a scheduled livestream or
+    # for LOGIN_REQUIRED when videoDetails element is not found because retrying won't help
+    if !{"LIVE_STREAM_OFFLINE", "LOGIN_REQUIRED"}.any?(playability_status) ||
+       playability_status == "LOGIN_REQUIRED" && !player_response.dig?("videoDetails")
       return {
         "version" => JSON::Any.new(Video::SCHEMA_VERSION.to_i64),
         "reason"  => JSON::Any.new(reason),


SamantazFox:
In essence, this change was added in #3238, and it fixes #3402 already (sorry, I forgot to close the issue).

Long explanation:
previously, `fetch_video` was calling `extract_video_info` one or two times. If `extract_video_info` returned a `reason` on the first try, it was assumed to be caused by age/login restriction, and a second attempt was done with an "embed" client.

The problem of this approach was the low efficiency, and the risk of not reacting well to a specific situation. We also had a high chance of getting `missing element "..."` errors (as in #3402).

What I've done is to remove that "retry" logic from `fetch_video` and instead properly handle the various situations in `extract_video_info` directly. So, if a video has a status of `LOGIN_REQUIRED`, we want to keep parsing the player response to get as many details available as possible (like the author, description, etc..), we want to get the next response too, to get related videos and comments, and finally, we'll try using various clients until we get playable video streams.
 
For reference, the relevant change was done in cc5c83333f2a51dc178b698a548b64f01a2ff453

unixfox:
> In essence, this change was added in #3238, and it fixes #3402 already (sorry, I forgot to close the issue).
> 

No it does not see the video still throw an error: https://test.invidious.io/watch?v=6Dm-jLBgbzE

SamantazFox:
> No it does not see the video still throw an error: https://test.invidious.io/watch?v=6Dm-jLBgbzE

Ah, that's annoying :/
I'll see what can be done about that, but in the meantime please don't remove `LOGIN_REQUIRED` in here, otherwise all age-restricted videos will break!

unixfox:
I've switched from not continuing when `videoDetails` is not present, this usually means that trying other clients won't do anything.

An age-restricted video has the element `videoDetails` but not a private video in order to keep the video description hidden, for example.

Is that a good method @SamantazFox?

SamantazFox:
> Is that a good method @SamantazFox?

Yup, that looks good to me!

SamantazFox (approved):

unixfox:
Ok I let you review the code and approve when I can merge.

Issue #2114, opened 2021-05-24
Ability to Import YouTube playlist to invidious playlists


Kreyren:
I have a playlist full of music that has 1000+ videos in it, but invidious `2021.05.13-75e5b49` only allows me to input the videos one at the time after it searched the name making this task impractical as it would take 12.5 hours to complete assuming that adding one video takes at least 45 seconds.

Proposing to implement import of youtube playlists in invidious playlists.

NOTE: Invidious can save a youtube playlist, but that makes it read-only which is not expected in this scenario.

PR #3595, opened 2023-01-28
User: Allow CSV Youtube playlists to be imported

thtmnisamnstr:
Added functionality to import Youtube playlists in CSV format exported through Google Takeout. Fixes Issue #2114. Docker image currently on Docker Hub at thtmnisamnstr/invidious:latest.


unixfox:
Can you remove all the comments with your name: `Gavin Johnson (thtmnisamnstr), 20230127: Call the Youtube playlist import function`?

We do not put the name of who worked on a piece of code.

thtmnisamnstr:
> Can you remove all the comments with your name: `Gavin Johnson (thtmnisamnstr), 20230127: Call the Youtube playlist import function`?
> 
> We do not put the name of who worked on a piece of code.

Pushed a commit removing all of my comments.

SamantazFox:
Do you have a sample CSV file to try this code one?

thtmnisamnstr:
Here are 2 examples.
- [BJJ - Butterfly Guard.csv](https://github.com/iv-org/invidious/files/10667636/BJJ.-.Butterfly.Guard.csv)
- [BJJ - Half Guard.csv](https://github.com/iv-org/invidious/files/10667637/BJJ.-.Half.Guard.csv)

Playlists can be exported from YouTube using Google Takeout following [these instructions](https://github.com/iv-org/documentation/blob/master/docs/export-youtube-subscriptions.md) but for Playlists instead of Subscriptions.

Note that this functionality only imports playlists. It doesn't update existing playlists or check for duplicate playlists in any way. It just creates the playlists you import, regardless of what other playlists exist.

SamantazFox:
Thanks for the files ^^ I'll try and review this PR as soon as possible!
Can you please fix the CI in the meantime? (Looks like it's mostly formatting. You can run `make format` to automatically correct this)

thtmnisamnstr:
Should be good to go now.

SamantazFox:
Regarding the YT export format: Is it always a single playlist per CSV file (so header line, playlist infos, empty line, header line, then a line per video), or can a CSV file contain more that one playlist?

thtmnisamnstr:
It's always a single playlist per csv. The export from Google Takeout gives you multiple csv files.

SamantazFox:
Ok, cool, thank!

I was thinking: we could split the input in two smaller CSV files (right before the string `Video Id,Time Added`) so the multiple conditionals that use `line_counter` can be removed, and the code can be much simpler:

```crystal
raw_head, raw_body = raw_input.split("\n\n", limit: 2, remove_empty: true)

csv_head = CSV.new(raw_head, headers: true)
csv_body = CSV.new(raw_body, headers: true)

# Short form, without loop, for a single entry
csv_head.next
title = csv_head[4]
# Description, etc

# Sort for to extract video IDs as an array
video_ids = csv_body.map &.[0]
```

thtmnisamnstr:
I'm currently working on implementing the suggested changes.

thtmnisamnstr:
@SamantazFox I updated the import to use an improved algorithm as per your suggestion (couldn't quite do 100% of what your pseudocode was implying, but it's better). Any other changes you'd recommend?

SamantazFox:
> @SamantazFox I updated the import to use an improved algorithm as per your suggestion (couldn't quite do 100% of what your pseudocode was implying, but it's better).

Neat ^^

> Any other changes you'd recommend?

In `parse_playlist_export_csv`, the `if title && privacy && user` condition is useless: `env.get` and `[]` (the array  index accessor) raise an exception if the operation can't succeed, and `privacy` [is always a truthy value](https://crystal-lang.org/reference/1.6/syntax_and_semantics/truthy_and_falsey_values.html).

Also, in this context `var = uninitialized <Type>` is not required. The compiler will tell you if you try to use a variable before assigning any value to it. As far as I know, this syntax is mostly used when doing advanced crystal shenanigans with pointers and external library bindings.

thtmnisamnstr:
I removed the unnecessary conditions and uninitialized variable declarations as you recommended. This is my first time writing Crystal, and I didn't know those specifics of the language. Thanks.

src/invidious/user/imports.cr
```diff
@@ -53,13 +47,8 @@ struct Invidious::User
...
-      end
-
-      if playlist && description
-        Invidious::Database::Playlists.update_description(playlist.id, description)
-      end
+      playlist = create_playlist(title, privacy, user)
+      Invidious::Database::Playlists.update_description(playlist.id, description)
 
       # Add each video to the playlist from the body content
       CSV.each_row(raw_body) do |row|
```
SamantazFox, line 54: I'd recommend to read the body using the following way:
```crystal
csv_body = CSV.new(raw_body, headers: true)
csv_body.each do |row|
  # Put the rest of the code here
end
```

By doing so, you remove a lot of useless checks (the first line is always headers, so it's skipped in the `.new` statement, and if there is no other CSV lines (no videos), `.each` will be skipped)

SamantazFox:
Can you please fix CI by the way? It's a simple line with extra spaces (see CI logs for a diff)

src/invidious/user/imports.cr
thtmnisamnstr: Updates made, checked, and commited.

thtmnisamnstr:
@SamantazFox I think everything should be updated as per your requests. Sorry for the delay.

SamantazFox (approved):

SamantazFox:
Testing done, everything seems to work as expected!

PR #3603, opened 2023-01-30
Auth: Add username to token callback

iBicha:
When getting a token to a 3rd party, we need to at least the username to display it to the user (so they can logout if this is not their profile)
This adds the username to the callback.


iBicha:
@unixfox thoughts?

unixfox (approved):
won't break existing implementations, so why not adding it

PR #3617, opened 2023-02-03
API: add endpoints for import/export

iBicha:


src/invidious/routes/api/v1/authenticated.cr
```diff
@@ -31,6 +31,55 @@ module Invidious::Routes::API::V1::Authenticated
...
+                      json.string video_id
+                    end
+                  end
+                end
+              end
+            end
+          end
+        end
+      end
+    end
```
SamantazFox, line 64: There is already the same export code in [src/invidious/routes/subscriptions.cr](https://github.com/iv-org/invidious/blob/master/src/invidious/routes/subscriptions.cr#L109). Can you please create a dedicated function in `src/invidious/user/exports.cr`, so that we don't get two copies that are likely to differ in the future!

SamantazFox (changes requested):

src/invidious/routes/api/v1/authenticated.cr
iBicha: That's a good idea

src/invidious/routes/subscriptions.cr
```diff
@@ -106,31 +106,7 @@ module Invidious::Routes::Subscriptions
...
-                        end
-                      end
-                    end
-                  end
-                end
-              end
-            end
-          end
-        end
+        return Invidious::User::Export.to_invidious(user)
```
SamantazFox, line 109: You can also remove the unnecessary DB call right above this line (I can't review outside of the code you changed) ^^

src/invidious/user/exports.cr
```diff
@@ -0,0 +1,35 @@
...
+          json.field "playlists" do
+            json.array do
+              playlists.each do |playlist|
+                json.object do
+                  json.field "title", playlist.title
+                  json.field "description", html_to_content(playlist.description_html)
+                  json.field "privacy", playlist.privacy.to_s
+                  json.field "videos" do
+                    json.array do
+                      Invidious::Database::PlaylistVideos.select_ids(playlist.id, playlist.index, limit: 500).each do |video_id|
```
SamantazFox, line 22: Hmm, that 500 video limit here is a potential problem (instance admins can configure playlists to have more than 500 videos now). I'm aware that this was present in the original code, so can you put a higher number for now?

src/invidious/routes/subscriptions.cr
iBicha: Good catch!

src/invidious/user/exports.cr
iBicha: I didn't know about the config, I will use that value instead

SamantazFox: Oh, right, that's even better ^^

SamantazFox (approved):

SamantazFox:
@iBicha Thanks for contributing ^^

Issue #3600, opened 2023-01-29
[Enhancement] normalize audio volume


JulianGro:
**Is your enhancement request related to a problem? Please describe.**
Currently, videos on Invidious all have very different audio volume.

**Describe the solution you'd like**
YouTube has a normalized volume in the player. I would hope that there is some way to access that normalization factor and apply it to the player in Invidious as well.

**Additional context**
Here is an example: https://www.youtube.com/watch?v=4Q5lHqSXWNo
The audio on that video is so loud that YouTube will tone it down to 44% maximum.
![Auswahl_052](https://user-images.githubusercontent.com/11144627/215346240-3565e173-c3cb-460e-b84f-37ad842da0ba.png)

Issue #3342, opened 2022-10-11
[Feature request] Build audio/webm and video/webm streams in DASH manifest


ndygo:
**Is your feature request related to a problem? Please describe.**
As part of #3011, the audio/webm and video/webm streams have been removed from the manifest build because of a videojs issue. This means that the VP9 and Opus streams are not used by clients.

I can personally tell whether I am listening to a YouTube AAC stream because of the audio quality in comparison with Opus. In addition, the VP9 streams on YouTube are higher quality than the H.264 streams, and there isn't always AV1 streams (which can be contained in mp4).

**Describe the solution you'd like**
I'd like to see a solution like unixfox originally suggested here: https://github.com/iv-org/invidious/pull/3011#issuecomment-1093084049

Where if the client requesting the manifest is not videojs, the audio/webm and video/webm streams are built. Perhaps this could be done by appending a parameter to the URL given to videojs.

**Describe alternatives you've considered**
None.

Issue #2988, opened 2022-03-25
[maybe Enhancement] adding 30fps/60fps frame rate selection for 720p dash


rezad1393:
**Is your enhancement request related to a problem? Please describe.**

yes.
the problem is with my internet bandwidth. I don't have much of it so when I use invidious to watch youtube videos, I limit the video to 480p dash, but many of the videos are a bit blurry and small text becomes a bit hard to read. so I want to go to 720p.

right now invidious doesn't support anything other than h264 it seems (or the video player that invidious uses doesn't support it)
so size of videos is already bigger than youtube own player (because youtube uses vp9 and av1 way more than h264)
when I watch a video that is ,say 118MB in size at 480p-h264, if I want to watch it at 720p-h264 then it size becomes 342MB for 30fps and 534MB for 60fps.
my laptop supports 60fps but I don't need that frame rate just 720p resolution.
I see that invidious uses the 60pfs one.
so I want to watch the video at 720p-30fps.
this way I can watch it without saturating my bandwidth and also when using proxy-video feature, the invidious instance is also less burdened with my download.

**Describe the solution you'd like**

if I understand correctly youtube only has this dual frame rate support only at 720p dash.
so in options instead of 720p dash, put two options :720p-30fps and 720p-60fps

this maybe a edge case but it would help a lot.

SamantazFox:
I feel your pain. Sadly, on many videos, only 60FPS is available, and we can't fall back to 30FPS.

For instance, on the latest video from PewDiePie, the quality jumps from `480p` to `720p60`:
```json
{
	"adaptiveFormats": [
		{
			"itag": "135",
			"type": "video/mp4; codecs=\"avc1.4d401f\"",
			"fps": 30,
			"container": "mp4",
			"encoding": "h264",
			"qualityLabel": "480p"
		},
		{
			"itag": "244",
			"type": "video/webm; codecs=\"vp9\"",
			"fps": 30,
			"container": "webm",
			"encoding": "vp9",
			"qualityLabel": "480p"
		},
		{
			"itag": "298",
			"type": "video/mp4; codecs=\"avc1.4d4020\"",
			"fps": 60,
			"container": "mp4",
			"encoding": "h264",
			"qualityLabel": "720p60"
		},
		{
			"itag": "302",
			"type": "video/webm; codecs=\"vp9\"",
			"fps": 60,
			"container": "webm",
			"encoding": "vp9",
			"qualityLabel": "720p60"
		},
		{
[...]

Issue #2647, opened 2021-11-19
[Bug] adaptiveFormats missing fields in some formats


megascrapper:
**Describe the bug**

Each entry in the `adaptiveFormats` field of `/api/v1/videos/:id` has different available fields. For example, audio streams are missing `resolution` and `qualityLabel`. This is currently not mentioned anywhere in the documentation, so it would be better to have these missing fields to return `null` instead, or at least mention it in the API docs.

**Steps to Reproduce**

have and API request on video page, e.g. `/api/v1/videos/xuCn8ux2gbs`
I found 3 types of entry:
Video streams (except AV1) : all fields available
```
    {
      "index": "739-3626",
      "bitrate": "135033",
      "init": "0-738",
      "url": "https://r1---sn-npoe7ne6.googlevideo.com/videoplayback?expire=1637316051&ei=cyGXYbrWEr6L9fwP0JuqyAM&ip=180.244.7.97&id=o-AN5_ndJiz7v7FbAdhORldKJjHAoE5dAfm8JO59de7FKl&itag=160&source=youtube&requiressl=yes&mh=_m&mm=31%2C26&mn=sn-npoe7ne6%2Csn-30a7rn7l&ms=au%2Conr&mv=m&mvi=1&pl=23&initcwndbps=550000&vprv=1&mime=video%2Fmp4&gir=yes&clen=10136982&dur=1165.800&lmt=1608432002286721&mt=1637293979&fvip=3&keepalive=yes&fexp=24001373%2C24007246&c=ANDROID&txp=5535432&sparams=expire%2Cei%2Cip%2Cid%2Citag%2Csource%2Crequiressl%2Cvprv%2Cmime%2Cgir%2Cclen%2Cdur%2Clmt&sig=AOq0QJ8wRgIhALECP_mGsuYkFXqFVSr1CJOAg1eQHhdJ0Nakcdz1yLbDAiEAn1z7Ty56CsD7deYVMyRoAEPXVjyi501uyHhU2eQDXec%3D&lsparams=mh%2Cmm%2Cmn%2Cms%2Cmv%2Cmvi%2Cpl%2Cinitcwndbps&lsig=AG3C_xAwRQIhALPOV9lqCO1hnRTUiJnVVCH6dCBMPuoaJDzNr7_Nt7T8AiA_celn8IFyFTmmZt-ewQzJQ27po_tQtrO2oULTmI7zhA%3D%3D&host=r1---sn-npoe7ne6.googlevideo.com",
[...]

PR #3620, opened 2023-02-05
Rewrite streams logic

SamantazFox:
WIP

This PR is meant to overhaul how the streams' data (`formats`/`adaptativeFormats`) is parsed.

Objectives:
 - support videos with multiple audio tracks (#2007)
 - provide a much more stable interface/API (#2513, #2647, #2988)
 - Fix various audio issues (#597, #2393, #3600)
 - Add mime types back to the DASH manifest on demand (#3342)

### DIFF
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -80,72 +80,67 @@ module Invidious::JSONify::APIv1
           video.adaptive_fmts.each do |fmt|
             json.object do
               # Only available on regular videos, not livestreams/OTF streams
-              if init_range = fmt["initRange"]?
-                json.field "init", "#{init_range["start"]}-#{init_range["end"]}"
-              end
-              if index_range = fmt["indexRange"]?
-                json.field "index", "#{index_range["start"]}-#{index_range["end"]}"
-              end
+              json.field "init", fmt.init_range.to_s if fmt.init_range
+              json.field "index", fmt.index_range.to_s if fmt.index_range
+
+              # Livestream chunk infos. Should be present when `init` and `index` aren't
+              json.field "targetDurationSec", fmt.target_duration if fmt.target_duration
+              json.field "maxDvrDurationSec", fmt.max_dvr_duration if fmt.max_dvr_duration
 
               # Not available on MPEG-4 Timed Text (`text/mp4`) streams (livestreams only)
-              json.field "bitrate", fmt["bitrate"].as_i.to_s if fmt["bitrate"]?
+              json.field "bitrate", fmt.bitrate.to_s if fmt.responds_to?(:bitrate)
 
               if proxy
                 json.field "url", Invidious::HttpServer::Utils.proxy_video_url(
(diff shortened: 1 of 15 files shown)


ChunkyProgrammer:
Here are some useful test videos for multi-audio tracks:
- https://youtu.be/YLt73w6criQ (english original + a lot of dubs)

PikachuEXE:
Also here is a video with missing attributes on some formats (AV1): https://youtu.be/8IFwpI7OT6I
From https://github.com/FreeTubeApp/FreeTube/issues/3382

Zero3K:
Would this fix https://github.com/iv-org/invidious/issues/3757?

unixfox:
> Would this fix #3757?

No I don't think because no changes have been done to support browsers that don't support AV1 codec.

Zero3K:
This supposedly covers one of the mentioned issues (the one that requests for  webm streams to be added to the DASH manifest).

unixfox:
> This supposedly covers one of the mentioned issues (the one that requests for webm streams to be added to the DASH manifest).

webm is not supported by videojs so that doesn't really impact the video player of invidious. this will still won't fix the issue on edge

src/invidious/jsonify/api_v1/video_json.cr
```diff
@@ -79,72 +79,67 @@ module Invidious::JSONify::APIv1
...
-                json.field "encoding", fmt_info["vcodec"]? || fmt_info["acodec"]
 
                 if fmt_info["height"]?
                   json.field "resolution", "#{fmt_info["height"]}p"
 
                   quality_label = "#{fmt_info["height"]}p"
-                  if fps > 30
-                    quality_label += "60"
-                  end
+                  quality_label += "60" if fmt.responds_to?(:video_fps) && fmt.video_fps > 30
```
absidue, line 128: Please use the fps value instead of hardcoding it to 60, as 50fps streams exist too.

absidue (commented):
Haven't checked anything else as this still seems to be work in progress, but I just thought I would mention that as I saw it while flying over the code.

absidue:
Could you please include the bitrate for the formatStreams in the API, I know it's not used by Invidious, but it's returned by YouTube and I would like to display it in FreeTube's debug/stats overlay.

src/invidious/views/components/player.ecr
```diff
@@ -11,19 +11,19 @@
...
                end
 
                audio_streams.each_with_index do |fmt, i|
-                src_url  = "/latest_version?id=#{video.id}&itag=#{fmt["itag"]}"
+                src_url  = "/latest_version?id=#{video.id}&itag=#{fmt.itag}"
                 src_url += "&local=true" if params.local
 
-                bitrate = fmt["bitrate"]
-                mimetype = HTML.escape(fmt["mimeType"].as_s)
+                bitrate = fmt.bitrate // 1000
```
hikkamorii, line 25: While this is an improvement, it's still ambiguous whether AAC or Opus is used. Also, the "k" will be appended later, and it will not be inline with how it works when selecting audio quality for video.

Issue #2982, opened 2022-03-22
[Bug] Trending API endpoint's type parameter does not work


llsc12:
**Describe the bug**

It appears that the endpoint returns the default trending data regardless of what type is specified. 

**Steps to Reproduce**
 
https://vid.puffyan.us/api/v1/trending?type=default

https://vid.puffyan.us/api/v1/trending?type=news

Both output the same trending data.
**Logs**

N/A
**Screenshots**

N/A
**Additional context**

Appears to be an issue on all instances.

PR #3622, opened 2023-02-07
Ignore casing for trending_type parameter

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/2982

Note: news is no longer a trending category and has been moved to a separate channel and should be removed from the api docs

### DIFF
--- a/src/invidious/trending.cr
+++ b/src/invidious/trending.cr
@@ -4,14 +4,12 @@ def fetch_trending(trending_type, region, locale)
 
   plid = nil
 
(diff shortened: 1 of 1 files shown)


SamantazFox (approved):
You can even compact the code by using a case/when statement:
```crystal
case trending_type.try &.downcase
when "aaa"
  # stuff
when "bbb"
  # other stuff
else
  # default, catches nil values
end
```

SamantazFox (approved):

Issue #3323, opened 2022-09-21
[Bug] 


SamantazFox:
**Describe the bug**
All v1 API endpoints should have `authorVerified` when available
Currently, only the `search` endpoint has it.

unixfox:
Please add a title

SamantazFox:
> Please add a title

Damn, I completely forgot. Sorry :X

syeopite:
Still an issue

Benjababe:
I am selfishly requesting the field to be added, if possible, to

1. /videos -> recommendedVideos

eg. /api/v1/videos/WE9NIZShfZY
![image](https://github.com/iv-org/invidious/assets/19561469/050470e9-1d04-4901-986d-830370f6519b)

2. /popular

eg. /api/v1/popular/
![image](https://github.com/iv-org/invidious/assets/19561469/78b8b7c7-22ed-47a5-9381-4cc596295054)

PR #3625, opened 2023-02-08
API: add authorVerified where possible + provide channel tabs list

ChunkyProgrammer:
Most endpoints should now return if the author is verified (I think playlist is missing it but there isn't any parsing for it on the web side so I did not implement it). 
https://github.com/iv-org/invidious/issues/3323

The channel endpoint also noe returns the tabs available which can be useful for third party clients.

I also noticed that comments have a 'verified' instead of 'authorVerified' property so it might be worth changing that for consistency.

### DIFF
--- a/src/invidious/helpers/serialized_yt_data.cr
+++ b/src/invidious/helpers/serialized_yt_data.cr
@@ -74,6 +74,7 @@ struct SearchVideo
       json.field "author", self.author
       json.field "authorId", self.ucid
       json.field "authorUrl", "/channel/#{self.ucid}"
+      json.field "authorVerified", self.author_verified
 
       json.field "videoThumbnails" do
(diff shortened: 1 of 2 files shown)


SamantazFox (approved):

SamantazFox:
> I also noticed that comments have a 'verified' instead of 'authorVerified' property so it might be worth changing that for consistency.

Uuuh, the comments are a mess, let's not touch that for now xD
(unless you feel adventurous enough to separate the parsing code from the front-end code!)

ChunkyProgrammer:
> Uuuh, the comments are a mess, let's not touch that for now xD (unless you feel adventurous enough to separate the parsing code from the front-end code!)

Maybe in the future, I'll look at it but not now

SamantazFox:

PR #3626, opened 2023-02-08
YT Comments: parse isPinned and strikethroughs

ChunkyProgrammer:
I only added api support for isPinned but strikethroughs can be displayed in Invidious with this PR:

Before:
image

After:
image

YouTube:
image

Link to comment: https://www.youtube.com/watch?v=OqiXFXlYFi8&lc=UgxpIdc1QlWErEUhbkZ4AaABAg

### DIFF
--- a/src/invidious/comments.cr
+++ b/src/invidious/comments.cr
@@ -181,6 +181,8 @@ def fetch_youtube_comments(id, cursor, format, locale, thin_mode, region, sort_b
               json.field "content", html_to_content(content_html)
               json.field "contentHtml", content_html
 
+              json.field "isPinned", (node_comment["pinnedCommentBadge"]? != nil)
+
               json.field "published", published.to_unix
               json.field "publishedText", translate(locale, "`x` ago", recode_date(published, locale))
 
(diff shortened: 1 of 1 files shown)


src/invidious/comments.cr
```diff
@@ -181,6 +181,8 @@ def fetch_youtube_comments(id, cursor, format, locale, thin_mode, region, sort_b
               json.field "content", html_to_content(content_html)
               json.field "contentHtml", content_html
 
+              json.field "isPinned", (node_comment["pinnedCommentBadge"]? != nil)
```
SamantazFox, line 184: We should probably check that `pinnedCommentBadge` is not equal to `false`, just in case.

SamantazFox:
```suggestion
              json.field "isPinned", (node_comment["pinnedCommentBadge"]?.try(&.as_bool) == true)
```

ChunkyProgrammer: Sounds good

SamantazFox (approved):

SamantazFox:

src/invidious/comments.cr
SamantazFox: My bad, the boolean check is causing issues because that element is actually not a boolean.
https://github.com/iv-org/invidious/pull/3631

ChunkyProgrammer: It's my fault too 😅, I thought `.try(&.as_bool)` would have returned the truthy/falsey value (similar to how JavaScript does it).

PR #3627, opened 2023-02-09
Make itag optional under /latest_version

iBicha:


unixfox:
Can you explain why you would want that?

iBicha:
> Can you explain why you would want that?

Yes, added some context in the PR description

unixfox:
I don't understand why you can't hardcode the fallback itag to 22 like it's already the case on invidious.

latest_version is an endpoint for serving non-dash video formats, I don't want it to randomly serve a dash video format...

iBicha:
> I don't understand why you can't hardcode the fallback itag to 22 like it's already the case on invidious.

Because 22 is not available on all videos (especially on older videos). 

> latest_version is an endpoint for serving non-dash video formats, I don't want it to randomly serve a dash video format...

It will not be used for DASH. It will be used to serve a video format, in case DASH fails

unixfox:
Mmh still need some test in order to be sure that it doesn't break anything.

SamantazFox (approved):

src/invidious/routes/video_playback.cr
```diff
@@ -277,7 +277,11 @@ module Invidious::Routes::VideoPlayback
       return error_template(500, ex)
     end
 
-    fmt = video.fmt_stream.find(nil) { |f| f["itag"].as_i == itag } || video.adaptive_fmts.find(nil) { |f| f["itag"].as_i == itag }
+    if itag.nil?
+      fmt = video.fmt_stream[-1]
```
SamantazFox, line 281:
```suggestion
      fmt = video.fmt_stream[-1]?
```

for reference:
 - https://play.crystal-lang.org/#/r/epbz (causes an IndexError Exception)
 - https://play.crystal-lang.org/#/r/epc0 (return nil)

SamantazFox (changes requested):

src/invidious/routes/video_playback.cr
SamantazFox: That's in the event where a video doesn't have a progressive HTTP itag available (e.g in progress or recently ended livestreams)

SamantazFox:
@unixfox Basically, if no itag is provided, it takes the last format from the progressive HTTP formats.
So there's no risk of providing a DASH stream.
Also, Invidious won't be impacted, as we always provide the itag when requesting that endpoint.

src/invidious/routes/video_playback.cr
iBicha: That's a good idea, returning a 404 if the list is empty. Thanks!

unixfox (approved):

PR #3651, opened 2023-02-19
Allow to set a label for exempting from staling

unixfox:
I'll squash the PR in order to not have a strange commit name, just too lazy to update it through the cmd.


SamantazFox:
Oh, sorry, forgot to approve changes

PR #3656, opened 2023-02-21
encode username on callback

iBicha:
Follow up to #3603, I realized the username is not encoded

### DIFF
--- a/src/invidious/routes/account.cr
+++ b/src/invidious/routes/account.cr
@@ -262,7 +262,7 @@ module Invidious::Routes::Account
       end
 
       query["token"] = access_token
-      query["username"] = user.email
+      query["username"] = URI.encode_path_segment(user.email)
       url.query = query.to_s
 
       env.redirect url.to_s


SamantazFox:
Oh, that wasn't needed! The `HTTP::Params` class [already have all the logic needed to encode the values](https://crystal-lang.org/api/1.7.2/URI/Params.html#to_s%28%2A%2Cspace_to_plus%3ABool%3Dtrue%29-instance-method).

iBicha:
> Oh, that wasn't needed! The `HTTP::Params` class [already have all the logic needed to encode the values](https://crystal-lang.org/api/1.7.2/URI/Params.html#to_s%28%2A%2Cspace_to_plus%3ABool%3Dtrue%29-instance-method).

I thought so too, but the encoding was replacing space with "+". This made it not possible to differentiate between a username that has a space in it or has a "+" in it. e.g. "user name" and "user+name", they both end up being "user+name"

This makes the arg encoded twice, but that's also true for the token anyway.

PR #3658, opened 2023-02-23
Add ability to add proxy domains via ENV

89Q12:
This PR adds the ability to add a list of proxy domains used for proxying videos, the domain is randomly chosen. If no domains are configured the `HOST_URL` is used.
The domains can be configured using in either the config.yaml or ENV
`proxy_domains: ["..","...",".."]`

Also this implementation assumes https to be available on the proxy domains.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -98,6 +98,22 @@ db:
 domain:
 
 ##
+## A list of domains where external proxy servers. This is used
+## to build the urls used for proxing videos.
+## When none is set invidious will be used for proxiying.
+## This generally only needs to be set for bigger instances,
+## a example proxy url would be like ['proxy.example.com'].
+## Using a list urls might look like this:
+## ['proxy-eu1.example.com','proxy-eu2.example.com']
+## Invidious would then choose at random which request gets one
+## of the proxy servers
+##
+## Accepted values: a list of fully qualified domain names (FQDN)
+## Default: [""]
+##
+proxy_domains: [""]
+
+##
 ## Tell Invidious that it is behind a proxy that provides only
 ## HTTPS, so all links must use the https:// scheme. This
 ## setting MUST be set to true if invidious is behind a
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -88,6 +88,8 @@ class Config
   property hmac_key : String?
   # Domain to be used for links to resources on the site where an absolute URL is required
   property domain : String?
+  # Domain list of external proxy servers to be used, when multiple are set they will be randomly choosen
+  property proxy_domains : Array(String) = [] of String
   # Subscribe to channels using PubSubHubbub (requires domain, hmac_key)
   property use_pubsub_feeds : Bool | Int32 = false
   property popular_enabled : Bool = true
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -7,6 +7,11 @@ module Invidious::Routes::API::Manifest
     local = env.params.query["local"]?.try &.== "true"
     id = env.params.url["id"]
     region = env.params.query["region"]?
+    proxy_url = HOST_URL
+
+    if CONFIG.proxy_domains.size > 0 && local
+      proxy_url = ("https:///" + CONFIG.proxy_domains.[Random.rand(CONFIG.proxy_domains.size)])
+    end
 
     # Since some implementations create playlists based on resolution regardless of different codecs,
     # we can opt to only add a source to a representation if it has a unique height within that representation
@@ -29,7 +34,7 @@ module Invidious::Routes::API::Manifest
 
         if local
           uri = URI.parse(url)
-          url = "#{HOST_URL}#{uri.request_target}host/#{uri.host}/"
+          url = "#{proxy_url}#{uri.request_target}host/#{uri.host}/"
         end
 
         "<BaseURL>#{url}</BaseURL>"
@@ -42,7 +47,7 @@ module Invidious::Routes::API::Manifest
 
     if local
       adaptive_fmts.each do |fmt|
-        fmt["url"] = JSON::Any.new("#{HOST_URL}#{URI.parse(fmt["url"].as_s).request_target}")
+        fmt["url"] = JSON::Any.new("#{proxy_url}#{URI.parse(fmt["url"].as_s).request_target}")
       end
     end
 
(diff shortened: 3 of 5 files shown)


SamantazFox:
When using an external proxy URL, it would be better to redirect `/lastest_version` and `/api/manifest/*` to the given proxy domain, rather than the individual streams. That way, you ensure that the proxy server will fetch the video streams by itself and make sure that they're bound to its own IP (I'm not entirely sure if the streams are geo-restricted to a country, or to a smaller region).

11Tuvork28:
I'm not sure what do you exactly mean by that, but [http3-ytproxy](https://github.com/TeamPiped/http3-ytproxy/) or [piped-proxy](https://github.com/TeamPiped/piped-proxy) have obviously no logic for  `/lastest_version`.  Sure invidious could be used itself when compiled with api only basically but its not as lightweight or fast as the a fore mentioned proxy.
I get the problem though, converting to draft for that reason.

unixfox:
Well once invidious will have redis support I think compiling invidious in API only will become a viable solution for running proxies.

Or one can also run alternatives software of Invidious that support the same endpoints like:
- https://github.com/unixfox/invidious-stripped-down
- https://github.com/11Tuvork28/rusted-invidious
- their own custom implementation?

Technically with this PR someone has the choice of running what they want.

----

Like @SamantazFox said I would also like to have as a feature the redirection to another domain for the endpoints `/latest_version` and `/api/manifest/*`?

11Tuvork28:
Okay kinda get it, well then the discussion on matrix was useless but sure will do that.

SamantazFox:
@11Tuvork28 What do you think of the following?

Mode 1 - default, use invidious' integrated proxy (current situation)
```yaml
video_proxy:
  redirect_mode: local
  # domains: <HOST_URL>
```

Mode 2 - redirects only the `/videoplayback` endpoint (e.g use Piped's `http3-ytproxy`)
```yaml
video_proxy:
  redirect_mode: partial
  domains:
    - a.example.com
    - b.example.com
```

Mode 3 - redirects all video endpoints (implies that the proxy server supports `/latest_version` and can provide DASH/HLS manifests)
```yaml
video_proxy:
  redirect_mode: full
  domains:
    - a.example.com
    - b.example.com
```

---

As an alternative:
```yaml
media_proxy:
  # Redirect for `/videoplayback`, and possibly `/latest_version` and `/api/manifest/*`
  video_redirect_mode: # like above
  video_proxy_domains:
    - a.example.com

  # Redirect for `/vi/*`, `/vi_webp/*`, `/sb/*`, `/ggpht/*`, `/s_p/*` and `/yts/img/*`
  images_redirect: # true/false
  images_proxy_domains:
    - b.example.com
```

unixfox:
I don't think there is a point for maintaining two kinds of redirect because **you need** to serve the content from the same IP address that generated the request to youtube servers.

11Tuvork28:
@SamantazFox 
That's a nice solution, I like it and would implement it that way including the image part.
I would however change the structure:
```yaml
media_proxy:
  # Redirect for `/videoplayback`, and possibly `/latest_version` and `/api/manifest/*`
  external_video_proxy: 
    enabled: true/false
    proxy_mode: # like above
    proxy_domains:
      - a.example.com

  # Redirect for `/vi/*`, `/vi_webp/*`, `/sb/*`, `/ggpht/*`, `/s_p/*` and `/yts/img/*`
  external_image_proxy:
    enabled: true/false
    append_host_to_uri: true/false
    image_proxy_domains:
      - b.example.com

```
Reason behind this is that the file structure is clearer in my opinion and `append_host_to_uri` would be needed for Piped's proxy implementations.
<hr>
@unixfox

No you don't, my instance produces requests from one of those IP's:
- 167.235.242.98(GER)
- 159.69.152.204(GER)
- 65.108.249.34(HEL)
- 167.235.202.120(GER)
- 116.203.183.22(GER)
- 95.217.189.191(HEL)

While the proxy server is on `45.83.107.127`(GER), it works great. Those IP's aren't even in the same country but it somehow works.
Here proof:
image
What you are seeing is a port forward to the pod in Helsinki with IP `65.108.249.34` making requests to  `45.83.107.127`, using dash that go through just fine.

unixfox:
> No you don't, my instance produces requests from one of those IP's:

Yes you need to serve it from the same IP, it's for encrypted videos (music videos and more) like this one: https://www.youtube.com/watch?v=pRpeEdMmmQ0. For example this video doesn't work on your server: https://watch.thekitty.zone/watch?v=pRpeEdMmmQ0

Related: https://github.com/iv-org/invidious/issues/2850

11Tuvork28:
I get that(Have read the issue before) and I tried this and a fw others I found in ytdl issues. They all work fine what can I say?
https://watch.thekitty.zone/watch?v=pRpeEdMmmQ0&quality=dash&local=1
https://watch.thekitty.zone/watch?v=qOtgxpDyGPw&quality=dash&local=1
https://watch.thekitty.zone/watch?v=AAbokV76tkU&quality=dash&local=1

In any case I still see the `image_proxy` as relevant, because invidious isn't MT and static content serving is better done via performant proxies.

SamantazFox:
> I don't think there is a point for maintaining two kinds of redirect because **you need** to serve the content from the same IP address that generated the request to youtube servers.
> 
> People who will use this functionality will want to serve a working experience to their users, so it doesn't make sense to redirect to the incorrect proxy resulting in a playback error.

Imo, both redirect options make sense. I see the following use cases:
 - `Invidious` + `http3-ytproxy` on the same machine, only to benefit from HTTP/3
 - `invidious` + multiple servers running `http3-ytproxy` (or similar) in the same rack/DC
 - `invidious` with a bunch of distant video servers, maybe in another country, likely managed by another pice of software (each server provides its own stream URLs)

> I would however change the structure:

Ah, sure, of course, it was only a quick draft. I definitely prefers your solution!

unixfox:
> * `Invidious` + `http3-ytproxy` on the same machine, only to benefit from HTTP/3

Why would you want to do that? HTTP3 is literally useful for high latency connections between a client and a server. That doesn't make any sense here.

> * `invidious` + multiple servers running `http3-ytproxy` (or similar) in the same rack/DC

In this configuration you can always have a reverse proxy that serve the requests for the endpoints `/videoplayback`, `/latest_version`, `dash manifest` (and more) to random servers.

Anyway, off to my real suggestion:

I would very much prefer, as a first batch, to keep this functionality very simple, simple redirect for both endpoints `/latest_version` and `dash manifest` to randomly chosen domains. We set the setting parameters so that they can be expanded later for much more advanced usages, like the suggestion written in https://github.com/iv-org/invidious/pull/3658#issuecomment-1454860052

This way we can provide a functionality that for the moment suit everyone and because it will be "simple" to code we can ship this functionality in the coming weeks.

89Q12:
FYI: Not sure when  if or when I will come around to actually implement it.

PR #3663, opened 2023-02-26
docker-compose fix default port

vinyll:
The default port is a very common and conflicts easily with services on the server.
Offering to serve it in a override.yml file allows the user to have an easy access to it while still permitting flexibility to the host.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,0 +1,6 @@
+version: "3"
+
+services:
+  invidious:
+    ports:
+      - "127.0.0.1:3000:3000"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,8 +12,8 @@ services:
       context: .
       dockerfile: docker/Dockerfile
     restart: unless-stopped
-    ports:
-      - "127.0.0.1:3000:3000"
+    expose:
+      - "3000"
     environment:
       # Please read the following file for a comprehensive list of all available
       # configuration options and their associated syntax:


unixfox:
Can't a user currently override these ports using the `docker-compose.override.yml` file? By the way, it's a development docker-compose.yml, so it is not recommended to use this one.

You should use this one instead: https://docs.invidious.io/installation/#docker-compose-method-production

vinyll:
Unfortunately a user can't overwrite the port, and that's probably why [the doc says](https://docs.invidious.io/installation/#docker-compose-method-production) "replace 127.0.0.1:3000:3000 with 3000:3000 under the ports: section.".
Overriding will the port set in the main file won't remove `127.0.0.1:3000:3000`; it will remain and still conflict if the port is used, just another port will be added.

You're right about the name, I therefore renamed following [the Docker documentation](https://docs.docker.com/compose/extends/#different-environments).
Wondering if a default port for development (therefore also having a `override.yaml` file) would make sense.

Any thought?

SamantazFox:
The container's port is 3000 because of the `EXPOSE` instruction in the `Dockerfile`.

But you can change your mapping in the docker-compose file like so:
```yaml
    ports:
      - "127.0.0.1:45678:3000"
```

So I don't see the point of an override file?

unixfox:
Closing as it's a development docker-compose, for production it is recommended to use the docker-compose from https://docs.invidious.io/installation/

Issue #3660, opened 2023-02-24
[Bug] Video descriptions sometimes don't load.


unnamed24:
It was noticed that at random times and it is not clear from what situations, the descriptions for the video are not loaded, which leaves an empty field instead of the description.
Video example:
https://inv.odyssey346.dev/watch?v=mPypEpPB2UQ
Screenshot of what the video description field looks like at the moment:
![unnamed](https://user-images.githubusercontent.com/124290392/221259313-059f23f5-cc84-41d7-99e1-4e073b8df976.png)
However, at the same time, on YouTube, the descriptions are loaded as expected.

----

Bounties for fixing this issue:
- @LennyPenny $5: `send me a message from a mastodon account or email linked in your github profile linked to the commit once merged`
- @99thim $5: `Same as LennyPenny, happy to offer $5 too to have this fixed. Would pay with bitcoin`

garoto:
Observing this behavior as well on four public instances and with different, popular videos.

SamantazFox:
Youtube has changed their description format ...

Here's how it looks now:
```json
"attributedDescription": {
  "content": "Save 10% and Free Worldwide Shipping at The Ridge by using offer code WAN at https://www.ridge.com/LINUS\n\nVisit https://www.squarespace.com/WAN and use offer code WAN for 10% off\n\nJumpCloud offers an Open Directory Platform that makes it possible to unify your technology stack across identity, access, and device management. Learn more here: https://lmg.gg/JumpCloud\n\nPodcast Download: TBD\n\nTimestamps: TBD",
  "commandRuns": [
    {
      "startIndex": 77,
      "length": 27,
      "onTap": {
        "innertubeCommand": {
          "commandMetadata": {
            "webCommandMetadata": {
              "url": "https://www.youtube.com/redirect?q=https%3A%2F%2Fwww.ridge.com%2FLINUS&v=_FqH1rqQrQg",
              "webPageType": "WEB_PAGE_TYPE_UNKNOWN",
              "rootVe": 83769
            }
          },
          "urlEndpoint": {
            "url": "https://www.youtube.com/redirect?q=https%3A%2F%2Fwww.ridge.com%2FLINUS&v=_FqH1rqQrQg",
            "target": "TARGET_NEW_WINDOW",
            "nofollow": true
          }
        }
      }
    },
    {
      "startIndex": 112,
      "length": 31,
      "onTap": {
        "innertubeCommand": {
          "commandMetadata": {
[...]

PR #3668, opened 2023-03-03
Fix description boxes smaller than 3 lines

amogusussy:
If a video has no description, (without this commit) the description box will still take up 8.3em, even if there's no content in it.

This PR fixes that issue.

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -515,7 +515,7 @@ hr {
 
 #descexpansionbutton ~ div {
   overflow: hidden;
-  height: 8.3em;
+  max-height: 8.3em;
 }
 
 #descexpansionbutton:checked ~ div {


amogusussy:
Example without this PR:
https://i.imgur.com/F7BoIfw.png

With this PR:
https://imgur.com/5j55WAG.png

pklion:
I think it's caused by [#3660](https://github.com/iv-org/invidious/issues/3660) and it shouldn't be fixed with CSS, I suppose.

SamantazFox:
> I think it's caused by #3660 and it shouldn't be fixed with CSS, I suppose.

Yes, this is indeed caused by #3660
for the record, here is a video that has actually no description: https://test.invidious.io/watch?v=iuevw6218F0

However, this fix can still be relevant for videos with a one or two lines description!

pklion:
> However, this fix can still be relevant for videos with a one or two lines description!
I agree.

unixfox (approved):

SamantazFox:

PR #3687, opened 2023-03-13
Implement sticky player box on mobile devices

AstudilloJuanF:


SamantazFox:
Hello! Can you please post screenshots showing the change?

AstudilloJuanF:
> Hello! Can you please post screenshots showing the change?

Hello! Sure

https://user-images.githubusercontent.com/57147770/225466276-8bbb1bf7-39b7-4793-bcb4-6f5659a777ef.mp4

unixfox:
Personally I would prefer to have that as a setting instead of being by default.

AstudilloJuanF:
> Personally I would prefer to have that as a setting instead of being by default.

I think the feature would be very convenient for mobile users. Perhaps it could be implemented as a setting, but active as default.

unixfox:
It's such a drastic change that I don't want it enabled by default. IMO it makes the navigation on small screens very unpleasant.

AstudilloJuanF:
We agree to disagree on UX. But I agree with you on the magnitude on the proposed change.

You are right, it should be disabled by default. I will close this pull request and create a new one with the sticky player box changes as an optional feature on preferences.

unixfox:
You didn't have to close this pull request. You can force push using the git command line.

PR #3692, opened 2023-03-16
API: Add hashtag endpoint

ChunkyProgrammer:
This PR will allow the hashtag endpoint to be parsed.

Example request:
`/api/v1/hashtag/hi`
Example response
image

Note:
I did not add support for pagination in the api as it seems to be broken when using the UI.

Example (page 1):
image

Example (page 2):
image

### DIFF
--- a/src/invidious/routes/api/v1/search.cr
+++ b/src/invidious/routes/api/v1/search.cr
@@ -59,10 +59,8 @@ module Invidious::Routes::API::V1::Search
   def self.hashtag(env)
     hashtag = env.params.url["hashtag"]
 
-    # page does not change anything.
-    # page = env.params.query["page"]?.try &.to_i?|| 1
+    page = env.params.query["page"]?.try &.to_i? || 1
 
-    page = 1
     locale = env.get("preferences").as(Preferences).locale
     region = env.params.query["region"]?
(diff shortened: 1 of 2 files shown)


src/invidious/routes/api/v1/search.cr
```diff
@@ -55,4 +55,30 @@ module Invidious::Routes::API::V1::Search
...
+    env.response.content_type = "application/json"
+
+    begin
+      results = Invidious::Hashtag.fetch(hashtag, page, region)
+    rescue ex
+      return error_json(400, ex)
+    end
+
+    JSON.build do |json|
+      json.array do
```
ChunkyProgrammer, line 77: should I make an `items` property to contain the videos incase we decide to show more hashtag information in the future or decide to use continuation tokens?

SamantazFox: Yes, that's a good idea! Maybe `results` would be more fitting, though?

SamantazFox:
Hashtag needs to be fixed first.
See: #3778

ChunkyProgrammer:
Switched to draft to add support for continuations

SamantazFox (approved):

Issue #3637, opened 2023-02-15
[Bug] Missing JSON element "reelPlayerHeaderSupportedRenderers" (BrokenTubeException)


SamantazFox:
**Describe the bug**
JSON parsing error on the "shorts" tab on some specific channel(s)

**Steps to Reproduce**
1. Go to `/channel/UC2ped54WsnVpSuKUw7HikOg/shorts`
2. See error

**Logs**
Date: `2023-02-15T20:34:14Z`
Version: `2023.01.21-4e3884ca @ master`

<details>
<summary>Backtrace</summary>
<p>
   
```
Missing JSON element "reelPlayerHeaderSupportedRenderers" (BrokenTubeException)
  from invidious.git/src/invidious/channels/videos.cr:159:9 in 'get_shorts'
  from invidious.git/src/invidious/routes/channels.cr:60:32 in 'shorts'
  from invidious.git/src/invidious/routing.cr:121:34 in '->'
  from /usr/share/crystal/src/primitives.cr:266:3 in '->'
  from /usr/share/crystal/src/primitives.cr:266:3 in 'process_request'
  from invidious.git/lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:212:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:94:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:145:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from invidious.git/src/invidious/helpers/handlers.cr:70:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
[...]

PR #3700, opened 2023-03-22
Fix parsing shorts on channel page

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/3637

As I am not 100% sure if they phased the old type out of all channels, I kept it in just in case. For now, we are able to get author name and author id from fallback but we cannot get published date.

image

Test:

go to: `/channel/UCghooJjCGtg2Kw2PSB9B0NA/shorts`
go to `/api/v1/channels/UCghooJjCGtg2Kw2PSB9B0NA/shorts`

### DIFF
--- a/src/invidious/channels/videos.cr
+++ b/src/invidious/channels/videos.cr
@@ -127,38 +127,15 @@ module Invidious::Channel::Tabs
   #  Shorts
   # -------------------
 
-  private def fetch_shorts_data(ucid : String, continuation : String? = nil)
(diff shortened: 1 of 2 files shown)


unixfox:

SamantazFox (approved):

SamantazFox:
view count extraction would need to be fixed, too (it's missing the multiplier)

unixfox (approved):
No issues on my side, I have been testing it for more than one week.

Issue #3706, opened 2023-03-27
[Bug] Missing hash key: "runs" (KeyError)


Perflyst:
**Describe the bug**
Error on video eBGIQ7ZuuiU (<- rick roll)

**Steps to Reproduce**
Open video eBGIQ7ZuuiU https://redirect.invidious.io/watch?v=eBGIQ7ZuuiU

**Logs**
2023-03-27 07:07:50 UTC [error] get_video: eBGIQ7ZuuiU : Missing hash key: "runs"

**Additional context**
<details>
<summary>Backtrace</summary>
<p>
   
```
Missing hash key: "runs" (KeyError)
  from /usr/share/crystal/src/hash.cr:1080:9 in '[]'
  from /usr/share/crystal/src/json/any.cr:139:5 in 'parse_video_info'
  from src/invidious/videos/parser.cr:92:12 in 'extract_video_info:video_id'
  from src/invidious/videos.cr:377:10 in 'fetch_video'
  from src/invidious/videos.cr:365:13 in 'get_video:region'
  from src/invidious/routes/watch.cr:63:15 in 'handle'
  from lib/kemal/src/kemal/route.cr:13:9 in '->'
  from src/invidious/helpers/handlers.cr:30:37 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
  from /usr/share/crystal/src/http/server.cr:500:5 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from ???
[...]

PR #3709, opened 2023-03-28
Videos: ignore empty music entries

ChunkyProgrammer:
There is an empty music track in the video provided in the issue (https://www.youtube.com/watch?v=eBGIQ7ZuuiU)
This causes an error to occur that prevents the page from rendering.
This PR will exclude the missing track from being parsed.

YouTube screenshot for context:
empty track followed by the correct "Never gonna give you up" track

closes https://github.com/iv-org/invidious/issues/3706


SamantazFox:
Thank you!

Issue #3704, opened 2023-03-24
[Enhancement] Allow autocomplete (from the browser history) in the search box


raphj:
**Is your enhancement request related to a problem? Please describe.**

I'd like to be able to use the browser history for the search bar. Especially on touchscreen where it is more difficult to type. It's annoying to have to type things completely again and again.

**Describe the solution you'd like**

This is caused by the `autocompete="off"` attribute in `views/components/search_box.ecr`.

```
                <input type="search" id="searchbox" autocomplete="off" autocorrect="off"
                autocapitalize="none" spellcheck="false" <% if autofocus %>autofocus<% end %>
                name="q" placeholder="<%= translate(locale, "search") %>"
                title="<%= translate(locale, "search") %>"
                value="<%= env.get?("search").try {|x| HTML.escape(x.as(String)) } %>">
```

I'd like that this attribute to be dropped.

@SamantazFox it seems this comes from your commit 68cbc1181 to fix #1107 (nice commit documentation! thanks for this). `spellcheck="false"` or `autocorrect="off"` seem necessary to avoid annoying autocorrection, do you remember why you introduced `autocomplete="off"` as well?

**Describe alternatives you've considered**

- a setting to disable `autocomplete=off`
- a local commit that disables it, since I'm currently building invidious myself (but that might not always be the case in the future)
- a user script that patches the search box
- an Invidious feature that remembers, per user, their search history but doesn't seem ideal to store such sensitive data server side.

**Additional context**

I can provide a PR if it helps.

unixfox:
Do note that there is no autocomplete system in the backend, so even if you were to enable it, nothing would be able to provide the autocompletion to the browser.

For the autocompletion to work, we need someone to implement this feature on the backend side.

raphj:
No no, when I remove the attribute, the browser suggests stuff I already typed there :-)

unixfox:
Then we are not talking about the same autocompletion, I'm talking about the autocompletion that YouTube servers would give. The same behavior that you would have when typing in the search bar of Google for example.

raphj:
I'm referring to the native feature of browsers that is disabled by `autocomplete=off`

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

Google would have it, but they disable it to provide their own (that indeed requires backend support).

raphj:
There's a risk browsers auto-fill the search bar with some previous search, I think it can be avoided by also setting `value=""`

[...]

PR #3724, opened 2023-04-02
Search: Allow browser suggestions

raphj:
`autocomplete=off` prevents the browser from suggesting previous searches. This PR removes this attribute.

Of course this does not suggests searches previously done on other devices / browsers, but suggestions from local search history are still very helpful, especially on touchscreens / small devices like phones where typing can be tedious, especially since autocorrect is disabled on this field. 

I have been using this for a couple weeks on Desktop Firefox on Linux (laptop and PinePhone) without seeing negative effects from this.

See #3704

### DIFF
--- a/src/invidious/views/components/search_box.ecr
+++ b/src/invidious/views/components/search_box.ecr
@@ -1,6 +1,6 @@
 <form class="pure-form" action="/search" method="get">
 	<fieldset>
-		<input type="search" id="searchbox" autocomplete="off" autocorrect="off"
+		<input type="search" id="searchbox" autocorrect="off"
 		autocapitalize="none" spellcheck="false" <% if autofocus %>autofocus<% end %>
 		name="q" placeholder="<%= translate(locale, "search") %>"
 		title="<%= translate(locale, "search") %>"


raphj:
Still relevant. This simple and tiny change makes Invidious much more usable for me.

SamantazFox (approved):

SamantazFox:

raphj:
Nice, thanks @SamantazFox :-)

ManiMatter:
Hi,

I find the autocomplete from the browser history rather annoying. It populates the search with completely unrelated stuff.

Two questions:
1) Can we make the "allow autofill" from browser history a flag that we can turn on / off via a config.yml* setting
2) What would actually be helpful is to see the youtube type-ahead suggestions 

What would be helpful: **Youtube type-ahead suggestions**
Bildschirmfoto 2023-07-27 um 14 40 12

What is useless (to me): **Browser history auto-fill**
Bildschirmfoto 2023-07-27 um 14 47 00

*with config.yml i refer to the settings that can be specified when launching invidious.
https://github.com/iv-org/invidious/blob/master/config/config.example.yml

Issue #446, opened 2019-03-27
private invidious?


johnwmail:
Hi, is it possible config invidious only allow watch/listen video for logged in user?
Thanks.

cyaniventer:
I don't think this feature exists but you can add a simple rule to disallow users which are not logged in. (@omarroth) This might be good feature request, some admins may want to restrict viewing to logged in users only.

johnwmail:
Hi, how to add "simple rule disallow users which are not logged in"?
thanks.

Perflyst:
What can **not** logged in users on such instances do? Why do you not just add a simple http auth?

johnwmail:
> What can **not** logged in users on such instances do? Why do you not just add a simple http auth?

1. Maybe just see the login page(input username and password).
2. Because http auth can not use cookie, and need to login everytime.
Thanks.

cyaniventer:
I don't know the language invidious is written in, maybe look for code where it checks if user is logged in and remove everything for `else` block and print you must be logged in :? 

You can also make it a Tor hidden service and protect it. See https://matt.traudt.xyz/p/FgbdRTFr.html & https://riseup.net/en/security/network-security/tor/onionservices-best-practices

look for `HiddenServiceAuthorizeClient`

omarroth:
Something like:
```diff
diff --git a/src/invidious.cr b/src/invidious.cr
index f93f17e..74b2a8e 100644
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -231,6 +231,12 @@ before_all do |env|
   end
 
   env.set "current_page", URI.escape(current_page)
+
+  unregistered_path_whitelist = {"/", "/login", "/licenses", "/privacy"}
+  if !env.get?("user") && !unregistered_path_whitelist.includes?(env.request.path)
+    env.response.headers["Location"] = "/"
+    halt env, status_code: 302
+  end
 end
 
 get "/" do |env|
```

Should produce the desired functionality.

You can modify the `unregistered_path_whitelist` as necessary, and the `env.response.headers["Location"] = "/"` to redirect all pages to e.g. `/login`.

I don't see something like this being merged into master since it's not very privacy friendly and I'm having trouble seeing a large enough use case to justify including it as a feature.

johnwmail:
> ```diff
> licenses
> ```

Hi, It work with redirect to /login for me, thanks. 

one more question, how to whitelist path like: "/vi/XU7YV1tWr10/mqdefault.jpg" ?
[...]

PR #3728, opened 2023-04-07
login_only

johnwmail:
PR refer to https://github.com/iv-org/invidious/issues/446

### DIFF
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -84,6 +84,7 @@ class Config
 
   # Used to tell Invidious it is behind a proxy, so links to resources should be https://
   property https_only : Bool?
+  property login_only : Bool?
   # HMAC signing key for CSRF tokens and verifying pubsub subscriptions
   property hmac_key : String?
   # Domain to be used for links to resources on the site where an absolute URL is required
--- a/src/invidious/routes/before_all.cr
+++ b/src/invidious/routes/before_all.cr
@@ -148,5 +148,11 @@ module Invidious::Routes::BeforeAll
     end
 
     env.set "current_page", URI.encode_www_form(current_page)
+
+    unregistered_path_whitelist = {"/", "/login", "/licenses", "/privacy"}
+    if !env.get?("user") && !unregistered_path_whitelist.includes?(env.request.path) && CONFIG.login_only
+      env.response.headers["Location"] = "/login"
+      haltf env, status_code: 302
+    end
   end
 end


unixfox:
Don't push other changes that have nothing to do with the feature.

You have added docker related things

johnwmail:
Sorry, this is my first time submit PR, meanwhile I am learning CI/CD things.
Anyway, should be reverted, thank you.

johnwmail:
Would you please review it again, thank you.

src/invidious/routes/before_all.cr
```diff
@@ -148,5 +148,11 @@ module Invidious::Routes::BeforeAll
     end
 
     env.set "current_page", URI.encode_www_form(current_page)
+
+    unregistered_path_whitelist = {"/", "/login", "/licenses", "/privacy"}
+    if !env.get?("user") && !unregistered_path_whitelist.includes?(env.request.path) && CONFIG.login_only
```
LennyPenny, line 153: don't know how crystal deals with this but its probably best to check the config flag first in order to not redundantly check this the whitelist on every request

johnwmail: Ok, I have updated the patch with your suggestion, thanks.

SamantazFox:
Thanks for the PR, and sorry for the delay it took me to review.

This subject is a tough one: By adding that check at the bottom of `before_all`, you still allow an external user to load images and videos. If you want to entirely restrict access, you'd need to put that code before `return if` (on line 64), which invloves re-thinking how auth is done (at the moment, we don't check if the user is authenticated on content proxy routes for performance reasons, at is saves us some DB calls).

Ideally, we should use a local cache to store user sessions (see the [kemalcr/kemal-session](https://github.com/kemalcr/kemal-session) shard), so that we don't have to call the DB all the time on every request.

SamantazFox:
bump?

johnwmail:
Sorry for late reply, I have updated the code before "return if (line 64)", please review, thank you.

rix1337:
I built the patch from this specific pull request and its working: [rix1337/docker-utube/general](https://hub.docker.com/repository/docker/rix1337/docker-utube/general) (arm image only)

Its broken however for clients like Yattee (even if logged in and with proxy enabled).

rix1337:
Also the health check from docker compose has now switched to unhealthy permanently:
```
    healthcheck:
      test: wget -nv --tries=1 --spider http://127.0.0.1:3000/api/v1/comments/jNQXAC9IVRw || exit 1
      interval: 30s
      timeout: 5s
      retries: 2
```

rix1337:
Related Yattee issue: https://github.com/yattee/yattee/issues/552

Robiqaz (approved):
please release. Was looking forward to this.

thank you.

src/invidious/config.cr
```diff
@@ -84,6 +84,7 @@ class Config
 
   # Used to tell Invidious it is behind a proxy, so links to resources should be https://
   property https_only : Bool?
+  property login_only : Bool?
```
Daviey, line 87: It would be good to add a descriptive comment on what this does, similar to the others.

Daviey (changes requested):
The change appears functionally good to me, except it does introduce an unhanded error when loading the loading page whilst not logged in:
```
2023-11-09 14:59:59 UTC [info] 302 GET / 441.37µs
Exception: Closed stream (IO::Error)
  from /usr/share/crystal/src/io.cr:121:5 in 'check_open'
  from /usr/share/crystal/src/compress/gzip/writer.cr:74:15 in 'write'
  from /usr/share/crystal/src/http/server/response.cr:97:7 in 'write'
  from /usr/share/crystal/src/io.cr:484:7 in 'write_string'
  from /usr/share/crystal/src/string.cr:5263:5 in 'to_s'
  from /usr/share/crystal/src/io.cr:177:5 in '<<'
  from /usr/share/crystal/src/io.cr:191:5 in 'print'
  from src/invidious/helpers/handlers.cr:40:5 in 'process_request'
  from lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:212:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:94:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:145:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:70:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/ext/kemal_static_file_handler.cr:162:16 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/exception_handler.cr:8:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/logger.cr:17:35 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'call'
  from /usr/share/crystal/src/http/server/request_processor.cr:51:11 in 'process'
  from /usr/share/crystal/src/http/server.cr:517:5 in 'handle_client'
  from /usr/share/crystal/src/http/server.cr:470:13 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:98:34 in '->'
  from ???

2023-11-09 14:59:59 UTC [info] 302 GET /feed/popular 10.91ms
2023-11-09 14:59:59 UTC [info] 200 GET /login 839.7µs
```

src/invidious/routes/before_all.cr
```diff
@@ -100,6 +88,24 @@ module Invidious::Routes::BeforeAll
       end
     end
 
+    unregistered_path_whitelist = {"/", "/login", "/licenses", "/privacy"}
```
stonerl, line 91:
```suggestion
    unregistered_path_whitelist = {"/", "/api/v1/comments/jNQXAC9IVRw", "/api/v1/videos/dQw4w9WgXcQ", "/login", "/licenses", "/privacy"}
```

Adding `/api/v1/comments/jNQXAC9IVRw` allows the health-check to succeed and `/api/v1/videos/dQw4w9WgXcQ` to add the instance to, e.g., Yattee.

stonerl:
There is another unhandled exeption:

```
2023-11-09 18:02:26 UTC [info] 302 POST /feed/webhook/v1:1699275109:79a13f9f:051c7b138f7ac06da40a268a20537f3039d1810a 37.35ms
Unhandled exception in spawn: Nil assertion failed (NilAssertionError)
  from /usr/share/crystal/src/nil.cr:113:7 in 'not_nil!'
  from /usr/share/crystal/src/nil.cr:109:3 in 'not_nil!'
  from src/invidious/routes/feeds.cr:412:14 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:98:34 in '->'
  from ???
```

stonerl:
> The change appears functionally good to me, except it does introduce an unhanded error when loading the loading page whilst not logged in:
> 
> ```
> 2023-11-09 14:59:59 UTC [info] 302 GET / 441.37µs
> Exception: Closed stream (IO::Error)
>   from /usr/share/crystal/src/io.cr:121:5 in 'check_open'
>   from /usr/share/crystal/src/compress/gzip/writer.cr:74:15 in 'write'
>   from /usr/share/crystal/src/http/server/response.cr:97:7 in 'write'
>   from /usr/share/crystal/src/io.cr:484:7 in 'write_string'
>   from /usr/share/crystal/src/string.cr:5263:5 in 'to_s'
>   from /usr/share/crystal/src/io.cr:177:5 in '<<'
>   from /usr/share/crystal/src/io.cr:191:5 in 'print'
>   from src/invidious/helpers/handlers.cr:40:5 in 'process_request'
>   from lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from src/invidious/helpers/handlers.cr:212:5 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from src/invidious/helpers/handlers.cr:94:12 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from src/invidious/helpers/handlers.cr:145:12 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from src/invidious/helpers/handlers.cr:70:5 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from src/ext/kemal_static_file_handler.cr:162:16 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from lib/kemal/src/kemal/exception_handler.cr:8:7 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from src/invidious/helpers/logger.cr:17:35 in 'call'
>   from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
>   from lib/kemal/src/kemal/init_handler.cr:12:7 in 'call'
>   from /usr/share/crystal/src/http/server/request_processor.cr:51:11 in 'process'
>   from /usr/share/crystal/src/http/server.cr:517:5 in 'handle_client'
>   from /usr/share/crystal/src/http/server.cr:470:13 in '->'
>   from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
>   from /usr/share/crystal/src/fiber.cr:98:34 in '->'
>   from ???
> 
> 2023-11-09 14:59:59 UTC [info] 302 GET /feed/popular 10.91ms
> 2023-11-09 14:59:59 UTC [info] 200 GET /login 839.7µs
> ```

This unhandled exception also happens when logged in.

src/invidious/routes/before_all.cr
```diff
@@ -100,6 +88,24 @@ module Invidious::Routes::BeforeAll
...
+    if CONFIG.login_only && !env.get?("user") && !unregistered_path_whitelist.includes?(env.request.path)
```
stonerl, line 92:
```suggestion
    unregistered_path_whitelist = {"/api/v1/comments/jNQXAC9IVRw", "/api/v1/videos/dQw4w9WgXcQ", "/feed/webhook/v1:", "/login", "/licenses", "/opensearch.xml", "/privacy"}
    if CONFIG.login_only && !env.get?("user") && !unregistered_path_whitelist.any? { |r| env.request.path.starts_with? r }
```

Changing it to this comes with the advantage, that the login page is opened first. It also eliminates some `302` and lets the health-check succeed. Furthermore, it allows adding the instance and logging in with an account in third-party apps like Yattee.

Daviey: This is good, but I also found that Clipious client requires `/api/v1/stat` when adding a new server to the client.  It loads this to check it is a valid server, and refuses if it doesn't work (currently it returns a 500 error, which should be a bug anyway).  However, even if this additional endpoint is added the client doesn't load Popular and Trending (default pane) because it makes these requests unauthenticated (will need to raise this as an issue with Clipious)

src/invidious/routes/before_all.cr
```diff
@@ -100,6 +88,24 @@ module Invidious::Routes::BeforeAll
...
+      env.response.headers["Location"] = "/login"
+      haltf env, status_code: 302
+    end
```
stonerl, line 95:
```suggestion
    unregistered_path_whitelist = {
      "/api/v1/comments/jNQXAC9IVRw",
      "/api/v1/stats",
      "/api/v1/videos/dQw4w9WgXcQ",
      "/feed/webhook/v1:",
      "/licenses",
      "/login",
      "/opensearch.xml",
      "/privacy"
      }

    if CONFIG.login_only && !env.get?("user") && !unregistered_path_whitelist.any? { |r| env.request.path.starts_with? r }
      env.response.headers["Location"] = "/login"
      haltf env, status_code: 401
    end
```

- Changing it to this comes with the advantage, that the login page is opened first.
- The API whitelisting allows third-clients to connect and add the instance and login to an existing account
- Changed the status code from 302 (Found) to 401 (Unauthorized) which is more appropriate in this case.

stonerl, line 95:
```suggestion
    unregistered_path_whitelist = {
      "/api/v1/comments/jNQXAC9IVRw",
      "/api/v1/stats",
      "/api/v1/videos/dQw4w9WgXcQ",
      "/feed/webhook/v1:",
      "/licenses",
      "/login",
      "/opensearch.xml",
      "/privacy"
      }

    if CONFIG.login_only && !env.get?("user") && !unregistered_path_whitelist.any? { |r| env.request.path.starts_with? r }
      env.response.headers["Location"] = "/login"
      haltf env, status_code: 401
    end
```

- Changing it to this comes with the advantage, that the login page is opened first.
- The API whitelisting allows third-clients to connect and add the instance and login to an existing account
- Changed the status code from 302 (Found) to 401 (Unauthorized) which is more appropriate in this case.

src/invidious/routes/before_all.cr
stonerl: Added your input to a new suggestion.

Daviey:
Third party apps using the API are depending on some endpoints being unauthenticated such as /api/v1/popular and /api/v1/trending, currently there are giving a 500 error (but no traceback), and I haven't investigated the cause.  These endpoints in https://github.com/iv-org/invidious/tree/master/src/invidious/routes/api/v1 are divided between Authenticated and others, where the others are implied to be Public.  I tried adding authentication headers to the requests in Clipious, but the server side component doesn't seem to be expecting auth'.

Whilst adding  "/api/v1/stats" facilitates login, the actual app' isn't really functional.  So I think we state that `api/v1` has a contract of having relatively open access, which makes me think that to implement this change, we either need to disable API access or create a v2.  But someone closer to the project should really add their informed view of this!

stonerl:
Well, these two endpoints could be added to the whitelist as well. Since video playback still wouldn't be possible without logging in.

Daviey:
> Well, these two endpoints could be added to the whitelist as well. Since video playback still wouldn't be possible without logging in.

True, I assumed the intent of login was to reduce data leak... but popular and trending isn't really sensitive.

rix1337:
> Third party apps using the API are depending on some endpoints being unauthenticated such as /api/v1/popular and /api/v1/trending, currently there are giving a 500 error (but no traceback), and I haven't investigated the cause. These endpoints in https://github.com/iv-org/invidious/tree/master/src/invidious/routes/api/v1 are divided between Authenticated and others, where the others are implied to be Public. I tried adding authentication headers to the requests in Clipious, but the server side component doesn't seem to be expecting auth'.
> 
> Whilst adding "/api/v1/stats" facilitates login, the actual app' isn't really functional. So I think we state that `api/v1` has a contract of having relatively open access, which makes me think that to implement this change, we either need to disable API access or create a v2. But someone closer to the project should really add their informed view of this!

Just adding that after I raised my issue with Yattee they were able to quickly produce a fix, so your findings may be isolated to Clipious.

The patch from this pr works perfectly in combination with the patch from Yattee, when I enable login_only.

stonerl:
> > Well, these two endpoints could be added to the whitelist as well. Since video playback still wouldn't be possible without logging in.
> 
> True, I assumed the intent of login was to reduce data leak... but popular and trending isn't really sensitive.

I added them to the suggestion, as @rix1337 pointed out, Yattee does show the Trending and Popular (if enabled) when `login_only: true`. So I guess Clipious might need to make some modifications to allow for this to work without having these endpoints exposed via the whitelist.

src/invidious/routes/before_all.cr
```diff
@@ -100,6 +88,24 @@ module Invidious::Routes::BeforeAll
       end
     end
 
+    unregistered_path_whitelist = {"/", "/login", "/licenses", "/privacy"}
+    if CONFIG.login_only && !env.get?("user") && !unregistered_path_whitelist.includes?(env.request.path)
+      env.response.headers["Location"] = "/login"
+      haltf env, status_code: 302
```
stonerl, line 94: I'm also not entirely sure about this part here. Since access is denied, because of a missing authentication, a `401`should be returned.

But sending a redirect somewhat defeats the purpose of returning a `401`.

Maybe a second configuration option should be added, like: `redirect_login_only`.

If `true` a `302`is returned and the user is redirected to `/login` otherwise, a `401`is sent and the nothing happens.

src/invidious/config.cr
stonerl, line 87:
```suggestion
 
   # Only allow usage of the instance with an account
  property login_only : Bool?
  # Redirect to the login page when login_only: true otherwise a 401 status code is returned.
  property redirect_login_only : Bool?
  
```

This allows to choose whether a non whitelisted request will be forwarded to `/login` or not.

stonerl:
I opened a separate PR (#4259), where I address all the missing bits of this one.

pievalentin:
That would be a cool feature to have! Yattee for instance doesn't support basic auth

rix1337:
This pr is still very much relevant

AlecKinnear:
Invidious is great software, just discovered it. I'm a YouTube Premium user who can't stand all the extras and autoplay and visual noise any more (even Premium isn't enough).

Please guys make Invidious for logged in users only. We are running it on a small VPS and if even a small group of five people were to come and start watching 4K videos, our VPS would collapse.

At this point, having read the related issue and the commits, it seems more a question of ill-will than a technical obstacle. That someone's stats program won't work when logged-in does not seem relevant. Mr Stats could just work on his own patch or not use the logged-in only feature.

Why don't you make logged-in only a paid feature? I'd happily pay $20 to support the project and to have this option. Yes, I could code it myself or hand apply these patches. But that doesn't help the project or others. It just means a whole lot of wasted duplicate effort.

src/invidious/routes/before_all.cr
Larslol140, line 91: I think `/videoplayback` needs to be added, otherwise I was not able to watch videos (DASH).

Larslol140:
I noticed that comments appear to be missing, but I couldn't narrow down the issue.

wes1993:
There are some news about this implementation?

ToddCrimson:
Any update on this? Would love to have this work only once a user logs in.

Issue #2281, opened 2021-07-27
Export Saved Playlists


ahangarha:
**Is your enhancement request related to a problem? Please describe.**
Exporting feature would export custom playlists but doesn't export saved ones.

**Describe the solution you'd like**
I think it should be easy to include Saved Playlists entry in "Invidious Data" export.

**Describe alternatives you've considered**
No alternative I can suggest

**Additional context**
Nothing

PR #3741, opened 2023-04-12
Added support for import and export of youtube playlists

dandimrod:
It has been bugging me that when you export and import user data it doesn't include any youtube playlists that the user is subscribed to, so every time I switch instances, I have to add my playlists by hand.

Unfortunately, I'm not a crystal expert and I don't have a working environment to test this, so some help on that regard would be appreciated.

I've added the title to the export for easier identification in case someone wants to remove some from the JSON by editing the file directly, but it can be removed either way, some input on this is also appreciated.

### DIFF
--- a/src/invidious/user/imports.cr
+++ b/src/invidious/user/imports.cr
@@ -101,6 +101,16 @@ struct Invidious::User
           end
         end
       end
+
+      if youtube_playlists = data["youtube_playlists"]?.try &.as_a?
+        playlists.each do |item|
+          begin
+            playlist = get_playlist(item["id"])
+            subscribe_playlist(user, playlist)
+          rescue ex
+          end
+        end
+      end
     end
 
     # -------------------
--- a/src/invidious/user/imports.cr
+++ b/src/invidious/user/imports.cr
@@ -105,8 +105,7 @@ struct Invidious::User
       if youtube_playlists = data["youtube_playlists"]?.try &.as_a?
         playlists.each do |item|
           begin
-            playlist = get_playlist(item["id"])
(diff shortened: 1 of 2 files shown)


dandimrod:
I think this would close https://github.com/iv-org/invidious/issues/2281 but I'm not sure since the description of that issue is quite sparse.

SamantazFox (commented):
Code looks good to me.
However, we're ending up with the same concerns as here:
 * https://github.com/iv-org/invidious/pull/3973#issuecomment-1633271008
 * https://github.com/iv-org/invidious/pull/3968#discussion_r1253508149

SamantazFox:
In regard to the issue mentioned above, can you remove the title and associated export logic?

dandimrod:
@SamantazFox Sorry for the delay, I've changed the code to prevent use of `get_playlist`

SamantazFox (approved):
Code looks good :)
Can you switch your PR to "ready for review" please?

SamantazFox:
Please fix the errors reported on by the CI build!

DOOMMARINE117:
so is it possible for it to e fixed? works for importing subscriptions. 

what about a fix for playlists and watch history importing?

src/invidious/user/imports.cr
```diff
@@ -101,6 +101,15 @@ struct Invidious::User
           end
         end
       end
+
+      if youtube_playlists = data["youtube_playlists"]?.try &.as_a?
+        playlists.each do |item|
```
RadoslavL, line 106: I think you meant to type `youtube_playlists.each do |item|`.

Issue #3731, opened 2023-04-10
[Bug] `Missing hash key: "lengthText" (KeyError)` when loading community tab


ahoge-enthusiast:
**Describe the bug**
I get this error when loading https://yewtu.be/channel/UCJ46YTYBQVXsfsp8-HryoUA/community I tried some other channels but none of them have the same issue.

**Logs**
> Title: `Missing hash key: "lengthText" (KeyError)`
> Date: `2023-04-10T17:25:31Z`
> Route: `/channel/UCJ46YTYBQVXsfsp8-HryoUA/community`
> Version: `2023.04.10-5091648 @ master`
> 
> <details>
> <summary>Backtrace</summary>
> <p>
>    
> ```
> Missing hash key: "lengthText" (KeyError)
>   from /usr/lib/crystal/core/json/any.cr:100 in '[]'
>   from /usr/lib/crystal/core/json/any.cr:102:10 in 'fetch_channel_community'
>   from src/invidious/routes/channels.cr:136:26 in 'community'
>   from lib/kemal/src/kemal/route.cr:13:9 in '->'
>   from lib/kemal/src/kemal/config.cr:92:7 in 'call'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call'
>   from /usr/lib/crystal/core/http/server/handler.cr:28:7 in 'call_next'
>   from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
>   from /usr/lib/crystal/core/http/server.cr:500:5 in '->'
>   from /usr/lib/crystal/core/fiber.cr:146:11 in 'run'
[...]

PR #3760, opened 2023-04-20
Fix live and upcoming community videos

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/3731

This pr does the following:

- skips video length + publish date if they don't exist
- parses owner verified
- parse liveNow and isUpcoming
- parses viewcount runs if simple text is not available

### DIFF
--- a/src/invidious/channels/community.cr
+++ b/src/invidious/channels/community.cr
@@ -146,7 +146,11 @@ def fetch_channel_community(ucid, continuation, locale, format, thin_mode)
                           Invidious::JSONify::APIv1.thumbnails(json, video_id)
                         end
 
-                        json.field "lengthSeconds", decode_length_seconds(attachment["lengthText"]["simpleText"].as_s)
+                        if length_text = attachment.dig?("lengthText", "simpleText").try &.as_s
+                          json.field "lengthSeconds", decode_length_seconds(length_text)
+                        else
+                          json.field "lengthSeconds", 0
+                        end
 
                         author_info = attachment["ownerText"]["runs"][0].as_h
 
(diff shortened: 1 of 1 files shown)


src/invidious/channels/community.cr
```diff
@@ -155,14 +157,22 @@ def fetch_channel_community(ucid, continuation, locale, format, thin_mode)
...
                         # TODO: json.field "authorThumbnails", "channelThumbnailSupportedRenderers"
-                        # TODO: json.field "authorVerified", "ownerBadges"
+                        json.field "authorVerified", has_verified_badge?(attachment["ownerBadges"]?).to_s
 
-                        published = decode_date(attachment["publishedTimeText"]["simpleText"].as_s)
+                        if published_time_text = attachment.dig?("publishedTimeText", "simpleText").try &.as_s
+                          published = decode_date(published_time_text)
+                          json.field "published", published.to_unix
+                          json.field "publishedText", translate(locale, "`x` ago", recode_date(published, locale))
+                        end
```
SamantazFox, line 166: For compatibility, can you please keep the fields, but put dummy values in here (0 for timestamp, and empty string for text). I'd prefer if we don't hard break other downstream projects!

src/invidious/channels/community.cr
```diff
@@ -155,14 +157,22 @@ def fetch_channel_community(ucid, continuation, locale, format, thin_mode)
...
 
-                        json.field "published", published.to_unix
-                        json.field "publishedText", translate(locale, "`x` ago", recode_date(published, locale))
+                        if thumbnail_overlay = attachment.dig?("thumbnailOverlays", 0, "thumbnailOverlayTimeStatusRenderer")
+                          overlay_style = thumbnail_overlay["style"].as_s
+                          json.field "liveNow", overlay_style == "LIVE"
+                          json.field "isUpcoming", overlay_style == "UPCOMING"
+                        end
```
SamantazFox, line 172: Ditto

SamantazFox (changes requested):

src/invidious/channels/community.cr
ChunkyProgrammer: Sounds good, i made the change for the lengthSeconds as well 🙂

ChunkyProgrammer:
Found a better way of doing this 😄

PR #3769, opened 2023-04-25
Ci bump crystal version

SamantazFox:


SamantazFox:
Note: I've purposefully stopped the ARM docker job.

PR #3770, opened 2023-04-26
Add proper exception handling for comments, rather than sliently failing

techmetx11:


techmetx11:
DO NOT PUSH, I left in the test exception

SamantazFox (changes requested):
As your function share all of it's code with `error_template_helper`, what I'd recommend you to do is to rename it so something more explicit (like `error_template_message`) and reuse it in `error_template_helper` like so:

```crystal
def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exception : Exception)
  if exception.is_a?(InfoException)
    return error_template_helper(env, status_code, exception.message || "")
  end

  locale = env.get("preferences").as(Preferences).locale

  env.response.content_type = "text/html"
  env.response.status_code = status_code

  error_message = error_template_message(env, exception)

  # Don't show the usual "next steps" widget. The same options are
  # proposed above the error message, just worded differently.
  next_steps = ""

  return templated "error"
end
```

techmetx11:
> As your function share all of it's code with `error_template_helper`, what I'd recommend you to do is to rename it so something more explicit (like `error_template_message`) and reuse it in `error_template_helper` like so:
> 
> ```crystal
> def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exception : Exception)
>   if exception.is_a?(InfoException)
>     return error_template_helper(env, status_code, exception.message || "")
>   end
> 
>   locale = env.get("preferences").as(Preferences).locale
> 
>   env.response.content_type = "text/html"
>   env.response.status_code = status_code
> 
>   error_message = error_template_message(env, exception)
> 
>   # Don't show the usual "next steps" widget. The same options are
>   # proposed above the error message, just worded differently.
>   next_steps = ""
> 
>   return templated "error"
> end
> ```

Thanks for the suggestion, I just threw this code together lazily. I'm gonna try making it better

SamantazFox (approved):

SamantazFox:
When no comments are found, this exception is raised, and this results in "undefined" to be printed: 
https://github.com/iv-org/invidious/blob/9cec83c1ffcb03bcfd2d43562528ed290a5546da/src/invidious/comments/youtube.cr#L43


Here is an screenshot:
image

SamantazFox:
Sorry, it took me forever to test :/

So I've deployed that on our test instance to try it out again:
<details><summary>Screenshot</summary>

image

</details>

Here are two things that needs to be addressed:

1. The current error template is a lot of text, this doesn't integrate well in the page. A smaller variant might be preferred. As an easy alternative, you could show a simple error message like "comments failed to load" then put the bug report under a foldable section (`<details>`).

2. When an error is returned from the comments API, the various "next steps" links shouldn't be present in the HTML (imo, trying to make the right links will be too difficult).

PR #3783, opened 2023-05-06
Community: Minor HTML/CSS fixes

SamantazFox:
Some minor HTML and CSS fixes:
 * Limit width of comments to about 800 CSS pixels
 * Make embedded videos look better
 * Remove some inline style
 * Wrap the "footer" (publication date + link + likes) in `<p></p>` to space it out from the content

result:
image

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -322,6 +322,16 @@ p.video-data   { margin: 0; font-weight: bold; font-size: 80%; }
 
 
 /*
+ * Comments & community posts
+ */
+
+#comments {
+  max-width: 800px;
+  margin: auto;
+}
+
+
+/*
  * Footer
  */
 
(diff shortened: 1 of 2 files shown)


SamantazFox:
Deployed at https://test.invidious.io

ChunkyProgrammer:
Tested on phone, i have to scroll down quite a bit to see the error message for videos with errors:

What I initially see:
Screenshots_2023-05-07-10-18-31

What i see after scrolling:
Screenshots_2023-05-07-10-18-53


This is not related to the PR but is there a reason why embed videos are used in the community page? It's different from how videos are displayed elsewhere.
Ex: 
Screenshots_2023-05-07-10-21-29

SamantazFox:
@ChunkyProgrammer Yeah, this is not related to this PR. However, the embed video page shouldn't show the whole webpage if the content is unavailable (in this case, members-only video). And the Iframe should probably be bigger anyway!

PR #3804, opened 2023-05-14
Search: Parse playlists when searching a channel

ChunkyProgrammer:
Test Url: `/search?q=channel%3AUCXuqSBlHAE6Xw-yeJA0Tunw+ces+2017`

Before (tested with an instance):
image

After (tested locally):
image


SamantazFox (approved):

SamantazFox:
(Might be a bit late to change the label, lol)

Works perfectly, thanks for the patch!

Issue #2344, opened 2021-08-22
[Bug] Channel search is gone.


SkMyPP:
An example of two instances that still have channel search (see the search bar at the top of the page):
https://ytprivate.com/channel/UCAL3JXZSzSm8AlZyD3nQdBA
https://invidious.noho.st/channel/UCAL3JXZSzSm8AlZyD3nQdBA

Every other public instance is missing this.

syeopite:
Seems like a commit after July 27th introduced the regression. Thanks!

TheFrenchGhosty:
A 10$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 20$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

matthewmcgarvey:
This _might_ be fixed already? I definitely see the search bar on the top of the page. Unless they mean specifically youtube's channel search which I don't see any evidence it ever existed.

SamantazFox:
Yep, this has been fixed. Though, I think that this issue is about the `channel:UC...` filter not showing up in the search bar on channel pages.

unixfox:
@SamantazFox fixed?

SamantazFox:
@unixfox Yes! It was fixed in #2956 (and #3033)

SamantazFox:
Nevermind, I misread the issue. Will be fixed in #2546

TheFrenchGhosty:
@SamantazFox Isn't it solved?

SamantazFox:
It has! I wonder why it wasn't automatically closed by #3699...

SkMyPP:
Wait, why was this closed? Channel search is still gone.

#3699 doesn't even work, and if it does it's not properly explained & extremely vague on how to search for a channel's videos.

Why not just populate the search bar with the channel code like it used to?

SkMyPP:
This is how it used to/should look like when visiting a channel:
<a href="https://imgbox.com/fubbl0a4" target="_blank"><img src="https://images2.imgbox.com/91/20/fubbl0a4_o.png" alt="image host"/></a>

SamantazFox:
> Why not just populate the search bar with the channel code like it used to?

Ah, yeah, sorry... I was planning to merge #2215 , which is supposed to add a search box on the channel page
But I haven't had the time to look that up...

PR #3805, opened 2023-05-14
Search: Populate search bar with ChannelId

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/2344

### DIFF
--- a/src/invidious/routes/channels.cr
+++ b/src/invidious/routes/channels.cr
@@ -278,6 +278,7 @@ module Invidious::Routes::Channels
       return error_template(500, ex)
     end
 
+    env.set "search", "channel:" + ucid + " "
     return {locale, user, subscriptions, continuation, ucid, channel}
   end
 end
--- a/src/invidious/routes/channels.cr
+++ b/src/invidious/routes/channels.cr
@@ -278,7 +278,7 @@ module Invidious::Routes::Channels
       return error_template(500, ex)
     end
 
-    env.set "search", "channel:" + ucid + " "
+    env.set "search", "channel:#{ucid} "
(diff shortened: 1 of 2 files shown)


src/invidious/routes/channels.cr
```diff
@@ -278,6 +278,7 @@ module Invidious::Routes::Channels
       return error_template(500, ex)
     end
 
+    env.set "search", "channel:" + ucid + " "
```
SamantazFox, line 281:
```suggestion
    env.set "search", "channel:#{ucid} "
```

src/invidious/routes/search.cr
```diff
@@ -65,7 +65,11 @@ module Invidious::Routes::Search
 
       redirect_url = Invidious::Frontend::Misc.redirect_url(env)
 
-      env.set "search", query.text
+      if query.type == Invidious::Search::Query::Type::Channel
+        env.set "search", "channel:" + query.channel + " " + query.text
```
SamantazFox, line 69:
```suggestion
        env.set "search", "channel:#{query.channel} #{query.text}"
```

SamantazFox (changes requested):
FYI: string interpolation (`"text #{var}"`) is faster than a concatenation:
https://crystal-lang.org/reference/1.8/guides/performance.html#use-string-interpolation-instead-of-concatenation

SamantazFox:
Btw, I'm blown away by the simplicity of the fix. Thanks for the PR :x

SamantazFox:
Thanks for your contribution!

SamantazFox:
And sorry for the erratic merge :x

Issue #3806, opened 2023-05-15
[Bug] YouTube playlist pagination shows videos from previous page


ChunkyProgrammer:
**Describe the bug**
When going to the next page in a playlist, it will copy the second half of the previous page into the new page.I think that was added to show previous videos in the playlist when on the watch page but this doesn't make much sense when on the playlist page or when using the API.

**Steps to Reproduce**
1. Go to: http://127.0.0.1:3000/playlist?list=PLbio2sNGiPZOu6C1XtpLKVHHsheIP2MsH&page=1
2. Notice last few videos
3. Go to page 2 http://127.0.0.1:3000/playlist?list=PLbio2sNGiPZOu6C1XtpLKVHHsheIP2MsH&page=2
4. Notice that the last half of page 1 appears around the middle of page 2

PR #3807, opened 2023-05-15
Playlists: Fix duplicated videos across pages

ChunkyProgrammer:
When going on a playlist with multiple pages, it would refetch a lot of the videos that were in the previous page. This PR will stop that from happening in some cases.

Example playlist: https://www.youtube.com/playlist?list=PLbio2sNGiPZOu6C1XtpLKVHHsheIP2MsH

closes https://github.com/iv-org/invidious/issues/3806

From what I can tell, this won't affect video urls with playlist ids in them

### DIFF
--- a/src/invidious/routes/playlists.cr
+++ b/src/invidious/routes/playlists.cr
@@ -410,8 +410,8 @@ module Invidious::Routes::Playlists
       return error_template(500, ex)
     end
 
-    page_count = (playlist.video_count / 100).to_i
-    page_count += 1 if (playlist.video_count % 100) > 0
+    page_count = (playlist.video_count / 200).to_i
+    page_count += 1 if (playlist.video_count % 200) > 0
 
     if page > page_count
       return env.redirect "/playlist?list=#{plid}&page=#{page_count}"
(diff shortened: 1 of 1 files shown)


SamantazFox:
Does Youtube return 200 videos per "page" now?

SamantazFox:
Nevermind, I'm stupid:
https://github.com/iv-org/invidious/blob/e238c08be5ad43c928a2f6cff17a46ef5d30b4b8/src/invidious/playlists.cr#L423-L430

I can't recall exactly what this was added for ...

ChunkyProgrammer:
If the logic to enforce 200 videos can be removed then that would probably be better. Youtube doesnt always return 100 videos (sometimes it returns 101) so there's still a chance of videos repeating in this code (it'll just be a couple videos instead of 100 videos that repeat)

SamantazFox:
> If the logic to enforce 200 videos can be removed then that would probably be better.

Yeah, I don't know why it was made such a large number in the first place.
I'd have to dig back in the PRs ^^

> Youtube doesnt always return 100 videos (sometimes it returns 101)

Ah, right, that's the reason why this `until ...` code was added!

SamantazFox (approved):

SamantazFox:

Issue #3035, opened 2022-04-17
[Enhancement] Title under videos for "Watch History"


DUOLabs333:
**Is your enhancement request related to a problem? Please describe.**
Currently, for "Watch History", there's no title under the videos, so I have to guess what video is which.

**Describe the solution you'd like**
Titles (optionally info like channel, views, etc.) under each video in "Watch History"

**Describe alternatives you've considered**

**Additional context**

What it looks like currently:
![image](https://user-images.githubusercontent.com/49244768/163697589-eeaef71b-7d94-40a6-a735-847d20754cc0.png)

chidera-ovuoba:
I think it will be better if the user  hover's on any  of the videos they will be an overlay on the video which describes the video details

Sidharth004:
@chidera-ovuoba the hovering over videos idea sounds cool, but I think providing a title below the videos, would look simple aesthetically but probably would provide a clear & precise context, without needing to hover over any / all video categories

spitfire:
Yup, it works like this in any other view and it's fine.

ShivamK2002:
Is this issue still open, if yes I would Iike to contribute. @DUOLabs333

DUOLabs333:
@ShivamK2002 Sure, you can work on it (I didn't assign myself the issue or anything, and I definitely do not have the time right now to work on it).

Issue #1356, opened 2020-08-29
non-original title for video in english


rezad1393:
this youtube video (his whole channel actually) https://www.youtube.com/watch?v=MFCNUvFEVFY has a Japanese title on invidious but a english one on youtube, and it is not a translated one.
also channel name too.
can this be added to invidious?
I know this is not a high priority but maybe add that for later releases?

TheFrenchGhosty:
This is a known issue, sometimes videos will report title in the wrong language in every page (search/channel) except the main video page.

According to omarroth this is a youtube issue and nothing can be done about it.

rezad1393:
> This is a known issue, sometimes videos will report title in the wrong language in every page (search/channel) except the main video page.
> 
> According to omarroth this is a youtube issue and nothing can be done about it.

thanks for adding the tags.
but I didnt get part of your answer.
in youtube the title is in english for all those channel's videos except for when using rss of youtube.
the main video page of video also is in english title (not the title at the toolbar that is in Japanese but the one under video).

but in invidious nowhere anything is in english for this channel (except the video main description which I think is in english from the source (not japanese at all))
not the video title under video, not the channel name.

nathgit:
That's different. I've seen the video title be wrong a few times too, but it's always been in the same language (english).

SamantazFox:
This problem comes from Youtube, there is not much we can do about it.

Basically, content creators have 3 choices:
 - Do not translate video titles
 - Propose their own translations for video titles
 - Opt in for automatic translation of the titles

In the last two cases, the data that youtube returns will contain a translated video title (except on the watch page, for some reason).

The reason why most of the time titles are in english is because this is the locale that invidious uses to fetch data from YT servers (with the `hl=` parameter), as it is the easiest to parse and pretty consistent. However, it has the side effect of always providing english titles when an english version of the title is available.

The only solution so far would be to fetch the `/watch` page for every videos of a playlist or in the search results, as this the only place that (still) provides un-translated titles. But this isn't feasible in reality, as it'll make way too much request to youtube servers, leading to instance being blocked more often and a increase in bandwidth use.

TheFrenchGhosty:
"""""feature"""""

PR #3815, opened 2023-05-21
Use Redis for caching

### LINKED ISSUE #3628: [Enhancement] Moving the cache to Redis
This issue is somewhat related to #2620, and is mostly here for informatoional purposes.

Invidious currently uses Postgres to store data. All of its data. And most notably, the video metadata (`/watch` endpoint) and the channel videos (for the notifications). This data is huge, and accessed/updated very often, which causes a lot of disk accesses and disk wear, as Postgres tries to immediately write on disk to avoid data losses.

On large instances, this is a huge performance drain because invidious is constantly waiting for DB to read various things from disk. It's also writing a lot to the disk (some instance owners have reported a few terabytes of data written to disk after a few months).

We tried to do some damage control in various PRs (like #2778, #3294, or more recently #3473), but the problem is still there, and only gets worse as the user base grows.

The solution is to use a proper high performance data store like [Redis](https://redis.io/) in order to store that data.

### Technical notes

Current shard candidates are [stefanwille/crystal-redis](https://github.com/stefanwille/crystal-redis) and [jgaskins/redis](https://github.com/jgaskins/redis).

SamantazFox:
Closes #3628

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -44,6 +44,24 @@ db:
 
 #########################################
 #
+#  Cache configuration
+#
+#########################################
+
+cache:
+  ##
+  ## URL of the caching server. To not use a caching server,
+  ## set to an empty string or leave empty.
+  ##
+  ## Note: The same "long" format as the 'db' parameter is
+  ## also supported.
+  ##
+  url: ""
+
+
+
+#########################################
+#
 #  Server config
 #
 #########################################
--- a/shard.lock
+++ b/shard.lock
@@ -32,6 +32,10 @@ shards:
     git: https://github.com/will/crystal-pg.git
     version: 0.24.0
 
+  pool:
+    git: https://github.com/ysbaddaden/pool.git
+    version: 0.2.4
+
   protodec:
     git: https://github.com/iv-org/protodec.git
     version: 0.1.5
@@ -39,6 +43,10 @@ shards:
   radix:
     git: https://github.com/luislavena/radix.git
     version: 0.4.1
+
+  redis:
+    git: https://github.com/stefanwille/crystal-redis.git
+    version: 2.8.3
 
   spectator:
     git: https://github.com/icy-arctic-fox/spectator.git
--- a/shard.yml
+++ b/shard.yml
@@ -10,24 +10,34 @@ targets:
     main: src/invidious.cr
 
 dependencies:
+  # Database
   pg:
     github: will/crystal-pg
     version: ~> 0.24.0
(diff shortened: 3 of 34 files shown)


syeopite:
What do you think about splitting the video cache into two? One for the streaming urls and the other for the metadata. We'd be able to persist the metadata for a lot longer than 6 hours then.

syeopite:
> I guess you want the metadata to be used for imports or playlists? 

Yep. Invidious also often calls `#get_video` to fetch video metadata only so splitting the cache can probably also save us quite a few calls to YouTube.

SamantazFox:
In the end, I think I'll go with that separation:

### `videos:<id>:title`

* Stores: the video title (raw text).
* Updated from: watch page
* Retention: as long as possible (2 months?)


### `videos:<id>:release_date`

* Stores: the exact date at which the video was released
* Format: ISO date
* Retention: as long as possible (1 year?)
* Updated from: watch page or channel RSS feed

### `videos:<id>:playback`

* Stores: the video's playback infos (captions/storyboards/streams)
* Format: JSON
* Retention: 5h, to keep the links fresh, except for livestreams where this is reduced to 1h.

### `videos:<id>:metadata`

* Stores: all the miscellaneous metadata for the watch page
* Format: JSON
* Retention: a few hours if the video was released recently, to keep views, likes and description fresh. After that, keep for a day or two.

syeopite:
Titles should probably just be included as a part of `metadata` imo. It's something that changes fairly frequently (especially when the video is new) and in some cases frequent title changes are the entire point of the video such as this one https://redirect.invidious.io/watch?v=BxV14h0kFs0

As for `playback`, you can probably split off storyboards from it. It's not something that should change often if at all.

SamantazFox:
> Titles should probably just be included as a part of `metadata` imo. It's something that changes fairly frequently (especially when the video is new) and in some cases frequent title changes are the entire point of the video such as this one https://redirect.invidious.io/watch?v=BxV14h0kFs0

I know that titles may change (especially true when the video was recently released, to boost views). The main goal is to address #1356 and/or #3959 + #3035, which requires a long cache time.

> As for `playback`, you can probably split off storyboards from it. It's not something that should change often if at all.

So you'd wand it with the metadata? Or separated? In all cases, all data will be refreshed whenever the video needs to be played.

syeopite:
> So you'd wand it with the metadata? Or separated?

Either or is fine. I'm just not entirely sure why captions and storyboards are placed under `videos:<id>:playback`

Issue #3714, opened 2023-03-30
[Enhancement] Add the ability to add custom text at the footer - contact a maintainer


unixfox:
It's a simple enhancement.

Add a way to configure some text that would include an email address or a link to a contact page. Could be used to add anything else. 

This text would be added to every page in the footer.

We already have a way to include some text at the very top of the page and this functionality already allows using HTML code, ideally this new feature should provide the same behavior.

PR #3826, opened 2023-05-29
Add custom text at the bottom - contact a maintainer

auralglow:
Hii y'all,

This is my first pull request so please be gentle ^^

I added a configuration where you can add a footer at the bottom of every page. It also supports HTML.
Should close issue #3714

Please feel free to ask for changes if need be.

I deployed this PR on a test instance: https://yt-dg-test.auralglow.me

Preview:
image

### DIFF
--- a/src/invidious/views/template.ecr
+++ b/src/invidious/views/template.ecr
@@ -153,6 +153,11 @@
                         <span><%= translate(locale, "Current version: ") %> <%= CURRENT_VERSION %>-<%= CURRENT_COMMIT %> @ <%= CURRENT_BRANCH %></span>
                     </div>
                 </div>
+            <% if CONFIG.footer %>
+            <div>
+               <h4 style="text-align: center;"><%= CONFIG.footer %></h3>
+            </div>
+            <% end %>
             </footer>
 
         </div>
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -436,6 +436,16 @@ jobs:
 #banner:
 
 ##
+## Custom footer displayed on the bottom of every page.
+## Similar to the banner option but at the bottom of the page
+## Useful for putting contact info of a/the maintainer for example.
+##
+## Accepted values: any string. HTML is accepted.
+## Default: <none>
+##
+#footer:
+
+##
 ## Subscribe to channels using PubSubHub (Google PubSubHubbub service).
 ## PubSubHub allows Invidious to be instantly notified when a new video
 ## is published on any subscribed channels. When PubSubHub is not used,
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -106,6 +106,8 @@ class Config
   property cache_annotations : Bool = false
   # Optional banner to be displayed along top of page for announcements, etc.
   property banner : String? = nil
+  # Optional footer banner to be displayed along the bottom of the page. Can be used for maintainer contact info, etc.
+  property footer : String? = nil
   # Enables 'Strict-Transport-Security'. Ensure that `domain` and all subdomains are served securely
   property hsts : Bool? = true
   # Disable proxying server-wide: options: 'dash', 'livestreams', 'downloads', 'local'


src/invidious/views/template.ecr
```diff
@@ -153,6 +153,11 @@
                         <span><%= translate(locale, "Current version: ") %> <%= CURRENT_VERSION %>-<%= CURRENT_COMMIT %> @ <%= CURRENT_BRANCH %></span>
                     </div>
                 </div>
+            <% if CONFIG.footer %>
+            <div>
+               <h4 style="text-align: center;"><%= CONFIG.footer %></h3>
```
ChunkyProgrammer, line 158: Any reason you chose an h4 over a span element? Also, the end tag doesnt match the start tag (h4/h3)

auralglow: I also tried with a span element. It doesn't look right imo:
image
I would think that it would fit better with the banner message and the rest of the page.
I also tried both h3 and h4 and forgot to change it on the other side (oops).

src/invidious/views/template.ecr
```diff
@@ -153,6 +153,11 @@
                         <span><%= translate(locale, "Current version: ") %> <%= CURRENT_VERSION %>-<%= CURRENT_COMMIT %> @ <%= CURRENT_BRANCH %></span>
                     </div>
                 </div>
+            <% if CONFIG.footer %>
+            <div>
+               <h4 style="text-align: center;"><%= CONFIG.footer %></h4>
```
SamantazFox, line 158: Please consider using a `<div>` with a unique class/id and CSS styling, rather than a sub-title (`<h4>`) element. This is in order to make invidious more accessible to blind users ^^

auralglow: Oh ok sure. To be honest, I couldn't think of a way to make the footer spatious enough. Thanks for the heads-up ^^

src/invidious/views/template.ecr
```diff
@@ -153,6 +153,11 @@
                         <span><%= translate(locale, "Current version: ") %> <%= CURRENT_VERSION %>-<%= CURRENT_COMMIT %> @ <%= CURRENT_BRANCH %></span>
                     </div>
                 </div>
+            <% if CONFIG.footer %>
+            <div class="h-box" style="padding-top: 1em; padding-bottom: 0.5em;">
+               <div id="footer-info" style="text-align: center;"><%= CONFIG.footer %></div>
```
realcharmer, line 158: I'd avoid using inline CSS, that's why you added classes and/or IDs. Or you might as well add `v-box` class to the div and avoid specifying custom styles entirely. The `footer` HTML5 semantic element already has `text-align: center` set in  https://github.com/iv-org/invidious/blob/master/assets/css/default.css#LL358C3-L358C22, which should get inherited here.
Your indentation is wrong as well.

realcharmer (changes requested):

src/invidious/views/template.ecr
```diff
@@ -153,6 +153,11 @@
...
+            </div>
+            <% end %>
```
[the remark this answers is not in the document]
SamantazFox: @auralglow please address this ^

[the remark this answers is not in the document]
auralglow: Just did ^^

src/invidious/views/template.ecr
syeopite: https://github.com/iv-org/invidious/blob/701f03d5e88a8f3c92575965e39b8f96b59f71f9/src/invidious/views/template.ecr#L108-L112

Do you think we should also change the element in the banner section?

src/invidious/views/template.ecr
```diff
@@ -153,6 +153,11 @@
                         <span><%= translate(locale, "Current version: ") %> <%= CURRENT_VERSION %>-<%= CURRENT_COMMIT %> @ <%= CURRENT_BRANCH %></span>
                     </div>
                 </div>
+                <% if CONFIG.footer %>
+                    <div class="h-box v-box" id="footer-info">
```
syeopite, line 157: I don't know if this is intentional or not but if you'd like to preserve the bold text seen in the initial version feel free to add custom styling in default.css.  

```css
#footer-info {
  font-weight: bold;
}
```

syeopite (approved):

src/invidious/views/template.ecr
auralglow: I decided at first that the bold text would be ok. Then I decided that the user should make this choice if they want it to through HTML. Thank you for the tip though :3

unixfox (approved):
Good for merging

Issue #3540, opened 2023-01-06
[Bug/Enhancement] Cache the random HMAC_KEY


iBicha:
**Is your enhancement request related to a problem? Please describe.**
I noticed that in a lot of instances, tokens stop working ("Invalid signature")
I investigated a bit, and I noticed that if HMAC_KEY will be random after reach restart, if it was not configured.
This means that all tokens will be immediately invalid after each restart. 
This is strange, since it is recommended from people who host Invidious to restart often. 

**Describe the solution you'd like**
I would like to have the HMAC_KEY (the randomly generated one - `HMAC_KEY = CONFIG.hmac_key || Random::Secure.hex(32)`) to be cached between restarts, so that tokens do not randomly stop working.
If we do not want to do that for some reason, then this config should be mandatory (e.g. invidious would not start without it - but that's an overkill probably)

**Describe alternatives you've considered**
I do not see any other way to prevent tokens from going invalid.

**Additional context**

This is not for the web frontend, this is for another app ([Playlet](https://github.com/iBicha/playlet)) so we're using the token not the cookie. For that reason, it's important to keep the token alive. This has been a blocker and painful to find the account logged out every time.

unixfox:
Just set the HMAC_KEY and you won't have any issue: https://github.com/iv-org/invidious/blob/master/config/config.example.yml#L465

iBicha:
@unixfox the problem is not an instance I control, it's instances under https://docs.invidious.io/instances/
This is not documented clearly enough for them. It's important piece that most people might miss.

unixfox:
Then send a message to each instance maintainer to set this value, I'll add that in the docs though.

All the maintainers have opened an issue here so feel free to ping them here: https://github.com/iv-org/documentation/issues

iBicha:
@unixfox Kinda stumbling on this issue again (https://github.com/iBicha/playlet/issues/77)

I really think this is a simple change that we can introduce to Invidious that persists the randomly generated key in the case it was not manually configured. Saving it to a file in a known location and checking it later is fine.
Otherwise Invidious should fail to start to bring the issue to the attention of the maintainer.

And there isn't a good way to know if an instance if properly configured or not, until it starts randomly failing to maintain sessions, or to authorize tokens. 
This feels like a low-hanging fruit to me

iBicha:
> No there is easy way to cache the hmac key.

Are you saying there's an easy way, or no easy way to cache the key?

unixfox:
[...]

PR #3828, opened 2023-05-30
Add hmacKeyConfigured field to stats api

iBicha:
see #3540

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -57,9 +57,9 @@ end
 # Simple alias to make code easier to read
 alias IV = Invidious
 
-CONFIG   = Config.load
+CONFIG              = Config.load
 HMAC_KEY_CONFIGURED = CONFIG.hmac_key != nil
-HMAC_KEY = CONFIG.hmac_key || Random::Secure.hex(32)
+HMAC_KEY            = CONFIG.hmac_key || Random::Secure.hex(32)
 
 PG_DB       = DB.open CONFIG.database_url
 ARCHIVE_URL = URI.parse("https://archive.org")
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -58,6 +58,7 @@ end
 alias IV = Invidious
 
 CONFIG   = Config.load
+HMAC_KEY_CONFIGURED = CONFIG.hmac_key != nil
 HMAC_KEY = CONFIG.hmac_key || Random::Secure.hex(32)
 
 PG_DB       = DB.open CONFIG.database_url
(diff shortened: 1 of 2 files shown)


unixfox:
Can you post the results of the API?

iBicha:
> Can you post the results of the API?

```
{
   "version":"2.0",
   "software":{
      "name":"invidious",
      "version":"2023.05.30-b56e55c8",
      "branch":"feature/hmac-key-check"
   },
   "openRegistrations":true,
   "usage":{
      "users":{
         "total":7,
         "activeHalfyear":7,
         "activeMonth":0
      }
   },
   "metadata":{
      "updatedAt":1685484461,
      "lastChannelRefreshedAt":1685467825,
      "hmacKeyConfigured":true
   }
}
```

SamantazFox:
I'd rather have a startup check for this config that prevents invidious from running if not set.
The docs should also be updated to shown a sample value in the example config/compose files, like Searx does:

https://github.com/searxng/searxng-docker/blob/e76656a162c40c04ecc4b119d2a28ec2104c5275/searxng/settings.yml#L3-L7

unixfox:
> I'd rather have a startup check for this config that prevents invidious from running if not set.
> The docs should also be updated to shown a sample value in the example config/compose files, like Searx does:
> 
> https://github.com/searxng/searxng-docker/blob/e76656a162c40c04ecc4b119d2a28ec2104c5275/searxng/settings.yml#L3-L7
> 

Do you understand that this will temporarily break thousands of existing installations?

I'm also in favor of this way but that's quite disruptive to do that right now.

SamantazFox:
@iBicha what's your original motivation behind this PR?
Is it because you're having issues on playlet with public instances?

> Do you understand that this will temporarily break thousands of existing installations?

Yes, I'm aware. Usually, I try to always make changes that won't hard break invidious, but I've been thinking about adding back semantic versioning and starting to make some heavier changes (That could deserve a separate issue, tho).

Regarding this specific PR, I'd like to take the opportunity to remove an entry from the FAQ, too.
We had many users complaining about the broken authentication in the past, so by making the hmac key mandatory, we could ensure that instances don't give a bad user experience because of misconfiguration.

iBicha:
> @iBicha what's your original motivation behind this PR? Is it because you're having issues on playlet with public instances?

Yes, for Playlet. Unconfigured key causes the tokens to invalidate somewhat randomly, and leaves people confused because there's no indication on why. It also affects the /authorize part. It takes a few retries to get a valid token in the first place.

I totally agree with you, this needs to be a startup check, with a clear error message on what was wrong.
Although disruptive, have Invidious failing to start will get the attention of maintainers and get them to fix this necessary config, and it's a one time thing.

If we really want to go through the less disruptive route, my original idea in #3540 is to cache if hmac_key between restarts. Of course the caching would happen only in the case the key was not configured by the instance owner.

The current PR is my least favorite option, but at least it takes out the guess work of why an instance "keeps logging me out"

unixfox:
I'm for introducing this breaking change but would prefer to show a warning message for a couple of weeks (2 weeks) everywhere before introducing the change. So that less people get "shocked" by this new change.

unixfox:
Closing in favor of https://github.com/iv-org/invidious/issues/3854

PR #3843, opened 2023-06-02
Chore: Bump various constants

SamantazFox:
Update a few constants (User Agent, client version) across the app.

### DIFF
--- a/src/invidious/yt_backend/connection_pool.cr
+++ b/src/invidious/yt_backend/connection_pool.cr
@@ -8,8 +8,9 @@
 
 def add_yt_headers(request)
   if request.headers["User-Agent"] == "Crystal"
-    request.headers["User-Agent"] ||= "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36"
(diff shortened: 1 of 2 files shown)


src/invidious/yt_backend/connection_pool.cr
```diff
@@ -8,8 +8,9 @@
 
 def add_yt_headers(request)
   if request.headers["User-Agent"] == "Crystal"
-    request.headers["User-Agent"] ||= "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36"
+    request.headers["User-Agent"] ||= "Mozilla/5.0 (Windows NT 10.0; rv:113.0) Gecko/20100101 Firefox/113.0"
```
unixfox, line 11: I feel like it would be better to keep chromium user agent as there are more chromium users than firefox user, this way we can "more" blend into the crowd of users around the world.

unixfox: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36`

unixfox (approved):

Issue #3854, opened 2023-06-04
[Announcement] hmac_key parameter in config.yml mandatory by July 1st


unixfox:
In the effort of fixing issues like in #3540, by the July 1st we are going to make the `hmac_key` mandatory in the `config.yml` file. After this date, Invidious will not start without this parameter configured.

As a reminder, `hmac_key` is a key that needs to be generated from random characters using the command `pwgen 20 1` or `openssl rand -hex 20`. This is explained in the documentation: https://docs.invidious.io/installation/#post-install-configuration

SamantazFox:
Note: this can be set with the environment variable `INVIDIOUS_HMAC_KEY` too!

PR #3855, opened 2023-06-04
Warn about hmac key deadline

unixfox:
Related to #3854

To be removed on the 1st July and replaced with the implementation that force the hmac_key parameter.

Screenshot of the message displayed only for the administrator(s):
image

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -57,8 +57,9 @@ end
 # Simple alias to make code easier to read
 alias IV = Invidious
 
-CONFIG   = Config.load
-HMAC_KEY = CONFIG.hmac_key || Random::Secure.hex(32)
+CONFIG              = Config.load
+HMAC_KEY_CONFIGURED = CONFIG.hmac_key != nil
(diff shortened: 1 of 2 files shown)


src/invidious.cr
```diff
@@ -230,6 +231,10 @@ Kemal.config.host_binding = Kemal.config.host_binding != "0.0.0.0" ? Kemal.confi
 Kemal.config.port = Kemal.config.port != 3000 ? Kemal.config.port : CONFIG.port
 Kemal.config.app_name = "Invidious"
 
+if !HMAC_KEY_CONFIGURED
+  LOGGER.warn("Please configure hmac_key by 1st July, see more here: https://github.com/iv-org/invidious/issues/3854")
```
SamantazFox, line 235:
```suggestion
  LOGGER.warn("Please configure hmac_key by July 1st, see more here: https://github.com/iv-org/invidious/issues/3854")
```

unixfox: Corrected.

SamantazFox (approved):

PR #3864, opened 2023-06-06
Hashtag API: Parse header when getting the first page

ChunkyProgrammer:
When requesting the first page of a hashtag (without continuations), YouTube will include a hashtag header that includes info such as the # of videos and # of channels in the hashtag. 

This PR will parse those properties (it also creates separate structs for the hashtag page + header)

### Before

```json
{
  "results": [
    {
      "type": "video",
      ...
    }
  ]
}
```

### After

Page 1:
```json
{
  "type": "hashtagPage",
  "header": {
    "type": "hashtag",
    "title": "#cat",
    "url": "/hashtag/cat",
    "channelCount": 8100000,
    "videoCount": 1100000
  },
  "results": [
    {
      "type": "video",
      ...
    }
  ],
  "hasNextPage": true
}

```

Page 2:

```json
{
  "type": "hashtagPage",
  "results": [
    {
      "type": "video",
      ...
    }
  ],
  "hasNextPage": true
}
```

Tests:
go to : `/api/v1/hashtag/cat?page=1` or `/api/v1/hashtag/cat` and see the header in the json response
go to `/api/v1/hashtag/cat?page=2` and no longer see the header on  page 2

### DIFF
--- a/src/invidious/hashtag.cr
+++ b/src/invidious/hashtag.cr
@@ -5,15 +5,15 @@ module Invidious::Hashtag
     include DB::Serializable
 
     property videos : Array(SearchItem) | Array(Video)
-    property header : HashtagHeader?
+    property header : SearchHashtag?
     property has_next_continuation : Bool
 
     def to_json(locale : String?, json : JSON::Builder)
       json.object do
-        json.field "type", "hashtag"
+        json.field "type", "hashtagPage"
         if self.header != nil
           json.field "header" do
-            self.header.to_json(json)
+            self.header.try &.as(SearchHashtag).to_json(locale, json)
           end
         end
         json.field "results" do
@@ -24,39 +24,6 @@ module Invidious::Hashtag
           end
         end
         json.field "hasNextPage", self.has_next_continuation
-      end
-    end
-
-    # TODO: remove the locale and follow the crystal convention
-    def to_json(locale : String?, _json : Nil)
-      JSON.build do |json|
-        to_json(locale, json)
-      end
-    end
-
-    def to_json(json : JSON::Builder)
-      to_json(nil, json)
-    end
-  end
-
-  struct HashtagHeader
-    include DB::Serializable
-
-    property tag : String
-    property channel_count : Int64
-    property video_count : Int64
-
-    def to_json(json : JSON::Builder)
-      json.object do
-        json.field "hashtag", self.tag
-        json.field "channelCount", self.channel_count
-        json.field "videoCount", self.video_count
-      end
-    end
-
-    def to_json(_json : Nil)
-      JSON.build do |json|
-        to_json(json)
       end
     end
   end
@@ -72,8 +39,8 @@ module Invidious::Hashtag
     else
       # item browses the first page (including metadata)
       response = YoutubeAPI.browse("FEhashtag", params: item, client_config: client_config)
-      if item_contents = response.dig?("header", "hashtagHeaderRenderer")
-        header = parse_hashtag_renderer(item_contents)
+      if item_contents = response.dig?("header")
+        header = parse_item(item_contents).try &.as(SearchHashtag)
       end
     end
 
@@ -118,21 +85,5 @@ module Invidious::Hashtag
       .try { |i| Protodec::Any.from_json(i) }
       .try { |i| Base64.urlsafe_encode(i) }
       .try { |i| URI.encode_www_form(i) }
-  end
-
-  def parse_hashtag_renderer(item_contents)
-    info = extract_text(item_contents.dig?("hashtagInfoText")) || ""
-
-    regex_match = /(?<videos>\d+\S)\D+(?<channels>\d+\S)/.match(info)
-
-    hashtag = extract_text(item_contents.dig?("hashtag")) || ""
-    videos = short_text_to_number(regex_match.try &.["videos"]?.try &.to_s || "0")
-    channels = short_text_to_number(regex_match.try &.["channels"]?.try &.to_s || "0")
-
-    return HashtagHeader.new({
-      tag:           hashtag,
-      channel_count: channels,
-      video_count:   videos,
-    })
   end
 end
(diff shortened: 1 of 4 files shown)


src/invidious/helpers/serialized_yt_data.cr
```diff
@@ -274,4 +274,62 @@ struct Continuation
   end
 end
 
+struct HashtagPage
```
ChunkyProgrammer, line 277: Not sure if these structs should be moved elsewhere?

SamantazFox:
Are these infos really useful? The numbers looks very approximated.

ChunkyProgrammer:
> Are these infos really useful? The numbers looks very approximated.

It was mentioned in my PR to add hashtag pags to FreeTube https://github.com/FreeTubeApp/FreeTube/pull/3483#issuecomment-1537138092 

@efb4f5ff-1298-471a-8973-3d47447115dc any thoughts on this?

efb4f5ff-1298-471a-8973-3d47447115dc:
Sorry for the delay. So i mentioned the possibility to add this because with the stats the user knows what they are getting into. A hashtag with low amount of approximate videos could make the user go through all of them and a high amount can make the user refrain from watching too much and falling into hours of watching videos.

SamantazFox:
Ah, I see, thanks!

src/invidious/helpers/serialized_yt_data.cr
SamantazFox: In the future, that could be a good idea. this file (as well as the extractors one) start to be quite clogged up!

src/invidious/helpers/serialized_yt_data.cr
```diff
@@ -274,4 +274,63 @@ struct Continuation
...
+      json.field "videoCount", self.video_count
+    end
+  end
+
+  def to_json(_json : Nil)
+    JSON.build do |json|
+      to_json(json)
+    end
+  end
+end
```
SamantazFox, line 334: Maybe we could use a generic search metadata struct here? The idea is to reuse it for regular search (we'd add `estimatedResults`).

src/invidious/yt_backend/extractors.cr
```diff
@@ -550,6 +551,39 @@ private module Parsers
       return {{@type.name}}
     end
   end
+
+  # Parses an InnerTube hashtagHeaderRender into a HashtagHeaderRender.
+  # Returns nil when the given object isn't a hashtagHeaderRender.
+  #
+  # hashtagHeaderRender contains metadate of the hashtag page such as video count and channel count
```
SamantazFox, line 558:
```suggestion
  # hashtagHeaderRender contains metadata of the hashtag page such as video and channel count
```

src/invidious/yt_backend/extractors.cr
```diff
@@ -550,6 +551,39 @@ private module Parsers
...
+        tag:           hashtag,
+        channel_count: channels,
+        video_count:   videos,
+      })
+    end
+
+    def self.parser_name
+      return {{@type.name}}
+    end
+  end
```
SamantazFox, line 586: I'm still undecided whether making a dedicated extractor module is needed or not, as this renderer is only used for hashtags. IMO, one-off parsing makes more sense, and desn't impact the other consumers of `extract_items` (I'm worried that adding too much renderers module could impact performance long term).

src/invidious/hashtag.cr
```diff
@@ -1,15 +1,42 @@
...
+        "1:string" => hashtag,
+        "2:varint" => 0_i64,
+        "3:varint" => 1_i64,
+      },
+    }
+
+    return object.try { |i| Protodec::Any.cast_json(i) }
+      .try { |i| Protodec::Any.from_json(i) }
+      .try { |i| Base64.urlsafe_encode(i) }
+      .try { |i| URI.encode_www_form(i) }
```
SamantazFox, line 39: Could you merge that with the `generate_continuation` function below? The `93:embedded` object is the same between the two, so you can easily share the code.

Note: don't mind that `:2`, it's a leftover of protodec's parser. `93:2:embedded` and `93:embedded` are essentially the same thing.

src/invidious/helpers/serialized_yt_data.cr
ChunkyProgrammer: I'm not sure i understand the changes you'd like here. Do i just add estimatedResults and change the struct name?

src/invidious/yt_backend/extractors.cr
ChunkyProgrammer: Moved the parsing to hashtag.cr

src/invidious/hashtag.cr
ChunkyProgrammer: I have moved them to the same function but couldn't figure out how to reuse the object

src/invidious/hashtag.cr
```diff
@@ -1,42 +1,138 @@
...
 
-  def fetch(hashtag : String, page : Int, region : String? = nil) : Array(SearchItem)
-    cursor = (page - 1) * 60
-    ctoken = generate_continuation(hashtag, cursor)
+  struct HashtagPage
+    include DB::Serializable
+
+    property videos : Array(SearchItem) | Array(Video)
+    property header : HashtagHeader?
+    property has_next_continuation : Bool
```
ChunkyProgrammer, line 9: Added this because it'd be useful in determining if there's a next page or not

src/invidious/hashtag.cr
```diff
@@ -1,42 +1,138 @@
...
+        to_json(locale, json)
+      end
+    end
+
+    def to_json(json : JSON::Builder)
+      to_json(nil, json)
+    end
+  end
+
+  struct HashtagHeader
```
ChunkyProgrammer, line 42: TODO: remove this struct in favor of SearchHashtag (to be added in #3989 )

ChunkyProgrammer:
Updated and working again

ChunkyProgrammer:
Code is still working, I also added some tests to the description

syeopite (approved):

PR #3873, opened 2023-06-08
ApiOnly CI

### LINKED ISSUE #3871: [Enhancement] Provide api_only docker image
Currently it's hard to deploy api_only (no UI) invidious with docker, because the images don't include a version without UI.

Addition to CI to build images with the api_only flag.

kubo6472:
Closes #3871


unixfox:
In my opinion I think it's better to build the image inside the same workflow: https://github.com/iv-org/invidious/blob/master/.github/workflows/container-release.yml

kubo6472:
are you thinking another Dockerfile with [this](https://github.com/iv-org/invidious/blob/545a5937d87d31622e87bb2ba8151f8aecd66c81/docker/Dockerfile#L28) changed to include that apionly flag?

kubo6472:
Hey, I did move the `-Dapi_only` from GH actions to the Dockerfiles. I suppose this can't be merged this way into master, right?

SamantazFox:
If you put `-Dapi_only` in the container file, *all* builds will be API-only !

@unixfox I think it should be preferable to use `make` in the container. What do you think?
That way we could pass build args from docker directly to make, without all that `if` logic.

unixfox:
> If you put `-Dapi_only` in the container file, _all_ builds will be API-only !
> 
> @unixfox I think it should be preferable to use `make` in the container. What do you think? That way we could pass build args from docker directly to make, without all that `if` logic.

Good idea, I have just been lazy to implement it ^^.

unixfox:
So the goal is to use make in the dockerfile, make the Makefile options configurable then in github actions for each building step use the correct Makefile options for each kind of build (normal build, api only build, quic build).

kubo6472:
Tried using make to build docker images, probably wrong, may be on the right track. Please do take a look. Thanks

docker/Dockerfile
```diff
@@ -25,20 +26,13 @@ RUN crystal spec --warnings all \
...
-        --static --warnings all \
-        --link-flags "-lxml2 -llzma"; \
+        make RELEASE=1 \
+    elif [[ "${api_only}" == 1 ]] ; then \
+        make RELEASE=1 API_ONLY=1  \
     else \
-        crystal build ./src/invidious.cr \
-        --static --warnings all \
-        --link-flags "-lxml2 -llzma"; \
+        make RELEASE=0 \
```
SamantazFox, line 35: you could simplify all of this to a single line:
`RUN make "RELEASE=${release}" "DISABLE_QUIC=${disable_quic}" "API_ONLY=${api_only}"`

kubo6472:
If I understand that error correctly, probs means that Makefile isn't in that place from which the make command is run?

kubo6472:
Please do take a look when you'll have time. Thanks!

unixfox:
Can you squash your commits?

.github/workflows/ci.yml
```diff
@@ -124,5 +124,21 @@ jobs:
...
+  
+  build-docker-api-only:
+
+          runs-on: ubuntu-latest
+      
+          steps:
+            - uses: actions/checkout@v3
+      
+            - name: Build Docker
+              run: docker-compose build --build-arg release=0 --build-arg api_only=1
```
SamantazFox, line 136: Is there a reason why `with:` ... `build-args:` can't be used here?

SamantazFox (changes requested):

kubo6472:
Due to the merges I am unable to squash further. I hope this isn't an issue. I've made the changes that were last requested.

unixfox:
You need to have just one commit remaining in https://github.com/iv-org/invidious/pull/3873/commits.
Check this guide: https://www.git-tower.com/learn/git/faq/git-squash

Or you can do it the ugly way, like I sometimes do. Copy somewhere else the changes that you made, remove the branch locally, create the same branch name and replace with the changes that you copied somewhere else.

kubo6472:
best I could do, please let me know if it is okay now. for the future, I'm sure I'll prefer using branches like feature-apionly to not mess up the outdated master with the changes 😅

kubo6472:
Are the conflicts the problem, please? Thanks

docker/Dockerfile.arm64
```diff
@@ -20,26 +21,12 @@ COPY ./.git/ ./.git/
...
-        crystal build ./src/invidious.cr \
-        --release \
-        --static --warnings all \
-        --link-flags "-lxml2 -llzma"; \
-    else \
-        crystal build ./src/invidious.cr \
-        --static --warnings all \
-        --link-flags "-lxml2 -llzma"; \
-    fi
+RUN make "RELEASE=${release}" "DISABLE_QUIC=${disable_quic}" "API_ONLY=${api_only}"
```
SamantazFox, line 29:
```suggestion
RUN make "RELEASE=${release}" "API_ONLY=${api_only}"
```

SamantazFox: We've removed QUIC support since then ^^

SamantazFox (changes requested):

docker/Dockerfile
```diff
@@ -20,26 +21,12 @@ COPY ./.git/ ./.git/
...
-        crystal build ./src/invidious.cr \
-        --release \
-        --static --warnings all \
-        --link-flags "-lxml2 -llzma"; \
-    else \
-        crystal build ./src/invidious.cr \
-        --static --warnings all \
-        --link-flags "-lxml2 -llzma"; \
-    fi
+RUN make "RELEASE=${release}" "DISABLE_QUIC=${disable_quic}" "API_ONLY=${api_only}"
```
SamantazFox, line 29: That one still remains ^^
Can you also rebase your changes?

kubo6472:
Should be good to go :) @SamantazFox

kubo6472:
If I take the energy to update this again, will it be merged?

PR #3903, opened 2023-06-11
Temp explanation about video not available issue

unixfox:
Temporary message so that we get less duplicated issues and also explain a bit better the reasoning behind this error.

I have not tested if it gives a way to switch to another instance.

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -78,7 +78,11 @@ def extract_video_info(video_id : String, proxy_region : String? = nil)
   elsif video_id != player_response.dig("videoDetails", "videoId")
     # YouTube may return a different video player response than expected.
     # See: https://github.com/TeamNewPipe/NewPipe/issues/8713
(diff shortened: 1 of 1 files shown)


locales/en-US.json
```diff
@@ -484,5 +484,6 @@
     "channel_tab_streams_label": "Livestreams",
     "channel_tab_playlists_label": "Playlists",
     "channel_tab_community_label": "Community",
-    "channel_tab_channels_label": "Channels"
+    "channel_tab_channels_label": "Channels",
+    "temporary_video_not_available_explanation": "Can't load the video. YouTube is currently trying to block Invidious instances. <a href=\"`x`\">Click here for more info about the issue.</a>"
```
SamantazFox, line 488: I wouldn't add a translation string, as it will be picked up by weblate, and it'll take a while to translate in all langues, then to remove when this is (hopefully) over.

unixfox:
@SamantazFox I have instead switched to the "reason message" logic.

SamantazFox (approved):

PR #3917, opened 2023-06-13
Added API Authentication

SpongeManiac:
The current routes for authentication (`/register`, `/login`, and `/signout`) are not designed to handle API requests in order to authenticate. I have created the respective routes for authenticating through purely the API: `/api/v1/api_register`, `/api/v1/api_login/`, and `/api/v1/signout`. Any feed back would be appreciated. I tried to keep the general structure and methods consistent with the rest of the project, however I ran into a scope conflict when using a route like `/api/v1/register`, since these are already pointing to the front-end. I also added a few non-breaking changes to `Invidious::Helpers::Tokens.validate_request` and `Invidous::User::Captcha.generate_text` to make the authentication possible through the API.

As this is my first time ever creating a pull request, let alone touching Crystal Lang and Ruby, feel free to make changes and advice.

Issue #473, opened 2019-04-08
[RFC] Authentication and Authorization API


omarroth:
Related: #427, #469

For various applications it is desirable to be able to read or write user data. This proposal describes a format for tokens, a method for authenticating requests, and endpoints for authorizing tokens.

## Token format

Each token is a JSON object with the following schema:

```javascript
{
  "session": String,
  "expire": Int32?,
  "scopes": Array(String)
  "signature": String
}
```

Tokens will be created with a given `session`, which allows them to be revoked by the user.

Tokens may be issued with an `expire` timestamp after which the token will be considered invalid.

Tokens will have a list of one or more `scopes` determining their permissions. The format for `scopes` is described below.

Tokens will have a `signature` to ensure integrity of the other fields.

Example token:

```json
{
  "session": "v1:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
  "expires": 1554680038,
  "scopes": [
    ":notifications"
    ":subscriptions/*",
    "GET:tokens*",
  ],
  "signature": "p5RnTayE6Cy9Zxb/fP5n+r7zIQMC4oFtnkz/QzyI2qY="
}
```

## Authentication

All sub-routes of `/api/v1/auth` **must** be authenticated.

Endpoints requiring authentication are authenticated with an `Authentication: Bearer <token>` header.

Tokens are validated by taking all key-value pairs, minus `signature`, and creating the following string:

```
key1=value1
key2=value1,value2,value3
```

Values are alpha-sorted. Keys are alpha-sorted and joined by newlines. There is no trailing newline.

The `signature` is created by signing the string using SHA256-HMAC with the instance's `HMAC_KEY` and Base64 encoded.

Using the above token as an example:

```
expires=1554680038
scopes=GET:tokens*,:notifications,:subscriptions/*
session=v1:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

Signed using `SECRET_KEY` provides the signature `p5RnTayE6Cy9Zxb/fP5n+r7zIQMC4oFtnkz/QzyI2qY=`.

## Scoping

In order to provide a means of limiting permissions for an application, tokens must provide one or more `scopes`.

Each `scope` refers to a sub-route of `/api/v1/auth`. Scopes are defined as zero or more HTTP methods (semicolon-separated), colon (:), followed by an endpoint and optional wildcard for matching sub-routes.

[...]

PR #3923, opened 2023-06-16
Added Authentication API

SpongeManiac:
Adds the ability to perform authentication by purely using the API, JSON, and Cookie Headers.

An updated and squashed version of [my previous pull request](https://github.com/iv-org/invidious/pull/3917).
I am still learning pull requests, so any advice and pointers would be helpful. Thank you!

Added authentication.cr
Updated routes.cr
Updated captcha.cr
Updated tokens.cr

### DIFF
--- a/src/invidious/helpers/tokens.cr
+++ b/src/invidious/helpers/tokens.cr
@@ -86,7 +86,13 @@ def validate_request(token, session, request, key, locale = nil)
   end
 
   scopes = token["scopes"].as_a.map(&.as_s)
-  scope = "#{request.method}:#{request.path.lchop("/api/v1/auth/").lstrip("/")}"
+  if request.path.includes?("auth")
+    scope = "#{request.method}:#{request.path.lchop("/api/v1/auth/").lstrip("/")}"
+  elsif request.path.includes?("v1")
+    scope = "#{request.method}:#{request.path.lchop("/api/v1/").lstrip("/")}"
+  else
+    scope = "#{request.method}:#{request.path.lstrip("/")}"
+  end
   if !scopes_include_scope(scopes, scope)
     raise InfoException.new("Invalid scope")
   end
--- a/src/invidious/routes/api/v1/authentication.cr
+++ b/src/invidious/routes/api/v1/authentication.cr
@@ -1,0 +1,184 @@
+require "json"
+
+module Invidious::Routes::API::V1::Authentication
+  def self.api_register(env)
+    env.response.content_type = "application/json"
+    body_io = env.request.body
+    body_json = "{}"
+    if body_io
+      body_json = env.request.body.not_nil!.gets_to_end
+    end
+    if CONFIG.registration_enabled
+      creds = nil
+      begin
+        creds = Credentials.from_json(body_json)
+      rescue JSON::SerializableError
+      end
+      # get user info
+      if creds
+        locale = env.get("preferences").as(Preferences).locale
+        username = creds.username.downcase
+        password = creds.password
+        username = "" if username.nil?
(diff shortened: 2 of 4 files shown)


unixfox:
You technically don't have to create a new PR for squashing commits :) (next time it would be great to do it under the same PR).

SpongeManiac:
@unixfox I could not squash my commits in that branch. I tried several different ways, but to no avail. I'm not an expert at Github, so the easiest solution I could find was to cherry pick my commits without a merge, then squash, then merge. For whatever reason, Github would not allow me to squash these commits on that branch. My apologies if I'm doing something the round about way, I have never done a pull request, or had a pull request give me these errors. If there is a way to open the same PR but for a different branch, I would love to know.

SamantazFox:
What I can recommend for the next time you need to squash commits:
1. backup your repo (in case you make a mistake! That saved me countless times)
2. rebase your changes interactively: `git rebase -i master`
3. When you are happy with the code, push changes `git push --force-with-lease your_fork your-branch`

SamantazFox:
I have a few questions regarding this PR:
* Why is this feature needed for? Could you provide some use cases?
* Why do you want to automate registration using the API?
* Why do you need to log-in using cookies? Why can't you use `Authentication: Bearer <token>`?
* Have you looked at https://github.com/iv-org/invidious/issues/473 ?

SpongeManiac:
@SamantazFox I've looked through the API on the official website. Not only is it out of date, it is poorly documented. I was having troubles using token authentication because of the following issues:
1. There is no way to register, login, and create tokens through purely JSON
2. When using bearer tokens, you must url encode a JSON object, which is unnecessary and is frankly inefficient when you could just use JSON
3. The path `/authorize_token` and the path `/api/v1/auth/tokens/register` requires you to have authorized somehow already. This would mean I would have to use cookies, then generate a token. Why use the token if I could just use the cookie sid? Again, this route for some reason uses www encoded json, which seems inefficient considering you could just use JSON.
4. I am coding a cross-platform application to use the invidious API. To make API requests from one domain to another, CORS headers must be added. Authentication headers are not allowed in CORS requests, meaning I cannot use bearer token.

These routes I implemented are the easiest way to accomplish my task; a collection of endpoints that operate on purely JSON to perform registration and authentication. If you would like, I could create a separate PR to attempt to alter existing API to fit my needs while still functioning identically. The only reason I did not do this is because it would change how the Front End sends requests, and how the back end processes the requests.

SamantazFox:
yes, I know about the bad state of our API docs. Sorry for that :/

Not being able to register through the API is normal. I'm not aware of any website/service that allows to do so (they all require the user to manually create an account first). As for the login, it's not required once you have an API token. An SID cookie is simply an API token with a wildcard scope (`:*`). If you can't use the `Authentication` header, then you can still pass that token as the SID cookie.

But I'm wondering why you couldn't use the `Authentication` header. Its use is limited by CORS in the same way as the `Cookie` header. So if you can't use one, you can't use the other ([See MDN page for reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#requests_with_credentials)).

As far as my knowledge goes, the best option is to use your own invidious instance (so that you can control CORS response headers) or use some middleware to do the request, rather than handling that on the client side.

As for the `/authorize_token` route, like the `/login` one, you're supposed to redirect your user there, where they can grant access to your app, or not. PR #3556 has streamlined the authentication/authorization flow: you can redirect your user to `/authorize_token`, invidious will ask them to login, grant access, and redirect back to your site all at once.

I admit that the token management UI is terrible, and would need a refresh.

Issue #3909, opened 2023-06-12
[Enhancement] Support crystal 1.8.x


tiekoetter:
We should add the support for crystal 1.8.x

--------------------

**Describe the bug**
```
Done checking player dependencies, now compiling Invidious...
In src/invidious/views/search_homepage.ecr:7:4

 7 | <% end %>
        ^
Error: expecting token 'EOF', not 'end'
```

After trying to compile Invidious I get this error message. I have tried deleting and cloning the repository again but it still shows this error. 

**Steps to Reproduce**
1. Compile Invidious `shards install --production && crystal build src/invidious.cr --release`

**Additional context**
 - OS: Debian 11

tiekoetter:
Thank you @unixfox!

I switched to a local install of crystal and am now using version 1.4.1

markozajc:
I am currently using Crystal 1.8.2 to compile Invidious (latest master at the time of speaking) on Debian 12 and it works without problems (I did have to clear the cache after upgrading Debian but that was due to a libssl version change). I am using Crystal's opensuse package (from https://download.opensuse.org/repositories/devel:/languages:/crystal/Debian_11)
```
$ crystal -v
Crystal 1.8.2 [7aa5cdd86] (2023-05-09)

LLVM: 15.0.7
Default target: x86_64-unknown-linux-gnu
$ shards install --production && crystal build -s --no-debug --release -o
```
I don't get any compile-time errors.

unixfox:
Thank you for the heads-up @markozajc, you made me realized that crystal 1.8.x was working fine for me too.

@tiekoetter what was the crystal version that you were using?

OrvilleRed:
I get the exact same error. I am on Fedora 37 using the RPM crystal-1.8.2-5.1.x86_64

`Crystal 1.8.2 [7aa5cdd86] (2023-05-09)

LLVM: 15.0.7
Default target: x86_64-unknown-linux-gnu`

markozajc:
I just ran into this error after upgrading crystal (`1.8.2-1+5.1 -> 1.8.2-1+5.3` from the Debian_Testing repository on OBS). Turns out removing `~/.cache/{crystal,shards}` seems fix it, but I don't know what exactly causes it.

OrvilleRed:
Confirmed, moving back up to crystal-1.8.2 and recompiling Invidious worked just fine after deleting `~/.cache/{crystal,shards}`

[...]

PR #3936, opened 2023-06-25
workaround for https://github.com/iv-org/invidious/issues/3909

OrvilleRed:
If people get stuck with weird compiler errors after a Crystal upgrade, they can just do `make distclean` and re-build, then it should Just Work


Makefile
```diff
@@ -86,6 +86,7 @@ clean:
 
 distclean: clean
 	rm -rf libs
+	rm -rf rm -rf ~/.cache/{crystal,shards}
```
SamantazFox, line 89:
```suggestion
	rm -rf ~/.cache/{crystal,shards}
```

PR #3947, opened 2023-06-28
Fix searchbar twitch

AHOHNMYC:
`input` in `searchbar` has 1px margin out of focus. When it is on focus, this line causes element resize and twitch all below this bar.

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -237,7 +237,6 @@ div.watched-indicator {
 }
 
 .searchbar input[type="search"]:focus {
-	margin: 0;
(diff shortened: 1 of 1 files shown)


SamantazFox:
On what browser?

AHOHNMYC:
@SamantazFox This is case for Chrome 114.0.5735.199 and yewtu.be (can't repeat for obvious reasons~).
Upd. Whoops. Looks like, I've committed too fast as on other instances including test one I can't see any glitch 😅 Sorry for bothering

PR #3955, opened 2023-07-01
Config: Mandatory hmac key

### LINKED ISSUE #3854: [Announcement] hmac_key parameter in config.yml mandatory by July 1st

SamantazFox:
Closes #3854

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -455,13 +455,17 @@ jobs:
 #use_pubsub_feeds: false
 
 ##
-## HMAC signing key used for CSRF tokens and pubsub
+## HMAC signing key used for CSRF tokens, cookies and pubsub
 ## subscriptions verification.
+##
+## Note: This parameter is mandatory and should be a random string.
+## Such random string can be generated on linux with the following
+## command: `pwdgen 20 1`
 ##
 ## Accepted values: a string
 ## Default: <none>
 ##
-#hmac_key:
+hmac_key: "CHANGE_ME!!"
 
 ##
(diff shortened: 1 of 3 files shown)


config/config.example.yml
```diff
@@ -455,13 +455,17 @@ jobs:
...
 ##
+## Note: This parameter is mandatory and should be a random string.
+## Such random string can be generated on linux with the following
+## command: `pwdgen 20 1`
+##
 ## Accepted values: a string
 ## Default: <none>
 ##
-#hmac_key:
+hmac_key: "CHANGE_ME!!"
```
unixfox, line 468: My only fear is that people will keep `"CHANGE_ME!!"`. That's why there is a special case for searxng for the same reason: https://github.com/searxng/searxng/blob/8e8d8dabe9b17c9db8db7432c6bc063d9ae980d1/searx/webapp.py#L135

SamantazFox: Didn't think about that, thanks! It's fixed now.

SamantazFox:
Works as expected (tested on our test VPS).

unixfox (approved):

Issue #3959, opened 2023-07-02
[Enhancement] Search Watch History


TangentFoxy:
**Is your enhancement request related to a problem? Please describe.**
Sometimes I want to find a specific video in my watch history, but since I can only search YouTube itself, I can't find it easily.

**Describe the solution you'd like**
I would like the ability to search videos in watch history.

**Describe alternatives you've considered**
I mean, the only alternatives are non-viable in my opinion:
1. Manually clicking through watch history (and trying to guess video titles because #3035 doesn't exist either).
2. Searching all of YouTube (which because of Invidious' privacy, which I desire, returns video unrelated to me).

**Additional context**
I think making it similar to YouTube's own search of watch history would be ideal: A search box that only appears on the watch history page(s).

SamantazFox:
That's a good idea, but don't expect that to come soon.

Major pain points:
 * We'd need to store video titles (history is only video IDs at the moment)
 * This requires an full search indexer to be added (I don't know if we can use an existing lib for this), which could hurt performance a lot (invidious is currently single threaded, we have to iron out a lot of stuff to make multi-threading a thing).

s-h-a-d-o-w:
I don’t know how invidious stores data but given that YouTube’s own search is pretty bad (Time and time again, it doesn’t find videos that contain words X and Y in the title when searching for X. But when looking for Y, the video does show up.), even a simple substring match like databases usually offer in the title field would not just be great in general but also an improvement over whatever YouTube does. 😅

PR #3968, opened 2023-07-05
Improve rendering of the Watch History videos

### LINKED ISSUE #3035: [Enhancement] Title under videos for "Watch History"

sofiaritz:
Now the videos are rendered the same way as they are in the subscriptions feed.
This should fix #3035 and improve #3959.

This is my first contribution and my first time using Crystal, but this looks fine to me :)

Screenshots:
imagen

### DIFF
--- a/src/invidious/routes/feeds.cr
+++ b/src/invidious/routes/feeds.cr
@@ -128,6 +128,7 @@ module Invidious::Routes::Feeds
       watched = user.watched.reverse[(page - 1) * max_results, max_results]
     end
     watched ||= [] of String
+    watched = watched.map { |id| get_video(id) }
 
     templated "feeds/history"
   end
--- a/src/invidious/views/feeds/history.ecr
+++ b/src/invidious/views/feeds/history.ecr
@@ -31,20 +31,51 @@
     <% watched.each do |item| %>
         <div class="pure-u-1 pure-u-md-1-4">
             <div class="h-box">
-                <a style="width:100%" href="/watch?v=<%= item %>">
+                <a style="width:100%" href="/watch?v=<%= item.id %>">
                     <% if !env.get("preferences").as(Preferences).thin_mode %>
                         <div class="thumbnail">
-                            <img class="thumbnail" src="/vi/<%= item %>/mqdefault.jpg" alt="" />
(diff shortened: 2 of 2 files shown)


sofiaritz:
I've just realized that this removes the option to mark individual videos as unwatched, I'll fix that

src/invidious/routes/feeds.cr
```diff
@@ -128,6 +128,7 @@ module Invidious::Routes::Feeds
       watched = user.watched.reverse[(page - 1) * max_results, max_results]
     end
     watched ||= [] of String
+    watched = watched.map { |id| get_video(id) }
```
SamantazFox, line 131: Do mind that when you're calling `get_video`, you're asking invidious to get the entire watch page information from Youtube. Although these infos are cached, the chance of cache hit are thing, as the cache is regularly discarded to keep the video playback URLs "fresh" (these expire after 6h).

The biggest issue here is the potential privacy leak. On a very large instance, you likely hit the cache for some of these videos, so only a part of the watch history will be revealed, but on a small instance, you're pretty much guaranteed to send their entire watch history to Youtube whenever they load the page.

The second problem is performance. It works in reverse of the previous issue, meaning that this will impact large instances more. Even with the cache, imagine that if you have 200 users trying to see their 1000 videos-long watch history, that's 200 000 requests (well, actually 400 000) to Youtube! This will cripple down any medium or large sized instance.

SamantazFox:
I'm marking this PR as blocked/unfinished in regard of the comment made above.
Don't worry, it's just for us to keep track of things, you can still work on the PR :)

src/invidious/routes/feeds.cr
sofiaritz: Performance is definitively something that should be improved (maybe paging would be a good idea to improve this and the privacy issues?, I would like your opinions on that :).

About the privacy leak: when the client loads the thumbnails, the watch history is also sent to YouTube (although in a more "indirect" way). Is that any different to loading the page when talking about privacy?

sofiaritz:
I am not going to be able to work on this for now. I'm closing the PR for now.

PR #3973, opened 2023-07-08
Playlist import no refresh

iBicha:
Playlist imports are very slow because of video refresh (imagine many playlists, with 500 videos each)
I'm wondering if refreshing all metadata is necessary here?

Please let me know if this is a bad idea.

Thanks

### DIFF
--- a/src/invidious/user/imports.cr
+++ b/src/invidious/user/imports.cr
@@ -133,7 +133,7 @@ struct Invidious::User
             next if !video_id
 
             begin
-              video = get_video(video_id)
+              video = get_video(video_id, false)
             rescue ex
               next
             end


SamantazFox:
Unfortunately, that won't help much :/

The `refresh` parameter of `get_video` is useful only if you hit a cached entry. But on a playlist import (especially on a small instance), the chance of hitting the cache are low, so the metadata are fetched nonetheless.

The ideal solution would be to have a background job that fetches the metadata over the course of an hour, or something like that. But that requires a better cache (I'm working on it!) and to separate the import code from the DB/cache code (the import functions should return parsed data, and not touch the cache).

iBicha:
I see your point! There's lots of room to improve this.
What I did here will be especially useful to import a profile where most of the profile was already imported.

But my thought on playlists is that their video metadata should not be fetched on import at all, let that be on demand. Or maybe perhaps only the first video. Or maybe only fetch the first page.
If I'm importing 10 playlists of 500 videos, it is very unlike to watch most of it within the next 6 hours, so a lot of wasted bandwidth. So kinda advocating for selective caching instead of continuous caching, but I didn't dig enough to form an opinion

SamantazFox:
We're on the same wavelength here, yes.
There is a lot of work to improve that part of invidious!
And given the other PRs around imports/history, it's clearly becoming urgent to do something :/

iBicha:
@SamantazFox does this change make sense for now? It's a very small improvement that can be merged until something better lands

SamantazFox (approved):

PR #3978, opened 2023-07-12
Community: Parse quiz attachments

ChunkyProgrammer:
Parse quiz attachments: https://github.com/FreeTubeApp/FreeTube/issues/3746

### DIFF
--- a/src/invidious/channels/community.cr
+++ b/src/invidious/channels/community.cr
@@ -216,6 +216,22 @@ def fetch_channel_community(ucid, continuation, locale, format, thin_mode)
                     parse_item(attachment)
                       .as(SearchPlaylist)
                       .to_json(locale, json)
+                  when .has_key?("quizRenderer")
+                    json.object do
+                      attachment = attachment["quizRenderer"]
+                      json.field "type", "quiz"
+                      json.field "totalVotes", short_text_to_number(attachment["totalVotes"]["simpleText"].as_s.split(" ")[0])
+                      json.field "choices" do
+                        json.array do
+                          attachment["choices"].as_a.each do |choice|
+                            json.object do
+                              json.field "text", choice.dig("text", "runs", 0, "text").as_s
+                              json.field "isCorrect", choice["isCorrect"].as_bool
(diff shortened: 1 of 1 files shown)


src/invidious/frontend/comments_youtube.cr
```diff
@@ -89,6 +89,20 @@ module Invidious::Frontend::Comments
...
+              html << <<-END_HTML
+              <div class="pure-g video-iframe-wrapper">
+                <p>#{attachment["error"]}</p>
+              </div>
+              END_HTML
+            else
+              html << <<-END_HTML
+              <div class="pure-g video-iframe-wrapper">
+                <iframe class="video-iframe" src='/embed/#{attachment.dig?("videos", 0, "videoId")}?autoplay=0&list=#{attachment["playlistId"]}'></iframe>
+              </div>
```
SamantazFox, line 103: On Youtube it's showing the first video too?

ChunkyProgrammer: YouTube renders video and playlist attachments in the same way as in the `video` and `playlist` tab (so it wouldn't use embeds for either). I can remove the playlist rendering part of this code if preferred. I think in the future it'd be beneficial to be able to use item.ecr for the video and playlist attachments (Comments would need to be refactored even more if that were to happen)

SamantazFox: > I can remove the playlist rendering part of this code and only keep the quiz parsing if that's preferred.

Yes, please! I think a lot needs to be refactored in comments/community before we can cleanly adding new UI elements.

> I think in the future it'd be beneficial to be able to use item.ecr for the video and playlist attachments instead of using the embeds

Hum, I don't know... But that's a discussion for a separate issue ^^

SamantazFox (approved):

SamantazFox:
FYI: I'm working on #3988 because community tab was not loading on the test instance, so I wasn't able to test your PR!

SamantazFox:
Thanks for the PR :)

Issue #3933, opened 2023-06-23
[Bug] Redirect from short youtu.be URLs does not work


Trit34:
**Describe the bug**
`youtu.be/×××××` short URLs are usually converted as `[Invidious]/×××××` and they used to be redirected to the full URLs `[Invidious]/watch?v=×××××`.
As of today (but I don’t know since when exactly), it does not work anymore and such shortened URLs lead to the homepage of the instance instead.

**Steps to Reproduce**
Tested on FDN’s Invidious instance (updated each Sunday):
1. Go to https://invidious.fdn.fr/yWsS5UJf8yU (for example)
2. Instead of redirecting to https://invidious.fdn.fr/watch?v=yWsS5UJf8yU, it loads https://invidious.fdn.fr/feed/popular, the homepage.

**Additional context**
 - Browser (if applicable): Vivaldi
 - OS (if applicable): Arch Linux

unixfox:
Closing as it works fine on instances that are officially published on https://instances.invidious.io

Trit34:
> Closing as it works fine on instances that are officially published on https://instances.invidious.io

Not all of them, then: it works with yewtu.be (last updated on 11th June), but not with https://yt.artemislena.eu/, nor https://invidious.projectsegfau.lt/ (both last updated to 21st June, and both are instances on this list).

Maybe you don’t want to hear about it anymore since it (still?) works on the instances you tested with, but I wonder if there was a commit that break this redirect (since it seems to affect only instances updated circa after mid-June), and I’m not a programmer. Luckily, it’s a minor (but annoying) bug, but I guess that if other people notice it on their side, they will report it too.

unixfox:
Reopening as I can reproduce the issue on the test instance: https://test.invidious.io/yWsS5UJf8yU

I did close the issue before because it was working on an up to date instance: https://vid.puffyan.us/yWsS5UJf8yU

unixfox:
> There are suspicions that this is due to some kind of incorrect web server settings, but I don’t know with which ones.

No I don't think it's that. It used to work in the past and invidious never required any special configuration.

It could be that youtube is doing some A/B testing and an endpoint that we are using for detecting short URLs doesn't work anymore.

johackim:
I have this bug too.

I use this invidious configuration on my instance https://invidious.ethibox.fr.

The format https://invidious.ethibox.fr/watch?v=fmsoym8I-3o works correctly.

[...]

PR #3983, opened 2023-07-13
Update player.js

rnek0:
About issue : https://github.com/iv-org/invidious/issues/3933

I think it should be reviewed here: https://github.com/iv-org/invidious/blob/507bed6313b49564e53b69a5c9b4d072d1e05e4b/assets/js/player.js#L40 

short_url will be used here : https://github.com/iv-org/invidious/blob/507bed6313b49564e53b69a5c9b4d072d1e05e4b/assets/js/player.js#L119

since get embedCode() a little further down returns a correct code by calling the same function on line 127

I would suggest changing this: 
```var short_url = location.origin + '/' + video_data.id + embed_url.search;```
to this 
```var short_url = location.origin + '/watch?v=' + video_data.id + embed_url.search;```

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -37,7 +37,7 @@ if (player_data.aspect_ratio) {
 
 var embed_url = new URL(location);
 embed_url.searchParams.delete('v');
-var short_url = location.origin + '/' + video_data.id + embed_url.search;
+var short_url = location.origin + '/watch?v=' + video_data.id + embed_url.search;
 embed_url = location.origin + '/embed/' + video_data.id + embed_url.search;
 
 var save_player_pos_key = 'save_player_pos';


unixfox:
That's certainly not the correct way to fix this bug.
We should fix the original issue, which is that the redirection detection is broken, it used to work. Not introducing a workaround.

rnek0:
> That's certainly not the correct way to fix this bug. We should fix the original issue, which is that the redirection detection is broken, it used to work. Not introducing a workaround.

I thought it might come from routing on watch.cr, the only place that seemed a bit suspicious to me was here: https://github.com/iv-org/invidious/blob/507bed6313b49564e53b69a5c9b4d072d1e05e4b/src/invidious/routes/watch.cr#L30 
But it is possible that I am far from what cause this redirect, I know little about this code, I hope it can help.

Issue #3440, opened 2022-11-27
[Bug] Wrong tiles in story board


iBicha:
**Describe the bug**
/videos/:id returns wrong storyboardHeight

**Steps to Reproduce**

1. `/api/v1/videos/UhbixyxgsiU`
2. returns 3 storyboards. The second storyboard returns 

```
         "url":"/api/v1/storyboards/UhbixyxgsiU?width=60&height=45",
         "templateUrl":"https://i.ytimg.com/sb/UhbixyxgsiU/storyboard3_L1/M$M.jpg?sqp=-oaymwENSDfyq4qpAwVwAcABBqLzl_8DBgjBmdybBg%3D%3D&sigh=rs%24AOn4CLBujIIzxs6-nY318GTi_j8tt52HAw",
         "width":60,
         "height":45,
         "count":74,
         "interval":2000,
         "storyboardWidth":10,
         "storyboardHeight":10,
         "storyboardCount":1
```

The image is actually 10x8, not 10x10, making the storyboardHeight field wrong

**Logs**

**Screenshots**
<img width="599" alt="Screen Shot 2022-11-27 at 6 44 50 PM" src="https://user-images.githubusercontent.com/17722782/204166162-e9ed176d-07f3-4bda-ad8a-28080f3e93ee.png">

iBicha:
I see many more cases of this. 10x5 reported as 10x10
```
         "url":"/api/v1/storyboards/Gb4ghMzz0a4?width=25&height=45",
         "templateUrl":"https://i.ytimg.com/sb/Gb4ghMzz0a4/storyboard3_L1/M$M.jpg?sqp=-oaymwENSDfyq4qpAwVwAcABBqLzl_8DBgicmuCbBg%3D%3D&sigh=rs%24AOn4CLA7AtTjMrWA77A6PaTdlV8QL6jxNg",
         "width":25,
         "height":45,
         "count":48,
         "interval":1000,
         "storyboardWidth":10,
         "storyboardHeight":10,
         "storyboardCount":1
```
<img width="253" alt="Screen Shot 2022-11-27 at 8 04 49 PM" src="https://user-images.githubusercontent.com/17722782/204170944-2658d288-a327-46a4-a1fa-2206dc6e8c5d.png">

SamantazFox:
Oh, that's interesting! The code always assumed 10x10, but seems like things have changed (probably because of shorts and other newer formats?).

SamantazFox:
Actually, it seems normal?

[...]

PR #3987, opened 2023-07-15
Add thumbnails to dash

### LINKED ISSUE #3435: [Feature request] Thumbnails in DASH
I'm currently building a Roku frontend for Invidious ([Playlet](https://github.com/iBicha/playlet))
Roku expects thumbnails, aka storyboards (for "trick play" or scrubbing through the progress bar) to be included in the DASH-IF manifest ([DASH-standard thumbnail tiles](https://developer.roku.com/en-ca/docs/developer-program/media-playback/trick-mode/hls-and-dash.md#dash-standard-thumbnail-tiles))
So it would be cool to have an option where the storyboard included in the dash manifest, so that Roku can consume it properly

`/api/manifest/dash/id/:videoid` would return a DASH manifest that has the storyboard info included ([like this example](https://developer.roku.com/en-ca/docs/developer-program/media-playback/trick-mode/hls-and-dash.md#dash-manifest-example))

I've tried to feed the storyboard manually into the system, and I did not succeed.
My options are to recreate the thumbnail data somehow (which is not trivial) or generate an extra `<AdaptationSet>` section for the thumbnails in the DASH manifest.
The latter is the most doable solution, but it's still a bad solution because
- I will have to serve the DASH manifest manually in a local server, proxying the request for the manifest, and injecting the thumbnails section
- I should not modify the DASH manifest on the fly like that, because I risk breaking it

**Additional context**
Is this feature is to be considered (I hope) maybe it could be an option like `/api/manifest/dash/id/:videoid?storyboard=true` so that  the feature remains backward compatible with existing API

iBicha:
This is an attempt to close #3435 

This allows seeking with a preview in Playlet
Here's what it looks like
It's not perfect, but it's a start

Note: I made sure that this is behind a flag (thumbnails) so that it does not change the behaviour for anyone who did not opt-in

thumbnails


iBicha:
A little bump to draw the attention

iBicha:
@unixfox @SamantazFox sorry for the ping, I'm not sure if this PR is in some sort of queue, or forgotten
Thanks

SamantazFox:
@iBicha yeah, sorry, I haven't been much active this past month (the PR backlog have been filling up :c). I'll try to review and test that soon!

iBicha:
> @iBicha yeah, sorry, I haven't been much active this past month (the PR backlog have been filling up :c). I'll try to review and test that soon!

Thank you, I know finding time for FOSS is hard! Cheers

src/invidious/routes/api/manifest.cr
```diff
@@ -126,6 +127,62 @@ module Invidious::Routes::API::Manifest
...
+
+              storyboards.each do |storyboard|
+                interval = storyboard[:interval].to_i
+                # interval can be 0 https://github.com/iv-org/invidious/issues/3441
+                next if interval == 0
+
+                storyboard_height = storyboard[:storyboard_height]
+
+                # BUG: storyboardHeight can be wrong https://github.com/iv-org/invidious/issues/3440
+                # this fix/assumption is only if we have storyboard[:storyboard_count] = 1
```
iBicha, line 145: Based on discussion in https://github.com/iv-org/invidious/issues/3440, this is not the right assumption.
We need to "calculate" the image size based on which page we're on, storyboard count, and tile size

iBicha:
@SamantazFox I'll refactor this after #4153 is merged

iBicha:
Reflecting on this PR a bit, I will be closing this.
I'm trying to add DASH features primarily to support Playlet. I felt that the DASH generation belongs in Invidious, since it is a standard. But the more I dig into it, I discover limitations that are not part of the standard, and that are specific to Roku.
I asked Roku specifically about supporting tiles with different sizes (like in our case with the last page) and their response was in the lines of 
> ...There are all kinds of variations in DASH. Based on needs we could implement those but this needs to be based on a business-related need (like loss or revenue and so on).

So there will always be quirks or limitations, and they will not be prioritized unless they can justify it.

With that reasoning, I'll will be advocating to add things to Invidious that are not standards, or perhaps includes hacks and workarounds, so they fit the Roku OS. That does not sound right, that's why I'll be moving this piece to Playlet so I can cater to these cases (there are many differences between a standard dash player and Roku, including codecs, languages, containers and so on).

Sucks, but it is what it is.

Thank you!

PR #3988, opened 2023-07-16
Channels: Use innertube to fetch the community tab

SamantazFox:
This was long overdue: don't rely on HTML scraping anymore, as these endpoints are rate limited. So now the community tab is extracted using innertube, and extracted using `extract_items(&)` as everything else.

The required Parser module (in `src/invidious/yt_backend/extractors.cr`) is for a future PR.


SamantazFox:
Continuation extraction is broken

SamantazFox:
It's fixed! PR deployed on the test instance!

PR #3989, opened 2023-07-16
Search: Add hashtag result

SamantazFox:
See https://github.com/FreeTubeApp/FreeTube/pull/3483#issuecomment-1537516380


src/invidious/yt_backend/extractors.cr
```diff
@@ -210,6 +211,56 @@ private module Parsers
...
+  # It can be found when searching for any hashtag (e.g "#hi" or "#shorts")
+  module HashtagRendererParser
+    def self.process(item : JSON::Any, author_fallback : AuthorFallback)
+      if item_contents = item["hashtagTileRenderer"]?
+        return self.parse(item_contents)
+      end
+    end
+
+    private def self.parse(item_contents)
+      text = item_contents.dig("hashtag", "simpleText").as_s # E.g "#hi"
```
ChunkyProgrammer, line 227: Should `extract_text` be used instead of checking the simpleText?

SamantazFox: You're right, thanks!

ChunkyProgrammer:
I've done some testing of the api through FreeTube and it is working great. 

Testing with the Invidious frontend, it looks like a locale string is missing so it's not displaying the channel count on the invidious side Screenshots_2023-07-18-07-29-13

SamantazFox:
Nice catch, thank you!

SamantazFox:
Deployed on the test instance, works as expected!

Issue #3752, opened 2023-04-15
[Enhancement] Force disable quic in the code


unixfox:
Google disabled their old QUIC protocols on their server side, and now lsquic doesn't work anymore. It just timeout after 5 seconds.

So if you have force quic using `quic: true`, invidious will just won't work anymore. We should force the inability to use quic in the code.

Confirmed with the test code in https://github.com/iv-org/lsquic.cr#usage

PR #4008, opened 2023-07-26
Remove lsquic entirely

syeopite:
Seems to work fine though I have no way of making sure the storyboard functionalities are kept since I can't test that on my local instance. It should be tested just in case. 

Close https://github.com/iv-org/invidious/issues/3752

### DIFF
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -52,7 +52,7 @@ jobs:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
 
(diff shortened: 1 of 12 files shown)


syeopite:
Amoeba is broken on Crystal 1.9. See https://github.com/crystal-ameba/ameba/issues/372. Once v1.5 of Ameba comes out it should be fixed.

unixfox (approved):

SamantazFox (approved):
Code looks good!

unixfox:
I would like to merge this because I have been using invidious without quic compiled for months and this works great.

syeopite:
Safe to merge then?

unixfox:
Merging as it's has been stable for a long time now.

Issue #3953, opened 2023-06-30
[Bug] Unable to import subscriptions


wisheth:
**Describe the bug**
Sent to error page when attempting to import subscriptions from FreeTube in every compatible format other than NewPipe .json.

**Steps to Reproduce**
- Export from other service in a format other than NewPipe
- Log in
- From home page, go to Subscriptions -> Manage subscriptions -> Import/Export subscriptions
- Upload a file into any of the file uploads
- Click "import"

**Logs**
```
Expected Hash for #[]?(key : String), not Array(JSON::Any) (Exception)
  from /usr/share/crystal/src/json/any.cr:117:7 in '[]?'
  from src/invidious/user/imports.cr:95:10 in 'from_invidious'
  from src/invidious/routes/preferences.cr:303:11 in 'update_data_control'
  from lib/kemal/src/kemal/route.cr:12:9 in '->'
  from src/invidious/helpers/handlers.cr:30:37 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/ext/kemal_static_file_handler.cr:112:11 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
  from /usr/share/crystal/src/http/server.cr:500:5 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from ???
```

**Additional context**
Attempted on Firefox (latest), Windows 11.

ttsoares:
What is the exact format that a .csv file should have to be accepted by the import ?
I tried this process today with no avail... used just one line:
[...]

Issue #3732, opened 2023-04-10
[Bug] can't import freetube subscriptions - unexpected token '{' at line 2, column 1 (JSON::ParseException)


BjarneVerschorre:
**Describe the bug**
I went to import my freetube subscriptions (.db file) and I got this error
-->

**Logs**
Title: `Expected Hash for #[](key : String), not Array(JSON::Any) (Exception)`
Date: `2023-04-10T22:58:19Z`
Route: `/data_control?referer=%2Ffeed%2Fpopular`
Version: `2023.04.01-8abbfd2 @ master`

<details>
<summary>Backtrace</summary>
<p>
   
```
Expected Hash for #[](key : String), not Array(JSON::Any) (Exception)
  from /usr/share/crystal/src/json/any.cr:105:7 in '[]'
  from src/invidious/user/imports.cr:162:9 in 'update_data_control'
  from lib/kemal/src/kemal/route.cr:12:9 in '->'
  from src/invidious/helpers/handlers.cr:30:37 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from src/ext/kemal_static_file_handler.cr:112:11 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:28:7 in 'call_next'
  from lib/kemal/src/kemal/init_handler.cr:12:7 in 'process'
  from /usr/share/crystal/src/http/server.cr:500:5 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from ???
```
</p>
</details>

ChunkyProgrammer:
A workaround could be exporting the file in the New Pipe format in the FreeTube app

[...]

PR #4011, opened 2023-07-27
User: Fix importing FreeTube subscriptions with multiple profiles

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/3953
closes https://github.com/iv-org/invidious/issues/3732

Example subscription file (extract the zip):
[freetube-subscriptions-2023-07-26.zip](https://github.com/iv-org/invidious/files/12178655/freetube-subscriptions-2023-07-26.zip)

### DIFF
--- a/src/invidious/user/imports.cr
+++ b/src/invidious/user/imports.cr
@@ -228,8 +228,12 @@ struct Invidious::User
       subs = matches.map(&.["channel_id"])
 
       if subs.empty?
-        data = JSON.parse(body)["subscriptions"]
-        subs = data.as_a.map(&.["id"].as_s)
+        profiles = body.split('\n', remove_empty: true)
+        profiles.each do |profile|
+          if data = JSON.parse(profile)["subscriptions"]?
+            subs += data.as_a.map(&.["id"].as_s)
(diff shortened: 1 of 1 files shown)


src/invidious/user/imports.cr
```diff
@@ -228,8 +228,13 @@ struct Invidious::User
       subs = matches.map(&.["channel_id"])
 
       if subs.empty?
-        data = JSON.parse(body)["subscriptions"]
-        subs = data.as_a.map(&.["id"].as_s)
+        profiles = body.split('\n')
+        profiles.pop
```
SamantazFox, line 232:
```suggestion
        profiles = body.split('\n', remove_empty: true
```

By doing so, so can be sure that no "profile" will is an empty string, and you ensure that if there is a single profile, it won't be dropped.

SamantazFox (approved):

SamantazFox (approved):

SamantazFox:
Works like a charm ^^ thanks for providing the test file!

SamantazFox:
Thanks for the PR :)

PR #4033, opened 2023-08-06
Shards: Update ameba version and fix typo

kz6wk9:
a bug when using recent versions of the crystal language may appear when building with ameba as described in:
https://github.com/crystal-ameba/ameba/issues/371
https://github.com/crystal-ameba/ameba/issues/372

this is fixed in version 1.5.0 by:
https://github.com/crystal-ameba/ameba/pull/373

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -1,5 +1,9 @@
 version: 2.0
 shards:
+  ameba:
+    git: https://github.com/crystal-ameba/ameba.git
+    version: 1.5.0
+
   athena-negotiation:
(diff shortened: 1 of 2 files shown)


syeopite (approved):
LGTM! Though while you're at it could you also fix the typo here?

https://github.com/iv-org/invidious/blob/701f03d5e88a8f3c92575965e39b8f96b59f71f9/shard.yml#L6

SamantazFox (approved):

PR #4036, opened 2023-08-07
Helm chart enhancements

obeone:
Hi,

Here are some proposed enhancements to the existing helm chart : 
- PostgreSQL : 
  - Optional, with chart dependancy not loaded if embedded pgsql is disabled
  - Add an initContainer which run migration on every start. So it's populate the db on firt start, and avoid the user to have to create an initdb configMap. And if there is schema update, migration will be ran
  - If using embedded postgresql, the hostname is automatically discovered
- Add an ingress template
- Add the mandatory hmac_key in default config

Hope this is useful !


kubernetes/values.yaml
```diff
@@ -53,9 +66,12 @@ config:
...
-    host: invidious-postgresql
+    #-- Leave empty if you use embeded postgresql server
+    host:
     port: 5432
     dbname: invidious
   full_refresh: false
   https_only: false
   domain:
+  #-- Generate a key using `pwgen 20 1` for example
+  hmac_key: CHANGEME
```
ChunkyProgrammer, line 77: `CHANGE_ME!!` is checked for in the code so it might be better to use that
https://github.com/iv-org/invidious/blob/701f03d5e88a8f3c92575965e39b8f96b59f71f9/src/invidious/config.cr#L210

obeone: Oh yes, indeed !

obeone:
Hi,

You're absolutely right. The initdb method used in docker-compose or the original chart indeed only works with a fresh PostgreSQL install, making it impossible to use with an existing DB.

The "--migrate" method, on the other hand, seems to work well with both. At first glance, I don't see a significant difference between a fresh or an existing PG installation, hence this method simplifies the code by obviating the need for two different approaches.

However, one aspect that remains unclear to me is the handling of schema changes during an upgrade. Does Invidious manage this automatically? If so, I can adjust my approach to use a one-time init job. Nonetheless, I remain convinced that a "one method fits all" solution is more efficient and better overall.

unixfox:
> However, one aspect that remains unclear to me is the handling of schema changes during an upgrade. Does Invidious manage this automatically? If so, I can adjust my approach to use a one-time init job. Nonetheless, I remain convinced that a "one method fits all" solution is more efficient and better overall.

That's exactly what I said above, we have not yet defined a clear path for migration, whatever it is running on kubernetes or docker or on the host directly.

obeone:
So it will not being merged ? :/

unixfox:
No, see https://github.com/iv-org/invidious/pull/4036#issuecomment-1715788489

PR #4037, opened 2023-08-07
Use workaround for fetching streaming URLs

### LINKED ISSUE #4027: [Bug] The media could not be loaded
**Describe the bug**

most videos can play, but only a few videos can't play

**Steps to Reproduce**

xxx.com/watch?v=Srjk6_j8J04
**Logs**

The media could not be loaded, either because the server or network failed or because the format is not supported.
**Screenshots**

image

**Additional context**

syeopite:
Closes #4027

YouTube appears to be A/B testing some new integrity checks. Adding the parameter "CgIQBg" to InnerTube player requests appears to workaround the problem

See https://github.com/TeamNewPipe/NewPipeExtractor/pull/1084

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -55,8 +55,9 @@ def extract_video_info(video_id : String, proxy_region : String? = nil)
   client_config = YoutubeAPI::ClientConfig.new(proxy_region: proxy_region)
 
   # Fetch data from the player endpoint
-  # 8AEB param is used to fetch YouTube stories
-  player_response = YoutubeAPI.player(video_id: video_id, params: "8AEB", client_config: client_config)
+  # CgIQBg is a workaround for streaming URLs that returns a 403.
+  # See https://github.com/iv-org/invidious/issues/4027#issuecomment-1666944520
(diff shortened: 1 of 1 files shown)


src/invidious/videos/parser.cr
```diff
@@ -55,8 +55,9 @@ def extract_video_info(video_id : String, proxy_region : String? = nil)
   client_config = YoutubeAPI::ClientConfig.new(proxy_region: proxy_region)
 
   # Fetch data from the player endpoint
-  # 8AEB param is used to fetch YouTube stories
```
syeopite, line 58: We can safely remove the `8AEB` param since stories was entirely removed by YouTube back in late June

SamantazFox (approved):

SamantazFox:
Is there a difference with and without the padding bytes (%3D in base64)?
https://github.com/TeamNewPipe/NewPipe/issues/9038#issuecomment-1289756816

SamantazFox:
Nevermind, NP uses the non-padded version, let's use that too.

unixfox (approved):

SamantazFox:
For the record, I did integrate that change into the test instance the other day (before the weekend) and it sure does work. I was planning to make a PR, but thanks for doing it instead ^^

PR #4040, opened 2023-08-08
Add workaround for storyboards on local instances

### LINKED ISSUE #3117: [Bug] Player uses `/api/v1` for storyboards
**Describe the bug**
When fetching storyboards, the player uses `/api/v1/storyboards/sb/` instead of `/sb`.

**Steps to Reproduce**
I'm not sure what causes it.

**Logs**

**Screenshots**

**Additional context**

syeopite:
Closes #3117

Also fixes a regression from #4037 that broke storyboards for everyone. 

-----

workaround

This PR adds a quick workaround within the 404 handler to allow storyboards to function on instances without `domain`, `external_port` and `https_only` set.

We should probably consider switching to [the freetube fork here](https://github.com/FreeTubeApp/videojs-vtt-thumbnails)

### DIFF
--- a/src/invidious/routes/errors.cr
+++ b/src/invidious/routes/errors.cr
@@ -1,6 +1,6 @@
 module Invidious::Routes::ErrorRoutes
   def self.error_404(env)
-    # Workaround for # 3117
+    # Workaround for #3117
     if HOST_URL.empty? && env.request.path.starts_with?("/v1/storyboards/sb")
       return env.redirect "#{env.request.path[15..]}?#{env.params.query}"
     end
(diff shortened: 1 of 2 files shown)


SamantazFox (approved):

Fijxu:
Tested it, works good. If I comment/unset the 3 settings that @syeopite mentions, storyboards appear without problems when connecting locally to the instance and via the HTTPS domain. However if I set the 3 settings mentioned, storyboards **no longer work when connecting locally** but they work if I connect via the HTTPS domain. I guess this is the intended purpose of this pull request

image

SamantazFox:
@syeopite Thanks for the PR :)

PR #4001, opened 2023-07-23
Captions: Add ability to use Innertube's transcripts API

syeopite:
Closes #2567

If applicable, I'll like to get the bounty. 

Theoretically, this should reenable closed captions support on larger instances that would otherwise be rate-limited by YouTube. 

There are some minor differences between the text received via transcripts and the regular timed text but nothing too major according to my testing. In addition, this will likely block us off implementing subtitle styling, animation, etc seen on YouTube until YT's timedtext API becomes an InnerTube endpoint. 

**To enable, simply set the `use_innertube_for_captions` to true within the config.**

PR #4070, opened 2023-08-24
Refactor WebVTT building logic into WebVTT::Builder

syeopite:
https://github.com/iv-org/invidious/pull/4001#discussion_r1303860270

Syntax is similar to `JSON.build`. 

```crystal
string = WebVTT.build do |vtt|
  vtt.cue(Time::Span.new(seconds: 1), Time::Span.new(seconds: 2), "Line 1")
  vtt.cue(Time::Span.new(seconds: 2), Time::Span.new(seconds: 3), "Line 2")
end

string # => "WEBVTT\n\n00:00:01.000 --> 00:00:02.000\nLine 1\n\n00:00:02.000 --> 00:00:03.000\nLine 2\n\n"
```

### DIFF
--- a/src/invidious/helpers/webvtt.cr
+++ b/src/invidious/helpers/webvtt.cr
@@ -1,0 +1,67 @@
+# Namespace for logic relating to generating WebVTT files
+#
+# Probably not compliant to WebVTT's specs but it is enough for Invidious.
+module WebVTT
+  # A WebVTT builder generates WebVTT files
+  private class Builder
+    def initialize(@io : IO)
+    end
+
+    # Writes an vtt line with the specified time stamp and contents
+    def line(start_time : Time::Span, end_time : Time::Span, text : String)
+      timestamp(start_time, end_time)
+      @io << text
+      @io << "\n\n"
+    end
+
+    private def timestamp(start_time : Time::Span, end_time : Time::Span)
+      add_timestamp_component(start_time)
+      @io << " --> "
+      add_timestamp_component(end_time)
+
+      @io << '\n'
+    end
+
+    private def add_timestamp_component(timestamp : Time::Span)
+      @io << timestamp.hours.to_s.rjust(2, '0')
+      @io << ':' << timestamp.minutes.to_s.rjust(2, '0')
+      @io << ':' << timestamp.seconds.to_s.rjust(2, '0')
+      @io << '.' << timestamp.milliseconds.to_s.rjust(3, '0')
+    end
+
+    def document(setting_fields : Hash(String, String)? = nil, &)
+      @io << "WEBVTT\n"
+
+      if setting_fields
+        setting_fields.each do |name, value|
+          @io << "#{name}: #{value}\n"
+        end
+      end
+
+      @io << '\n'
+
+      yield
+    end
+  end
+
+  # Returns the resulting `String` of writing WebVTT to the yielded WebVTT::Builder
+  #
+  # ```
+  # string = WebVTT.build do |io|
+  #   vtt.line(Time::Span.new(seconds: 1), Time::Span.new(seconds: 2), "Line 1")
+  #   vtt.line(Time::Span.new(seconds: 2), Time::Span.new(seconds: 3), "Line 2")
+  # end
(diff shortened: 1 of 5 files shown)


src/invidious/helpers/webvtt.cr
```diff
@@ -0,0 +1,67 @@
...
+  # string = WebVTT.build do |io|
+  #   vtt.cue(Time::Span.new(seconds: 1), Time::Span.new(seconds: 2), "Line 1")
+  #   vtt.cue(Time::Span.new(seconds: 2), Time::Span.new(seconds: 3), "Line 2")
+  # end
+  #
+  # string # => "WEBVTT\n\n00:00:01.000 --> 00:00:02.000\nLine 1\n\n00:00:02.000 --> 00:00:03.000\nLine 2\n\n"
+  # ```
+  #
+  # Accepts an optional settings fields hash to add settings attribute to the resulting vtt file.
+  def self.build(setting_fields : Hash(String, String)? = nil, &)
```
syeopite, line 59: The `setting_fields` here is mostly for backwards compatibility. I'm pretty sure — or at least I couldn't find any reference to it — that the fields Invidious adds aren't actually apart of WebVTT

src/invidious/helpers/webvtt.cr
```diff
@@ -0,0 +1,67 @@
...
+      @io << "\n\n"
+    end
+
+    private def timestamp(start_time : Time::Span, end_time : Time::Span)
+      add_timestamp_component(start_time)
+      @io << " --> "
+      add_timestamp_component(end_time)
+
+      @io << '\n'
+    end
```
SamantazFox, line 23: Why making a dedicated function for that? Can't it be integrated into `cue()`?

src/invidious/helpers/webvtt.cr
```diff
@@ -0,0 +1,67 @@
...
+      @io << ':' << timestamp.seconds.to_s.rjust(2, '0')
+      @io << '.' << timestamp.milliseconds.to_s.rjust(3, '0')
+    end
+
+    def document(setting_fields : Hash(String, String)? = nil, &)
+      @io << "WEBVTT\n"
+
+      if setting_fields
+        setting_fields.each do |name, value|
+          @io << "#{name}: #{value}\n"
```
SamantazFox, line 37:
```suggestion
          @io << name << ": " << value << '\n'
```

src/invidious/helpers/webvtt.cr
SamantazFox: Yeah, it looks completely non-standard. I couldn't find anything about them [on MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API)

SamantazFox (approved):
You even wrote a spec for it <3

src/invidious/helpers/webvtt.cr
SamantazFox: Yeah, it looks completely non-standard. I couldn't find anything about them [on MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API)

src/invidious/helpers/webvtt.cr
syeopite: Oh interesting! I didn't know you could do that in Crystal

src/invidious/helpers/webvtt.cr
syeopite: No reason in particular. Imo it feels cleaner to treat the timestamp as a single entity in `#cue` rather than having it be made out of two timestamp components and an arrow; thus a function. Do you want me to get rid of `#timestamp`?

SamantazFox: Oh okay. Keep it like that, then!

SamantazFox:
Can you please rebase your changes?

syeopite:
Rebased

SamantazFox (approved):

SamantazFox:
It works like a charm :D

SamantazFox:
Thanks for your PR :D

Issue #4074, opened 2023-08-27
[Bug] On iOS: Screen turns off after screen timeout in Loop mode


138138138:
**Describe the bug**
On iOS devices, when playing videos, the screen turns off after some time, despite the fact that a video is playing.

**Steps to Reproduce**
1. In iOS Settings, Go to Display & Brightness, Set the Auto-Lock to 30 seconds
2. Go to Face ID & Passcode, Turn off Attention Aware Features
3. Turn on Always loop on Invidious
4. Play any video
5. Wait about 30 seconds
6. The screen turns off and phone is locked

PR #4075, opened 2023-08-27
Fix iOS screen timeout on video playback

138138138:
Fixes: https://github.com/iv-org/invidious/issues/4074

After declaring player.muted,
If player volume is set to greater than 0, the screen stays on and will not timeout.
If player volume is set to 0, the screen will not stay on until user press the speaker icon to unmute the video.

Setting player volume to 0 also allows video autoplay in muted on iOS.


138138138:
This method is not working 100% of the time. I need to further modify the code

PR #4076, opened 2023-08-27
Player: Fix iOS screen timeout in loop mode

138138138:
Fixes: https://github.com/iv-org/invidious/issues/4074

iOS seems to turn off the screen when video is in loop mode.
This workaround turns off builtin loop, and set position to 0 when video ends.


SamantazFox:
Deployed on the test instance!

SamantazFox:

PR #4083, opened 2023-08-31
Add blockage status endpoint 

### LINKED ISSUE #3957: [Feature request] API endpoint that informs if the instance has been blocked by Youtube
As an Invidious instance maintainer, it is hard to know whether my instance is really working. The invidious software might run just fine and return 200 responses even though it has been blocked by Youtube. It often takes some time to be aware that the instance is not loading Youtube videos. 

This could be fixed by implementing an API endpoint that tells if Youtube has blocked the instance. This endpoint could be used then by an external uptime monitoring software like [Uptime Kuma](https://github.com/louislam/uptime-kuma) or [Gatus](https://github.com/TwiN/gatus). To achieve compatibility with different kinds of software, I wish the endpoint had a clear status code informing the blockage state (for example [503 - service unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)).

The existing `/api/v1/stats` endpoint could be used for this. In my opinion the endpoint should also return some json data for other software indicating is the instance blocked, without needing to look into status codes. 

```json
{
   "blocked": <boolean>
}
```

syeopite:
Closes #3957 

This PR works by utilizing a background job that runs every 30 minutes to check for the blockage of video playback. The result is then shown on the `/api/v1/information/status/blocked` endpoint.

I have not a clue whether or not this actually works 😅 but theoretically it should show whether or not an instance has been blocked. Testing is definitely needed.

### DIFF
--- a/src/invidious/jobs/blockage_status_job.cr
+++ b/src/invidious/jobs/blockage_status_job.cr
@@ -14,7 +14,7 @@ class Invidious::Jobs::CheckBlockageStatusJob < Invidious::Jobs::BaseJob
     loop do
       begin
         # TODO find performant way of fetching a random video from the videos table.
-        video = fetch_video("zj82_v2R6ts", nil)
+        video = fetch_video("jNQXAC9IVRw", nil)
 
         if !video.nil?
           # Assume unblocked
--- a/src/invidious/jobs/blockage_status_job.cr
+++ b/src/invidious/jobs/blockage_status_job.cr
@@ -1,0 +1,51 @@
+class Invidious::Jobs::CheckBlockageStatusJob < Invidious::Jobs::BaseJob
+  private getter db : DB::Database
+
+  BLOCKAGE_STATUS = {
+    "version" => "1.0",
+    "blocked" => false,
+  }
+
+  def initialize(@db)
+  end
+
+  def begin
+    # Logic mostly taken from bypass_captcha_job.cr
+    loop do
+      begin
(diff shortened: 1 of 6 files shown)


unixfox:
Shouldn't this be listed inside /api/v1/stats?

src/invidious/jobs/blockage_status_job.cr
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize(@db)
+  end
+
+  def begin
+    # Logic mostly taken from bypass_captcha_job.cr
+    loop do
+      begin
+        # TODO find performant way of fetching a random video from the videos table.
+        video = fetch_video("zj82_v2R6ts", nil)
```
unixfox, line 17: Usually we take the first youtube video, the zoo one.

syeopite: I just took the video from the bypass captcha job. I'd go ahead and switch to the zoo one

TheFrenchGhosty:
"meta" might be better than "information"

absidue:
Probably also a good idea to only request a range of the video, as YouTube throttles requests without a range (that's why yt-dlp and youtube-dl download in 10MiB chunks). Probably best to avoid the blockage check job, contributing to the problem.

unixfox:
> Probably best to avoid the blockage check job, contributing to the problem.

That's also what I have also been thinking. We could do a passive check, a check based on what the users load as videos, but this will be much more difficult to implement.

PR #4085, opened 2023-09-01
Add playback success rate to `/api/v1/stats`

### LINKED ISSUE #3957: [Feature request] API endpoint that informs if the instance has been blocked by Youtube

syeopite:
https://github.com/iv-org/invidious/pull/4083#issuecomment-1702660189

This PR tracks the total amount of `/videoplayback` requests and the ones that are successful in order to create a playback success rate in the `/api/v1/stats` endpoint. 

By default it is an empty object:
```json
"playback" : {}
```

But as soon as a video is played:

```json
"playback": {
    "totalRequests": 19,
    "successfulRequests": 19,
    "ratio": 1.0
}
```

The caveat however is that with the way I have it currently set up the playback object gets reset back to an empty object each time the stats is refreshed. This is to prevent a success rate so high that 403 requests won't be able to budge the ratio at all. A better solution is probably needed there. 

Closes #3957

### DIFF
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -223,5 +223,5 @@ def get_playback_statistic
     Invidious::Jobs::StatisticsRefreshJob::STATISTICS["playback"] = tracker
   end
 
-  return tracker
+  return tracker.as(Hash(String, Int64 | Float64))
 end
--- a/src/invidious/helpers/helpers.cr
+++ b/src/invidious/helpers/helpers.cr
@@ -208,3 +208,20 @@ def proxy_file(response, env)
     IO.copy response.body_io, env.response
   end
 end
+
+# Fetch the playback requests tracker from the statistics endpoint.
+#
+# Creates a new tracker when unavailable.
+def get_playback_statistic
+  if (tracker = Invidious::Jobs::StatisticsRefreshJob::STATISTICS["playback"]) && tracker.as(Hash).empty?
+    tracker = {
+      "totalRequests"      => 0_i64,
+      "successfulRequests" => 0_i64,
+      "ratio"              => 0_f64,
+    }
+
+    Invidious::Jobs::StatisticsRefreshJob::STATISTICS["playback"] = tracker
+  end
+
+  return tracker
+end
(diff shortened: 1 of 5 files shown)


unixfox:
> The caveat however is that with the way I have it currently set up the playback object gets reset back to an empty object each time the stats is refreshed. This is to prevent a success rate so high that 403 requests won't be able to budge the ratio at all. A better solution is probably needed there.

Not even sure if we have a better solution than that? @SamantazFox thoughts?

I think this PR is critical if we want to show to our users which instance works correctly or not.

unixfox:
Very strange on my instance the stats are not getting updated, it stay at:
```
"playback" : {}
```

syeopite:
Huh that is weird. I can't replicate this on my end. Is another patch possibly interfering with it?

unixfox:
Oops I think I had an issue when testing the PR, it wasn't sending the requests to local /videoplayback endpoint.

Now it works:
```json
"playback":{"totalRequests":3,"successfulRequests":0,"ratio":0.0}
```

But the stats do not stay for very long. Like after 10 seconds they are gone.

unixfox (approved):

syeopite:
That can be mitigated if we bump the default stats refresh interval from one minute to something higher

unixfox:
> That can be mitigated if we bump the default stats refresh interval from one minute to something higher

Could you do this change in this PR?

syeopite:
I've bumped it to ten minutes

unixfox:
I have merged this PR because I consider it working. And we can always improve the PR later.

Also, this is really useful for the project https://github.com/iv-org/smart-ipv6-rotator

unixfox:
Now that I think of it. Wouldn't have been better to have a percentage instead of a ratio?

Percentage is much more appealing than a "ratio".

PR #4098, opened 2023-09-12
Added search button to search box

### LINKED ISSUE #529: [suggestion] search button
it would be great to have a button that starts the search next to the search box. in some situations a pointing device is the easier choice and a button would allow this way of usage. maybe put it inside the text field like it is done with the firefox search box

viciousvex:
Closes #529

Added a search button, because some mobile or TV browsers like Amazon Silk do not submit the search query after entering a value into the search box.

### DIFF
--- a/src/invidious/views/components/search_box.ecr
+++ b/src/invidious/views/components/search_box.ecr
@@ -5,5 +5,6 @@
 		name="q" placeholder="<%= translate(locale, "search") %>"
 		title="<%= translate(locale, "search") %>"
 		value="<%= env.get?("search").try {|x| HTML.escape(x.as(String)) } %>">
+		<input type="submit" value="<%= translate(locale, "search") %>">
 	</fieldset>
 </form>


src/invidious/views/components/search_box.ecr
```diff
@@ -5,5 +5,6 @@
 		name="q" placeholder="<%= translate(locale, "search") %>"
 		title="<%= translate(locale, "search") %>"
 		value="<%= env.get?("search").try {|x| HTML.escape(x.as(String)) } %>">
+		<input type="submit" value="<%= translate(locale, "search") %>">
```
syeopite, line 8:
```suggestion
		<button class="search-button" id="right-overlap-search-button" type="submit"><i class="icon ion-md-search"></i></button>
```

syeopite (changes requested):

src/invidious/views/components/search_box.ecr
ChunkyProgrammer: I like the suggestion but an `aria-label` would need to be added to that button to make it accessible for visually impaired users

`... type="submit" aria-label="<%= translate(locale, "search") %>"><i ...`

SamantazFox:
@viciousvex Can you please implement the changes proposed by the reviewers?

PR #4099, opened 2023-09-13
Routing: Add support for new routes

### LINKED ISSUE #3607: [Feature request] Support /live/ route
/live/ links, which should redirect to the /watch link of the VOD, instead redirect to /feed/popular

Example: https://vid.puffyan.us/live/gbLKoy6qvEk

### LINKED ISSUE #4091: [Enhancement] having /featured at the end of url redirects to instance homepage
many of youtube redirections that work with piped, don't work with invidious.

like this https://www.youtube.com/channel/UCaB9TDPxm_dpNx6sDpZTuzg/featured

because it have /features at the end, it goes to homepage of instance and not even the channel page.

can invidious make it so that for parts of youtube that it does not support, it will redirect to the channel or video page instead of instance homepage?

that way user at least doesnt lose the page that had that feature.
I open many youtube pages and I use redirect addon in firefox to redirect to invidious.
but later after I check them I see some of them redirected to homepage of instance and so I even lose the info which was the url that it redirected from.

SamantazFox:
Closes #3607
Closes #4091

### DIFF
--- a/src/invidious/routes/channels.cr
+++ b/src/invidious/routes/channels.cr
@@ -217,6 +217,11 @@ module Invidious::Routes::Channels
     env.redirect "/channel/#{ucid}"
   end
 
+  private KNOWN_TABS = {
+    "home", "videos", "shorts", "streams", "podcasts",
+    "releases", "playlists", "community", "channels", "about",
+  }
+
   # Redirects brand url channels to a normal /channel/:ucid route
   def self.brand_redirect(env)
     locale = env.get("preferences").as(Preferences).locale
@@ -227,7 +232,10 @@ module Invidious::Routes::Channels
     yt_url_params = URI::Params.encode(env.params.query.to_h.select(["a", "u", "user"]))
 
     # Retrieves URL params that only Invidious uses
-    invidious_url_params = URI::Params.encode(env.params.query.to_h.select!(["a", "u", "user"]))
+    invidious_url_params = env.params.query.dup
+    invidious_url_params.delete_all("a")
+    invidious_url_params.delete_all("u")
+    invidious_url_params.delete_all("user")
 
     begin
       resolved_url = YoutubeAPI.resolve_url("https://youtube.com#{env.request.path}#{yt_url_params.size > 0 ? "?#{yt_url_params}" : ""}")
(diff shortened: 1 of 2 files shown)


syeopite (approved):

SamantazFox:
I might want to do a redirect for all unknown channel routes...

unixfox:
> I might want to do a redirect for all unknown channel routes...

That's what I have been thinking we should do.

src/invidious/routing.cr
```diff
@@ -131,7 +131,7 @@ module Invidious::Routing
 
     # Channel catch-all, to redirect future routes to the channel's home
     # NOTE: defined last in order to be processed after the other routes
-    get "/channel/:ucid/*", Routes::Channels, :home
+    get "/channel/:ucid/*", Routes::Channels, :redirect_home
```
syeopite, line 134: What's the point of using a new handler here?

src/invidious/routes/channels.cr
```diff
@@ -1,6 +1,12 @@
 {% skip_file if flag?(:api_only) %}
 
 module Invidious::Routes::Channels
+  # Redirection for unsupported routes ("tabs")
+  def self.redirect_home(env)
+    ucid = env.params.url["ucid"]
+    return env.redirect "/channel/#{URI.encode_www_form(ucid)}"
```
syeopite, line 7:
```suggestion
    return env.redirect "/channel/#{URI.encode_path_segment(ucid)}"
```

syeopite (approved):

src/invidious/routing.cr
SamantazFox: To actually redirect to `/channel/UCxxx`.
The `home` handler displays the video page, but never redirects.
So if you go to `/channel/UCxxx/invalid` the URL won't change and the videos are still displayed, which could confuse users.

src/invidious/routes/channels.cr
SamantazFox: This is on purpose, to prevent path transversal attempts. All segments in `env.params.url` are `URI.decode`'d, so if we use that value directly or with `encode_path_segment` (which doesn't encode `/`), a malicious URI like `/channel/UCxxxx%2F..%2F..%2Fsomething` would redirect to `/channel/UCxxxx/../../something`.

In theory, there is no vulnerable path (all critical routes like `/delete_account` will show a confirmation prompt on a GET request), but I prefer to be on the safe side.

SamantazFox:
@unixfox good to merge?

unixfox:
I guess so we haven't received any bad feedback about that.

unixfox (approved):

PR #4100, opened 2023-09-14
CSS: Improve links contrast on dark theme

### LINKED ISSUE #672: Footer font color is hard to read when using Dark Mode

RadoslavL:
This closes #672.

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -449,12 +449,10 @@ footer {
 
 .light-theme footer a {
   color: #7c7c7c !important;
-/*text-decoration: underline;*/
 }
 
 .dark-theme footer a {
   color: #adadad !important;
-/*text-decoration: underline;*/
 }
 
 footer span {
@@ -566,7 +564,6 @@ span > select {
 
   .no-theme footer a {
     color: #7c7c7c !important;
-/*  text-decoration: underline;*/
   }
 
   .light-theme .pure-menu-heading {
@@ -694,7 +691,6 @@ body.dark-theme {
 
   .no-theme footer a {
     color: #adadad !important;
-  /*text-decoration: underline;*/
   }
 }
 
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -432,20 +432,19 @@ p.video-data   { margin: 0; font-weight: bold; font-size: 80%; }
  * Footer
  */
 
-.light-theme footer {
-  color: #7c7c7c;
+footer {
   margin-top: auto;
   padding: 1.5em 0;
   text-align: center;
   max-height: 30vh;
 }
 
+.light-theme footer {
+  color: #7c7c7c;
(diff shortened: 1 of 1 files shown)


syeopite (approved):
Looks good to me. But why did you make a new PR?

RadoslavL:
> Looks good to me. But why did you make a new PR?

In my first PR I didn't create a seperate branch, so I couldn't delete it after the PR is merged.

SamantazFox:
It needs testing, as it's impacting all links, not only titles.

Also, you'll need to change the value here, too (the one you changed is for the forced dark theme, and the one I've linked below is for the automatic dark theme, based on browser/system settings):
https://github.com/iv-org/invidious/blob/beec62cf0e45fe5620f9381050080c685f32070e/assets/css/default.css#L637-L640

RadoslavL:
I noticed that the footer was unaffected by this change, as it is set by `footer a` instead of `.dark-theme a`, so changing it would affect both dark and light theme. I think I can work around that.

Before:
screenshot_before

After:
screenshot_after

RadoslavL:
I completely misunderstood #672. I will still try to solve it.

RadoslavL:
I changed the footer color in dark theme from `#919191` to `#adadad`, and in light theme from `#919191` to `#7c7c7c`.

Light theme:

Before:
screenshot2_before

After:
screenshot2_after

Dark theme:

Before:
screenshot1_before

After:
screenshot1_after

assets/css/default.css
```diff
@@ -432,17 +432,30 @@ p.video-data   { margin: 0; font-weight: bold; font-size: 80%; }
...
  */
 
-footer {
-  color: #919191;
+.light-theme footer {
+  color: #7c7c7c;
   margin-top: auto;
   padding: 1.5em 0;
   text-align: center;
   max-height: 30vh;
```
syeopite, line 440: There's no need for you to duplicate these styles so many times across the file. You can just place them under a footer element selector once.

assets/css/default.css
```diff
@@ -433,16 +433,28 @@ p.video-data   { margin: 0; font-weight: bold; font-size: 80%; }
...
+  color: #7c7c7c;
+}
+
+.dark-theme footer {
+  color: #adadad;
+}
+
+.light-theme footer a {
+  color: #7c7c7c !important;
+/*text-decoration: underline;*/
```
SamantazFox, line 452: Can you remove the commented lines please?

SamantazFox:

PR #4060, opened 2023-08-18
Add feature of iframe api like youtube

bonjinnorenka:
I add feature of iframe embed control like youtube iframe api(https://developers.google.com/youtube/iframe_api_reference)
This feature fullfil part of #3705.
This feature can't execute on IE11.
But i didn't add all feature of youtube iframe api.
I didn't unimplemented these functions and events.
functiom
all deprecated function
player.cuePlaylist()
player.loadPlaylist()
player.getSphericalProperties()
player.setSphericalProperties(properties:Object)
player.nextVideo()
player.previousVideo()
player.playVideoAt(index:Number)
player.setLoop(loopPlaylists:Boolean)
player.setShuffle(shufflePlaylist:Boolean)
player.getVideoLoadedFraction()
player.getPlaylist()
player.getPlaylistIndex()

event
onPlaybackQualityChange
onError this event not return exactory reason like youtube iframe api
onApiChange

How to use.
use invidious instead of YT in javascript.

PR #4101, opened 2023-09-14
Frontend: Add video timestamp on external links

### LINKED ISSUE #3089: [Enhancement] Add timestamp when switching instances

I'm always frustrated when I switch an instance and have to find out what time I was at.

Add a query parameter to the url (like YouTube, e.g: `https://youtu.be/XamwuOdG3OA?t=3`)

**Additional context**

### LINKED ISSUE #3348: [Enhancement] YouTube Link in Invidious Embed (with Time Stamps)
When I'm watching a video on Invidious and want to share it with people who don't use Invidious, I typically just copy the "watch on Youtube" link and share that. When watching embeded videos on other sites however, this option is absent unless I open the video in Invidious proper. Likewise, I can't copy a time-stamped YouTube link from either Invidious or an Embed, forcing me to open in YouTube just to get a proper timestamp (or copy the Invidious Timestamp and then re-write that as a YouTube Link).

Just as embeded Invidious Videos features a link to open/share in Invidious, so too should there be links to open/share in YouTube, including with time-stamps
At present, the only solution for sharing YouTube links is either rewriting the copy/pasted Invidious Link, or opening in Invidious to copy/paste the Watch In YouTube Link. It's even worse to get timestamps, as besides re-writing links my only option is to open the video in YouTube itself.

**Additional context**
Maybe if Invidious were to improve their embeded videos on services like Discord or Whatsapp I'd feel more comfortable sharing Invidious links, but even then I'd have to explain what Invidious is to those not in the know. Furthermore, I'm confident that people like me in the know use extensions like Privacy Redirect anyway, so a raw YouTube link is no threat for us.

SamantazFox:
Closes #3089
Improves the situation for #3348

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -112,6 +112,21 @@ function addCurrentTimeToURL(url) {
     return urlUsed;
 }
 
+/**
+ * Timer that updates the timestamp on "watch on youtube" and "embed" links
+ */
+player.ready(function () {
+    let elem_watch = document.getElementById('link-yt-watch');
+    let elem_embed = document.getElementById('link-yt-embed');
+
+    let base_url_watch = elem_watch.getAttribute('data-base-url');
+    let base_url_embed = elem_embed.getAttribute('data-base-url');
+
+    setTimeout(() => { elem_watch.setAttribute('href') = addCurrentTimeToURL(base_url_watch); }, 5000);
+    setTimeout(() => { elem_embed.setAttribute('href') = addCurrentTimeToURL(base_url_embed); }, 5000);
+});
+
+
 var shareOptions = {
     socials: ['fbFeed', 'tw', 'reddit', 'email'],
 
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -116,14 +116,17 @@ function addCurrentTimeToURL(url) {
  * Timer that updates the timestamp on "watch on youtube" and "embed" links
  */
 player.ready(function () {
-    let elem_watch = document.getElementById('link-yt-watch');
-    let elem_embed = document.getElementById('link-yt-embed');
+    let elem_yt_watch = document.getElementById('link-yt-watch');
+    let elem_yt_embed = document.getElementById('link-yt-embed');
+    let elem_iv_embed = document.getElementById('link-iv-embed');
 
-    let base_url_watch = elem_watch.getAttribute('data-base-url');
-    let base_url_embed = elem_embed.getAttribute('data-base-url');
+    let base_url_yt_watch = elem_yt_watch.getAttribute('data-base-url');
+    let base_url_yt_embed = elem_yt_embed.getAttribute('data-base-url');
+    let base_url_iv_embed = elem_iv_embed.getAttribute('data-base-url');
 
-    setTimeout(() => { elem_watch.setAttribute('href') = addCurrentTimeToURL(base_url_watch); }, 5000);
-    setTimeout(() => { elem_embed.setAttribute('href') = addCurrentTimeToURL(base_url_embed); }, 5000);
+    setTimeout(() => { elem_yt_watch.setAttribute('href') = addCurrentTimeToURL(base_url_yt_watch); }, 5000);
+    setTimeout(() => { elem_yt_embed.setAttribute('href') = addCurrentTimeToURL(base_url_yt_embed); }, 5000);
+    setTimeout(() => { elem_iv_embed.setAttribute('href') = addCurrentTimeToURL(base_url_iv_embed); }, 5000);
 });
 
 
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -113,9 +113,12 @@ function addCurrentTimeToURL(url) {
 }
 
 /**
- * Timer that updates the timestamp on all external links
+ * Callback that updates the timestamp on all external links
  */
-player.ready(function () {
+player.on('timeupdate', function () {
+    // Only update once every 5 seconds
+    if ((Math.ceil(player.currentTime()) % 5) != 0) return;
+
     // YouTube links
 
     let elem_yt_watch = document.getElementById('link-yt-watch');
@@ -124,10 +127,8 @@ player.ready(function () {
     let base_url_yt_watch = elem_yt_watch.getAttribute('data-base-url');
     let base_url_yt_embed = elem_yt_embed.getAttribute('data-base-url');
 
-    setTimeout(() => {
-        elem_yt_watch.setAttribute('href') = addCurrentTimeToURL(base_url_yt_watch);
-        elem_yt_embed.setAttribute('href') = addCurrentTimeToURL(base_url_yt_embed);
-    }, 5000);
+    elem_yt_watch.setAttribute('href') = addCurrentTimeToURL(base_url_yt_watch);
+    elem_yt_embed.setAttribute('href') = addCurrentTimeToURL(base_url_yt_embed);
 
     // Invidious links
 
@@ -137,10 +138,8 @@ player.ready(function () {
     let base_url_iv_embed = elem_iv_embed.getAttribute('data-base-url');
     let base_url_iv_other = elem_iv_other.getAttribute('data-base-url');
 
-    setTimeout(() => {
-        elem_iv_embed.setAttribute('href') = addCurrentTimeToURL(base_url_iv_embed);
-        elem_iv_other.setAttribute('href') = addCurrentTimeToURL(base_url_iv_other);
-    }, 5000);
+    elem_iv_embed.setAttribute('href') = addCurrentTimeToURL(base_url_iv_embed);
+    elem_iv_other.setAttribute('href') = addCurrentTimeToURL(base_url_iv_other);
 });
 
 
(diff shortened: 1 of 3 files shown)


assets/js/player.js
```diff
@@ -112,6 +112,38 @@ function addCurrentTimeToURL(url) {
     return urlUsed;
 }
 
+/**
+ * Timer that updates the timestamp on all external links
```
syeopite, line 116: VideoJS has a `timeupdate` event that fires each time the playback position changes. It's probably a better idea to listen for that instead of using a timer.

assets/js/player.js
```diff
@@ -112,6 +112,38 @@ function addCurrentTimeToURL(url) {
...
+player.ready(function () {
+    // YouTube links
+
+    let elem_yt_watch = document.getElementById('link-yt-watch');
+    let elem_yt_embed = document.getElementById('link-yt-embed');
+
+    let base_url_yt_watch = elem_yt_watch.getAttribute('data-base-url');
+    let base_url_yt_embed = elem_yt_embed.getAttribute('data-base-url');
+
+    setTimeout(() => {
```
syeopite, line 127: I think you meant to use `setInterval` here. `setTimeout` just delays the execution of the code by the given delay. 

```suggestion
    setInterval(() => {
```

syeopite (changes requested):

assets/js/player.js
```diff
@@ -112,6 +112,37 @@ function addCurrentTimeToURL(url) {
     return urlUsed;
 }
 
+/**
+ * Callback that updates the timestamp on all external links
+ */
+player.on('timeupdate', function () {
+    // Only update once every 5 seconds
+    if ((Math.ceil(player.currentTime()) % 5) != 0) return;
```
syeopite, line 120: If I'm not mistaken here, the links will pretty much never get updated. Since `timeupdate` fires every 15 - 250ms and the `player.currentTime()` returns the current time in seconds, the chances of reaching a perfect multiple of 5 would be quite low. 

--- 

On a semirelated note,

Why should the link only get updated every 5 seconds? The performance impact of increasing it to every second or even to each occurrence of the `timeupdate` event should be negligible.

assets/js/player.js
```diff
@@ -112,6 +112,37 @@ function addCurrentTimeToURL(url) {
...
+    // YouTube links
+
+    let elem_yt_watch = document.getElementById('link-yt-watch');
+    let elem_yt_embed = document.getElementById('link-yt-embed');
+
+    let base_url_yt_watch = elem_yt_watch.getAttribute('data-base-url');
+    let base_url_yt_embed = elem_yt_embed.getAttribute('data-base-url');
+
+    elem_yt_watch.setAttribute('href') = addCurrentTimeToURL(base_url_yt_watch);
+    elem_yt_embed.setAttribute('href') = addCurrentTimeToURL(base_url_yt_embed);
```
syeopite, line 131:
```suggestion
    elem_yt_watch.href = addCurrentTimeToURL(base_url_yt_watch);
    elem_yt_embed.href = addCurrentTimeToURL(base_url_yt_embed);
```

assets/js/player.js
```diff
@@ -112,6 +112,37 @@ function addCurrentTimeToURL(url) {
...
+    // Invidious links
+
+    let elem_iv_embed = document.getElementById('link-iv-embed');
+    let elem_iv_other = document.getElementById('link-iv-other');
+
+    let base_url_iv_embed = elem_iv_embed.getAttribute('data-base-url');
+    let base_url_iv_other = elem_iv_other.getAttribute('data-base-url');
+
+    elem_iv_embed.setAttribute('href') = addCurrentTimeToURL(base_url_iv_embed);
+    elem_iv_other.setAttribute('href') = addCurrentTimeToURL(base_url_iv_other);
```
syeopite, line 142:
```suggestion
    elem_iv_embed.href = addCurrentTimeToURL(base_url_iv_embed);
    elem_iv_other.href = addCurrentTimeToURL(base_url_iv_other);
```

syeopite (changes requested):

assets/js/player.js
SamantazFox: I'm not sure how often that event is fired. I couldn't find the info in VideoJS' docs nor code. As for the 5s delay, I feel like every second is too often. Plus, by going a few seconds back, It's easier to catch up, imo.

But if you think it's not useful, I can remove that.

syeopite: > I'm not sure how often that event is fired. I couldn't find the info in VideoJS' docs nor code. 

It fires every 15 - 250ms as seen here https://docs.videojs.com/docs/api/player.html#Eventstimeupdate

> But if you think it's not useful, I can remove that.

I mostly just want to copy YouTube's behavior which by default sets the timestamp as the current second in their share widget. You can decide if you'll like to keep it at updating every 5 seconds.

assets/js/player.js
```diff
@@ -112,6 +112,37 @@ function addCurrentTimeToURL(url) {
...
+ * Callback that updates the timestamp on all external links
+ */
+player.on('timeupdate', function () {
+    // Only update once every 5 seconds
+    if ((Math.ceil(player.currentTime()) % 5) != 0) return;
+
+    // YouTube links
+
+    let elem_yt_watch = document.getElementById('link-yt-watch');
+    let elem_yt_embed = document.getElementById('link-yt-embed');
```
syeopite, line 125: Actually now that I think about it does it even make sense to include the timestamp in the embed link? YouTube doesn't support the `t=` url parameter and instead wants you to use their iframe API. 

Invidious currently does support the `t=` parameter on embeded videos but once #4060 gets merged maybe we should have the same behavior?

assets/js/player.js
SamantazFox: I'm not sure how to cleanly update only once a second.
That'd be better if we don't update all the links every 15ms x)

assets/js/player.js
SamantazFox: As stated in #3348, it's hard to manually add the timestamp to an URL. But if the timestamp is present, users can easily remove it.

As for iframes, I'm not very found of forcing the use of an API. Sure, this can be another way of doing it, but why not supporting both?

syeopite: > ... #3348

I believe that issue is more about adding a separate "Watch on YouTube" link with a timestamp on it to the embedded player.
 
> but why not supporting both?

Fair enough. I was mostly thinking of a weird edge case where the `t=` parameter could impact compatibility between the YouTube and Invidious embed players but that's probably something we don't have to worry about.

assets/js/player.js
syeopite: Ah, well lets keep it at 5 seconds for now then. A future PR can resolve this.

src/invidious/views/watch.ecr
```diff
@@ -112,19 +112,36 @@ we're going to need to do it here in order to allow for translations.
...
+                <a id="link-iv-other" data-base-url="<%= link_iv_other %>" href="<%= link_iv_other %>"><%= translate(locale, "Switch Invidious Instance") %></a>
             </p>
+
             <p id="embed-link">
-                <a href="<%= embed_link %>"><%= translate(locale, "videoinfo_invidious_embed_link") %></a>
+                <%-
+                    params_iv_embed = env.params.query.dup
+                    params_iv_embed.delete_all("v")
+
+                    link_iv_embed = URI.new(path: "/embed/#{id}")
```
syeopite, line 139: It seems like Javascript can't parse `/embed/#{id}` as valid a URL. It needs a domain for JS to parse it correctly at this line in the `addCurrentTimeToURL` function:

https://github.com/iv-org/invidious/blob/bb14f794969f62582917a39c2dd57bf92fa146a7/assets/js/player.js#L105

syeopite (changes requested):

src/invidious/views/watch.ecr
absidue: Adding `window.location.origin` as the second parameter for the URL constructor, fixes the issue, while also creating a valid URL.

SamantazFox: Fixed, thanks for your feedback :)

assets/js/player.js
SamantazFox: > I believe that issue is more about adding a separate "Watch on YouTube" link with a timestamp on it to the embedded player.

Oooh, you're right! I misunderstood the issue!
We probably need a much better menu for videoJS though :/

absidue (approved):
The JavaScript code looks good, I don't know crystal so I haven't reviewed that and I also don't have an Invidious development environment set up, so I haven't tested it either.

SamantazFox:
@syeopite In the end, I've done it properly. It now updates exactly once per second.
The previous code was bogus anyway, as it would update every 15ms for 1s, then don't update for 4s, and so on.

SamantazFox:
@absidue it's deployed on the test instance if you want.

absidue (approved):
Seems good to me on the test instance.

syeopite:
The timestamp parameter is ignored when automatic instance redirection is enabled.  I guess a quick and dirty fix would be to just automatically include any additional URL parameters in the `/redirect` endpoint.

Issue #2395, opened 2021-09-12
[Meta] Overhaul Invidious player for more YT features.


syeopite:
This is a meta-issue to track everything that needs to be done in order to fix the discrepancy between Invidious and YouTube's players

- [ ] End-cards
- [ ] Auto-generated captions (Follow up to: https://github.com/iv-org/invidious/issues/2187#issuecomment-866014531)
- [ ] Audio tracks (#2007)
- [ ] Channel Favicons 
- [ ] Custom playback speeds (#533)
- [ ]  Miniplayer (#583, #631)
- [ ] 3D 
- [ ] Frame-by-frame skipping (#2113),
- [ ] Fine scrubbing (mobile-only feature on YT)
- [ ] Chapters
- [ ] Theater mode? 
- [ ]  Related videos in full screen? (useful for embeds and binge watching.)

AlexMiniReal:
Is someone working on the chapters enhancement?

AlexMiniReal:
![Screenshot_7](https://user-images.githubusercontent.com/76444133/143689087-25160c89-5369-4a9a-afd7-0e3a52c8c01b.jpg)

unixfox:
> Is someone working on the chapters enhancement?

not that we know of

debpalash:
I was waiting for the UI overhaul for so long until made a stylus css myself for player pages now it has full viewport width player and home. [repo here](https://github.com/debpalash/InvidiousFullScreen-Stylus)

IronMaltese:
Is this able to be implemented in mainline invidious?

debpalash:
i guess one could fork and inject my css 🤔

arjun44:
@debpalash FYI: I forked your stylus css; I removed and simplified some things. It can be easily extended further. You can see the source [here](https://gist.github.com/arjun44/4700b1a629d0885a6dc564a3d67e6cca), it's very short.

debpalash:
@arjun44 nice 😂 . btw you need this to work on stylus `@-moz-document url-prefix("https://instance.tld")`

selurvedu:
I'm not seeing YouTube Cards support in the list (example [here](https://i.ytimg.com/vi/zQg3f-sVxiU/maxresdefault.jpg)). Is it planned?

syeopite:
Added. Thank you

PR #4109, opened 2023-09-19
Add chapters support to Invidious

syeopite:
This PR adds a chapters menu to the player control bar and a separate widget within the description to select chapters along with their thumbnail similar to what's on YouTube. See screenshots below. 

Unfortunately, we can't segment the progress bar into multiple chunks like what YouTube does yet. VideoJS doesn't support this natively and it doesn't look like there are any extensions that does this.

Related: https://github.com/videojs/video.js/discussions/8235

chapters
chapters widget

Closes chapters component of #2395


syeopite:
VideoJS renders all text as lowercase by default (see captions menu, or the new chapters button in this PR). We may want to override this with something like:

https://github.com/videojs/video.js/issues/3015#issuecomment-215895729
```css
.vjs-menu li {
  text-transform: initial;
}
```

syeopite:
Apologies everyone. I completely messed up on the rebase 😅

PR #4111, opened 2023-09-19
Add chapters support to Invidious

syeopite:
This PR adds a chapters menu to the player control bar and a separate widget within the description to select chapters along with their thumbnail similar to what's on YouTube. See screenshots below. 

Unfortunately, we can't segment the progress bar into multiple chunks like what YouTube does yet. VideoJS doesn't support this natively and it doesn't look like there are any extensions that does this.

Related: https://github.com/videojs/video.js/discussions/8235

chapters
chapters widget

Closes chapters component of #2395


absidue:
Is there a reason you decided to create a completely new JSON API endpoint instead of just including it in the existing `/api/v1/videos/{video id}`? Seems like it will add extra unnecessary load on Invidious, as apps like FreeTube will be making the request for every video, but it'll frequently be pointless, as most videos don't have any chapters.
Integrating it into the existing API response will make integration a lot easier.

syeopite:
I'm not actually sure why I did that... good catch! Data for chapters should now be available in the standard `/api/v1/videos/{video id}` endpoint

json

absidue:
Thanks!

absidue:
Thanks!

src/invidious/videos/chapters.cr
```diff
@@ -0,0 +1,125 @@
...
+        thumbnails: thumbnails,
+      )
+    end
+
+    return segments
+  end
+
+  # Converts an array of Chapter objects to a webvtt file
+  def self.chapters_to_vtt(chapters : Array(Chapter))
+    vtt = String.build do |vtt|
```
SamantazFox, line 52: Now that is merged, you can use your own `WebVTT::Builder` :P

src/invidious/views/components/description_chapters_widget.ecr
```diff
@@ -0,0 +1,34 @@
...
+        <h4><%=HTML.escape(translate(locale, "video_chapters_label"))%></h4>
+
+        <% if video.automatically_generated_chapters? %> 
+          <h5><%=HTML.escape(translate(locale, "video_chapters_auto_generated_label"))%> </h5>
+        <% end %>
+
+        <div class="description-chapters-content-container">
+            <% chapters.each do | chapter | %>
+                <%- start_in_seconds = chapter.start_ms.total_seconds.to_i %>
+                <a href="/watch?v=<%-= video.id %>&t=<%=start_in_seconds %>">
```
SamantazFox, line 13: The attributes `data-onclick="jump_to_time" data-jump-time="<seconds>"` should go on this `<a>` element. They don't need to go anywhere else!

SamantazFox: Also, we can pass a floating point number, with full ms precision, as the [function that decodes the `t` parameter](https://github.com/iv-org/invidious/blob/b16f66ef0003843c4561f7bb3124339e6b446b2b/src/invidious/helpers/utils.cr#L73) supports milliseconds precision (And on the JS side, [`player.currentTime()` does too](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime)).

src/invidious/views/components/description_chapters_widget.ecr
```diff
@@ -0,0 +1,34 @@
...
+                                <img loading="lazy" class="thumbnail" src="<%-=URI.parse(chapter.thumbnails[-1]["url"].to_s).request_target %>" alt="" data-onclick="jump_to_time" data-jump-time="<%=start_in_seconds%>"/>
+                            <%- else -%>
+                                <div class="thumbnail-placeholder" data-onclick="jump_to_time" data-jump-time="<%=start_in_seconds%>"></div>
+                            <%- end -%>
+                        </div>
+                        <%- if start_in_seconds > 0 -%>
+                            <p data-onclick="jump_to_time" data-jump-time="<%=start_in_seconds%>"><%-= recode_length_seconds(start_in_seconds) -%></p>
+                        <%- else -%>
+                            <p data-onclick="jump_to_time" data-jump-time="<%=start_in_seconds%>">0:00</p>
+                        <%- end -%>
```
SamantazFox, line 26: Can you instead change `recode_length_seconds()` to return `"0:00"` is the provided time is `0`? It also makes sense in all the other places that use that function.

src/invidious/routes/api/v1/videos.cr
```diff
@@ -363,4 +363,41 @@ module Invidious::Routes::API::V1::Videos
       end
     end
   end
+
+  def self.chapters(env)
+    id = env.params.url["id"]
+    region = env.params.query["region"]? || env.params.body["region"]?
```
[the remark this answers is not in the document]
syeopite: It isn't. A dedicated API endpoint is needed to fetch the WebVTT document that represents chapters, which is what VideoJS uses to populate the chapters button on the control bar.

src/invidious/views/components/description_chapters_widget.ecr
syeopite: I'm a bit hesitant to change that since Invidious also uses a 0 to represent a nil value when length information isn't available

assets/css/player.css
```diff
@@ -114,6 +114,10 @@ ul.vjs-menu-content::-webkit-scrollbar {
   order: 5;
 }
 
+.vjs-chapters-button {
+  order: 5;
```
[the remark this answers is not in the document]
syeopite: It shouldn't cause any problems 

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items#the_order_property

> ... If more than one item has the same integer value, then within that group the items are laid out as per source order.

src/invidious/routes/api/v1/videos.cr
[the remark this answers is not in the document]
SamantazFox: Oh, okay!

VnkCCENcXH0c3AhnijQ0y7SBeZx27:
any ETA on chapters within invidious environment?

shaedrich:
@VnkCCENcXH0c3AhnijQ0y7SBeZx27 They currently have other problems, like trying to keep Invidious alive, so I assume, this isn't very high up their agenda, understandably.

PR #4122, opened 2023-09-26
Add configuration option to control preloading of video data

### LINKED ISSUE #4110: [Feature request] Add config option like "Don't autoload video data"
Hello.

I would also like to see a configuration option like ["Don't autoload video data"](https://github.com/iv-org/invidious/issues/798) so that video data only starts loading when the user explicitly hits the "Play" button.

This would be useful for people who are using a cellular connection for their home network and do not want to waste any mobile data volume when some tabs in Firefox with Invidious URLs are re-loading on browser restart, for example.

It is possible [to configure Firefox ](https://superuser.com/a/1344698)so that only a small portion of the video will be buffered. However, this still consumes a few Megabytes on page load when the first part of the video stream is buffered in the background.

Thank you!
– Thomas

Nerdmind:
This PR should fix #4110 by adding a configuration option to control the preloading of video data on page load by making use of the HTML5 [`preload`](https://www.w3schools.com/tags/att_video_preload.asp) attribute of the `<video>` element.

The option is enabled by default, so the `preload` attribute will have the default value `auto`. If users wanting to prevent preloading of video data, they can change the option to `false` so that the attribute will have the value `none`.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -3,7 +3,6 @@ var player_data = JSON.parse(document.getElementById('player_data').textContent)
 var video_data = JSON.parse(document.getElementById('video_data').textContent);
 
 var options = {
-    preload: 'auto',
     liveui: true,
     playbackRates: [0.25, 0.5, 0.75, 1.0, 1.25, 1.5, 1.75, 2.0],
     controlBar: {
@@ -33,6 +32,10 @@ var options = {
 
 if (player_data.aspect_ratio) {
     options.aspectRatio = player_data.aspect_ratio;
+}
+
+if (player_data.preload) {
+    options.preload = player_data.preload
 }
 
 var embed_url = new URL(location);
--- a/src/invidious/views/components/player.ecr
+++ b/src/invidious/views/components/player.ecr
@@ -1,5 +1,6 @@
 <video style="outline:none;width:100%;background-color:#000" playsinline poster="<%= thumbnail %>"
     id="player" class="on-video_player video-js player-style-<%= params.player_style %>"
+    preload="<% if params.preload %>auto<% else %>none<% end %>"
     <% if params.autoplay %>autoplay<% end %>
     <% if params.video_loop %>loop<% end %>
     <% if params.controls %>controls<% end %>>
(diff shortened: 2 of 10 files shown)


src/invidious/views/components/player.ecr
```diff
@@ -73,7 +74,8 @@
     "title" => video.title,
     "description" => HTML.escape(video.short_description),
     "thumbnail" => thumbnail,
-    "preferred_caption_found" => !preferred_captions.empty?
+    "preferred_caption_found" => !preferred_captions.empty?,
+    "preload" => params.preload ? "auto" : "none"
```
syeopite, line 78:
```suggestion
    "preload" => params.preload

File: assets/js/player.js [hunk 2]
```diff
@@ -35,6 +34,10 @@ if (player_data.aspect_ratio) {
     options.aspectRatio = player_data.aspect_ratio;
 }
 
+if (player_data.preload) {
+    options.preload = player_data.preload
+}
+
```
syeopite, line 40:
```suggestion
if (player_data.preload) {
    options.preload = "auto"
} else {
    options.preload = "none"
};

```

syeopite (commented):
Is there any particular reason why you're having the preload parameter directly set the videojs option?

Nerdmind:
Hi. No, not a specific reason. But your question got me thinking about something else:

I've set the `preload` option in VideoJS explicitly because `preload: "auto"` is currently defined in the `player.js` file in the `master` branch and I didn't consider that this option could be *removed completely* from the`player.js` when the HTML `<video>` element already contains the `preload` attribute directly so that VideoJS can inherit it.

I just tested it locally when removing setting the `preload` option in `player.js` completely and it works as expected. I will add a new commit to this PR in a few minutes. This seems to be a cleaner solution with less redundancy.

syeopite (approved):

Nerdmind:
I just want to inform you that I'm actively using this feature on my private instance since about half a year now. It works flawlessly and never made a problem. Any chance this PR will get reviewed/merged soon?

syeopite:
@SamantazFox Any chance this can be merged in the next batch?

unixfox (approved):

SamantazFox (approved):

SamantazFox:

Issue #660, opened 2019-07-31
No "Previous Page" on the playlists list


TheFrenchGhosty:
There is no way to go back to the previous page of the playlists list.
![bug](https://user-images.githubusercontent.com/47571719/62250663-04bab500-b3dd-11e9-8657-d48bf6f1d3c5.png)
(this is the third and last page of the playlists of Linux Tech Tips)

omarroth:
This is a limitation of how playlists are provided by YouTube. It's easier to see when looking at the [respective API endpoint](https://invidio.us/api/v1/channels/playlists/UCXuqSBlHAE6Xw-yeJA0Tunw?pretty=1). 

Essentially, YouTube provides a random token on each request that can be used to get the next page of playlists. AFAIK, YouTube doesn't provide the previously used token, so it's not possible to go back without using controls provided by the browser.

leonklingele:
That's because YouTube only supports infinite-scrolling without a offering "Next page" / "Previous page" link. 
Why not simply add a new URL parameter then, e.g. `previous=OLD_TOKEN`?

TheFrenchGhosty:
> That's because YouTube only supports infinite-scrolling without a offering "Next page" / "Previous page" link.
> Why not simply add a new URL parameter then, e.g. `previous=OLD_TOKEN`?

If it's not possible just having a "first page" might also be better than having to go back in the channel and clicking playlists again

omarroth:
> Why not simply add a new URL parameter then, e.g. `previous=OLD_TOKEN`?

Unfortunately this won't solve the problem, since the previous page won't be able to get the `previous` for itself, if that makes sense. It'd be necessary to have a `previous=OLD_TOKEN1,OLD_TOKEN2,...`.

I think having a "first page" link is probably the best option here. It may make sense to refactor this to use the same code as `/community` since the implementation is extremely similar, although it wouldn't work without Javascript.

elypter:
you could cache those tokens as cookies

TheFrenchGhosty:
Still an issue.

TheFrenchGhosty:
Still an issue

Issue #2130, opened 2021-06-05
[Enhancement] Persist &listen=1 between suggested videos (for when listening to music)


SylveonBottle:
**Is your enhancement request related to a problem? Please describe.**

When I'm watching videos using Invidious, sometimes I'll get into a music-mood and decide to look up some tunes. I'll find the video I'd like to listen to, then click the music symbol to watch the video with &listen=1.
After the song is finished, I'll let invidious autoplay the next track OR I'll find another song on the suggested videos on the right.
However, upon clicking these videos &listen=1 is lost.

**Describe the solution you'd like**

- Add &listen=1 to videos on the rightside recommendation list if the current page is also in &listen=1
- Add the `icon ion-md-headset` to videos on the rightside recommendation list
- Toggle &listen=1 until the user switches back using icon ion-ios-videocam

**Describe alternatives you've considered**

While I could simply enter my settings and change to Listen To Default (#876), I feel as though this would be a hassle constantly switching back and forth.
If video quality is persistent between videos, #1450 could be an alternative.

**Additional context**

This may be related to #932

SamantazFox:
Related: #2036, #2021 and #1610

SylveonBottle:
This appears to be implemented? 
As seen [here](https://yewtu.be/watch?v=LSZ1xygUbOE&list=PL3zIYvF1XjLmRLZEIdjD-eB8tnknbwBC2&index=0&listen=1&t=7m22s), the next video automatically plays and starts with listen=1.

SamantazFox:
> This appears to be implemented?

URL parameters are inconsistently handled. I should probably change the title.

PR #4123, opened 2023-09-27
Frontend: Add a first page and previous page buttons for channel navigation

RadoslavL:
Fixes: #660

This works in all channel content categories (videos, shorts, playlists, etc). I also added translations of the button text to a couple of locales.

Update: The code now supports both a previous page button (using Javascript) and a first page button. When the browser doesn't support Javascript, the first page button is used instead of the previous page button. All credit for the previous page button goes out to @syeopite for providing the Javascript code. (I modified it a bit to make the variable names the same as the ones used in my old code, but the structure is exactly the same.)

### DIFF
--- a/assets/js/pagination.js
+++ b/assets/js/pagination.js
@@ -32,14 +32,22 @@ function button_press(){
     // On the first page, the stored continuation token is null.
     if (prev_ctoken === null) {
         sessionStorage.removeItem(CONT_CACHE_KEY);
-        window.location.href = window.location.href.split('?')[0];
+        let url = window.location.href.split('?')[0];
+        let params = window.location.href.split('?')[1];
+        let url_params = new URLSearchParams(params);
+        url_params.delete('continuation');
+        window.location.href = `${url}?${url_params.toString()}`;
 
         return;
     }
 
     sessionStorage.setItem(CONT_CACHE_KEY, JSON.stringify(prev_data));
+    let url = window.location.href.split('?')[0];
+    let params = window.location.href.split('?')[1];
+    let url_params = new URLSearchParams(params);
+    url_params.set("continuation", prev_ctoken);
 
-    window.location.href = `${window.location.pathname}?continuation=${prev_ctoken}`;
+    window.location.href = `${url}?${url_params.toString()}`;
 };
 
 addEventListener('DOMContentLoaded', function(){
--- a/assets/js/pagination.js
+++ b/assets/js/pagination.js
@@ -43,7 +43,7 @@ function button_press(){
 };
 
 addEventListener('DOMContentLoaded', function(){
-    const pagination_data = JSON.parse(document.getElementById('pagination_locale_strings').textContent);
+    const pagination_data = JSON.parse(document.getElementById('pagination-data').textContent);
     const next_page_containers = document.getElementsByClassName("page-next-container");
 
     for (let container of next_page_containers){
--- a/assets/js/pagination.js
+++ b/assets/js/pagination.js
@@ -68,7 +68,7 @@ addEventListener('DOMContentLoaded', function(){
         const prev_page_containers = document.getElementsByClassName("page-prev-container")
 
         for (let container of prev_page_containers) {
-            if (pagination_data.is_locale_rtl) {
+            if (pagination_data.is_rtl) {
                 container.innerHTML = `<button class="pure-button pure-button-secondary">${pagination_data.prev_page}&nbsp;&nbsp;<i class="icon ion-ios-arrow-forward"></i></button>`
             } else {
                 container.innerHTML = `<button class="pure-button pure-button-secondary"><i class="icon ion-ios-arrow-back"></i>&nbsp;&nbsp;${pagination_data.prev_page}</button>`
--- a/assets/js/pagination.js
+++ b/assets/js/pagination.js
@@ -32,23 +32,37 @@ function button_press(){
     // On the first page, the stored continuation token is null.
     if (prev_ctoken === null) {
         sessionStorage.removeItem(CONT_CACHE_KEY);
-        let url = window.location.href.split('?')[0];
-        let params = window.location.href.split('?')[1];
-        let url_params = new URLSearchParams(params);
-        url_params.delete('continuation');
-        window.location.href = `${url}?${url_params.toString()}`;
+        let url = set_continuation();
+        window.location.href = url;
 
         return;
     }
 
     sessionStorage.setItem(CONT_CACHE_KEY, JSON.stringify(prev_data));
+    let url = set_continuation(prev_ctoken);
+
+    window.location.href = url;
+};
+
+// Method to set the current page's continuation token
+// Removes the continuation parameter when a continuation token is not given
+function set_continuation(prev_ctoken = null){
     let url = window.location.href.split('?')[0];
     let params = window.location.href.split('?')[1];
     let url_params = new URLSearchParams(params);
-    url_params.set("continuation", prev_ctoken);
 
-    window.location.href = `${url}?${url_params.toString()}`;
-};
+    if (prev_ctoken) {
+        url_params.set("continuation", prev_ctoken);
+    } else {
+        url_params.delete('continuation');
+    };
+
+    if(Array.from(url_params).length > 0){
+        return `${url}?${url_params.toString()}`;
+    } else {
+        return url;
+    }
+}
 
 addEventListener('DOMContentLoaded', function(){
     const pagination_data = JSON.parse(document.getElementById('pagination-data').textContent);
--- a/assets/js/pagination.js
+++ b/assets/js/pagination.js
@@ -1,0 +1,71 @@
+'use strict';
+
+const CURRENT_CONTINUATION = (new URL(document.location)).searchParams.get("continuation");
+const CONT_CACHE_KEY = `continuation_cache_${encodeURIComponent(window.location.pathname)}`;
+
+function get_data(){
+    return JSON.parse(sessionStorage.getItem(CONT_CACHE_KEY)) || [];
+}
+
+function save_data(){
+    const prev_data = get_data();
+    prev_data.push(CURRENT_CONTINUATION);
+
+    sessionStorage.setItem(CONT_CACHE_KEY, JSON.stringify(prev_data));
+}
+
+function button_press(){
+    let prev_data = get_data();
+    if (!prev_data.length) return null;
+
+    // Sanity check. Nowhere should the current continuation token exist in the cache 
+    // but it can happen when using the browser's back feature. As such we'd need to travel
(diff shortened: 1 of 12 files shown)


syeopite:
> I don't think adding a previous page button is possible

You actually can! It's just a bit more complicated than a first page button though.

As said in the issue, you'll just have to cache the continuation tokens. Though something that's not mentioned is that you'll probably also need a sanity check on the continuation tokens if you aren't caching it in the URL due to the browsers "back" feature.

I have a [branch that implements this functionality through JS and session storage here.](https://github.com/syeopite/invidious/tree/add-playlist-prev-page-btn) Feel free to take what you want in there if you'd like to implement this 😁

src/invidious/frontend/pagination.cr
```diff
@@ -72,12 +90,18 @@ module Invidious::Frontend::Pagination
     end
   end
 
-  def nav_ctoken(locale : String?, *, base_url : String | URI, ctoken : String?)
+  def nav_ctoken(locale : String?, *, base_url : String | URI, ctoken : String?, first_page : String?)
```
syeopite, line 93: Just a slight nitpick, `first_page` should be just a bool instead.

RadoslavL:
> > I don't think adding a previous page button is possible
> 
> You actually can! It's just a bit more complicated than a first page button though.
> 
> As said in the issue, you'll just have to cache the continuation tokens. Though something that's not mentioned is that you'll probably also need a sanity check on the continuation tokens if you aren't caching it in the URL due to the browsers "back" feature.
> 
> I have a [branch that implements this functionality through JS and session storage here.](https://github.com/syeopite/invidious/tree/add-playlist-prev-page-btn) Feel free to take what you want in there if you'd like to implement this 😁

Yes, I knew there was a way to do it with cookies and/or session storage. The main problem with that is that this requires JavaScript. I thought Invidious was supposed to be JS free, that's why I never considered that solution. Thank you for the reply! I will see if I can implement this using JavaScript.

src/invidious/frontend/pagination.cr
RadoslavL: Yes, this was the first test that worked, so I just kept the code untouched. I will replace it in the function parameters.

syeopite:
Invidious is supposed to work without Javascript but (at least in my opinion) pure quality of life features can be safely locked behind JS. A previous page button would certainly fall under that category.

src/invidious/frontend/pagination.cr
```diff
@@ -72,12 +90,18 @@ module Invidious::Frontend::Pagination
     end
   end
 
-  def nav_ctoken(locale : String?, *, base_url : String | URI, ctoken : String?)
+  def nav_ctoken(locale : String?, *, base_url : String | URI, ctoken : String?, first_page : Bool?)
```
syeopite, line 93:
```suggestion
  def nav_ctoken(locale : String?, *, base_url : String | URI, ctoken : String?, first_page : Bool)
```

`first_page` shouldn't ever be nil

src/invidious/frontend/pagination.cr
```diff
@@ -72,12 +90,18 @@ module Invidious::Frontend::Pagination
...
     return String.build do |str|
       str << %(<div class="h-box">\n)
       str << %(<div class="page-nav-container flexible">\n)
 
-      str << %(<div class="page-prev-container flex-left"></div>\n)
+      str << %(<div class="page-prev-container flex-left">)
+
+      if !first_page?
```
syeopite, line 100:
```suggestion
      if !first_page
```
Wrong variable name

syeopite (changes requested):

src/invidious/frontend/pagination.cr
RadoslavL: Sorry for that, I copied the code from a different local repository.

syeopite (approved):

RadoslavL:
@syeopite I created the previous page button in my fork. Do I create another PR for the new code, or wait for this one to get merged?

syeopite:
> @syeopite I created the previous page button in my fork. Do I create another PR for the new code, or wait for this one to get merged?

You can just put it into this PR.

> Update...

Exactly. If I'm not mistake caching the continuation tokens client side with Javascript is really the only way to accomplish this. Sending the data back and forth through various means will not only impact performance significantly, you'll also reach the maximum length limitations when you aren't caching it in the body directly. 

Unless you are caching it in session storage I'll recommend just keeping with the first page button.

If you are going to cache it in session storage then feel free to just integrate my implementation (you can have the bounty) or something like it into this PR and do something like the following to merge it together smoothly:

1. Make the previous page button optional through the preferences page 
2. Use a first page button when the aforementioned is turned off or when JavaScript is disabled

RadoslavL:
@syeopite I added a user preferences option, that chooses which navigation button to use.

src/invidious/frontend/pagination.cr
```diff
@@ -3,6 +3,30 @@ require "uri"
 module Invidious::Frontend::Pagination
   extend self
 
+  private def first_page(str : String::Builder, locale : String?, url : String, previous_page_button : Bool)
+    if previous_page_button
+      str << %(<noscript>)
+    end
```
SamantazFox, line 9: Why are you adding a `<noscript>` tag here? Mutating the "first-page" button to a "previous-page" button on the client side would be much simpler, and would reduce the complexity server side.

src/invidious/routes/preferences.cr
```diff
@@ -140,6 +140,10 @@ module Invidious::Routes::PreferencesRoute
     notifications_only ||= "off"
     notifications_only = notifications_only == "on"
 
+    previous_page_button = env.params.body["previous_page_button"]?.try &.as(String)
+    previous_page_button ||= "off"
+    previous_page_button = previous_page_button == "on"
+
```
SamantazFox, line 146: As mentioned in other PRs, I'm not found of adding more config options, especially when we don't have any way in the UI to explain in more details what a config option does ("Previous page button (requires Javascript)" is not very explicit).

I'd prefer if that feature was always enabled for clients with JS.

In essence, show a "first page" button by default, and if some continuations are saved, show a "previous page" button.

assets/js/pagination.js
```diff
@@ -0,0 +1,71 @@
...
+    if (conflict_at != -1) {
+        prev_data.length = conflict_at;
+    }
+
+    const prev_ctoken = prev_data.pop();
+
+    // On the first page, the stored continuation token is null.
+    if (prev_ctoken === null) {
+        sessionStorage.removeItem(CONT_CACHE_KEY);
+        window.location.href = window.location.href.split('?')[0];
```
SamantazFox, line 35: If I understand that correctly, you're nuking all query parameters from the URL.
I'm pretty sure there are ways to selectively remove the `continuation` param only.

RadoslavL: The `removeItem` function only removes `CONT_CACHE_KEY` which is the current continuation array. It is defined as `continuation_cache_${encodeURIComponent(window.location.pathname)}`.

src/invidious/frontend/pagination.cr
RadoslavL: Yes, but this code needed to have `<noscript>` on the server side, because it checked the `previous_page_button` preferences option.

assets/js/pagination.js
SamantazFox: I'm talking about the `window.location.href` part ^^

RadoslavL: There is a way to selectively remove just the `continuation` parameter, but it's not needed. It took me way too long to realize this, but when we go to the next page, we are already nuking all of the query parameters. Only the continuation parameter is saved. Let me know if you still want this changed.

SamantazFox: That behavior is actually a bug, tracked in #2130.
So that would be great if you could help make it less bad x)

RadoslavL: I fixed it (for the channel page only) in [e8c2388](https://github.com/iv-org/invidious/pull/4123/commits/e8c2388589494f1370231154ffb2030c3871291b).

assets/js/pagination.js
```diff
@@ -0,0 +1,79 @@
...
+            next_page_button[0].addEventListener("click", save_data);
+        }
+    }
+
+    // Only add previous page buttons when not on the first page
+    if (CURRENT_CONTINUATION) {
+        const prev_page_containers = document.getElementsByClassName("page-prev-container")
+
+        for (let container of prev_page_containers) {
+            if (pagination_data.is_locale_rtl) {
```
syeopite, line 71:
```suggestion
            if (pagination_data.is_rtl) {
```

Typo

assets/js/pagination.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    const prev_ctoken = prev_data.pop();
+
+    // On the first page, the stored continuation token is null.
+    if (prev_ctoken === null) {
+        sessionStorage.removeItem(CONT_CACHE_KEY);
+        let url = window.location.href.split('?')[0];
+        let params = window.location.href.split('?')[1];
+        let url_params = new URLSearchParams(params);
```
syeopite, line 37: You can consolidate the two cases of these into a function. Maybe something like:


```js
// Some pseudo js code

// Method to set the current page's continuation token
// Removes the continuation parameter when a continuation token is not given
function set_continuation(prev_ctoken = null){
    let url = window.location.href.split('?')[0];
    let params = window.location.href.split('?')[1];
    let url_params = new URLSearchParams(params);

    if (continuation) {
        url_params.set("continuation", prev_ctoken);
    } else {
        url_params.delete('continuation');
    };
}

````

syeopite (changes requested):

assets/js/pagination.js
RadoslavL: Thank you for the suggestion!

syeopite (approved):

RadoslavL:
Bump.

RadoslavL:
I'm pretty sure it's still applicable.

RadoslavL:
If applicable, I'd like to receive the bounty associated with this pull request!

syeopite:
@TheFrenchGhosty

RadoslavL:
I've been ghosted by The French Ghosty 😥

Perflyst:
@RadoslavL please send your bitcoin or xmr address and I will take care of the bounty

RadoslavL:
Thank you very much! 🩵

bc1qpq7prf3qs7hvprdljtxtcrrmx8dpr4l70h9a93

Let me know if there's anything I need to do! I'm quite new to receiving btc transactions.

Perflyst:
The bounty has been paid. Transaction fd82c8b90abff3ce379176fb36a99bb3c93b7732f0ef5fc367cc10f643dd40b5

Issue #1708, opened 2021-01-25
[Bug] Next page button is missing


lunainvictum:
tldr The next page button is missing and you can't view more than 19 videos through the search page.

------

Hi,

if i search for "speedrun", i get only 19 Results. On Youtube millions of Results. 

This Problem occurs on every instance. And every instance delivers the same 19 Videos. 

And another Problem is, if Invidious thinks there is more, the next page shows the same videos from the first page. And on the third, fourth, and so on. 

Sry for my bad english. im native german.

hackerncoder:
This appears to be a duplicate of #1480. If so, please comment on the appropriate existing issue instead. If you believe this is not really a duplicate, please leave a comment briefly explaining why. Thank you.

lunainvictum:
The Secound Problem seems to be a duplicate.

But the first one not. It is only possible to find 19 Videos (the same, unique ones) on every instance.

TheFrenchGhosty:
A 20$ bounty has been added to this issue.

Anyone opening a PR fixing this issue, will receive 20$ (in BTC) from the Invidious project.

More details: https://github.com/iv-org/invidious/issues/1898

ralf1307:
Is this fixed by adding "Next Page" and "Previous Page" buttons on the first page or am I misunderstanding the issue? As adding &page=2 results in going to the next page with the buttons shown (with different results from the first page of course).

tagener-noisu:
The problem is that the search view only shows the next page button if the number of videos is greater than 19, but the search sometimes returns just 19 videos instead of 20. What's more interesting, the number of results doesn't depend on the search query, if you search for "games" for example and then reload page several times, you can see that sometimes the next page button is present and sometimes not.
I'm not sure if the problem is with the youtube API, or processing of the results inside invidious. Is there a way to see the raw youtube search results through the API?

PR #4126, opened 2023-09-28
Fix: Search not showing next page button when a next page exists

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/1708

Similar to (part of): https://github.com/iv-org/invidious/pull/3864

This PR checks for a continuation token to determine whether a next page button should be shown. A good example to test this with is when adding videos to a playlist.

Test case:
- create a playlist
- go to page to add videos to playlist
- search "yo mama"
- See next page button if using this PR, don't see it if not using this pr

Whether there's a next page or not is not included in the api as it would have to be a breaking change

### DIFF
--- a/src/invidious/yt_backend/extractors.cr
+++ b/src/invidious/yt_backend/extractors.cr
@@ -717,14 +717,14 @@ private module Extractors
     end
 
     private def self.extract(target)
-      raw_items = [] of Array(JSON::Any)
+      raw_items = [] of JSON::Any
 
       target.dig("primaryContents", "sectionListRenderer", "contents").as_a.each do |node|
         if new_node = node["itemSectionRenderer"]?
-          raw_items << new_node["contents"].as_a
-        end
-        if node["continuationItemRenderer"]?
-          raw_items.push([node])
+          raw_items += new_node["contents"].as_a
+        elsif node["continuationItemRenderer"]?
+          # we put the node in an array so the ContinuationItemRendererParser is able to parse it
+          raw_items << node
         end
       end
 
--- a/src/invidious/yt_backend/extractors.cr
+++ b/src/invidious/yt_backend/extractors.cr
@@ -720,8 +720,11 @@ private module Extractors
       raw_items = [] of Array(JSON::Any)
 
       target.dig("primaryContents", "sectionListRenderer", "contents").as_a.each do |node|
-        if node = node["itemSectionRenderer"]?
-          raw_items << node["contents"].as_a
+        if new_node = node["itemSectionRenderer"]?
+          raw_items << new_node["contents"].as_a
+        end
+        if node["continuationItemRenderer"]?
+          raw_items.push([node])
         end
       end
 
--- a/src/invidious/routes/api/v1/channels.cr
+++ b/src/invidious/routes/api/v1/channels.cr
@@ -440,7 +440,7 @@ module Invidious::Routes::API::V1::Channels
     query.channel = env.params.url["ucid"]
 
     begin
-      search_results = query.process
+      search_results, _ = query.process
     rescue ex
       return error_json(400, ex)
     end
--- a/src/invidious/routes/api/v1/search.cr
+++ b/src/invidious/routes/api/v1/search.cr
@@ -8,7 +8,7 @@ module Invidious::Routes::API::V1::Search
     query = Invidious::Search::Query.new(env.params.query, :regular, region)
 
     begin
-      search_results = query.process
+      search_results, _ = query.process
     rescue ex
       return error_json(400, ex)
     end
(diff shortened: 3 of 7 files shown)


src/invidious/yt_backend/extractors.cr
```diff
@@ -718,8 +718,11 @@ private module Extractors
       raw_items = [] of Array(JSON::Any)
 
       target.dig("primaryContents", "sectionListRenderer", "contents").as_a.each do |node|
-        if node = node["itemSectionRenderer"]?
-          raw_items << node["contents"].as_a
+        if new_node = node["itemSectionRenderer"]?
+          raw_items << new_node["contents"].as_a
+        end
+        if continuation = node["continuationItemRenderer"]?
+          raw_items.push([node])
```
syeopite, line 725: The `continuation` assign is not needed

src/invidious/yt_backend/extractors.cr
```diff
@@ -585,14 +585,14 @@ private module Parsers
   #
   module ContinuationItemRendererParser
     def self.process(item : JSON::Any, author_fallback : AuthorFallback)
-      if item_contents = item["continuationItemRenderer"]?
+      if item_contents = (item.dig?("continuationItemRenderer", "continuationEndpoint") || item.dig?("continuationEndpoint"))
```
syeopite, line 588: If I'm not mistaken since you're just passing in `node` here

https://github.com/iv-org/invidious/blob/3c6742be50d40fb1e393d4ad731c673530ab4181/src/invidious/yt_backend/extractors.cr#L724-L725

I don't believe the changes to `ContinuationItemRendererParser` is necessary

src/invidious/yt_backend/extractors.cr
```diff
@@ -718,8 +718,11 @@ private module Extractors
...
 
       target.dig("primaryContents", "sectionListRenderer", "contents").as_a.each do |node|
-        if node = node["itemSectionRenderer"]?
-          raw_items << node["contents"].as_a
+        if new_node = node["itemSectionRenderer"]?
+          raw_items << new_node["contents"].as_a
+        end
+        if node["continuationItemRenderer"]?
+          raw_items.push([node])
         end
```
syeopite, line 726:
```suggestion
        if new_node = node["itemSectionRenderer"]?
          raw_items << new_node["contents"].as_a
        elsif node["continuationItemRenderer"]?
          raw_items.push([node])
        end
```

ChunkyProgrammer: Iirc they're both in the same object which is why i didnt do an elsif. I can probably double check in a few days (not at my computer right now)

SamantazFox: I'm wondering, what's the use case for that?
If I understand correctly, it will extract the continuation for shelf renderers and alike?

ChunkyProgrammer: Yeah, for the search page, the continuation seems to be in the same json object as the itemSectionRenderer

SamantazFox: @syeopite improving upon your suggestion with the use of `<<`
```suggestion
        if new_node = node["itemSectionRenderer"]?
          raw_items << new_node["contents"].as_a
        elsif node["continuationItemRenderer"]?
          raw_items << node
        end
```

src/invidious/yt_backend/extractors.cr
```diff
@@ -722,9 +722,9 @@ private module Extractors
       target.dig("primaryContents", "sectionListRenderer", "contents").as_a.each do |node|
         if new_node = node["itemSectionRenderer"]?
           raw_items << new_node["contents"].as_a
-        end
-        if node["continuationItemRenderer"]?
-          raw_items.push([node])
+        elsif node["continuationItemRenderer"]?
+          # we put the node in an array so the ContinuationItemRendererParser is able to parse it
+          raw_items << [node]
```
SamantazFox, line 727: That's not needed ^^ We flatten the array right after this loop ^^
(See four lines after this patch)

ChunkyProgrammer: I get this error when I remove the square brackets `Error: expected argument #1 to 'Array(Array(JSON::Any))#<<' to be Array(JSON::Any), not JSON::Any`

SamantazFox: Oh! I see the problem! you need to do the following:

```crystal
raw_items += new_node["contents"].as_a
```

Then everything will be the same type!

ChunkyProgrammer:
There seems to still be continuation tokens but no next page... (the button displays but clicking the button leads to a blank page)

ChunkyProgrammer:
There's some issues with this PR so I'll be closing it for the time being, I might open another PR in a few weeks that should hopefully work better

PR #4146, opened 2023-10-05
Search: Add support for Youtube URLs

### LINKED ISSUE #3300: [Enhancement] [Search] Open video pasting YouTube URL
## Proposal

Improve the search field, in order to accept a YouTube URL instead of a generic search term. Example: you can paste this:

https://www.youtube.com/watch?v=5mgYnwhM4eg

To be redirected there:

https://yewtu.be/watch?v=5mgYnwhM4eg

Some interesting related documents:

* https://stackoverflow.com/questions/6903823/regex-for-youtube-id

Pro:

* you don't need a rewrite extension
    * useful when you are on another computer, or from a smartphone, or from a TV
    * useful to help iOS users in migrating from YouTube (they usually cannot use extensions)
* you don't need to manually rewrite the URL
    * usually longer than copy-paste
    * allows to avoid typo/mistakes

Example use case:

Somebody sends a YouTube URL in a chat. You have an Invidio.us tab already opened (or you have a bookmark to it). You could just copy that, paste in the search bar, and see the video.

SamantazFox:
Closes #3300

### DIFF
--- a/src/invidious/search/query.cr
+++ b/src/invidious/search/query.cr
@@ -58,9 +58,9 @@ module Invidious::Search
       # Remove surrounding whitespaces. Mostly useful for copy/pasted URLs.
       @raw_query = _raw_query.strip
 
-      # Check for smart features (ex: URL search) inhibitor (exclamation mark).
+      # Check for smart features (ex: URL search) inhibitor (backslash).
       # If inhibitor is present, remove it.
-      if @raw_query.starts_with?('!')
+      if @raw_query.starts_with?('\\')
         @inhibit_ssf = true
         @raw_query = @raw_query[1..]
       end
--- a/src/invidious/search/query.cr
+++ b/src/invidious/search/query.cr
@@ -20,6 +20,9 @@ module Invidious::Search
     property region : String?
     property channel : String = ""
 
+    # Flag that indicates if the smart search features have been disabled.
+    @inhibit_ssf : Bool = false
+
     # Return true if @raw_query is either `nil` or empty
     private def empty_raw_query?
       return @raw_query.empty?
@@ -54,6 +57,13 @@ module Invidious::Search
 
       # Remove surrounding whitespaces. Mostly useful for copy/pasted URLs.
       @raw_query = _raw_query.strip
+
+      # Check for smart features (ex: URL search) inhibitor (exclamation mark).
+      # If inhibitor is present, remove it.
+      if @raw_query.starts_with?('!')
+        @inhibit_ssf = true
+        @raw_query = @raw_query[1..]
+      end
 
       # Get the page number (also common to all search types)
       @page = params["page"]?.try &.to_i? || 1
@@ -140,6 +150,9 @@ module Invidious::Search
 
     # Checks if the query is a standalone URL
     def is_url? : Bool
+      # If the smart features have been inhibited, don't go further.
+      return false if @inhibit_ssf
+
       # Only supported in regular search mode
       return false if !@type.regular?
 
--- a/src/invidious/search/query.cr
+++ b/src/invidious/search/query.cr
@@ -48,11 +48,12 @@ module Invidious::Search
     )
       # Get the raw search query string (common to all search types). In
       # Regular search mode, also look for the `search_query` URL parameter
-      if @type.regular?
-        @raw_query = params["q"]? || params["search_query"]? || ""
-      else
-        @raw_query = params["q"]? || ""
-      end
+      _raw_query = params["q"]?
+      _raw_query ||= params["search_query"]? if @type.regular?
+      _raw_query ||= ""
+
+      # Remove surrounding whitespaces. Mostly useful for copy/pasted URLs.
+      @raw_query = _raw_query.strip
 
       # Get the page number (also common to all search types)
       @page = params["page"]?.try &.to_i? || 1
@@ -85,7 +86,7 @@ module Invidious::Search
           @filters = Filters.from_iv_params(params)
           @channel = params["channel"]? || ""
 
-          if @filters.default? && @raw_query.includes?(':')
+          if @filters.default? && @raw_query.index(/\w:\w/)
             # Parse legacy filters from query
             @filters, @channel, @query, subs = Filters.from_legacy_filters(@raw_query)
           else
@@ -135,6 +136,20 @@ module Invidious::Search
       params["channel"] = @channel if !@channel.empty?
 
       return params
+    end
+
+    # Checks if the query is a standalone URL
+    def is_url? : Bool
+      # Only supported in regular search mode
+      return false if !@type.regular?
+
+      # If filters are present, that's a regular search
+      return false if !@filters.default?
+
+      # Simple heuristics: domain name
+      return @raw_query.starts_with?(
+        /(https?:\/\/)?(www\.)?(m\.)?youtu(\.be|be\.com)\//
+      )
     end
   end
 end
--- a/src/invidious/yt_backend/url_sanitizer.cr
+++ b/src/invidious/yt_backend/url_sanitizer.cr
@@ -1,0 +1,121 @@
+require "uri"
+
+module UrlSanitizer
+  extend self
+
+  ALLOWED_QUERY_PARAMS = {
+    channel:  ["u", "user", "lb"],
+    playlist: ["list"],
+    search:   ["q", "search_query", "sp"],
+    watch:    [
+      "v",                                  # Video ID
+      "list", "index",                      # Playlist-related
+      "playlist",                           # Unnamed playlist (id,id,id,...) (embed-only?)
+      "t", "time_continue", "start", "end", # Timestamp
+      "lc",                                 # Highlighted comment (watch page only)
+    ],
+  }
+
+  # Returns wether the given string is an ASCII word. This is the same as
+  # running the following regex in US-ASCII locale: /^[\w-]+$/
+  private def ascii_word?(str : String) : Bool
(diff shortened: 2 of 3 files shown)


src/invidious/yt_backend/url_sanitizer.cr
```diff
@@ -0,0 +1,121 @@
...
+
+        bc
+      end
+
+    # If nothing remains, it's either a legit URL to the homepage
+    # (who does that!?) or because we filtered some junk earlier.
+    return new_uri if breadcrumbs.empty?
+
+    # Replace the original query parameters with the sanitized ones
+    case unsafe_uri.host.not_nil!
```
syeopite, line 99: Could you rewrite this section to not use `not_nil!`? I'd like to slowly remove Invidious's dependence on it since it is generally frowned upon (see https://crystal-lang.org/api/1.10.1/Object.html#not_nil!-instance-method) and would allow for easier diagnosis of the various NilAssertionError seen across Invidious when designed right.

syeopite:
Would it be possible to make this configurable (either in this PR or maybe after #2377)? Since this does prevent some search queries such as `youtube.com` from going through and instead just being redirected.

src/invidious/yt_backend/url_sanitizer.cr
```diff
@@ -0,0 +1,121 @@
...
+    return new_uri if (unsafe_uri.host.nil? || unsafe_uri.path.nil?)
+
+    breadcrumbs = unsafe_uri.path
+      .split('/', remove_empty: true)
+      .compact_map do |bc|
+        # Exclude attempts at path trasversal
+        next if bc == "." || bc == ".."
+
+        # Non-alnum characters are unlikely in a genuine URL
+        next if !ascii_word?(bc)
```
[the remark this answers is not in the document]
SamantazFox: The problem of `#ascii_only?` is that it returns true for *any* ASCII characters, even control characters, spaces, etc... :
https://github.com/crystal-lang/crystal/blob/1908c816f/src/string.cr#L5189-L5198

src/invidious/yt_backend/url_sanitizer.cr
SamantazFox: If I do the following
```crystal
    # Replace the original query parameters with the sanitized ones
    case unsafe_uri.host
    when .ends_with?("youtube.com")
      # ...
    end
```

then the compiler complains that `Nil` doesn't have a `#ends_with?` method, even if I checked that it wasn't nil beforehand:

```crystal
# Redirect to homepage for bogus URLs
return new_uri if (unsafe_uri.host.nil? || unsafe_uri.path.nil?)
```

SamantazFox: I can make an intermediary variable (before the `if` check) if you want!
That's the only other way to make the compiler happy.

src/invidious/yt_backend/url_sanitizer.cr
```diff
@@ -0,0 +1,121 @@
...
+    watch:    [
+      "v",                                  # Video ID
+      "list", "index",                      # Playlist-related
+      "playlist",                           # Unnamed playlist (id,id,id,...) (embed-only?)
+      "t", "time_continue", "start", "end", # Timestamp
+      "lc",                                 # Highlighted comment (watch page only)
+    ],
+  }
+
+  # Returns wether the given string is an ASCII word. This is the same as
```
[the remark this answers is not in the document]
SamantazFox: Ah, thanks! Idk why I'm always skipping that first `h` ...

SamantazFox:
Searching `youtube.com` is skipped for this exact reason:
https://github.com/iv-org/invidious/pull/4146/files#diff-19bdcb6c58e32fcfbe37cf190733371fc06272002c38b847e114e9705b3ce649R149-R152

However, searching for `youtube.com/` (note the trailing `/`) will trigger URL redirection

src/invidious/yt_backend/url_sanitizer.cr
syeopite: Yes, please do. Intermediate variables seems to be the intended way to handle nilable properties within Crystal. There's an interesting discussion here about it https://forum.crystal-lang.org/t/is-casting-the-idiomatic-way-of-dealing-with-nil/1693

syeopite:
Oops I missed that.

But my point still stands. In general this PR blocks certain search queries from going through to YouTube due to the URL redirection functionality. 

`youtube.com`
`youtube.com/`
`youtube.com/watch?v=`

etc all returns different search results on YouTube. And although this may be an edge case, it still does "block" certain search results from showing up on Invidious and that's something I'd like to avoid if possible.

SamantazFox:
Is that edge case worth the added complexity of another config option?
IIRC Piped has such a URL search feature and no toggle.

syeopite:
I think its worth it. Blocking search terms from going through to YouTube is something Invidious should never do imo edgecase or not.

syeopite:
This is one of the few times where I disagree on doing something automatically for the sake of convenience. To give a personal anecdote, I've accidentally encountered the automatically redirection feature on Libreddit more than once when just trying to search for a subreddit name. Its way less of an issue on YouTube but it does exist. 

There are many popular videos titled with just the Never Gonna Give You Up link for example. This feature (without an toggle) would make it harder to locate those videos through the search function.

SamantazFox:
What do you think of a DDG "bang" style trigger, instead? In essence, the URL search would trigger only if the search field contains exactly one valid URL, without any extra search terms, and preceded by `!`.

syeopite:
Won't that end up being the exact same problem? The search results for a link prefixed with a exclamation mark and one that isn't is slightly different. It should occur less than with just a link but it doesn't really solve the problem. 

If a configuration option is too complex what about a "smart search" button directly within the search bar? Here's an example mock up with a placeholder icon.
smart search

Enabled by default, since this is an edge case, it can be disabled by toggling the button prior to a search result. The only additional complexity then would be a single check for a query parameter in the search handler.

SamantazFox:
Well, the probability that a user searches for an exact, valid Youtube URL, with a leading exclamation mark (that seem to be ignored by youtube anyway?) without any other search terms, is very thin, if not to say null.

I like the idea of a smart search thing, but I don't know how to integrate that nicely with the rest (a search button, for sure, and maybe a toggle for suggestions). I'm worried it'll clutter the search bar...

syeopite:
Actually now that I think about it, I'll be fine with a bang system as long as the `!` (and the backslash) can be escaped.

SamantazFox:
@syeopite As you requested, I added some logic to bypass this feature.

By default, if the search query is a valid youtube URL, it redirects to that page.

If the query starts with an excalmation mark (`!`), then the search returns to regular "dumb" search. If you still want to search for `!www.youtube.com/etc..`, you have to type a second `!`, like so: `!!www.youtube.com/etc..`.

Is that good for you?

syeopite:
Is there a way to also escape the second bang? If so that's fine with me.

syeopite:
Yeah that's fine by me.

src/invidious/search/query.cr
```diff
@@ -48,10 +51,18 @@ module Invidious::Search
...
-      if @type.regular?
-        @raw_query = params["q"]? || params["search_query"]? || ""
-      else
-        @raw_query = params["q"]? || ""
+      _raw_query = params["q"]?
+      _raw_query ||= params["search_query"]? if @type.regular?
+      _raw_query ||= ""
+
+      # Remove surrounding whitespaces. Mostly useful for copy/pasted URLs.
+      @raw_query = _raw_query.strip
```
syeopite, line 59: Shouldn't this be under an `@inhibit_ssf` check as so we're just sending the query as-is when its off?

src/invidious/search/query.cr
```diff
@@ -48,10 +51,18 @@ module Invidious::Search
...
+
+      # Check for smart features (ex: URL search) inhibitor (exclamation mark).
+      # If inhibitor is present, remove it.
+      if @raw_query.starts_with?('!')
+        @inhibit_ssf = true
+        @raw_query = @raw_query[1..]
```
syeopite, line 65: Using `!` as an inhibitor seems a bit odd to me, especially when it functions as an escape. What do you think of using a backslash instead?

syeopite (approved):

src/invidious/search/query.cr
SamantazFox: I'm neutral on this

src/invidious/search/query.cr
SamantazFox: if the query starts with a space, the check below becomes useless, so we need to remove at least the leading spaces.

syeopite: That's a good point, with the way as is though is sending a raw query like  `" youtube.com/watch?v=dQw4w9WgXcQ"` impossible then?

src/invidious/search/query.cr
syeopite: If it doesn't make parsing more difficult then I'm in favor of using a backslash instead of an exclamation mark. Since usually an `!` is to _use_ a special smart functionality rather than negate it

src/invidious/search/query.cr
SamantazFox: Yes, though Youtube pretty much ignores the extra space. Search result for "query" and " query" are identical on average (they change from request to request, even with the exact same query)

SamantazFox: I realized that you can search for `\ youtube.com/xxx`, tho. It will disable smart search, then remove the backslash, and pass the rest of the string.

PR #4149, opened 2023-10-07
API: Add a 'published' video parameter for related videos

### LINKED ISSUE #3058: [Enhancement] API ``/api/v1/videos/:id`` - Return ``published`` for ``recommendedVideos``
I'm currently implementing Invidious support into NewPipe (https://github.com/TeamNewPipe/NewPipe/pull/8220) and noticed that there is no published-date/timestamp for related/recommended videos.

The timestamp should be exposed via the API.

Currently you have to switch to the NewPipe-native YT extraction to see the publish-date/time of a related video.

**Additional context**
N/A

RadoslavL:
This closes #3058.

It could also allow recommended videos to have their upload date shown on the `/watch` route. Let me know if you want me to add that as well.

Important thing to note - this way of adding the parameter wouldn't be very precise for old videos. New videos may show up as "1 hour ago" and when that happens, the date would be accurate to the hours. Old videos may show up as "1 year ago" and that would mean that the date would only be accurate to the years. The month and day would just be chosen to be the current time's month and day, so it wouldn't even be close. But without making more API requests I don't see how this can work more accurately.

### DIFF
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -246,7 +246,11 @@ module Invidious::JSONify::APIv1
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
-                json.field "publishedTimeText", translate(locale, "`x` ago", rv["publishedText"].to_s.gsub(" ago", ""))
+                if !rv[published].nil?
+                  json.field "publishedText", translate(locale, "`x` ago", recode_date(rv[published], locale))
+                else
+                  json.field "publishedText", translate(locale, "`x` ago", "NaN")
+                end
               end
             end
           end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -246,7 +246,7 @@ module Invidious::JSONify::APIv1
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
-                if !rv["published"].nil?
+                if !rv["published"]?.nil?
                   json.field "publishedText", translate(locale, "`x` ago", recode_date(Time.parse_rfc3339(rv["published"].to_s), locale))
                 else
                   json.field "publishedText", ""
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -245,7 +245,7 @@ module Invidious::JSONify::APIv1
                 json.field "lengthSeconds", rv["length_seconds"]?.try &.to_i
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
-                json.field "published", rv["published"]?.try { |t| Time.parse(t, "%Y-%m-%d", Time::Location::UTC).to_unix } || Time.utc
+                json.field "published", rv["published"]?
               end
             end
           end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -247,9 +247,9 @@ module Invidious::JSONify::APIv1
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
                 if !rv["published"].nil?
-                  json.field "publishedText", translate(locale, "`x` ago", recode_date(Time.unix(rv["published"].to_i), locale))
+                  json.field "publishedText", translate(locale, "`x` ago", recode_date(Time.parse_rfc3339(rv["published"].to_s), locale))
                 else
-                  json.field "publishedText", translate(locale, "`x` ago", "NaN")
+                  json.field "publishedText", ""
                 end
               end
             end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -246,8 +246,8 @@ module Invidious::JSONify::APIv1
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
-                if !rv[published].nil?
-                  json.field "publishedText", translate(locale, "`x` ago", recode_date(rv[published], locale))
+                if !rv["published"].nil?
+                  json.field "publishedText", translate(locale, "`x` ago", recode_date(rv["published"], locale))
                 else
                   json.field "publishedText", translate(locale, "`x` ago", "NaN")
                 end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -246,7 +246,7 @@ module Invidious::JSONify::APIv1
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
-                json.field "publishedTimeText", rv["publishedText"]?
+                json.field "publishedTimeText", translate(locale, "`x` ago", rv["publishedText"].to_s)
               end
             end
           end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -247,7 +247,7 @@ module Invidious::JSONify::APIv1
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
                 if !rv["published"].nil?
-                  json.field "publishedText", translate(locale, "`x` ago", recode_date(rv["published"], locale))
+                  json.field "publishedText", translate(locale, "`x` ago", recode_date(Time.unix(rv["published"].to_i), locale))
                 else
                   json.field "publishedText", translate(locale, "`x` ago", "NaN")
                 end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -246,6 +246,7 @@ module Invidious::JSONify::APIv1
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
+                json.field "publishedTimeText", rv["publishedText"]?
               end
             end
           end
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -246,7 +246,7 @@ module Invidious::JSONify::APIv1
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
                 json.field "published", rv["published"]?
-                json.field "publishedTimeText", translate(locale, "`x` ago", rv["publishedText"].to_s)
+                json.field "publishedTimeText", translate(locale, "`x` ago", rv["publishedText"].to_s.gsub(" ago", ""))
               end
             end
           end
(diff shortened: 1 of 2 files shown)


absidue:
Please refactor this to query the database instead of making 20+ API requests every single time. Ideally parse the published date that is already in the watch next/recommended videos feed instead of making any extra requests. This current approach will get every instance, other than small ones, rate limited by YouTube very quickly, which will result in it being impossible to play videos at all on those instances.

RadoslavL:
> Please refactor this to query the database instead of making 20+ API requests every single time. Ideally parse the published date that is already in the watch next/recommended videos feed instead of making any extra requests. This current approach will get every instance, other than small ones, rate limited by YouTube very quickly, which will result in it being impossible to play videos at all on those instances.

When I made this I thought that a couple of API calls wouldn't really matter, but now that I think about it, this is a very big problem. I can't get the published date from the related videos' info hash, because they only include the simpleText version. This means that it will show up as "1 hour ago" instead of "2023-10-07". Unless I can fetch the video data separately or compute it manually, I wouldn't be able to solve this issue. I will try parsing the date using the timezone data tomorrow. I will also make sure to use the database when possible. Thank you for the reply!

RadoslavL:
I removed the need for more API calls, but keep in mind that this method wouldn't be very precise. If the video is an hour or a day old, the date would be right, but if it is a year old, it might be off by months.

src/invidious/videos/parser.cr
```diff
@@ -157,7 +158,86 @@ def try_fetch_streaming_data(id : String, client_config : YoutubeAPI::ClientConf
   end
 end
 
-def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any)) : Hash(String, JSON::Any)
+def parse_published_string(item)
```
SamantazFox, line 161: We already have an helper function to do that ^^
Here is the relevant code: https://github.com/iv-org/invidious/blob/069e91d2a609b1df94bdfbb693335969b2a281d8/src/invidious/helpers/utils.cr#L94-L141

RadoslavL: Thank you!

src/invidious/videos/parser.cr
```diff
@@ -157,7 +158,7 @@ def try_fetch_streaming_data(id : String, client_config : YoutubeAPI::ClientConf
   end
 end
 
-def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any)) : Hash(String, JSON::Any)
+def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any), proxy_region : String? = nil) : Hash(String, JSON::Any)
```
syeopite, line 161: Please remove the extra `proxy_region` argument. Seems to be leftover from your other changes.

src/invidious/videos/parser.cr
```diff
@@ -235,7 +236,14 @@ def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any
...
     if item = element["compactVideoRenderer"]?
-      related_video = parse_related_video(item)
+      time_text = item["publishedTimeText"]?
+      if !time_text.nil?
+        time = decode_date(item["publishedTimeText"].to_s)
+        published1 = JSON::Any.new(time.to_unix.to_s)
+      else
+        published1 = JSON::Any.new("")
+      end
+      related_video = parse_related_video(item, published1)
```
syeopite, line 246:
```suggestion
      if item["publishedTimeText"]?
        rv_decoded_time = decode_date(item["publishedTimeText"].to_s)
        rv_published_timestamp = JSON::Any.new(rv_decoded_time.to_unix.to_s)
      else
        rv_published_timestamp = JSON::Any.new("")
      end
      related_video = parse_related_video(item, rv_published_timestamp)
```

src/invidious/videos/parser.cr
```diff
@@ -250,7 +258,7 @@ def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any
 
     player_overlays.try &.as_a.each do |element|
       if item = element["endScreenVideoRenderer"]?
-        related_video = parse_related_video(item)
+        related_video = parse_related_video(item, JSON::Any.new(""))
```
syeopite, line 261:
```suggestion
        related_video = parse_related_video(item, published: JSON::Any.new(""))
```

src/invidious/videos/parser.cr
```diff
@@ -6,7 +6,7 @@ require "json"
 #
 # TODO: "compactRadioRenderer" (Mix) and
 # TODO: Use a proper struct/class instead of a hacky JSON object
-def parse_related_video(related : JSON::Any) : Hash(String, JSON::Any)?
+def parse_related_video(related : JSON::Any, published) : Hash(String, JSON::Any)?
```
syeopite, line 9: You can probably just make `published` into an optional argument here. Also, I'd prefer if you can just initialize the JSON::Any below. 

The related video data struct can then look something like 
` "published"        => JSON::Any.new(published || "0"),`

syeopite (changes requested):

src/invidious/videos/parser.cr
```diff
@@ -47,6 +47,8 @@ def parse_related_video(related : JSON::Any) : Hash(String, JSON::Any)?
     "view_count"       => JSON::Any.new(view_count || "0"),
     "short_view_count" => JSON::Any.new(short_view_count || "0"),
     "author_verified"  => JSON::Any.new(author_verified),
+    "published"        => JSON::Any.new(published || ""),
+    "publishedText"    => JSON::Any.new(publishedText || ""),
```
syeopite, line 51:
```suggestion
    "publishedText"    => JSON::Any.new(translate(locale, "`x` ago", recode_date(video.published, locale)) || ""),
```

src/invidious/videos/parser.cr
```diff
@@ -235,7 +237,14 @@ def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any
     .dig?("secondaryResults", "secondaryResults", "results")
   secondary_results.try &.as_a.each do |element|
     if item = element["compactVideoRenderer"]?
-      related_video = parse_related_video(item)
+      if item["publishedTimeText"]?
+        rv_published_time_text = item["publishedTimeText"].to_s
+        rv_decoded_time = decode_date(item["publishedTimeText"].to_s)
```
syeopite, line 242:
```suggestion
      if rv_published_time_text = item["publishedTimeText"]?
        rv_published_time_text = rv_published_time_text.as_s
        rv_decoded_time = decode_date(rv_published_time_text)
```

syeopite (approved):
Just a couple more nitpicks

src/invidious/videos/parser.cr
```diff
@@ -47,6 +47,8 @@ def parse_related_video(related : JSON::Any) : Hash(String, JSON::Any)?
     "view_count"       => JSON::Any.new(view_count || "0"),
     "short_view_count" => JSON::Any.new(short_view_count || "0"),
     "author_verified"  => JSON::Any.new(author_verified),
+    "published"        => JSON::Any.new(published || ""),
+    "publishedText"    => JSON::Any.new(translate(locale, "`x` ago", recode_date(video.published, locale)) || ""),
```
SamantazFox, line 51: The translations should be handled on the API side. The `locale` variable never reaches the parser's scope.

src/invidious/videos/parser.cr
```diff
@@ -235,7 +237,14 @@ def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any
     .dig?("secondaryResults", "secondaryResults", "results")
   secondary_results.try &.as_a.each do |element|
     if item = element["compactVideoRenderer"]?
-      related_video = parse_related_video(item)
+      if rv_published_time_text = item["publishedTimeText"]?
```
SamantazFox, line 240: if `publishedTimeText` is part of the `compactVideoRenderer`, this should be handled inside `parse_related_video`. Overall, variables should be kept in they function and not passed around as much as possible!

src/invidious/videos/parser.cr
syeopite: That was my suggestion. Oops. I'm surprised the CI didn't catch that though

src/invidious/videos/parser.cr
RadoslavL: Sorry for the delay. It's solved in 68611482906f3b70e9c105fec5c57436c877a3b0

RadoslavL:
Bump.

ashley0143:
doesnt work after the new youtube update

src/invidious/jsonify/api_v1/video_json.cr
```diff
@@ -245,6 +245,8 @@ module Invidious::JSONify::APIv1
                 json.field "lengthSeconds", rv["length_seconds"]?.try &.to_i
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
+                json.field "published", rv["published"]?
+                json.field "publishedTimeText", translate(locale, "`x` ago", rv["publishedText"].to_s.gsub(" ago", ""))
```
SamantazFox, line 249: Two things here:
1. The JSON field should be `publishedText`, to match the rest of the API
2. The text should be translated from the time we parsed earlier. The text from the innertube (internal Youtube) API is always in english, and can't be used directly.

src/invidious/videos/parser.cr
```diff
@@ -47,6 +55,8 @@ def parse_related_video(related : JSON::Any) : Hash(String, JSON::Any)?
     "view_count"       => JSON::Any.new(view_count || "0"),
     "short_view_count" => JSON::Any.new(short_view_count || "0"),
     "author_verified"  => JSON::Any.new(author_verified),
+    "published"        => JSON::Any.new(published || ""),
+    "publishedText"    => JSON::Any.new(published_time_text || ""),
```
SamantazFox, line 59: Please remove that: internally, we shouldn't store any "published" text, especially since the Youtube side is only in english.

SamantazFox (changes requested):

src/invidious/jsonify/api_v1/video_json.cr
RadoslavL: I had trouble fully understanding the second request, but I improved my old code. I thought about using `Time.unix(0)` when `rv[published]` was `nil`, but decided to make it more obvious that something is wrong, when it happens.

I also greatly apologize for the 3 month delay! I took a very long break, because of personal issues in my life.

src/invidious/videos/parser.cr
RadoslavL: Got it. Sorry!

src/invidious/jsonify/api_v1/video_json.cr
```diff
@@ -245,6 +245,12 @@ module Invidious::JSONify::APIv1
                 json.field "lengthSeconds", rv["length_seconds"]?.try &.to_i
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
+                json.field "published", rv["published"]?
+                if !rv["published"].nil?
+                  json.field "publishedText", translate(locale, "`x` ago", recode_date(Time.unix(rv["published"].to_i), locale))
+                else
+                  json.field "publishedText", translate(locale, "`x` ago", "NaN")
+                end
```
SamantazFox, line 253: I'd rather not show anything if that information is unknown. It makes for a better user experience than `Il y a NaN` ("NaN ago", translated in french, for instance)

SamantazFox, line 253: I think it's better to not return anything (or nil) rather than sending a broken text to the API client (it makes for a bad user experience)

src/invidious/videos/parser.cr
```diff
@@ -36,6 +36,13 @@ def parse_related_video(related : JSON::Any) : Hash(String, JSON::Any)?
 
   LOGGER.trace("parse_related_video: Found \"watchNextEndScreenRenderer\" container")
 
+  if published_time_text = related["publishedTimeText"]?
+    decoded_time = decode_date(published_time_text["simpleText"].to_s)
+    published = decoded_time.to_unix.to_s
```
SamantazFox, line 41: The rest of the API uses the RFC 3339 format:
```suggestion
    published = decoded_time.to_rfc3339
```

src/invidious/jsonify/api_v1/video_json.cr
SamantazFox: > I also greatly apologize for the 3 month delay! I took a very long break, because of personal issues in my life.

No problem :) Your life is much more important than anything else!

RadoslavL: Thank you so much for saying that :)

src/invidious/jsonify/api_v1/video_json.cr
```diff
@@ -245,6 +245,12 @@ module Invidious::JSONify::APIv1
                 json.field "lengthSeconds", rv["length_seconds"]?.try &.to_i
                 json.field "viewCountText", rv["short_view_count"]?
                 json.field "viewCount", rv["view_count"]?.try &.empty? ? nil : rv["view_count"].to_i64
+                json.field "published", rv["published"]?
+                if !rv["published"].nil?
```
SamantazFox, line 249: The '?' is required to prevent an error when loading previously cached video data after an update!
```suggestion
                if !rv["published"]?.nil?
```

SamantazFox (approved):

syeopite:
Thank you so much for your PR 🎉! Our deepest apologies about just how long it has taken for this to get merged 😅

RadoslavL:
Don't worry about it @syeopite 😅

I actually figured out I was bi and then trans, all before you merged the pull request 😭

But don't worry!!! It's all okay! ❤️

It's just a little funny that it happened this way 😅

syeopite:
Hey congratulations! I'm so happy for you 🥳

This really puts into perspective the state of Invidious' development when all that can happen in the span of a single PR's lifecycle 😆

PR #4152, opened 2023-10-08
Captions: Use 'fmt=vtt' instead of 'format=vtt'

SamantazFox:
Thanks @absidue for making me aware that the URL parameter changed ^^


unixfox:
Does this fixes a bug? If so then I approve the PR.

absidue:
YouTube removed/ignored the `format` parameter so currently Invidious is always requesting the XML and has to transform it to a WebVTT, switching to the `fmt` will make YouTube return a WebVTT file, which saves Invidious some work.

PR #4156, opened 2023-10-09
update params for fetching stream data

unixfox:
the param has changed and it's now this value that is used by all the alternative clients to youtube.

example in youtube js: https://github.com/LuanRT/YouTube.js/blob/main/src/core/endpoints/PlayerEndpoint.ts#L46

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -137,9 +137,9 @@ end
 
 def try_fetch_streaming_data(id : String, client_config : YoutubeAPI::ClientConfig) : Hash(String, JSON::Any)?
   LOGGER.debug("try_fetch_streaming_data: [#{id}] Using #{client_config.client_type} client.")
-  # CgIQBg is a workaround for streaming URLs that returns a 403.
+  # 2AMBCgIQBg is a workaround for streaming URLs that returns a 403.
   # See https://github.com/iv-org/invidious/issues/4027#issuecomment-1666944520
-  response = YoutubeAPI.player(video_id: id, params: "CgIQBg", client_config: client_config)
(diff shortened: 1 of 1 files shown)


unixfox:
> Not NewPipe apparently? [TeamNewPipe/NewPipeExtractor#1115 (comment)](https://github.com/TeamNewPipe/NewPipeExtractor/issues/1115#issuecomment-1754526593)

newpipe is a client that is used by one person at a time, whereas ours is being used by many people at the same time.

on yewtu.be I'm using this new parameter and it has worked great.

I don't agree with point that it will easier to detect us as a 3rd party client because we are already being detected as every X weeks youtube block the IP addresses of all the public instances.

src/invidious/videos/parser.cr
```diff
@@ -137,9 +137,9 @@ end
 
 def try_fetch_streaming_data(id : String, client_config : YoutubeAPI::ClientConfig) : Hash(String, JSON::Any)?
   LOGGER.debug("try_fetch_streaming_data: [#{id}] Using #{client_config.client_type} client.")
-  # CgIQBg is a workaround for streaming URLs that returns a 403.
+  # 2AMBCgIQBg is a workaround for streaming URLs that returns a 403.
   # See https://github.com/iv-org/invidious/issues/4027#issuecomment-1666944520
```
SamantazFox, line 141: Can you update (or remove) that link?

unixfox: done

SamantazFox (approved):

cgmastertecnology:
hi all, my docker instance is very slow
how can i fix it?
thank you

PR #4161, opened 2023-10-10
Workflow: use docker-meta to generate versioned docker tags & OCI compatible labels

unbelauscht:
I updated the workflow to have additional tags based on git tags and [OCI compatible labels](https://github.com/opencontainers/image-spec/blob/main/annotations.md).

This way an administrator can get images by Invidious version tag.

Here is an example tag [0.20.4](https://github.com/unbelauscht/invidious/actions/runs/6473340341) and the [resulting images](https://hub.docker.com/r/jan4971/invidious/tags).

### DIFF
--- a/.github/workflows/container-release.yml
+++ b/.github/workflows/container-release.yml
@@ -64,7 +64,7 @@ jobs:
             quay.expires-after=12w
 
       - name: Build and push Docker AMD64 image for Push Event
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: docker/Dockerfile
@@ -90,7 +90,7 @@ jobs:
             quay.expires-after=12w
 
       - name: Build and push Docker ARM64 image for Push Event
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v5
         with:
           context: .
           file: docker/Dockerfile.arm64


unbelauscht:
I modified it to use the current date. Feel free to close the pull request, if that's not what you want :)

unixfox:
If you could have the date and the commit id that would be awesome.

Like we have at the bottom of every invidious page:

```
Current version: 2023.10.11-e8641cb @ master
```

unbelauscht:
Luckily, `docker/metadata-action@v5` supports `type=sha,format=short`, so I added the date as prefix which results in the format displayed in Invidious.

This is the [resulting metadata](https://github.com/unbelauscht/invidious/actions/runs/6480032448/job/17594904096) for my fork:
```
  {
    "tags": [
      "***/invidious:latest",
      "***/invidious:2023.10.11-817a9c9"
    ],
    "labels": {
      "org.opencontainers.image.created": "2023-10-11T08:40:36.664Z",
      "org.opencontainers.image.description": "Invidious is an alternative front-end to YouTube",
      "org.opencontainers.image.licenses": "AGPL-3.0",
      "org.opencontainers.image.revision": "817a9c9b961773b4e19f0f1581e38e877051ea74",
      "org.opencontainers.image.source": "https://github.com/unbelauscht/invidious",
      "org.opencontainers.image.title": "invidious",
      "org.opencontainers.image.url": "https://github.com/unbelauscht/invidious",
      "org.opencontainers.image.version": "latest",
      "quay.expires-after": "12w"
    }
  }
```

.github/workflows/container-release.yml
```diff
@@ -52,28 +51,52 @@ jobs:
...
+          images: quay.io/invidious/invidious
+          tags: |
+            type=sha,format=short
+            type=sha,format=short,prefix={{date 'YYYY.MM.DD'}}-,enable=${{ github.ref == format('refs/heads/{0}', 'master') }}
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'master') }}
+          labels: |
+            quay.expires-after=12w
+
       - name: Build and push Docker AMD64 image for Push Event
-        if: github.ref == 'refs/heads/master'
```
SamantazFox, line 56: Why would you remove that? We don't want to build and push containers for other branches!

unixfox:
Could you rebase against master? Thank you

unixfox (approved):

.github/workflows/container-release.yml
```diff
@@ -52,28 +51,52 @@ jobs:
...
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v3
```
SamantazFox, line 67: Can you please revert this change?

unixfox: Same for building arm image

TheFrenchGhosty:
@unbelauscht This PR seems to cause the creation of 2 "useless" branch that have "sha" as prefix:

image


Other than that, it seems to work perfectly

unbelauscht:
Yes, that's true. Do we want to keep these? They used to be long hashes without prefix.

Screenshot 2023-12-06 at 13 41 30

If we want to keep / remove these tags, we may need to reopen this PR.

TheFrenchGhosty:
@unbelauscht In my opinion the date-sha1 is the better replacement, those sha-sha1 should be removed.

I'll let you open another PR for that if you can do it

PR #4163, opened 2023-10-12
Added support for `preservesPitch`

Windows81:
I hadn't been able to test it properly, but changes should be minimal.  Since Video.js doesn't natively support the `preservesPitch` property, I used this snippet of code in `player.js` to get the video player to support the property:
```js
player.tech_.el_.preservesPitch = video_data.params.preserves_pitch;
```

I copied the existing preference structure from `video_loop`, except that the default setting is `true`.  I also provided a brief summary in `config.example.yaml` and added localisation strings in English and Spanish, along with a few other languages.


unixfox:
Can you explain what is "preservesPitch" feature?

Windows81:
If `preservesPitch` is set to `false` on a `<video>` element, then the pitch goes up in proportion with the speed.  That means that if you play a video at 1.5× speed, for example, the pitch goes up by a major fifth.  By default, it's set to `true`.

TheFrenchGhosty:
As someone who watch a lot of videos at  1.25x, 1.5x or 2x speed I can tell you that the pitch doesn't change with the speed.

So I don't see any point in this honestly

SamantazFox:
While I'm not against adding new features, I'm not found of adding more config options. There are already a lot of settings, and imho this can be quite confusing for newer users. It is also getting hard to maintain (the preferences page is slowly becoming an incoherent mess).

As this setting will only be used by a very small minority of our users, in addition to being specific to VideoJS, I'd prefer if it was handled 100% client side. I'm not sure how to present that to the user, but maybe adding a dedicated VideoJS "advanced settings" overlay or similar would be nice!

PR #4209, opened 2023-10-28
ci: update github actions

armujahid:


SamantazFox (approved):

unixfox:
@armujahid 

ideally it would have been so cool if you could update the version of all the other actions :). that would save us a lot of time

armujahid:
Sure. Will check

armujahid:
@ unixfox  done

unixfox:
@armujahid did you check that all the tags exist because there is an error at the compilation: https://github.com/iv-org/invidious/actions/runs/6884034885/job/19054420702?pr=4209

armujahid:
@unixfox  I have reverted my change related to install-crystal. it should pass now.

PR #4211, opened 2023-10-29
Channels sorted by last published video

iBicha:
This adds a query param to `/api/v1/auth/subscriptions`: `sorted=true`.

What it does is it sorts the channels you are subscribed to by the date of their last published video.

The reason is I want to experiment with a view of the channels (not the subscription feed, which is the videos) where the more "active" channels are showed first. but at the same time, if some channels are too active (e.g. multiple videos a day) they do not "hog" the interface, they still occupy a single cell in the screen.

It's a way to see at a glance the channels that you follow who recently published videos, and so you can decide which channels you want to explore further, but also keeping the channels that haven't published anything in years at the back of the list.


SamantazFox:
I like the idea! Can you make that `sort=last` though? That way, it can be expanded in the future, and it matches the other endpoints' behavior.

iBicha:
> I like the idea! Can you make that `sort=last` though? That way, it can be expanded in the future, and it matches the other endpoints' behavior.

Makes sense, done

iBicha:
@SamantazFox Look at this interesting visual representation https://github.com/iBicha/playlet/issues/260#issuecomment-1894143161

Channels are sorted by their last upload, and their latest videos feed are also visible on screen. But since each channel only fill a row on screen, they are all equally represented.

src/invidious/database/channels.cr
```diff
@@ -85,6 +85,25 @@ module Invidious::Database::Channels
...
+  # Select channels sorted by last published video
+  def select_sorted(ids : Array(String)) : Array(InvidiousChannel)?
+    return [] of InvidiousChannel if ids.empty?
+
+    request = <<-SQL
+      WITH max_published AS (
+        SELECT ucid, MAX(published) AS last_published
+        FROM channel_videos
+        GROUP BY ucid
+      )
```
syeopite, line 98: I'm not very familiar with SQL so please correct me if I'm wrong but doesn't this fetch every single row from the channels_videos column regardless of which channel IDs were requested? That seems inefficient

iBicha: Might be, it's been a while so I don't know what I was thinking (perhaps I looked it up and it was performant, or perhaps I thought I'll validate the idea first and optimize it after)
Either way, I think this will never merge, I'll just close the PR
Thanks for the review (and of course if this is useful in anyway please feel free to take over)

PR #4221, opened 2023-11-01
API: Fix URL for vtt subtitles

karelrooted:
for fmt=vtt to work the fmt parameter in the original caption api url need to be replaced

### DIFF
--- a/src/invidious/routes/api/v1/videos.cr
+++ b/src/invidious/routes/api/v1/videos.cr
@@ -136,7 +136,11 @@ module Invidious::Routes::API::V1::Videos
           end
         end
       else
-        webvtt = YT_POOL.client &.get("#{url}&fmt=vtt").body
+        uri = URI.parse(url)
+        query_params = uri.query_params
+        query_params["fmt"] = "vtt"
+        uri.query_params = query_params
+        webvtt = YT_POOL.client &.get(uri.request_target).body
 
         if webvtt.starts_with?("<?xml")
           webvtt = caption.timedtext_to_vtt(webvtt)


absidue:
In all the API responses I've seen, YouTube doesn't include the `fmt` value by default.

As I'm guessing there is a reason why you opened this pull request, it's probably better to change this to proper query parameter handling, instead of a find and replace that won't always work.

karelrooted:
> In all the API responses I've seen, YouTube doesn't include the `fmt` value by default.
> 
> As I'm guessing there is a reason why you opened this pull request, it's probably better to change this to proper query parameter handling, instead of a find and replace that won't always work.

the YoutubeAPI::ClientType::Android client request response with fmt=srv3,  I didn't dig deep into the code, but from the request recorded by the mitm proxy , there is two request for the videoid info, the second android one has fmt=srv3 and the  second one is being used

I will change the code to  handle query parameter properly later

src/invidious/routes/api/v1/videos.cr
```diff
@@ -136,7 +136,11 @@ module Invidious::Routes::API::V1::Videos
           end
         end
       else
-        webvtt = YT_POOL.client &.get("#{url}&fmt=vtt").body
+        uri = URI.parse("#{url}")
```
SamantazFox, line 139: You don't need to use string extrapolation here!

```suggestion
        uri = URI.parse(url)
```

SamantazFox (changes requested):

src/invidious/routes/api/v1/videos.cr
karelrooted: fixed, the commits is squashed

SamantazFox (approved):
Code looks good! If you could provide an example case where this code is required (= the relevant sample JSON response) that'd be perfect!

karelrooted:
> Code looks good! If you could provide an example case where this code is required (= the relevant sample JSON response) that'd be perfect



> Code looks good! If you could provide an example case where this code is required (= the relevant sample JSON response) that'd be perfect!

fix in the commit subject might not be correctly word, invidious api always return corrent vtt subtitle (the xml fallback check code is working), I have a local branch with bilingual  subtitle support , when merging with newest upstream main branch , found the problem , so submit the PR

as json example , the following curl request shows append fmt=vtt to a already exist fmt=srv3 url did not work

append example (fmt=srv3&fmt=vtt , return xml)
```bash
curl 'https://www.youtube.com/api/timedtext?v=isY5MCWPi_0&ei=PW1UZb6tGNXM2roPxd2r2A4&opi=112496729&xoaf=5&hl=en&ip=0.0.0.0&ipbits=0&expire=1700057005&sparams=ip,ipbits,expire,v,ei,opi,xoaf&signature=EB364571027A63836BB97ABE4EF93C73790B56F3.2B2D8D9E9143255F38CF415CBA9960AB619DED87&key=yt8&lang=en-US&fmt=srv3&tlang=&fmt=vtt' -H 'Host: www.youtube.com' -H 'User-Agent: Crystal' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7' -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' -H 'Accept-Language: en-us,en;q=0.5' -H 'Cookie: ; CONSENT=PENDING+125' --compressed
```

replace example (fmt=vtt , return vtt)
```bash
curl 'https://www.youtube.com/api/timedtext?v=isY5MCWPi_0&ei=PW1UZb6tGNXM2roPxd2r2A4&opi=112496729&xoaf=5&hl=en&ip=0.0.0.0&ipbits=0&expire=1700057005&sparams=ip,ipbits,expire,v,ei,opi,xoaf&signature=EB364571027A63836BB97ABE4EF93C73790B56F3.2B2D8D9E9143255F38CF415CBA9960AB619DED87&key=yt8&lang=en-US&fmt=vtt&tlang=' -H 'Host: www.youtube.com' -H 'User-Agent: Crystal' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7' -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' -H 'Accept-Language: en-us,en;q=0.5' -H 'Cookie: ; CONSENT=PENDING+125' --compressed
```

SamantazFox:
Thanks for contributing to invidious :)

PR #4222, opened 2023-11-01
Private Invidious

### LINKED ISSUE #446: private invidious?

rix1337:
I want to see the feature go live, so I created this PR based on https://github.com/iv-org/invidious/issues/446#issuecomment-1342296218
* Update config.cr
* Update before_all.cr

This should resolve / fix #446


unixfox:
Thank you but there is already a pending PR for that: https://github.com/iv-org/invidious/pull/3728 and with the same code but improved.

Issue #4045, opened 2023-08-10
[Bug] Constant 403 errors while trying to load videoplayback


TheEdgeOfRage:
**Describe the bug**
I host my own invidious instance and some videos just won't load on it. The thumbnail loads, but the video itself does not. There's just a spinner and the timeline says 0:00/0:00. This does not happen with all videos, but I haven't spotted a pattern yet.

When looking at the network tab, I can see that the browser is constantly sending requests to `/videoplayback` and getting a 403 status with no response body. It sends about 10 requests per second and doesn't stop until I close the tab. For reference, I use DASH on my instance.

**Steps to Reproduce**

1. Go to https://vid.theedgeofrage.com/watch?v=RBo1Mq1JMh8
2. Try to play video

**Logs**
Logs don't show anything besides the failing request log itself:

```
[...]

PR #4225, opened 2023-11-01
force resolve for everything

unixfox:
Previously, only www.youtube.com was forced resolved, but now all the traffic will be force resolved to whatever IP version the user specify.

This is for being sure that all the proxy to googlevideo.com is being sent in the correct specified IP version. This will help for the issue https://github.com/iv-org/invidious/issues/4045

I did test and this doesn't break anything. When nothing is forced, the default value is used: https://github.com/iv-org/invidious/blob/master/src/invidious/config.cr#L122

### DIFF
--- a/src/invidious/yt_backend/connection_pool.cr
+++ b/src/invidious/yt_backend/connection_pool.cr
@@ -37,7 +37,7 @@ struct YoutubeConnectionPool
         conn.close
         conn = HTTP::Client.new(url)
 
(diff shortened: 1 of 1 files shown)


unixfox:
Confirmed on matrix that it's critical for sending all the traffic in the correct IP address so that the public instance gets unblocked.

syeopite (commented):
LGTM. But this should also probably be included in the stdlib override at `crystal_class_overrides.cr` to force resolve connections that aren't made by `make_client` or the yt pool

SamantazFox (approved):

PR #4229, opened 2023-11-03
Fixed typo on loop not allowing subsciption import from YT

Novocaine85:
Fixing this solved import of subscription from G YT takeout.

### DIFF
--- a/src/invidious/user/imports.cr
+++ b/src/invidious/user/imports.cr
@@ -16,7 +16,7 @@ struct Invidious::User
       rows.each do |row|
         # Limit to 1200
         row_counter += 1
-        break if row_counter > 1_200
+        break if row_counter > 1200
 
         # Channel ID is the first column in the csv export we can't use the header
         # name, because the header name is localized depending on the


unixfox:
@SamantazFox I think it's still the same int in both cases (before and after)?

Novocaine85:
> @SamantazFox I think it's still the same int in both cases (before and after)?

I thought so, but after strugling importing a subscrition file (I get an error everytime i tried) i've tried to edit that file and rebuild and it worked. 

I've no idea why.

absidue:
The underscore is not a typo, it is an allowed character in numbers in many programming languages not just crystal. The code before an after this pull request compiles into the exact same output.

I'm pretty sure it's just a placebo that it worked with this change.

Even the official crystal docs use underscores to make long numbers readable: https://crystal-lang.org/api/1.10.1/Number.html#humanize%28io%3AIO%2Cprecision%3D3%2Cseparator%3D%27.%27%2Cdelimiter%3D%27%2C%27%2C%2A%2Cbase%3D10%2A%2A3%2Csignificant%3Dtrue%2Cprefixes%3AIndexable%3DSI_PREFIXES%29%3ANil-instance-method

unixfox:
That's what I thought. Closing as it's the same result before and after this PR.

SamantazFox:
> @SamantazFox I think it's still the same int in both cases (before and after)?

Sorry, I didn't see that PR before.

Yes, both values are the same. The underscore is added fore legibility, this is a Crystal feature. Rust has that too, for instance.

PR #4249, opened 2023-11-10
Improve depends_on docker-compose

killua99:
Checking the service is up and healthy before start the service to avoid issue first boot.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,9 @@ services:
       timeout: 5s
       retries: 2
     depends_on:
-      - invidious-db
+      invidious-db:
+        condition: service_healthy
+        restart: true
 
   invidious-db:
     image: docker.io/library/postgres:14
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,6 @@ services:
     depends_on:
       invidious-db:
         condition: service_healthy
-        restart: true
 
   invidious-db:
     image: docker.io/library/postgres:14


unixfox:
Thank you for the CI is failing: https://github.com/iv-org/invidious/actions/runs/6823141172/job/18558101386?pr=4249

killua99:
Hi, interesting perhaps is something with the compose version.

I'm going to double check.

https://docs.docker.com/compose/compose-file/05-services/#depends_on

ChunkyProgrammer:
> Thank you for the CI is failing: https://github.com/iv-org/invidious/actions/runs/6823141172/job/18558101386?pr=4249

@unixfox I believe this PR fixes the CI issue https://github.com/iv-org/invidious/pull/4242

killua99:
Oh that's look promising.

But the option restart isn't available on Ubuntu LTS image from the CI, it require a modern docker compose component.

SamantazFox (approved):
looks good to me! I've that deployed on my private instance, it works like a charm!

unixfox:
I'm still not sure if we should use the new format because then the docker-compose file will not work with the older versions or compose.

killua99:
I didn't alter the version requiere for docker compose to work, so version 3 and above (the current version you support) should be fine and able to run it without issue.

Actually, I should update documentation right ?

unixfox:
But
```
depends_on:
      invidious-db:
        condition: service_healthy
```

is invalid in older docker-compose versions. that's what I'm talking about

killua99:
you have set `version: "3"` that's means should be supported, or am I missing something? gonna double check anyway

unixfox:
I'm not entirely sure, but I think this is only supported by the new Compose V2, the old Compose V1 do not support this depends_on syntax.

killua99:
then, perhaps you will like to add some tests for V1 composer ? (which BTW is no longer supported)

I can add tests for composer V1

unixfox:
It is not about tests in the CI but about if we should break existing setups that use Compose V1 or not. Just for the sake of a "better start" of Invidious.

Do note though that the docker-compose.yml file is not being used for production, it's only for development. The docker-compose file that is the most used is located at https://github.com/iv-org/documentation/blob/master/docs/installation.md#docker-compose-method-production

killua99:
> It is not about tests in the CI but about if we should break existing setups that use Compose V1 or not. Just for the sake of a "better start" of Invidious.
> 
> Do note though that the docker-compose.yml file is not being used for production, it's only for development. The docker-compose file that is the most used is located at https://github.com/iv-org/documentation/blob/master/docs/installation.md#docker-compose-method-production

Indeed in a previous comment I mentioned if could I upgrade the documentation (I didn't mention production)

But again docker compose V1 is outdated and without security patches is risky keep /maintained outdated OSS

unixfox:
> Indeed in a previous comment I mentioned if could I upgrade the documentation (I didn't mention production)

Yes everyone can!

> But again docker compose V1 is outdated and without security patches is risky keep /maintained outdated OSS

We tend to support old things, in a sense invidious allows to use YouTube in old devices. So if the functionality is not a dealbreaker we usually try to not add it if it means dropping support for old devices, old software and so on.

unixfox:
> I agree that we should support old browsers, but on the server side, we should try to incentivize the use of updated software. In addition to that, I've personally never seen a distro shipping compose v1/v2, so I don't see what edge case we are covering here.

OK that is a good idea. @killua99 could you submit the same change to the documentation please?

PR #4259, opened 2023-11-13
support for private instances

stonerl:
This is a modification of PR #3728. And addresses #446

Server admins can set the instance to be private. Which means it is only accessible with a registered user account.

The endpoints `/api/v1/popular` and `/api/v1/trending` are whitelisted because some clients expect them to be open.

### DIFF
--- a/src/invidious/routes/before_all.cr
+++ b/src/invidious/routes/before_all.cr
@@ -88,11 +88,11 @@ module Invidious::Routes::BeforeAll
       end
     end
 
-    # TODO: popular and trending are here for clients that require these endpoints to be accessible e.g. Clipious
+    # TODO: popular and trending are whitelisted for clients that require these endpoints to be accessible e.g. Clipious
     # can be removed as soon as those clients can handele these request on private instances
     unregistered_path_whitelist = {
       "/login",
-      "/privacy"
+      "/privacy",
       "/api/v1/stats",
       "/api/v1/popular",
       "/api/v1/trending",
--- a/src/invidious/routes/before_all.cr
+++ b/src/invidious/routes/before_all.cr
@@ -88,12 +88,12 @@ module Invidious::Routes::BeforeAll
       end
     end
 
+    # TODO: popular and trending are here for clients that require these endpoints to be accessible e.g. Clipious
+    # can be removed as soon as those clients can handele these request on private instances
     unregistered_path_whitelist = {
       "/login",
       "/privacy"
       "/api/v1/stats",
-      # TODO: popular and trending are here for clients that require these endpoints to be accessible e.g. Clipious
-      # can be removed as soon as those clients can handele these request on private instances
       "/api/v1/popular",
       "/api/v1/trending",
       "/feed/webhook/v1:",
--- a/src/invidious/routes/before_all.cr
+++ b/src/invidious/routes/before_all.cr
@@ -61,18 +61,6 @@ module Invidious::Routes::BeforeAll
       env.response.headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains; preload"
     end
 
-    return if {
-                "/sb/",
-                "/vi/",
-                "/s_p/",
-                "/yts/",
-                "/ggpht/",
-                "/api/manifest/",
-                "/videoplayback",
-                "/latest_version",
-                "/download",
-              }.any? { |r| env.request.resource.starts_with? r }
-
     if env.request.cookies.has_key? "SID"
       sid = env.request.cookies["SID"].value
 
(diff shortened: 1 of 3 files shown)


unixfox:
What is the exact difference between "private_instance" and "redirect_login"?

stonerl:
> What is the exact difference between "private_instance" and "redirect_login"?

When an instance is set to private, the administrator can choose whether they want to redirect to the login page.

E.g. A user accesses the page `https://my.invidious.instance/`, since `/` is not whitelisted, The user gets redirected to `https://my.invidious.instance/login` when `redirect_login` is set to `true`, if set to `false` the user just sees a white page and does not get redirected. Without `private_instance: true` `redirect_login` has no effect.

`redirect_login: true`
redirect

`redirect_login: false`
no_redirect

unixfox:
Thank you for the feedback with the differences.

But what's the use case for the parameter "private_instance" here? It returns a blank page. How are the users supposed to use the instance if it returns a blank page for the home page?

stonerl:
Foremost. Sorry, for the failing workflows. That was me being stupid …

Regarding your question. Registered users can still access the page via `https://my.invidious.instance/login` since the login route is on the whitelist.

It's more or less: Security through obscurity. People end up on one's instance for whatever reason. But it is not clear from the beginning whether the site is working or not. So they might think, “Oh, it's broken” and go their merry way. I know that this isn't really secure, but it obfuscates the instance at least a little.

unixfox:
> It's more or less: Security through obscurity. People end up on one's instance for whatever reason. But it is not clear from the beginning whether the site is working or not. So they might think, “Oh, it's broken” and go their merry way. I know that this isn't really secure, but it obfuscates the instance at least a little.

Well for me this functionality shouldn't be in invidious. It's something that you can do on the reverse proxy side.

I do understand why the log in functionality is implemented in invidious because you can't do this on the reverse proxy side. But the idea to hide the home page is something so niche that I don't think this should be included into the core of invidious.

rix1337:
Also Security by obscurity will not help against real threats.

The rest of the feature is looking good

stonerl:
Okay, I removed `redirect_login`.

And linting should work now 😆 I ran: `crystal tool format`

syeopite:
The before_all handler isn't called for the error routes. See https://github.com/kemalcr/kemal/issues/663. Is that likely to impact this PR? 

On those endpoints, Invidious will always be open to those without an account.  I'm not sure if that that has any implications.

stonerl:
> The before_all handler isn't called for the error routes. See [kemalcr/kemal#663](https://github.com/kemalcr/kemal/issues/663). Is that likely to impact this PR?
> 
> On those endpoints, Invidious will always be open to those without an account. I'm not sure if that that has any implications.

No, since non-existing routes are not whitelisted, these requests will always be forwarded to the login page.

rix1337:
What further changes are expected here?
Are we ready to merge?

stonerl:
From my side, it is good to go.

syeopite:
The user is being redirected but it seems like the handlers for the various endpoints are still being called in the end. 

Try increase the log level to trace and query the /search endpoint. You'd see that the user got redirected to the /login page but the handler is still called and the query is still sent to YouTube, parsed etc.

stonerl:
> The user is being redirected but it seems like the handlers for the various endpoints are still being called in the end.

Okay, but I assume that is something that needs to be handled in another PR? Since this is an issue not directly related to private instances…

rix1337:
Agreed. Let’s not creep more features into this and instead release.

syeopite:
It very much is directly related and not a feature creep. It is a bug of the way private instances are implemented in this PR.

This PR redirects to /login for all endpoints when the user is not logged in. But the logic for those routes are still being ran anyways.

Not only is this not right for how a redirect should work, this bug:

- Increases the load to what a public instance would see even on private instances as the same logic and requests are being done even when the user is not allowed to access the page

- Increases the chances of Invidious being blocked due to all of the requests getting sent to YouTube on each redirect. Once again even when all the user sees is a /login page.

This needs to be fixed.

stonerl:
I'll take a look at this.

src/invidious/routes/before_all.cr
```diff
@@ -100,6 +88,38 @@ module Invidious::Routes::BeforeAll
...
+    # TODO: popular and trending are whitelisted for clients that require these endpoints to be accessible e.g. Clipious
+    # can be removed as soon as those clients can handele these request on private instances
+    unregistered_path_whitelist = {
+      "/login",
+      "/privacy",
+      "/api/v1/stats",
+      "/api/v1/popular",
+      "/api/v1/trending",
+      "/feed/webhook/v1:",
+      "/api/v1/videos/dQw4w9WgXcQ",
```
rix1337, line 100: @stonerl  why whitelist this one?

unixfox: `/api/v1/videos/dQw4w9WgXcQ` is for the healthcheck in docker I think
`/feed/webhook/v1:` for keeping pubsub working

sigulete:
I'm looking for this patch. Is it going to be merged soon?

unixfox:
> I'm looking for this patch. Is it going to be merged soon?

this comment is still the reason why this PR hasn't been merged yet: https://github.com/iv-org/invidious/pull/4259#issuecomment-1828573067

stonerl:
I'm currently short on time. I'll update the PR as soon as possible.

rix1337:
Activity

unixfox:
We will reopen the PR once https://github.com/iv-org/invidious/pull/4259#issuecomment-1828573067 has been tackled.

PR #4265, opened 2023-11-15
Misc: Fix logic for setting user agent

ChunkyProgrammer:


src/invidious/yt_backend/connection_pool.cr
```diff
@@ -1,6 +1,6 @@
 def add_yt_headers(request)
   if request.headers["User-Agent"] == "Crystal"
-    request.headers["User-Agent"] ||= "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36"
+    request.headers["User-Agent"] = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36"
```
syeopite, line 3: This is intentional. The `||=` operator assigns that user-agent when the key doesn't exist.

ChunkyProgrammer: Wouldn't the if statement prevent that from happening?

syeopite: Ah you're right! Oops. In that case I think the if statement should just be removed. 

There's a couple cases where the user agent is completely empty when the add_yt_headers function is called. And because of that the ff agent would never be added and Crystal probably defaults back to the "Crystal" user agent or an empty one.  

I do wonder if this is one method of how YouTube is detecting Invidious...

syeopite: Also, please replace the `||=` operator with a call to this method instead. https://crystal-lang.org/api/1.10.1/Hash.html#put_if_absent%28key%3AK%2C%26%3AK-%3EV%29%3AV-instance-method 

It should be more performant.

ChunkyProgrammer: I was having some issues with put_if_absent so i didnt include that change

syeopite (approved):

SamantazFox (approved):
Ah, neat, good catch!

unixfox (approved):

SamantazFox:
Thanks!

Issue #301, opened 2019-01-19
Allow to use proxies (HTTP(s) & socks)


dimqua:
This could be useful for those having a self-hosted instance. The idea is about adding an option to specify a proxy server (e.g. tor socks proxy), which can be used when a server are going to make a request to YouTube to fetch the data.

Perflyst:
Maybe it is also possible to proxy the youtube media through the server itself so that the users ip address is hidden to google. If this will be done, maybe also add the option to enable/disable this feature, because some hosters have traffic limit on their servers.

Nutomic:
I honestly thought this is what invidious does. I was a bit surprised that it still connects me directly to youtube. This should definitely be possible, at least with self-hosting.

omarroth:
To be clear, the instance itself proxies requests, and then uses a proxy in another region as necessary to circumvent geo-restrictions (see #92). See my comment [here](https://github.com/omarroth/invidious/issues/453#issuecomment-477574597) for some clarification on what Invidious does.

There appears to be a [socks](https://github.com/wontruefree/socks) shard that should make this possible, I'll have to look more into it to ensure it can be used here.

leonklingele:
Running the invidious binary with `torsocks` should proxy requests through Tor.

rakamanja:
I wish I could have replied to this earlier but our microsoft overlords decided they don't like my ISP.

`torsocks` overwrites calls to `libc` functions. Therefore it only works if your software is using libc. I believe the crystal network library is not implemented on top if `libc`, but I haven't tested nor do I know crystal. Source on that information (second and third paragraphs): https://gitweb.torproject.org/torsocks.git/tree/README.md

If this information is correct, the best way to tunnel the requests of your local instance through tor would be to either:
1. Run it on a VM an route all the VM's traffic through tor;
2. Run invidious under a dedicated user and use iptables to redirect every package from that user through a transparent proxy. In that case you have to make sure that your system is not using your ISP's DNS servers.

kVxQeXZbQAufBnQqA:
It seems this issue has not received much attention. I would like to explore using the [sox](https://github.com/wontruefree/sox) shard, previously called "socks". My main worry with implementing this is that Google generally does not like IP addresses of Tor exit nodes and will either completely block these addresses or result in an endless Captcha loop.

alexw1982:
as a private self-hoster of an instance I would love to run the requests to youtube/google also via a socks proxy. Would be awesome!

[...]

PR #4270, opened 2023-11-17
Add support for using Invidious through a HTTP Proxy

syeopite:
Partially addresses #301

To use just set at least the host and port configuration here in the Invidious configs.
```yaml
##
## Configuration for using a HTTP proxy
##
## If unset, then no HTTP proxy will be used.
## 
http_proxy:
  user:
  password:
  host:
  port:
```

### DIFF
--- a/src/invidious/helpers/crystal_class_overrides.cr
+++ b/src/invidious/helpers/crystal_class_overrides.cr
@@ -18,6 +18,40 @@ end
 class HTTP::Client
   property family : Socket::Family = Socket::Family::UNSPEC
 
+  # Override stdlib to automatically initialize proxy if configured
+  #
+  # Accurate as of crystal 1.10.1
+
+  def initialize(@host : String, port = nil, tls : TLSContext = nil)
+    check_host_only(@host)
+
+    {% if flag?(:without_openssl) %}
+      if tls
+        raise "HTTP::Client TLS is disabled because `-D without_openssl` was passed at compile time"
+      end
+      @tls = nil
+    {% else %}
+      @tls = case tls
+             when true
+               OpenSSL::SSL::Context::Client.new
+             when OpenSSL::SSL::Context::Client
+               tls
+             when false, nil
+               nil
+             end
+    {% end %}
+
+    @port = (port || (@tls ? 443 : 80)).to_i
+
+    self.proxy = make_configured_http_proxy_client() if CONFIG.http_proxy
+  end
+
+  def initialize(@io : IO, @host = "", @port = 80)
+    @reconnect = false
+
+    self.proxy = make_configured_http_proxy_client() if CONFIG.http_proxy
+  end
+
   private def io
     io = @io
     return io if io
--- a/src/invidious/helpers/crystal_class_overrides.cr
+++ b/src/invidious/helpers/crystal_class_overrides.cr
@@ -20,7 +20,7 @@ class HTTP::Client
 
   # Override stdlib to automatically initialize proxy if configured
   #
-  # Accurate as of crystal 1.10.1
+  # Accurate as of crystal 1.12.1
 
   def initialize(@host : String, port = nil, tls : TLSContext = nil)
     check_host_only(@host)
--- a/src/invidious/yt_backend/connection_pool.cr
+++ b/src/invidious/yt_backend/connection_pool.cr
@@ -26,13 +26,15 @@ struct YoutubeConnectionPool
 
   def client(&block)
     conn = pool.checkout
+    # Proxy needs to be reinstated every time we get a client from the pool
     conn.proxy = make_configured_http_proxy_client() if CONFIG.http_proxy
+
     begin
       response = yield conn
     rescue ex
       conn.close
-      conn = HTTP::Client.new(url)
 
+      conn = HTTP::Client.new(url)
       conn.proxy = make_configured_http_proxy_client() if CONFIG.http_proxy
       conn.family = CONFIG.force_resolve
       conn.family = Socket::Family::INET if conn.family == Socket::Family::UNSPEC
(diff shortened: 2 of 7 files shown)


syeopite:
Now that there's "proper" support for HTTP proxies, I suppose Invidious should also be removing its own proxy implementation at `HTTPClient` and `HTTPProxy`

SamantazFox:
Won't that collide with the code in `src/invidious/yt_backend/proxy.cr`?
It might be a good idea to remove all of that proxy code (the IP list hasn't been updated in ages anyway) and let the instance owner use their own.

ms-jpq:
I think it might be helpful to add the http proxy for pulling in reddit comments as well, provided it s not alot more work on top of this.

Reddit just recently IP banned alot of VPS IP ranges, so currently the easiest work around is to tunnel things either via TOR or cloudflare warp

syeopite:
Reddit comments should already be fetched through the proxy in this PR

mk-pmb:
Is this patch meant to also address the install step that says

> Checking player dependencies, this may take more than 20 minutes... If it is stuck, check your internet connection.

?

syeopite:
The part that fetches the videojs dependencies is technically not a part of the main Invidious program so requests in that don't go through the configured proxy.

unixfox:
> The part that fetches the videojs dependencies is technically not a part of the main Invidious program so requests in that don't go through the configured proxy.

Maybe we should? If one configure a proxy he expect all the requests to go through the proxy.

mk-pmb:
Re-downloading the exact same files from npm registry everytime the docker image is rebuilt is an anti-pattern anyway, so we should make a directory where we keep the downloaded files. And once we have that, people behind proxies can pre-populate it with proper npm (#4361).

mk-pmb:
A lot of `fetch-player-dependencies.cr` is just calling OS commands already, so it may be easier to just invoke `wget` or `curl` rather than implementing proxy support in its crystal parts. I even wonder if it might be worth converting the script to bash entirely.

mk-pmb:
As for the PR itself, I think ce4c236046e2448d2f73caa3d03047321c2c49c3 and 1a2760d11021d308a545d113bed29213a415e83d should both be amended into their parent commits.

unixfox:
Please use the edit button, don't triple post...

mk-pmb:
I deleted the off-topic comment and merged the other two. I'll delete this one soon, too; just notifying you so you can reconsider the hiding.

mk-pmb:
I researched a bit more about how to best pass proxy settings to the build process and found that doing it securely (not revealing proxy hostname and credentials in the built image) is more complicated than I had hoped. So build-time proxy stuff should best be left for a 3rd-party project.

@syeopite, I found a few more things that I'd prefer switched around in git history, and some whitespace that seems accidential. Please allow me to propose this [alternative history](https://github.com/mk-pmb/invidious/commits/http-proxy-amended/) for the same effect as your current PR. The last commit is just so you can easily verify that files in both branches are otherwise identical. If you like, feel free to hard-reset your branch to my proposal's 2nd commit.

Does Crystal have an elegant way to use a common function to de-duplicate the `conn.proxy = …` assignments?
A project-wide function to set default config for all HTTP(S) connections might also turn out to be useful if one day we want to configure more aspects than just proxy.

src/invidious/yt_backend/connection_pool.cr
```diff
@@ -26,17 +26,23 @@ struct YoutubeConnectionPool
   end
 
   def client(region = nil, &block)
-    if region
+    if !CONFIG.http_proxy && region
       conn = make_client(url, region)
+      conn.proxy = make_configured_http_proxy_client() if CONFIG.http_proxy
```
ChunkyProgrammer, line 31: This will never run because of line 29

mk-pmb: Good catch. The proxy decision should probably be fully independent from the region decision.

SamantazFox:
Could you please rebase your changes?

src/invidious/helpers/crystal_class_overrides.cr
```diff
@@ -18,6 +18,40 @@ end
 class HTTP::Client
   property family : Socket::Family = Socket::Family::UNSPEC
 
+  # Override stdlib to automatically initialize proxy if configured
```
syeopite, line 21: On second thought, I think it'd be better to just initialize the proxy on each http request rather than override stdlib.

syeopite:
It seems like the upstream library no longer works on Crystal 1.12

ChunkyProgrammer:
@syeopite https://github.com/mamantoha/http_proxy/blob/master/CHANGELOG.md#0102 I think you could update the library to a newer version

syeopite:
Oops! I didn't see the newest update 😅

src/invidious/helpers/crystal_class_overrides.cr
SamantazFox: Yes, It would be better indeed!

src/invidious/helpers/crystal_class_overrides.cr
```diff
@@ -18,6 +18,40 @@ end
...
+  #
+  # Accurate as of crystal 1.10.1
```
SamantazFox, line 23: Does it work with other crystal versions?

syeopite: Yep the initialize function wasn't touched

SamantazFox (approved):

mk-pmb:
I'd say let's get this rolling for people whose proxy can accept insane amounts of connections, and upgrade to the pooling solution as soon as it becomes available.

mooleshacat:
It'd be great to get movement on this. I'd like to see it merged into master. Great work syeopite 👍 

I have compiled syeopite/invidious on the http-proxy branch and tested. It appears to be working.

unixfox:
I vote on merging this very awaited PR. Since the YouTube stream management will soon be dedicated to a separate app, the issue described in https://github.com/iv-org/invidious/pull/4270#issuecomment-2241670792 won't be an issue anymore.

unixfox (approved):

mooleshacat:
> I vote on merging this very awaited PR. Since the YouTube stream management will soon be dedicated to a separate app, the issue described in [#4270 (comment)](https://github.com/iv-org/invidious/pull/4270#issuecomment-2241670792) won't be an issue anymore.

Please do. I have 4 patched instances that each have their own proton-privoxy. Playback works & stable for me (incl. dash) as long as you have sig_helper and some method to update po_tokens. No more problems for me 👍  Merge to master means less patching 👍

SamantazFox:
@Sommerwiesel @mooleshacat thanks for your feedback :)

PR #4276, opened 2023-11-20
API: Remove the `fields` parameter

SamantazFox:
## Rationale

Multiple users have reported that the `fields` parameter is slowing down API response times significantly.

As most API endpoints are already optimized to make as few requests as possible to Youtube, there is no point in limiting the output. Furthermore, the added processing might be part of the memory leak problem (#1438).

In addition, I'm certain that the small increase in data output is not much of an issue when we have a huge video proxy next to it.

## Possible caveats

This might break some people's bash scripts that relied on this parameter to reduce the received JSON output. In my opinion, this is not a bit concern, we can recommend using `jq` for that purpose.


syeopite:
https://github.com/iv-org/invidious/actions/runs/6933226444/job/18858928182?pr=4276

APIHandler is still called

SamantazFox:
@syeopite thanks!

absidue:
Just adding a note here, as a reminder that the docs will need updating after this is merged.

src/invidious/helpers/handlers.cr
```diff
@@ -134,74 +134,6 @@ class AuthHandler < Kemal::Handler
...
-
-    env.response.headers["Access-Control-Allow-Origin"] = "*"
-
-    # Since /api/v1/notifications is an event-stream, we don't want
-    # to wrap the response
-    return call_next env if exclude_match? env
-
-    # Here we swap out the socket IO so we can modify the response as needed
-    output = env.response.output
-    env.response.output = IO::Memory.new
```
ChunkyProgrammer, line 155: Do IO operations need to be explicitly closed to free up memory in crystal? If yes then this PR will most likely fix a memory leak. (Unless kemal closes it)
https://crystal-lang.org/api/1.10.1/IO/Memory.html#close%3ANil-instance-method

src/invidious/helpers/handlers.cr
```diff
@@ -134,74 +134,6 @@ class AuthHandler < Kemal::Handler
...
-  {% for method in %w(GET POST PUT HEAD DELETE PATCH OPTIONS) %}
-  only ["/api/v1/*"], {{method}}
-  {% end %}
-  exclude ["/api/v1/auth/notifications"], "GET"
-  exclude ["/api/v1/auth/notifications"], "POST"
-
-  def call(env)
-    return call_next env unless only_match? env
-
-    env.response.headers["Access-Control-Allow-Origin"] = "*"
```
iBicha, line 147: @SamantazFox This broke access to Playlet web app
```
Access to fetch at 'http://192.168.1.119:8095/api/v1/trending?region=CA&fields=title%2CvideoId%2CvideoThumbnails%2ClengthSeconds%2CviewCount%2Cauthor%2CauthorId%2CpublishedText%2CliveNow%2CisUpcoming%2CpremiereTimestamp' from origin 'http://192.168.1.107:8888' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

iBicha: That's alright, things happen! Thanks

Issue #4002, opened 2023-07-24
[Bug] Letterboxed playlist thumbnails


syeopite:
**Describe the bug**

A recent commit seems to have caused letterboxing on playlist thumbnails. See screenshots below.

**Steps to Reproduce**

1. Go to a channel
2. Click on the playlists tab
3. See letterboxing on all playlist thumbnails 

**Logs**

**Screenshots**

Channel playlists tab:
![letterbox](https://github.com/iv-org/invidious/assets/70992037/8cfabcbf-47f6-4b9b-b664-ab833fa59187)

Search results:
![letterboxed playlists](https://github.com/iv-org/invidious/assets/70992037/8168e18b-d930-49bf-9eaa-70f7f1dabb40)

**Additional context**

SamantazFox:
Probably caused by https://github.com/iv-org/invidious/pull/3763

SamantazFox:
Uh, FF devtools shows that the image is indeed letterboxed. Did YT change something again?
![image](https://github.com/iv-org/invidious/assets/52980486/7a209aa6-4403-4e68-99c1-e678c94196da)

PR #4278, opened 2023-11-21
CSS: Fix thumbnails' aspect ratio to prevent CLS

cornedor:
Closes https://github.com/iv-org/invidious/issues/4002

Static fixed aspect ratio to prevent CLS when thumbnails load.

Old situation:

Screen Recording 2023-11-21 at 08 17 41

New situation:

Screen Recording 2023-11-21 at 08 16 57

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -197,6 +197,7 @@ img.thumbnail {
   display: block; /* See: https://stackoverflow.com/a/11635197 */
   width: 100%;
   object-fit: cover;
+  aspect-ratio: 16 / 9;
 }
 
 .thumbnail-placeholder {


unixfox:
To be fair I actually liked the lazy load of the images. That they load one at a time.
But I do understand that it is strange.

One thing is that I'm not entirely sure if it's how it should be done in CSS.

cornedor:
> To be fair I actually liked the lazy load of the images. That they load one at a time. But I do understand that it is strange.

I think this is because I used different servers for the examples. The images are still lazy loaded, and the change should not have any impact on how images are loaded, just the height of the images before they are loaded.

SamantazFox (approved):

SamantazFox:
That needs to be tested with the new, larger playlist thumbnails (See #4002)

SamantazFox:
It solves #4002, nice!

PR #4284, opened 2023-11-22
Control speed of playback using mousewheel

hikkamorii:
This change will allow user to control playback speed by hovering over "playback rate" button and scrolling on mouse. 

Clicking on the button will still increment speed in pre-defined steps, unless the speed is custom.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -580,11 +580,29 @@ function toggle_fullscreen() {
 }
 
 function increase_playback_rate(steps) {
-    const maxIndex = options.playbackRates.length - 1;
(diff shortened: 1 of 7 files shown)


unixfox:
Nice but I can see this being activated by mistake.

hikkamorii:
I personally never had issues with accidentally changing speed by this, but if that's a concern, I can try adding it as a toggle in preferences.

hikkamorii:
This build issue is unrelated to my changes, see https://github.com/crystal-lang/crystal/commit/2173371b4485db90a750d4a47752b2dab9e067a1

Issue #4235, opened 2023-11-07
[Enhancement] audio codec representation


foxjaw:
**Is your enhancement request related to a problem? Please describe.**

YES. This is related to a problem. See the below screenshot:
![Screenshot_20231030-183141](https://github.com/iv-org/invidious/assets/117489882/076b86b3-0dd7-48cd-888c-9c85788844ee)

YouTube uses two audio codecs **aac** & **opus**. One has **m4a** container, & another's **webm**. If you listen carefully, opus stream with least quality (typically at 35 to 55 kbps) has amazing audio quality compared to the highest aac ones.

### The problem is
In above screenshot, look at the audio quality list. There's no indication of codec type. There's only bitrate. 

**Describe the solution you'd like**

I suggest remove that 5/6 digit number & implement something like m4a128k, webm55k, m4a64k, webm150k, etc. It's very useful.

**Describe alternatives you've considered**

Instead, you can also segregate streams into two categories inside the list. One with m4a & another with webm.

**Additional context**

Related to this, can you also look into #4181 and implement default audio quality setting, like how default video quality exists ?

unixfox:
Related: https://github.com/iv-org/invidious/issues/3012

foxjaw:
@unixfox It's not. #3012 is related to setting a default codec. This enhancement is showing the codec type in the list.

gsuberland:
Also seems like there's a bug here - 135502k would be 135Mbps, which is definitely not correct for an audio stream. It's probably supposed to be 135502 bits per second, i.e. 135.5Kbps.

PR #4287, opened 2023-11-23
Enchanced audio selection

hikkamorii:
This change should address issues #4235 and #2513 

To be exact:

- When listening to audio in listening mode, codec and quality are specified in the list (i.e. "opus @ 64k", "mp4 @ 128k")
- When viewing video in DASH mode, audio quality can be selected based on the bitrate (it's not necessary to have codec listed, as, at least for now, support for Opus is not enabled) (i.e. "64k", "128k", etc.)


unixfox:
May conflict with https://github.com/iv-org/invidious/pull/3620

Issue #4021, opened 2023-07-29
[Bug] Playlist atom feeds return HTTP 200 for non-existent playlists


absidue:
**Describe the bug**
Instead of returning an HTTP 500 status code like the channel atom feed or a 404 like YouTube does for both feeds, Invidious returns an HTTP 200 status code for non-existent playlists, with the content type: `application/atom+xml` but with this body:

```
<!DOCTYPE html>
<html/>
```

Instead of pretending everything is fine, Invidious should either return a 500 with an error response like the channel atom feed or 404 like YouTube does.

**Steps to Reproduce**
https://test.invidious.io/feed/playlist/ZZZ

absidue:
Not stale

PR #4298, opened 2023-11-30
RSS: Channel + Playlist improvements

ChunkyProgrammer:
Channel feed: don't make an extra request, deprecate thumbnail field

Playlist feed: return 404 when playlist isn't found

As RSS clients can make a lot of requests pretty quickly, I removed fetching the channel to get the thumbnail for the RSS feed.
The rest of the fixes are mainly so FreeTube can detect if a channel doesn't exist via RSS

Related FreeTube PR: https://github.com/FreeTubeApp/FreeTube/pull/4407

closes: https://github.com/iv-org/invidious/issues/4021

### DIFF
--- a/src/invidious/routes/feeds.cr
+++ b/src/invidious/routes/feeds.cr
@@ -146,7 +146,11 @@ module Invidious::Routes::Feeds
     env.response.headers["Content-Type"] = "application/atom+xml"
     env.response.content_type = "application/atom+xml"
 
-    ucid = env.params.url["ucid"]
+    if env.params.url["ucid"].matches?(/^[\w-]+$/)
+      ucid = env.params.url["ucid"]
+    else
+      return error_atom(400, InfoException.new("Invalid channel ucid provided."))
+    end
 
     params = HTTP::Params.parse(env.params.query["params"]? || "")
 
--- a/src/invidious/routes/feeds.cr
+++ b/src/invidious/routes/feeds.cr
@@ -309,8 +309,9 @@ module Invidious::Routes::Feeds
     end
 
     response = YT_POOL.client &.get("/feeds/videos.xml?playlist_id=#{plid}")
-    document = XML.parse(response.body)
+    return error_atom(404, NotFoundException.new("Playlist does not exist.")) if response.status_code == 404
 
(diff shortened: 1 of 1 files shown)


src/invidious/routes/feeds.cr
```diff
@@ -143,32 +143,21 @@ module Invidious::Routes::Feeds
   # RSS feeds
 
   def self.rss_channel(env)
-    locale = env.get("preferences").as(Preferences).locale
-
     env.response.headers["Content-Type"] = "application/atom+xml"
     env.response.content_type = "application/atom+xml"
 
     ucid = env.params.url["ucid"]
```
SamantazFox, line 149: Please add a sanity check on this user input, like `.matches?(/^\w+$/)`. You're using it directly in the generated response, which could cause an XSS vulnerability.

absidue: As channel IDs can also contain `-` (just like video IDs and the protobuf payloads, they use use base64url), this would be a more appropriate regex `.matches?(/^[\w-]+$/)` :).

ChunkyProgrammer:
This PR is working and still applicable.

SamantazFox (changes requested):

src/invidious/routes/feeds.cr
```diff
@@ -200,30 +193,33 @@ module Invidious::Routes::Feeds
...
         "xml:lang": "en-US") do
         xml.element("link", rel: "self", href: "#{HOST_URL}#{env.request.resource}")
-        xml.element("id") { xml.text "yt:channel:#{channel.ucid}" }
-        xml.element("yt:channelId") { xml.text channel.ucid }
-        xml.element("icon") { xml.text channel.author_thumbnail }
-        xml.element("title") { xml.text channel.author }
-        xml.element("link", rel: "alternate", href: "#{HOST_URL}/channel/#{channel.ucid}")
+        xml.element("id") { xml.text "yt:channel:#{ucid}" }
+        xml.element("yt:channelId") { xml.text ucid }
+        xml.element("icon") { xml.text "" }
```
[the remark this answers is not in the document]
ChunkyProgrammer: Done! :smiley:

PR #4318, opened 2023-12-07
Add support for displaying channel watermark/favicon on videos

syeopite:
watermark

Since this can be considered a user hostile feature, the icon is fully togglable within the preferences. 

Closes channel favicon component of #2395

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -824,7 +824,7 @@ default_user_preferences:
   #save_player_pos: false
 
   ##
-  ## Show the channel's watermark on the bottom-rght corner of videos
+  ## Show the channel's watermark on the bottom-right corner of videos
   ##
   ## Accepted values: true, false
   ## Default: false
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -823,6 +823,15 @@ default_user_preferences:
   ##
   #save_player_pos: false
 
+  ##
+  ## Show the channel's watermark on the bottom-rght corner of videos
+  ##
+  ## Accepted values: true, false
+  ## Default: false
+  ##
+  #show_channel_watermark: true
+
+
   # -----------------------------
   #  Subscription feed
   # -----------------------------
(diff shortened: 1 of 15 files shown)


src/invidious/config.cr
```diff
@@ -44,6 +44,7 @@ struct ConfigPreferences
   property vr_mode : Bool = true
   property show_nick : Bool = true
   property save_player_pos : Bool = false
+  property show_channel_watermark : Bool = true
```
ChunkyProgrammer, line 47:
```suggestion
  property show_channel_watermark : Bool = false
```

I believe this is supposed to be false?

config/config.example.yml
```diff
@@ -823,6 +823,15 @@ default_user_preferences:
   ##
   #save_player_pos: false
 
+  ##
+  ## Show the channel's watermark on the bottom-right corner of videos
+  ##
+  ## Accepted values: true, false
+  ## Default: false
+  ##
+  #show_channel_watermark: true
```
ChunkyProgrammer, line 832:
```suggestion
  #show_channel_watermark: false
```

This too, unless you wanted the default to be true?

syeopite: Yeah those are intentional. I'd like for it to be enabled by default with an option to disable it for those who don't want it.

But if enough people want it be disabled by default, I'll change that.

ChunkyProgrammer: @syeopite ah okay, i wasnt sure if it was intentional or not as it says `Default: false` on line 830 in thia file (which would need to change in that case)

syeopite: Oops missed that. Thanks!

SamantazFox:
I'm not really sure if that's a necessary "feature". As far as I know, it was never requested to us (I can't find a related issue) and it adds a lot of code, for something that might be left disabled.

As overlays goes, I'd prefer to see the end screen overlays (with a toggle in the preferences).

PR #4321, opened 2023-12-07
Use HTTP::Client directly for automatic redirects

syeopite:
The client created by `make_client` is affected by the `force_resolve` configuration option. However, as `api.invidious.io` does not support ipv6, it leads to an error whenever `force_resolve` is set to only use ipv6.

The normal `HTTP::Client` however is not affected by the force_resolve configuration and therefore remains unaffected

### DIFF
--- a/src/invidious/helpers/utils.cr
+++ b/src/invidious/helpers/utils.cr
@@ -325,7 +325,7 @@ end
 
 def fetch_random_instance
   begin
-    instance_api_client = make_client(URI.parse("https://api.invidious.io"))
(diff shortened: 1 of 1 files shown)


absidue:
Probably worth adding a comment so that people know why it doesn't use `make_client` in the future and don't revert it back.

unixfox:
At first sight, seems great. Will need to be tested if this does not break IPv6 only setups.

HTTP builtin in Crystal should support dual stack so it should work fine.

syeopite:
`api.invidious.io` now supports IPv6

PR #4335, opened 2023-12-15
Add parameter to disable `force_resolve` in `make_client`

syeopite:
Invidious depends on external websites for various parts of its functionality, such as Reddit for Reddit comments and textcatpcha for well text captchas.

The problem however is that some of these services do not support IPv6 and as such when Invidious is forcing all requests from the client made by `make_client` to use IPv6, requests to these services fail and in turn the corresponding features in Invidious fails as well.

### DIFF
--- a/src/invidious/routes/video_playback.cr
+++ b/src/invidious/routes/video_playback.cr
@@ -42,7 +42,7 @@ module Invidious::Routes::VideoPlayback
       headers["Range"] = "bytes=#{range_for_head}"
     end
 
-    client = make_client(URI.parse(host), region)
+    client = make_client(URI.parse(host), region, force_resolve = true)
     response = HTTP::Client::Response.new(500)
     error = ""
     5.times do
@@ -57,7 +57,7 @@ module Invidious::Routes::VideoPlayback
           if new_host != host
             host = new_host
             client.close
-            client = make_client(URI.parse(new_host), region)
+            client = make_client(URI.parse(new_host), region, force_resolve = true)
           end
 
           url = "#{location.request_target}&host=#{location.host}#{region ? "&region=#{region}" : ""}"
@@ -71,7 +71,7 @@ module Invidious::Routes::VideoPlayback
         fvip = "3"
 
         host = "https://r#{fvip}---#{mn}.googlevideo.com"
-        client = make_client(URI.parse(host), region)
+        client = make_client(URI.parse(host), region, force_resolve = true)
       rescue ex
         error = ex.message
       end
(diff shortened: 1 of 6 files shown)


unixfox (approved):

SamantazFox (approved):

src/invidious/yt_backend/connection_pool.cr
```diff
@@ -59,9 +59,14 @@ struct YoutubeConnectionPool
   end
 end
 
-def make_client(url : URI, region = nil)
+def make_client(url : URI, region = nil, force_resolve : Bool = true)
```
unixfox, line 62: Shouldn't this be set to `force_resolve : Bool = false`? We do not want force_resolve by default.

unixfox (changes requested):

syeopite:
Should be disabled by default now.  This should be tested to ensure that I've explicitly enabled force resolve on all YouTube requests.

src/invidious/routes/video_playback.cr
```diff
@@ -42,7 +42,7 @@ module Invidious::Routes::VideoPlayback
       headers["Range"] = "bytes=#{range_for_head}"
     end
 
-    client = make_client(URI.parse(host), region)
+    client = make_client(URI.parse(host), region, force_resolve = true)
```
unixfox, line 45: Same for all the `force_resolve = true` or I am missing something? Why is this set to true?

unixfox (changes requested):

src/invidious/routes/video_playback.cr
syeopite: That's a request to googlevideos so it should go through IPv6

unixfox: No. One should still be able to choose the IP protocol.

syeopite: Oops I messed up on my explanation there. Setting `force_resolve = true` just means that the request will be affected by the force_resolve option in the configs. 

https://github.com/iv-org/invidious/blob/4f9332f652eca71bb38c0079fbd3c23003805ced/src/invidious/yt_backend/connection_pool.cr#L66-L68

Which if I'm not mistaken is what we want for requests to YouTube and Google to circumvent blockage

As long as the config option is unset then those request can freely go through either IPv4 and IPv6.

unixfox: Ha ok I did not understood the code like this.

SamantazFox (approved):

PR #4339, opened 2023-12-16
Improve playback statistics

syeopite:
- Latches previous playback stats object to ensure that Invidious won't ever return an empty object for that endpoint (unless the process has just started)
 - Increment `successfulRequests` when YouTube returns the correct video in InnerTube

### DIFF
--- a/src/invidious/jobs/statistics_refresh_job.cr
+++ b/src/invidious/jobs/statistics_refresh_job.cr
@@ -27,6 +27,12 @@ class Invidious::Jobs::StatisticsRefreshJob < Invidious::Jobs::BaseJob
     "playback" => {} of String => Int64 | Float64,
   }
 
+  # Latches the playback success stats from before statistics gets refreshed
+  # Used to ensure that the object won't get reset back to an empty object
+  LATCHED_PLAYBACK_STATS = {
+    "playback" => {} of String => Int64 | Float64,
+  }
+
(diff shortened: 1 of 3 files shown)


unixfox (approved):

src/invidious/jobs/statistics_refresh_job.cr
```diff
@@ -65,6 +71,7 @@ class Invidious::Jobs::StatisticsRefreshJob < Invidious::Jobs::BaseJob
     }
 
     # Reset playback requests tracker
+    LATCHED_PLAYBACK_STATS["playback"] = STATISTICS["playback"].as(Hash(String, Int64 | Float64))
     STATISTICS["playback"] = {} of String => Int64 | Float64
```
SamantazFox, line 75: Why not just moving this line (the reset) to `load_initial_stats` and then never touch that element again?

syeopite: `load_initial_stats` is only called once and we need to latch the old playback stats each time the statistics is refreshed.

https://github.com/iv-org/invidious/blob/6a88c098f08e54e696b87e3aa478637afdfe4101/src/invidious/jobs/statistics_refresh_job.cr#L41-L49

unixfox:
Closing as playback tracking is now in companion.

PR #4346, opened 2023-12-20
Ensure IO is properly closed when importing NewPipe subscriptions

ChunkyProgrammer:
File.open accepts a second argument which will close the initial stream (`IO::Memory.new(body)`) when the File stream closes.

I also made it so that the code is a bit more safe with closing connections. ex: if an error occurs. This may also fix some potential memory leaks

### DIFF
--- a/src/invidious/user/imports.cr
+++ b/src/invidious/user/imports.cr
@@ -291,40 +291,38 @@ struct Invidious::User
 
     def from_newpipe(user : User, body : String) : Bool
       Compress::Zip::File.open(IO::Memory.new(body), true) do |file|
-        file.entries.each do |entry|
-          entry.open do |file_io|
-            next if entry.filename != "newpipe.db"
+        entry = file.entries.find { |file_entry| file_entry.filename == "newpipe.db" }
+        return false if entry.nil?
+        entry.open do |file_io|
+          # Ensure max size of 4MB
+          io_sized = IO::Sized.new(file_io, 0x400000)
 
-            # Ensure max size of 4MB
-            io_sized = IO::Sized.new(file_io, 0x400000)
-
-            begin
-              temp = File.tempfile(".db") do |tempfile|
-                begin
-                  File.write(tempfile.path, io_sized.gets_to_end)
-                rescue
-                  return false
-                end
-
-                DB.open("sqlite3://" + tempfile.path) do |db|
-                  user.watched += db.query_all("SELECT url FROM streams", as: String)
-                    .map(&.lchop("https://www.youtube.com/watch?v="))
-
-                  user.watched.uniq!
-                  Invidious::Database::Users.update_watch_history(user)
-
-                  user.subscriptions += db.query_all("SELECT url FROM subscriptions", as: String)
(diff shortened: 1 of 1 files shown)


src/invidious/user/imports.cr
```diff
@@ -290,42 +290,38 @@ struct Invidious::User
...
-              return false
-            end
-
-            db = DB.open("sqlite3://" + tempfile.path)
+            # Ensure max size of 4MB
+            io_sized = IO::Sized.new(file_io, 0x400000)
 
-            user.watched += db.query_all("SELECT url FROM streams", as: String)
-              .map(&.lchop("https://www.youtube.com/watch?v="))
+            temp = File.tempfile(".db") do |tempfile|
```
ChunkyProgrammer, line 301: Would we maybe want to randomly generate part of the file file's name? (ex: .db-{UUID}) I imagine this might cause some issues if invidious starts to use Concurrency

SamantazFox: Yes, that might be a huge problem! The ideal option would be ti use in-memory databases and use [`sqlite3_serialize`](https://www.sqlite.org/c3ref/serialize.html), but that requires to make a PR to [crystal-sqlite3](https://github.com/crystal-lang/crystal-sqlite3)

ChunkyProgrammer: Yes, that would be even better! I imagine there might need to be a PR made here as well (if serialization is supported by other databases): [crystal-db](https://github.com/crystal-lang/crystal-db)

syeopite: In memory-databases are already supported in crystal-sqlite3 like so:

```crystal
DB.open "sqlite3::memory" do | db |
  db.exec "create table contacts (name text, age integer)"
  db.exec "insert into contacts values (?, ?)", "John Doe", 30
end
```

The only thing that needs to be added is a way to actually access an in-memory database.

There's no need for a PR directly imo, an issue would be enough.

SamantazFox (approved):

src/invidious/user/imports.cr
SamantazFox: Done: https://github.com/crystal-lang/crystal-sqlite3/issues/95

src/invidious/user/imports.cr
```diff
@@ -290,42 +290,38 @@ struct Invidious::User
...
-            Invidious::Database::Users.update_subscriptions(user)
+                user.subscriptions.uniq!
+                user.subscriptions = get_batch_channels(user.subscriptions)
 
-            db.close
-            tempfile.delete
+                Invidious::Database::Users.update_subscriptions(user)
+              end
+            end
+            temp.delete
```
SamantazFox, line 324: I realized that the file will not be deleted in the case of a database error (i.e if `DB.open` raises an exception). You'll need to wrap the entire thing in a `begin` / (`rescue`) / `ensure`.

ChunkyProgrammer: Done :smile:

syeopite:
Thank you for your work here! This should help make Invidious more stable on large instances 🎉

Issue #4190, opened 2023-10-21
[Feature request] Automatic migration system


broquemonsieur:
**Is your feature request related to a problem? Please describe.**

Since 2021, a host of PRs have been blocked because there is no way to migrate the currently-running Invidious instances' database to the database proposed by the PRs.

**Describe the solution you'd like**

A two-part solution
1. Establish a protocol: 

- For each PR, determine the following

   - For each new table or column, decide what the default and initial values are

2. A script to carry out the migration. Each script will be coded on a per PR basis.

**Describe alternatives you've considered**

Ask all GitHub contributors to include a script with their PR should it contain database changes

**Additional context**

We can unblock some of the best that the Invidious community has to offer! #4032 #2653 #2469 #2254 not to mention the countless future PRs that can continue keeping Invidious the game-changer that it is today!

unixfox:
Related: https://github.com/iv-org/invidious/issues/2620

SamantazFox:
For the record, there is already the required logic for migrations in invidious' code. Now the problem is mostly "how to make everyone use it, without hurdles?"

In essence, you have a bunch of instance who deployed their DB using the old `.sql` scripts located in `config/sql`; I want them to transition to that new migration system/process before merging any PR that requires migrations.

For people not using docker, this is trivial: save the DB, run the migration, check integrity, and you're good to go.

However, for docker, it's a bit more involved. We have to provide clear instructions on how to backup the DB and restore it in case of a failure, we have to decide whether the migration is automatically ran (so automatic backups?) and probably other things.

I won't lie, containers are still quite new to me (I started using docker only because invidious relies on it), and I haven't had the time to look deeper into that subject... Help is appreciated :D

broquemonsieur:
I am offering a bounty of 40 USD worth of BTC

unixfox:
Ideas from matrix:
Basically one can run multiple invidious processes for the same postgresql database. Running automatic migration may break their database.
So I had the idea to detect if multiple postgresql clients not related to the main invidious process would be detected. If detected we do not do the automatic migration.

And for the backup we could copy the schema "public" to another schema. Only when the database is less than 10GB in order to avoid filling up the disk, if more than 10GB we abort the automatic migration.

broquemonsieur:
I have successfully enabled backup-creation whenever new migrations are detected: https://github.com/iv-org/invidious/compare/master...broquemonsieur:invidious:invidious_migrations?expand=1

Before:
 
[...]

PR #4359, opened 2023-12-28
Invidious migrations

broquemonsieur:
https://github.com/iv-org/invidious/issues/4190#issuecomment-1866020412

### DIFF
--- a/src/invidious/database/migrator.cr
+++ b/src/invidious/database/migrator.cr
@@ -10,9 +10,17 @@ class Invidious::Database::Migrator
     versions = load_versions
 
     ran_migration = false
+    backed_up = false
     load_migrations.sort_by(&.version)
       .each do |migration|
         next if versions.includes?(migration.version)
+
+      
+        if !backed_up
+          puts "New migration(s) found: creating database backup"
+          back_up_database
+          backed_up = true
+        end  
 
         puts "Running migration: #{migration.class.name}"
         migration.migrate
@@ -46,4 +54,33 @@ class Invidious::Database::Migrator
       )
     SQL
   end
+
+  private def back_up_database
+    table_names_request = <<-SQL
+      SELECT tablename FROM pg_catalog.pg_tables
+      WHERE schemaname = 'public'
+    SQL
+    
+    table_names = @db.query_all(table_names_request, as: String)
+
+    table_names.try &.each do |name|
+      copy_table(name)
+    end  
+  end  
+
+  private def copy_table(table_name : String)
+    @db.exec <<-SQL
+      CREATE TABLE IF NOT EXISTS backup.#{table_name} (
+        id bigserial PRIMARY KEY
+      )
+    SQL
+
+    @db.exec("DROP TABLE backup.#{table_name}")
+  
+    @db.exec <<-SQL
+      SELECT * INTO backup.#{table_name}
+      FROM public.#{table_name}
+    SQL
+  end  
+
 end
--- a/src/invidious/database/migrator.cr
+++ b/src/invidious/database/migrator.cr
@@ -15,12 +15,11 @@ class Invidious::Database::Migrator
       .each do |migration|
         next if versions.includes?(migration.version)
 
-      
         if !backed_up
           puts "New migration(s) found: creating database backup"
           back_up_database
           backed_up = true
-        end  
+        end
 
         puts "Running migration: #{migration.class.name}"
         migration.migrate
@@ -60,13 +59,13 @@ class Invidious::Database::Migrator
       SELECT tablename FROM pg_catalog.pg_tables
       WHERE schemaname = 'public'
     SQL
-    
+
     table_names = @db.query_all(table_names_request, as: String)
 
     table_names.try &.each do |name|
       copy_table(name)
-    end  
-  end  
+    end
+  end
 
   private def copy_table(table_name : String)
     @db.exec <<-SQL
@@ -76,11 +75,10 @@ class Invidious::Database::Migrator
     SQL
 
     @db.exec("DROP TABLE backup.#{table_name}")
-  
+
     @db.exec <<-SQL
       SELECT * INTO backup.#{table_name}
       FROM public.#{table_name}
     SQL
-  end  
-
+  end
 end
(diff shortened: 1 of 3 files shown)


src/invidious/database/migrator.cr
```diff
@@ -46,4 +53,32 @@ class Invidious::Database::Migrator
...
+  end
+
+  private def copy_table(table_name : String)
+    @db.exec <<-SQL
+      CREATE TABLE IF NOT EXISTS backup.#{table_name} (
+        id bigserial PRIMARY KEY
+      )
+    SQL
+
+    @db.exec("DROP TABLE backup.#{table_name}")
```
SamantazFox, line 77: The problem I see with that is that it might nuke a perfectly valid backup if invidious is interrupted mid-migration and the user tries to run it again immediately after.

broquemonsieur: so there's stage 1,2,3. the migration can get interrupted pre-1,pre-2,pre-3 & post-3.

pre-1 : The migration wouldn't be run (hopefully upon re-attempting migration the migration is question will still be queued up)

pre-2 : At this point, there exists a table backup.cool_table - it is empty. An interuption now, would leave a empty table that'd be re-populated upon re-attempt

pre-3 : at this point, there is a backup.cool_table and a lack of backup.cool_table (this latter one being cool_tables version right before migration). At this point, critical data is destroyed and we must replenish it ASAP - an interuption now would be fatal

Ok - I see what you're saying, what if I created a copy of backup.cool_table (Pre-Migration's Version) earlier on?

src/invidious/database/migrator.cr
```diff
@@ -46,4 +53,32 @@ class Invidious::Database::Migrator
...
+      copy_table(name)
+    end
+  end
+
+  private def copy_table(table_name : String)
+    @db.exec <<-SQL
+      CREATE TABLE IF NOT EXISTS backup.#{table_name} (
+        id bigserial PRIMARY KEY
+      )
+    SQL
```
broquemonsieur, line 75: stage 1

src/invidious/database/migrator.cr
broquemonsieur, line 77: stage 2

src/invidious/database/migrator.cr
```diff
@@ -46,4 +53,32 @@ class Invidious::Database::Migrator
...
+
+    @db.exec("DROP TABLE backup.#{table_name}")
+
+    @db.exec <<-SQL
+      SELECT * INTO backup.#{table_name}
+      FROM public.#{table_name}
+    SQL
```
broquemonsieur, line 82: stage 3

src/invidious/database/migrator.cr
```diff
@@ -46,4 +53,32 @@ class Invidious::Database::Migrator
       )
     SQL
   end
+
+  private def back_up_database
+    table_names_request = <<-SQL
+      SELECT tablename FROM pg_catalog.pg_tables
+      WHERE schemaname = 'public'
+    SQL
```
SamantazFox, line 61: Also, does this copy views, data types, and other database elements?

src/invidious/database/migrator.cr
broquemonsieur: The following is based off the assumption that former broquemonsieur would write SQL that looks redundant/hacky but actually is doing the right thing due to SQL's odd and vague limitations where you cannot do something that should be simple or the only way to achieve a simple operation (that seemingly should be a one-liner) is the combo of two ops (a.k.a. two-liner). Present day broquemonsieur has failing memory but can vouch for this ^ :)
---
Meet Sonic, a new Invidious host. At this point in her users have a few playlists saved and Sonic has not migrated her DB...ever. Her users playlists look like this:

| title      | author      |
| ------------- | ------------- |
| Mark Hamill Moments | kilowattwomxn |
| Awesome Trailers | movieDad |

Sonic migrates her DB:
Stage 1 - Creates a table `backup.playlists`. It is empty
Stage 2 - This op done nothing (it fails silently). (If Sonic were to have a backup of her users' playlists already there, it would be destroyed now to make way for a snapshot of her user's present-day playlists) tl;dr This stages deletes the prior version of  `backup.playlists`, not the one you just read out in Stage 1 (Yes, SQL forums agree this is weird that SQL has to be written this way lol)
Stage 3 - This op populates  `backup.playlists` with

| title      | author      |
| ------------- | ------------- |
| Mark Hamill Moments | kilowattwomxn |
| Awesome Trailers | movieDad |
---
Meet Amy, a jaded Invidious host who already has a backup for her users' playlists.

Amy migrates her DB:
you get the idea :)
---

That should clear it up (mainly I was trying to explain for my recollection to myself lol). But [basically](https://github.com/iv-org/invidious/pull/4359#discussion_r1685807683), the system is vulnerable . At time this comment is sent, I begin efforts to prevent memory loss given an asteroid strike at any point in time

src/invidious/database/migrator.cr
broquemonsieur: That SQL query retrieves only table names from the pg_tables view in PostgreSQL. It does not include views, data types, or other database elements.

Views: Views are stored in the pg_views catalog.
Data Types: Data types are found in the pg_type catalog.
Other Database Elements: Other elements, such as indexes or sequences, are in different catalog tables like pg_indexes and pg_sequences.

If you need to include views or other database elements, additional queries targeting the respective catalog tables would be required.

src/invidious/database/migrator.cr
broquemonsieur: Hi! Reposting back from my coding session. Now, we only destroy the "second most recent" backup. Whenever an asteroid strikes, the "most recent" backup will always be preserved.

broquemonsieur: Asteroid strikes are useless analogies. 
Meet Dale, he is chilling in Mexican beach propping a MacBook Air on pool tube. He is SSH'd to his Invidious server and tries to do DB migration - a major part of which is DB backup.

The hotel Wifi could go out, the MacBook could get wet or the Server itself could shutdown at any point during this critical time and **still** Dale's users would not lose 

1. "most recent" backup 
2. "2nd most recent" backup
3. regular DB tables

Issue #4362, opened 2023-12-30
[Bug] (DB::PoolRetryAttemptsExceeded) -- self hosted using docker-compose.yaml config and containers


zestysoft:
Hey all, since [this issue](https://github.com/iv-org/invidious/issues/1582) is closed, I'm creating a new issue in an attempt to further help to troubleshoot this.

**Describe the bug**
Unhandled exception in spawn:  (DB::PoolRetryAttemptsExceeded)
  from lib/db/src/db/query_methods.cr:46:7 in 'query:args'
  from lib/db/src/db/query_methods.cr:61:12 in 'begin'
  from src/invidious/jobs.cr:37:15 in '->'
  from /usr/share/crystal/src/fiber.cr:146:11 in 'run'
  from ???

**Steps to Reproduce**
Start the latest invidious container

**Logs**
Postgresql:
```
2023-12-30 09:00:50.534 UTC [1] LOG:  database system is ready to accept connections
2023-12-30 09:00:59.868 UTC [40] LOG:  connection received: host=[local]
2023-12-30 09:00:59.869 UTC [40] LOG:  connection authorized: user=kemal database=invidious application_name=pg_isready
2023-12-30 09:00:59.873 UTC [40] LOG:  disconnection: session time: 0:00:00.005 user=kemal database=invidious host=[local]
2023-12-30 09:01:09.877 UTC [47] LOG:  connection received: host=[local]
2023-12-30 09:01:09.877 UTC [47] LOG:  connection authorized: user=kemal database=invidious application_name=pg_isready
2023-12-30 09:01:09.879 UTC [47] LOG:  disconnection: session time: 0:00:00.002 user=kemal database=invidious host=[local]
2023-12-30 09:01:19.869 UTC [54] LOG:  connection received: host=[local]
2023-12-30 09:01:19.870 UTC [54] LOG:  connection authorized: user=kemal database=invidious application_name=pg_isready
2023-12-30 09:01:19.871 UTC [54] LOG:  disconnection: session time: 0:00:00.002 user=kemal database=invidious host=[local]
2023-12-30 09:01:29.868 UTC [62] LOG:  connection received: host=[local]
2023-12-30 09:01:29.869 UTC [62] LOG:  connection authorized: user=kemal database=invidious application_name=pg_isready
2023-12-30 09:01:29.870 UTC [62] LOG:  disconnection: session time: 0:00:00.001 user=kemal database=invidious host=[local]
[...]

PR #4366, opened 2023-12-31
Limit jobs to 5 and wrap in exception handler to release

zestysoft:
@unixfox I've narrowed down the edits to this.  If I revert these changes I see the errors again.

#4362

### DIFF
--- a/src/invidious/jobs.cr
+++ b/src/invidious/jobs.cr
@@ -1,5 +1,6 @@
 module Invidious::Jobs
-  JOBS = [] of BaseJob
+  JOBS      = [] of BaseJob
+  SEMAPHORE = ::Channel(Nil).new(5) # Max 5 jobs running at once
 
   # Automatically generate a structure that wraps the various
   # jobs' configs, so that the following YAML config can be used:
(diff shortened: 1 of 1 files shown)


SamantazFox:
Can you please explain what this code does?

zestysoft:
> Can you please explain what this code does?

@SamantazFox I've added inline comments.

SamantazFox:
Oooh, okay, I see what you did here!

For your information, Jobs in invidious aren't scheduled in any way. Each job is running in it's own "fiber" (co-routine), and is basically a `While True` loop, with a delay (`sleep`) on each iteration. This sleep allows the internal Crystal scheduler to jump between fibers. And the role of `job.begin` is to spawn that fiber, and exiting right away.

zestysoft:
So I started with this purely to help in troubleshooting the issue, but somehow after implementing this change, I could not and still have not reproduced the issue.  If I revert the change, the issue comes back though, so this is reproducible.

Btw, I'm not interested in the bounty at all (assuming that's even on the table).

zestysoft:
The details in the issue I created shows that invidious was only creating a few connections to the DB, but that, for reasons still unknown, had something similar to jobs locking up (race condition?) until it got to the point where it just sat there not talking to the DB at all.  It def wasn't using the default 100 connections that the DB was configured for, and the DB wasn't overwhelmed with requests -- it had answers waiting for invidious to pick up that it just wasn't.

SamantazFox:
> So I started with this purely to help in troubleshooting the issue, but somehow after implementing this change, I could not and still have not reproduced the issue. If I revert the change, the issue comes back though, so this is reproducible.

I think that it might have to do with the error handling rather than the semaphore part.
Can you test different variations (no semaphore, no rescuing, less/more channels)?

SamantazFox (changes requested):

PR #4376, opened 2024-01-06
Stats: Fix two swapped function names

vojkovic:
The function names 'count_users_active_6m' and 'count_users_active_1m' have been swapped.


SamantazFox (approved):
Thanks, good catch!

unixfox (approved):

PR #4402, opened 2024-01-21
Enable authenticating via reverse proxy headers

corvec:
This feature allows an instance to be configured such that, if Reverse Proxy Auth is present, the user will be logged into the account matching the email address present in the configured header (by default, `'Remote-Email'`, which is used by Authelia) upon navigating to the `/login` page.

It also supports automatically registering users who don't yet have accounts, as well as using the username and a supplied domain to build an email instead of using the email header. For example, if you configured it as follows:

```yaml
        reverse_proxy_auth_enabled: true
        reverse_proxy_registration_enabled: true
        reverse_proxy_auth_header: Remote-User
        reverse_proxy_email_header: Remote-Email
        reverse_proxy_auth_domain: noreply.github.com
        reverse_proxy_auth_prefer_user: true
```

and the request had the following headers:

```json
{
  "Remote-User": "corvec"
  "Remote-Email": ""
}
```

then it would log me in as `"corvec@noreply.github.com"`. If that account didn't exist, it would create it, setting my password to a randomly generated max-length string.

I've tested this with Traefik and Authelia using the above config (with my own domain) using Docker Compose as part of an existing stack. It works well for me but I haven't tested it with other configurations - e.g., KeyCloak, Authentik, nginx, Caddy, etc..

### DIFF
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -113,6 +113,13 @@ class Config
   # Enable the user notifications for all users
   property enable_user_notifications : Bool = true
 
+  # Enable logging in via headers that are set by your reverse proxy
+  property reverse_proxy_auth_enabled : Bool = false
+  # Enable auto-registration when logging in with reverse proxy auth
+  property reverse_proxy_registration_enabled : Bool = false
+  # Header that will contain the username, email address,
+  property reverse_proxy_auth_header : String = "Remote-User"
+
   # URL to the modified source code to be easily AGPL compliant
   # Will display in the footer, next to the main source code link
   property modified_source_code_url : String? = nil
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -118,7 +118,20 @@ class Config
   # Enable auto-registration when logging in with reverse proxy auth
   property reverse_proxy_registration_enabled : Bool = false
   # Header that will contain the username, email address,
-  property reverse_proxy_auth_header : String = "Remote-User"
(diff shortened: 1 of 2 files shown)


unixfox:
Would be more in favor of https://github.com/iv-org/invidious/pull/3164 instead.

corvec:
@unixfox thanks for looking over it. I checked out #3164 and agree that I'd like to see it merged as well.

For my purposes either this or #3164 would have worked, but using the headers does have some advantages over OAuth, like not needing an extra click to authorize the application. AFAIK it's just a matter of preference, as I don't know of any SSO provider that currently supports this style but that doesn't support OAuth, but I've nonetheless seen a few applications with support for both styles of login. #3164 looks like it was written with the intent of making it easy to add more login styles going forward, but I don't see an explicit carve-out for this type of auth. That all said, just trying to clarify:

Do you mean that you'd rather have *only* OAuth of these two, or do you mean that you'd rather have this implemented in the same style as #3164? If the latter, would it make sense to base a follow-up to this on #3164's branch and wait until it's merged to submit it?

unixfox:
I would rather not complicate the current code and add a system that is more flexible than the one that you propose here. So for me oauth is better.

not sure what you think @SamantazFox and @syeopite

syeopite:
I agree with Unixfox here. While I don't see any inherent reason why we can't have both, Invidious is already starved enough for maintenance as-is. Prefering OAuth seems like the better solution.

unixfox:
Closing as I would prefer to not have many ways to implement external authentication.

Issue #4398, opened 2024-01-18
[Bug] No search results for self-harm related queries


Theta-Dev:
YouTube is showing a panel with crisis hotline information if the user is searching for certain self-harm related terms. Examples for these terms are:

- `suicide`
- `cutting`
- `blue whale`

Since Invidious cannot display this panel, the search simply returns no results. To circumvent the information panel and get to the search results, the search filter parameter `30` needs to be set to `1` (varint). Here is the decoded search filter parameter:

```bash
$ echo "8AEB" | protodec -bp      
{
  "30:0:varint": 1
}
```

NewPipe has already identified and fixed this issue, here is the commit for reference.

https://github.com/TeamNewPipe/NewPipeExtractor/commit/6ba8251be19f5008ff6654467224b6e5069298c0

colleirose:
I am working on fixing this right now

I'll submit a PR once I fix the issue

PR #4403, opened 2024-01-24
fix: prevent censoring of self-harm related search queries

Theta-Dev:
fix: #4398


syeopite (approved):

unixfox:
in testing on yewtu.be

SamantazFox (approved):

PR #4405, opened 2024-01-27
CSS: expand #contents width on small screens

toabr:
Hi there, i made the **#contents** div take the full width on small screens less than 1280px, so all page elements have a little more room especially the video titles.

original | improved
--- | ---
invidious_narrow | invidious_wide

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -13,6 +13,7 @@ body {
   display: flex;
   flex-direction: column;
   min-height: 100vh;
+  margin: auto;
 }
 
 .h-box {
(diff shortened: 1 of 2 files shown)


SamantazFox (approved):

Fijxu:
Tested, looks good 


https://github.com/iv-org/invidious/assets/61166695/672c153d-82c5-4365-874f-5edc1dc4f62a

SamantazFox:
Thanks for the PR!

toabr:
Nice. I really would like to help a little on the frontend. But i find it hard to see changes i made without recompiling the hole thing. At the moment i just tinker in the browser console and then carry the changes over. Is there a better way for testing purposes? Thanks

SamantazFox:
@toabr Testing with the developer tools and porting changes over is the best thing you can do. Unfortunately, there is no other option than recompiling available at the moment.

Issue #3522, opened 2022-12-30
[Enhancement] Add support for carousel community posts


ChunkyProgrammer:
**Is your enhancement request related to a problem? Please describe.**
Some community posts can contain multiple images. 
Ex: https://www.youtube.com/post/Ugkxt21IuGwAASZbIZjcev04GN4alJlYK3T5

In invidious api it will say "unknown attachment type"
https://redirect.invidious.io/api/v1/channels/UCnkp4xDOwqqJD7sSM3xdUiQ/community?pretty=1

**Describe the solution you'd like**
Add support for parsing posts with multiple images

**Describe alternatives you've considered**
None

SamantazFox:
Partially implemented by #3599
This still requires a front-end implementation.

ChunkyProgrammer:
Looks like this is an accessible carousel using only css and html:
codepen: https://codepen.io/jwjertzoch/pen/JjyGeRy 
article: https://levelup.gitconnected.com/how-to-make-a-fully-accessible-css-only-carousel-40e8bd62032b

PR #4412, opened 2024-01-30
Channels: Add support for multi-image community posts

ChunkyProgrammer:
closes https://github.com/iv-org/invidious/issues/3522

This implementation functions with JavaScript disabled and is fairly accessible
image
image

### DIFF
--- a/assets/css/carousel.css
+++ b/assets/css/carousel.css
@@ -1,0 +1,119 @@
+/*
+Copyright (c) 2024 by Jennifer (https://codepen.io/jwjertzoch/pen/JjyGeRy)
+
+Permission is hereby granted, free of charge, to any person 
+obtaining a copy of this software and associated documentation 
+files (the "Software"), to deal in the Software without restriction,
+ including without limitation the rights to use, copy, modify, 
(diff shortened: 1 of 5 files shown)


src/invidious/helpers/i18n.cr
```diff
@@ -78,7 +78,7 @@ def load_all_locales
   return locales
 end
 
-def translate(locale : String?, key : String, text : String | Nil = nil) : String
+def translate(locale : String?, key : String, text : String | Nil = nil, texts : Hash(String, String) | Nil = nil) : String
```
SamantazFox, line 81: You can do the following, and use `text.is_a?(Type)` to differentiate between the type of inputs:
```suggestion
def translate(locale : String?, key : String, text : String | Hash(String, String) | Nil = nil) : String
```

assets/css/carousel.css
```diff
@@ -0,0 +1,119 @@
+/*
```
SamantazFox, line 1: Please fix trailing whitespaces in this file!

ChunkyProgrammer: Done 🙂

SamantazFox:
Thanks for your help ^^

Issue #3814, opened 2023-05-20
[Bug] Inconsistent handling of empty channel tabs


absidue:
**Describe the bug**
A channel tab being empty is handled differently depending on the tab. The playlists tab displays nothing on the website and returns an empty API response `{"playlists":[],"continuation":null}`, the community tab however shows `This channel hasn't posted yet` on the website and returns a HTTP 500 API response with this error: `{"error":"This channel hasn't posted yet"}`.

It's weird that Invidious treats the playlists tab being empty as okay, but the community tab being empty as an error. The `This channel hasn't posted yet` is displayed on YouTube too, however Invidious happily ignores the `This channel has no playlists.` message on the playlists tab.

**Steps to Reproduce**
Webpage:
1. Visit `{instance}/channel/UCUKPG1-r4WFyxpyhIuCs86Q/playlists`
2. Visit `{instance}/channel/UCUKPG1-r4WFyxpyhIuCs86Q/community`
3. Compare the pages

API:
1. Visit `{instance}/api/v1/channels/UCUKPG1-r4WFyxpyhIuCs86Q/playlists`
2. Visit `{instance}/api/v1/channels/UCUKPG1-r4WFyxpyhIuCs86Q/community`
3. Compare the responses

**Screenshots**
Playlists tab:
![web-playlists](https://github.com/iv-org/invidious/assets/48293849/798666cc-5ea4-4113-96e7-30081bfc8ff2)

Community tab:
![web-community](https://github.com/iv-org/invidious/assets/48293849/80acab24-54e8-417c-8ab6-1591b306aae2)

SamantazFox:
For the community tab, that's an "error" returned by youtube's API. In addition, the community tab still relies on older code, which hasn't been cleaned, so it isn't handled the same way as the others.

absidue:
If that error is expected, does that mean you'll be returning errors for the other channel tab endpoints too or is the goal to have the community tab API endpoint behave the same as the other ones and return an empty response?

Ideally from an API user's perspective, it would be best to only return errors if there is an actual error, an empty channel tab is normal on YouTube, so shouldn't return an error. Would also align it with the behaviour of other endpoints e.g. you don't return an error when there are no search results for a search query.

SamantazFox:
Yes, I agree with you here: the community endpoint shouldn't give an error if there is no posts. Just an empty array.

syeopite:
[...]

PR #4413, opened 2024-01-30
don't error when a channel doesn't have any community posts

ChunkyProgrammer:
we now return an empty array when there are no posts.

closes https://github.com/iv-org/invidious/issues/3814
https://github.com/FreeTubeApp/FreeTube/issues/3499

### DIFF
--- a/src/invidious/channels/community.cr
+++ b/src/invidious/channels/community.cr
@@ -57,6 +57,17 @@ def extract_channel_community(items, *, ucid, locale, format, thin_mode, is_sing
       .try &.as_s || ""
     if error_message == "This channel does not exist."
       raise NotFoundException.new(error_message)
+    elsif error_message == "This channel hasn't posted yet"
+      response = JSON.build do |json|
+        json.object do
+          json.field "authorId", ucid
+          json.field "comments" do
+            json.array do
+            end
+          end
+        end
+      end
+      return response
     else
       raise InfoException.new(error_message)
     end
--- a/src/invidious/channels/community.cr
+++ b/src/invidious/channels/community.cr
@@ -61,10 +61,7 @@ def extract_channel_community(items, *, ucid, locale, format, thin_mode, is_sing
       response = JSON.build do |json|
         json.object do
           json.field "authorId", ucid
-          json.field "comments" do
-            json.array do
-            end
-          end
+          json.field "comments", json.array { }
         end
       end
       return response
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -496,5 +496,6 @@
     "toggle_theme": "Toggle Theme",
     "carousel_slide": "Slide {{current}} of {{total}}",
     "carousel_skip": "Skip the Carousel",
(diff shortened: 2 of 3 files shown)


syeopite (approved):

Fijxu:
I wanted to test this but I get "This channel hasn't posted yet" even if I don't have this pull request merged into my debug instance and that string doesn't even exists on the Invidious code.

image


image


image

However if I merge this pull request into my debug instance I don't even get the message This channel hasn't posted yet on Invidious. What is going on?

image

syeopite:
That's the point. This PR makes Invidious response to blank responses from YouTube the same by displaying an empty page instead of an error message. 

I'm not sure if I agree with not showing a message though. 
https://github.com/iv-org/invidious/issues/3814#issuecomment-1918129165

ChunkyProgrammer:
> I'm not sure if I agree with not showing a message though. [#3814 (comment)](https://github.com/iv-org/invidious/issues/3814#issuecomment-1918129165)

I can add a message to the invidious ui for the empty tabs (that way we can also include it in weblate). I don't think it would be that beneficial to include that message in the api imo

src/invidious/channels/community.cr
```diff
@@ -57,6 +57,17 @@ def extract_channel_community(items, *, ucid, locale, format, thin_mode, is_sing
       .try &.as_s || ""
     if error_message == "This channel does not exist."
       raise NotFoundException.new(error_message)
+    elsif error_message == "This channel hasn't posted yet"
+      response = JSON.build do |json|
+        json.object do
+          json.field "authorId", ucid
+          json.field "comments" do
+            json.array do
+            end
```
SamantazFox, line 66:
```suggestion
            json.array {}
```

src/invidious/channels/community.cr
```diff
@@ -57,6 +57,17 @@ def extract_channel_community(items, *, ucid, locale, format, thin_mode, is_sing
...
+          end
```
SamantazFox, line 67:
```suggestion
          json.field "comments", json.array {}
```

syeopite:
> > I'm not sure if I agree with not showing a message though. [#3814 (comment)](https://github.com/iv-org/invidious/issues/3814#issuecomment-1918129165)
> 
> I can add a message to the invidious ui for the empty tabs (that way we can also include it in weblate). I don't think it would be that beneficial to include that message in the api imo

Yes please, that'd be great. This PR just comes with an unintentional side-effect of also removing the error message on the frontend side of things.

ChunkyProgrammer:
When a community tab is empty it will display the message in the UI now :smile:

SamantazFox (approved):

ChunkyProgrammer:
build failure is unrelated to this PR and caused by changes in crystal nightly

SamantazFox (approved):

ChunkyProgrammer:
YouTube seems to be hiding these channels now

ChunkyProgrammer:
Yeah, this PR shouldn't be needed now.

PR #4348, opened 2023-12-21
API: Fix missing wildcards after login redirect

lamarios:
fix #4200

PR #4418, opened 2024-02-03
Fix referer decoding

iBicha:
This fixes an issue where
- User is redirected from Playlet to `/authorize_token`
- User is not logged in, so they are redirected again to `/login`
- The token scope has the `*` char

The token would lose scope in this case, see https://github.com/iBicha/playlet/issues/277#issuecomment-1925001220

Kinda related to https://github.com/iv-org/invidious/pull/3556, but I didn't have scope with `*` at the time, so I didn't detect this issue until now.

This is causing people to be logged out (since we're verifying the scope, and we're detecting that it's missing permissions)

### DIFF
--- a/src/invidious/helpers/utils.cr
+++ b/src/invidious/helpers/utils.cr
@@ -262,7 +262,7 @@ def get_referer(env, fallback = "/", unroll = true)
   end
 
   referer = referer.request_target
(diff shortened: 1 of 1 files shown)


absidue:
Duplicate of https://github.com/iv-org/invidious/pull/4348

iBicha:
Thanks @absidue

PR #4423, opened 2024-02-08
Fix pubsub feed parsing

tleydxdy:
similar to what's done in #3793, this is causing an assert on my instance


tleydxdy:
still testing this on my instance, it didn't crash and burn right away tho

Fijxu:
Works on my instance without problems. https://inv.nadeko.net/feed/popular

https://git.nadeko.net/Fijxu/invidious/commit/49ad660de82a96917a7a1b858b25d91d68a77734

SamantazFox (approved):

tleydxdy:
now this code has been running for 3 days, I no longer need to restart my instance every night. so I guess it worked

PR #4431, opened 2024-02-13
API: Add APIHandler back

SamantazFox:
This handler should no have been removed in #4276, as it adds the required CORS header (Access-Control-Allow-Origin) for public acces to the API.

[Thanks to @iBicha for noticing this!](https://github.com/iv-org/invidious/pull/4276/files#r1487086161)


syeopite (approved):

iBicha:
This works, thanks for the quick fix!

PR #4437, opened 2024-02-18
Preferences: Fix handling of modified source code URL

nooptek:
Currently, setting the modified code repo URL through the preferences page in Invidious is broken.

- the HTML input tag for this field has invalid type "input" (though browser falls back on text input)
- the URL is used to set the "checked" property and not as a plain value, which makes no sense for a text-based input (and results in a blank field)
- when the submitted field is empty, the retrieved value is an empty String instead of nil, causing the "modified source code URL" being an empty href link which just points to the current page


SamantazFox (approved):
Thanks a lot for the fix! And sorry for taking so long to review ^^

SamantazFox:
Thanks for the PR :D

PR #4438, opened 2024-02-19
Overhaul VideoJS dependency manager

syeopite:
- Refactors code in the script
- Simplify workarounds for dependencies with differing structure
- Adds check for missing dependency files in assets/videojs/* in case of improper installation
- Use native Crystal digest for checksum instead of system command
- Don't clear the cache each time the script is, instead there's the parameter `--clear-cache` or the compile time flag `clear_player_dependencies_cache` when ran through Invidious.
- Changes the default cache directly to be relative to the Invidious repo
- Adds ability to use a custom NPM registry
- Adds ability to change the cache directory
- Allows using pre-existing downloaded tarballs of the dependencies
- Logs checksum when validation fails. 

- Allows skipping checksum with the `--skip-checksum` parameter when calling the script directly, or the `skip_player_dependencies_checksum` compile-time flag when used through Invidious

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /invidious
 /sentry
 /config/config.yml
+/invidious-videojs-dep-install
--- a/scripts/fetch-player-dependencies.cr
+++ b/scripts/fetch-player-dependencies.cr
@@ -72,6 +72,7 @@ class PlayerDependenciesConfig
   include YAML::Serializable
 
   property version : String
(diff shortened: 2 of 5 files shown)


mk-pmb:
Is there something I can do to help get this back on track?

syeopite:
I think all there's left to do is to add an `npm` section within the dependency file as to allow for downloading dependencies from other sources in the future.

PR #4441, opened 2024-02-20
add a refresh param to /api/v1/videos/:id

iBicha:
I found myself needing to fetch video info without caring about the freshness of the details.

Use cases are many, but includes casting from YouTube app (Lounge) https://github.com/iBicha/playlet/pull/276
When user adds a video to the queue, only the video id(s) are sent, so we'd need to make a request to be able to display the video queue.

It would be nice to have an API endpoint (perhaps in v2) where the caller can indicate the details and/or accuracy is not needed, so in that case, older cached videos in the DB are fine, and also, there would be no need to hit the `next` endpoint. So fetching would need a single request when calling `fetch_video`

But until then, I thought adding a `refresh` arg to to the endpoint would be already an improvement and should save processing time and bandwidth.

PS: ~~I haven't tested this (yet)~~ yes I did

### DIFF
--- a/src/invidious/routes/api/v1/videos.cr
+++ b/src/invidious/routes/api/v1/videos.cr
@@ -8,8 +8,12 @@ module Invidious::Routes::API::V1::Videos
     region = env.params.query["region"]?
     proxy = {"1", "true"}.any? &.== env.params.query["local"]?
 
+    refresh = env.params.query["refresh"]?
+    refresh ||= "1"
+    refresh = refresh == "1" || refresh == "true"
+
     begin
-      video = get_video(id, region: region)
+      video = get_video(id, refresh: refresh, region: region)
     rescue ex : NotFoundException
       return error_json(404, ex)
     rescue ex


unixfox:
Huh I can see this being abused to take down instances by asking to refresh continuously.

iBicha:
> Huh I can see this being abused to take down instances by asking to refresh continuously and exhaust the resources.

I don't see how this is possible. The default arg value for `refresh` in `get_video` is `true`, so it was always asking for a fresh video before this PR. This only adds a chance to ask for a less fresh video

unixfox:
I have removed my comment, I thought it was the reverse, asking to always refresh but I guess that's already the case? I thought invidious would refresh only if the cache for the video is older than 10 minutes or something: https://github.com/iv-org/invidious/blob/master/src/invidious/videos.cr#L366

iBicha:
@unixfox I'm looking again at https://github.com/iv-org/invidious/blob/e8a36985aff1a5b33ddf9abea85dd2c23422c2f7/src/invidious/videos.cr#L361 and is it me, or does the `get_video` function not use the `region` parameter at all? It looks to me that if the region is provided, it is guaranteed to not return any db hit, and instead do a fetch always. On top of that, if the region is specified, the fetched video is not added to the DB. I don't see the `fetch_video` function use the region in any meaningful way.

iBicha:
This is now tested and works well https://github.com/iBicha/playlet/pull/300
Saves a lot of unnecessary fetching when older data is fine

SamantazFox (approved):

iBicha:
Unstale

iBicha:
It's been a year - closing

syeopite:
Why close it? The PR will still get merged eventually it's just Invidious development speed is slow

iBicha:
> Why close it? The PR will still get merged eventually it's just Invidious development speed is slow

A couple of months is slow, a year is likely no one is going to get to this, and it wasn't important enough to prioritize. Hoarding PRs and issues that will never be merged or solved doesn't achieve anything and adds triage burden. Plus, after a year a code base changes enough that a PR like this might need validation/testing again. I keep it a habit to try and tidy things up.

That being said, if anyone wants to pick this up again, by all means please do.

PR #4450, opened 2024-02-24
Allow embedding videos in local HTML files

### LINKED ISSUE #4448: [Bug] Content Security Policy violation when trying to embed in a local HTML file
**Describe the bug**

1. Create a new HTML file with this content:

```
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8">
	</head>
<body>
	<div>
		<object type="text/html" data="https://yewtu.be/embed/n3vQbR1GhmA"></object>
	</div>
</body>
</html>
```

2. Open the HTML in a browser.

**Logs**

```
Refused to frame 'https://yewtu.be/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'self' http: https:".
```

**Additional context**

The specific Invidious instance doesn't seem to matter, and also the behaviour seems the same in all Web browsers (e.g. Chromium, Firefox, Palemoon, etc.).

YouTube itself has no problems with embedding videos, and also Piped, the other YouTube alternative viewer, embeds with no issues this way.

tomasz1986:
routes: Allow embedding videos in local HTML files (fixes #4448)

The current Content Security Policy does not allow to embed videos
inside local HTML files which are viewed in the browser via the file
protocol. This commit adds the file protocol to the allowed frame
ancestors, so that the embedded videos load correctly in local HTML
files.

This behaviour is consistent which how the official YouTube website
allows to embed videos from itself.

Signed-off-by: Tomasz Wilczyński <twilczynski@naver.com>

### DIFF
--- a/src/invidious/routes/before_all.cr
+++ b/src/invidious/routes/before_all.cr
@@ -30,7 +30,7 @@ module Invidious::Routes::BeforeAll
 
     # Only allow the pages at /embed/* to be embedded
     if env.request.resource.starts_with?("/embed")
-      frame_ancestors = "'self' http: https:"
+      frame_ancestors = "'self' file: http: https:"
     else
       frame_ancestors = "'none'"
     end


SamantazFox (approved):
Sounds good to me!

SamantazFox (approved):

SamantazFox:
test case used:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
  </head>
<body>
  <div>
    <!-- Test 1: instance without this patch = does not load -->
    <iframe id='ivplayer' width='640' height='360' src='https://{domainA}/embed/n3vQbR1GhmA?quality=dash'></iframe>

    <!-- Test 2: instance with this patch applied = loads as expected -->
    <iframe id='ivplayer' width='640' height='360' src='https://{domainA}/embed/n3vQbR1GhmA?quality=dash'></iframe>
  </div>
</body>
</html>
```

SamantazFox:

PR #4480, opened 2024-03-09
feat: refactor kubernetes helm chart

JuniorJPDJ:
- `ConfigMap` -> `Secret` (we are holding passwords and `hmac_key` here)
- `Ingress` support
- standard labels
- better object naming - to name objects `invidious` instead of `invidious-invidious` if helm release is called `invidious`
- ability to disable internal postgres and use external one
- smaller image for `pg_isready` use (`wait_for_postgres` initContainer)
- `envFrom` instead of `env` with config (simpler syntax)
- more options to specify `securityPolicy` (and `podSecurityPolicy` separatly)
- `tolerations`, `affinity` and `nodeSelector` support
- sort the config with original config order


kubernetes/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami/
-  version: 12.11.1
-digest: sha256:3c10008175c4f5c1cec38782f5a7316154b89074c77ebbd9bcc4be4f5ff21122
-generated: "2023-09-14T22:40:43.171275362Z"
+  version: 14.3.1
```
perennialtech, line 4: The major version bump here might cause issues for existing deployments

JuniorJPDJ: Hmmm.. That's true, but we also bumped major helm version.
What should we do? Leave very old helm chart?

perennialtech: Perhaps just leave it as `version: 12.11.1`, but add a note saying that a newer Postgres version should be used if it's a fresh deployment?

JuniorJPDJ: I also bumped the image from postgres:13 to postgres:16

perennialtech: Should probably change that as well if getting rid of the version bump in the Postgres chart

JuniorJPDJ: True, but only if we want to get rid of it, I'll wait for more comments.
I bumped helm version to new major as it's not strict-compatible with older version - `securityPolicy` is the next changed thing.

kubernetes/templates/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "invidious.fullname" . }}
+  labels:
+    {{- include "invidious.labels" . | nindent 4 }}
+stringData:
+  INVIDIOUS_CONFIG: |
```
perennialtech, line 8: Wouldn't it better to leave `INVIDIOUS_CONFIG` as a configmap, with secrets being templated in as needed? Having a separate `secrets.yaml` would also sensitive info to actually be encrypted, for example by using the [`helm-secrets`](https://github.com/jkroepke/helm-secrets) plugin, without making changes to `INVIDIOUS_CONFIG` needlessly difficult.

For example:
```yaml
# secrets.yaml
postgresql:
  auth:
    username: kemal
    password: kemal
  primary:
    initdb:
      username: kemal
      password: kemal

config:
  db:
    user: kemal
    password: kemal
  hmac_key: hmac_key
````


```yaml
# values.yaml
postgresql:
  enabled: true
  image:
    tag: 16
  auth:
    # username: SECRET
    # password: SECRET
    database: invidious
  primary:
    initdb:
      # username: SECRET
      # password: SECRET
      scriptsConfigMap: invidious-postgresql-init

config:
  db:
    # user: kemal
    # password: kemal
    host: invidious-postgresql
    port: 5432
    dbname: invidious
  # hmac_key: SECRET
```

kubernetes/Chart.lock
unixfox: OK with the postgresql version bump.

We do not maintain an helm repository yet so people should look at the new features before applying the new helm chart.

Now that I think about it, I think we should extract the helm chart from the repository and do it like with searxng in a separate repo: https://github.com/searxng/searxng-helm-chart

kubernetes/templates/secret.yaml
unixfox: I do agree on that.

JuniorJPDJ: Well, I wanted to do the same but I lack the idea how to do this properly. It would be easier if config options would be a separate env vars, but this is one env var and we are unable to concatenate it in helm chart before pushing it to the pod/container.

JuniorJPDJ: Also the solution here allows the usage of helm-secret without any difficulty - it's an object, not a string in the values, so you can split the secret parts to other values file and encrypt it with helm-secrets using SOPS or provide it externally using vals.

JuniorJPDJ: `Secret`s aren't encrypted by default, but `ConfigMap`s cannot be encrypted at all.
If we have an idea how we can split part of config to go to `ConfigMap` and `Secret` we can then roll it back to `ConfigMap` and selectively move parts of config values to `Secret`. If we have no idea how to do this, I'd personally prefer to have whole `INVIDIOUS_CONFIG` env var in the `Secret` due to sensitive values inside.

JuniorJPDJ:
Hello @unixfox, are you still there? ;)

unixfox:
hello @JuniorJPDJ,

Very sorry for the lack of replies lately! I really lack of free time.

I made a big decision, I have moved the helm chart to a new dedicated repository: https://github.com/iv-org/invidious-helm-chart

I have invited you as a maintainer of the repository. I think you will do great in order to keep this helm chart active if you agree.

We now even have a dedicated helm repository for the helm chart located at http://charts-helm.invidious.io/.

Feel free to continue creating pull requests on that repository. The code there is the one from your GitHub repository, so consider this pull request "merged".

Thank you. I'll close this pull request.

JuniorJPDJ:
Thanks! That's nice!

PR #4486, opened 2024-03-15
Batch user notifications together

999eagle:
This PR changes how user notifications are handled. Both the `RefreshChannelsJob` and incoming pubsub webhooks are changed here to not directly persist notifications or the feed update flag to the database. A new job (or rather, a new fiber in the `NotificationJob` as I thought that'd be a good place to put this) handles all notifications and periodically writes them to the database batched by channels. With this change the `RefreshChannelsJob` now also correctly (at least that's my understanding) notifies the `/api/v1/auth/notifications` endpoint through the PostgreSQL `NOTIFY/LISTEN` mechanism already in use.
This greatly reduces the write load on the database when a channel has multiple video updates within the job timeout (i.e. 1 minute currently). From my testing this does seem to happen pretty often, especially when the `RefreshChannelsJob` runs at instance restart after the previous instance's backoff time for the job got too high.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -184,8 +184,9 @@ if CONFIG.popular_enabled
   Invidious::Jobs.register Invidious::Jobs::PullPopularVideosJob.new(PG_DB)
 end
 
-CONNECTION_CHANNEL = ::Channel({Bool, ::Channel(PQ::Notification)}).new(32)
-Invidious::Jobs.register Invidious::Jobs::NotificationJob.new(CONNECTION_CHANNEL, CONFIG.database_url)
+NOTIFICATION_CHANNEL = ::Channel(VideoNotification).new(32)
+CONNECTION_CHANNEL   = ::Channel({Bool, ::Channel(PQ::Notification)}).new(32)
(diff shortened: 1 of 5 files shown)


unixfox:
Could you take a look at @syeopite and @SamantazFox :)?

999eagle:
not stale, still using this pr on iv.catgirl.cloud and waiting for a review

SamantazFox (changes requested):

999eagle:
thanks for the suggestions, applied those! not sure if the mutex is done correctly now, no idea how mutexes should be used in crystal really..

SamantazFox:
> thanks for the suggestions, applied those! not sure if the mutex is done correctly now, no idea how mutexes should be used in crystal really..

My knowledge on the subject is quite new, but I think that [`.synchronize`](https://github.com/crystal-lang/crystal/blob/04998c0c7/src/mutex.cr#L140) is better as it handles errors nicely and ensures that the Mutex is always unlocked. It is also much clearer where the unsafe section of code starts and ends:
```crystal
some_mutex.synchronize do
  # Stuff
end
```

999eagle:
ah, thanks! changed the mutex to use `.synchronize`

SamantazFox (approved):

PR #4549, opened 2024-03-31
Make invidious use HTML5 innertube client if the Android client fails

techmetx11:
This makes Invidious fallback on the HTML5 client if fetching the streaming data from Android fails.
Fixes the current issue with YouTube returning a "Video Not Available" video on the outdated Android client version that Invidious uses (also makes the code log a warning in case this happens too)

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -102,16 +102,24 @@ def extract_video_info(video_id : String, proxy_region : String? = nil)
 
   new_player_response = nil
 
-  if reason.nil?
-    # Fetch the video streams using an Android client in order to get the
-    # decrypted URLs and maybe fix throttling issues (#2194). See the
-    # following issue for an explanation about decrypted URLs:
-    # https://github.com/TeamNewPipe/NewPipeExtractor/issues/562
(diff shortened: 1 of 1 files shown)


SamantazFox:
The problem with WEB/TV clients is that the streams are throttled.
To fix that, we'd have to implement a JS sandbox for the n-function.

It's good as a temporary fix for a small instance, but it can't scale for public instances.

SamantazFox:
Still, thanks for the PR ^^

unixfox:
I still think we should fallback to WEB even if we are throttled.

It is better to provide a video feed than having an error.

SamantazFox:
I don't know. I thing it's more frustrating to have a video that takes forver to load than a straight out error.

I'm really reconsidering adding some JS parser for the n-function.

raxod502:
Even if the JS parser isn't added, it might be warranted to allow server administrators to enable fallback to HTML5 streaming. After all, some servers are intended for personal or small-group use, and the tradeoff might be considered superior in those cases. The administrator of the server in question would be best positioned to make that determination, hence my suggestion that this might make sense as a server-side config option.

PR #4555, opened 2024-04-02
Update pagination logic when searching to add videos to a playlist

ChunkyProgrammer:
The old logic would filter the results and use the count of the filtered results to determine if a next page should be shown. If there were playlists or channels then the next page button would not be shown even if there was a next page.

### DIFF
--- a/src/invidious/routes/playlists.cr
+++ b/src/invidious/routes/playlists.cr
@@ -254,9 +254,12 @@ module Invidious::Routes::Playlists
 
     begin
       query = Invidious::Search::Query.new(env.params.query, :playlist, region)
-      items = query.process.select(SearchVideo).map(&.as(SearchVideo))
+      processed_query = query.process
+      total_items = processed_query.size
+      items = processed_query.select(SearchVideo).map(&.as(SearchVideo))
     rescue ex
       items = [] of SearchVideo
+      total_items = 0
     end
 
     # Pagination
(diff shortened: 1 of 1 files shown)


SamantazFox (approved):

SamantazFox:
That's much simpler than what I had in mind ^^ (which involved returning the continuation token from very far in the call stack)

SamantazFox:
Still doesn't work.

~~That's probably because we skip the categories (e.g "Latest (shorts) from [channel]" and "People also watched").~~
I'd bet it's caused by "Latest posts from [channel]", as this type of search result are not supported yet.

Here I tested searching for [`linus tech tips`](https://test.invidious.io/search?q=linus+tech+tips) on the test instance, and there is no button:

image

ChunkyProgrammer:
@SamantazFox yeah, looks like this PR isnt working because https://github.com/iv-org/invidious/issues/4583 is no longer working

ChunkyProgrammer:
Also, I think you're using the search page in the screenshot (which was not modified by this code, I only modified the screen where you search videos to add to a playlist `/add_playlist_items?q=SEARCH_QUERY`)

SamantazFox:
> Also, I think you're using the search page in the screenshot (which was not modified by this code, I only modified the screen where you search videos to add to a playlist `/add_playlist_items?q=SEARCH_QUERY`)

You're right .-.

SamantazFox (approved):

ChunkyProgrammer:
> That's much simpler than what I had in mind ^^ (which involved returning the continuation token from very far in the call stack)

Btw this will show the next page button even if there's no next page: https://github.com/iv-org/invidious/pull/4126/ which is why I closed this PR

ChunkyProgrammer:
Closing as this might not be a proper fix (as long as regular search's pagination isn't working)

PR #4561, opened 2024-04-05
Use Trending API for health checks

ChunkyProgrammer:
Using the Trending API by default for health checks is a nice workaround for heath checks failing due to the new comment format. I also imagine that going to the Trending API every 30 seconds might appear to be more normal than getting comments from a specific video every 30 seconds (even if that video is very popular).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,7 +32,7 @@ services:
         # statistics_enabled: false
         hmac_key: "CHANGE_ME!!"
     healthcheck:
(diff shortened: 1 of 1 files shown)


ashley0143:
just wait until comments get fixed tbh

syeopite (approved):
This is still a good change nonetheless even if comments weren't broken.

Besides, it is probably better to have the health check working right now rather than wait for comments to get fixed.

ashley0143:
true, i am sorry

unixfox (approved):
Will probably need to change on the documentation repo too.

PR #4562, opened 2024-04-05
API: Add 'authorVerified' field on recommended videos

ChunkyProgrammer:
https://github.com/iv-org/invidious/issues/3323

### DIFF
--- a/src/invidious/jsonify/api_v1/video_json.cr
+++ b/src/invidious/jsonify/api_v1/video_json.cr
@@ -227,6 +227,7 @@ module Invidious::JSONify::APIv1
                 json.field "author", rv["author"]
                 json.field "authorUrl", "/channel/#{rv["ucid"]?}"
                 json.field "authorId", rv["ucid"]?
+                json.field "authorVerified", rv["author_verified"]
                 if rv["author_thumbnail"]?
                   json.field "authorThumbnails" do
                     json.array do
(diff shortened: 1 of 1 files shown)


syeopite (changes requested):
This is currently returning as a string instead of a bool.

You'll have to modify the `video.related_videos` property as well as each attribute under it is converted into a string.

https://github.com/iv-org/invidious/blob/08390acd0c17875fddb84cabba54197a5b5740e4/src/invidious/videos.cr#L91-L93

ChunkyProgrammer:
> `rv["author_verified"]` is currently returning as a string instead of a bool.
> 
> You'll have to modify the `video.related_videos` property as well as each attribute under it is currently converted into a string.
> 
> https://github.com/iv-org/invidious/blob/08390acd0c17875fddb84cabba54197a5b5740e4/src/invidious/videos.cr#L91-L93

Other parts of the code assumes that author_verified is a string so I just switched to checking if it's equal to "true" for the api instead of refactoring

SamantazFox:
> Other parts of the code assumes that author_verified is a string so I just switched to checking if it's equal to "true" for the api instead of refactoring

Yes, you can do `if rv["author_verified"]? == "true"` like it's done on the HTML side.

SamantazFox (approved):

SamantazFox:
@syeopite One day this will be refactored, but today is not that day x)

PR #4570, opened 2024-04-07
Remove legacy proxy code

syeopite:
https://github.com/iv-org/invidious/pull/4270#issuecomment-1858876952

Also fixes the build on nightly as the offending code was removed


SamantazFox:
Everything still works like before, that's neat :)

PR #4571, opened 2024-04-07
Fix CORS on /latest_version

iBicha:
I noticed that `/latest_version` is missing `["Access-Control-Allow-Origin"] = "*"`
While other endpoints after the redirect (like `/videoplayback`) have the right header, this one is missing

Tested the endpoint before and after the change (before the change blocked, after the change, video plays)

### DIFF
--- a/src/invidious/routes/video_playback.cr
+++ b/src/invidious/routes/video_playback.cr
@@ -298,6 +298,7 @@ module Invidious::Routes::VideoPlayback
       url += "&title=#{URI.encode_www_form(title, space_to_plus: false)}" if title
     end
 
+    env.response.headers["Access-Control-Allow-Origin"] = "*"
     return env.redirect url
   end
 end


unixfox:
Not entirely sure if we want to allow any website on the internet to use the latest_version endpoint.

I can see this being abused. I'm OK to allow CORS when local proxy is not being used but not when it's being used.

People have to use the embed feature not bypassing it through latest_version.

iBicha:
> Not entirely sure if we want to allow any website on the internet to use the latest_version endpoint.
> 
> I can see this being abused. I'm OK to allow CORS when local proxy is not being used but not when it's being used.
> 
> People have to use the embed feature not bypassing it through latest_version.

What I just did does not introduce any new access to the `/latest_version`. You can achieve the same by copy pasting the following to a blank html file and opening it in the browser:

```
<!DOCTYPE html>
<html>
  <head>
    <title>Test</title>
  </head>
  <body>
    <script>
      (async () => {
        var videoId = "dQw4w9WgXcQ";
        var invidiousInstance = "https://invidious.fdn.fr";

        var url = invidiousInstance + "/api/v1/videos/" + videoId;
        var response = await fetch(url);
        var videoInfo = await response.json();

        var streamUrl = videoInfo.formatStreams.find(
          (stream) => stream.itag === "22"
        ).url;

        streamUrl = new URL(streamUrl);
        streamUrl.host = invidiousInstance.split("://")[1];

        var video = document.createElement("video");
        video.src = streamUrl.href;
        video.controls = true;
        video.autoplay = true;
        document.body.appendChild(video);
      })();
    </script>
  </body>
</html>
```

That streams the video through the instance without CORS issues. So it looks like this header missing is a bug.

I already see a guard against using this in the first place here https://github.com/iv-org/invidious/blob/a05d9bf0aaf6f294b1e5078de8e2c51d1e6eb8e9/src/invidious/routes/video_playback.cr#L273

iBicha:
> People have to use the embed feature not bypassing it through latest_version.

I was experimenting with something different when I found out about this: using a video in a 3d context (like a webgl application) where the video is projected on 3d mesh or something (Can be done with Three.js, Unity, and so on)
In that case, the embed feature does not work

unixfox:
For me the endpoint latest_version should only be used by invidious itself. Any request for an API usage from an external program should use the /api endpoints.

If things are missing from the API endpoint then please request them.

unixfox:
> > Not entirely sure if we want to allow any website on the internet to use the latest_version endpoint.
> > 
> > I can see this being abused. I'm OK to allow CORS when local proxy is not being used but not when it's being used.
> > 
> > People have to use the embed feature not bypassing it through latest_version.
> 
> What I just did does not introduce any new access to the `/latest_version`. You can achieve the same by copy pasting the following to a blank html file and opening it in the browser:
> 
> ```
> <!DOCTYPE html>
> <html>
>   <head>
>     <title>Test</title>
>   </head>
>   <body>
>     <script>
>       (async () => {
>         var videoId = "dQw4w9WgXcQ";
>         var invidiousInstance = "https://invidious.fdn.fr";
> 
>         var url = invidiousInstance + "/api/v1/videos/" + videoId;
>         var response = await fetch(url);
>         var videoInfo = await response.json();
> 
>         var streamUrl = videoInfo.formatStreams.find(
>           (stream) => stream.itag === "22"
>         ).url;
> 
>         streamUrl = new URL(streamUrl);
>         streamUrl.host = invidiousInstance.split("://")[1];
> 
>         var video = document.createElement("video");
>         video.src = streamUrl.href;
>         video.controls = true;
>         video.autoplay = true;
>         document.body.appendChild(video);
>       })();
>     </script>
>   </body>
> </html>
> ```
> 
> That streams the video through the instance without CORS issues. So it looks like this header missing is a bug.
> 
> I already see a guard against using this in the first place here https://github.com/iv-org/invidious/blob/a05d9bf0aaf6f294b1e5078de8e2c51d1e6eb8e9/src/invidious/routes/video_playback.cr#L273

That's different. An instance can block API access but not the latest_version endpoint. Does not mean that this instance want to be used as an API through this other mean.

iBicha:
That's a fair take, thanks

Issue #4566, opened 2024-04-05
[Bug] Comments won't load on all videos


DestinyofYeet:
**Describe the bug**

Comments won't load on some videos

**Steps to Reproduce**

1. Open video with video ID q0zCZpyv8HI
2. Scroll down to the comments
3. Be greeted by an infinite loading circle instead of the comment section

**Logs**

The GET request to `api/v1/comments/VIDEOID?format=html&hl=en-US&thin_mode=false` fails with a 500 error code.

```json
{
	"error": "Missing hash key: \"commentRenderer\"",
[...]

PR #4576, opened 2024-04-09
Comments: Add support for new format

ChunkyProgrammer:
Closes https://github.com/iv-org/invidious/issues/4566

This should fix the issues with most comments but there are still some more changes that would need to be made like adding support for formatting (bold, italic, underline) and channel emojis. The new format for these changes is similar to how descriptions of commandRuns.
image

### DIFF
--- a/src/invidious/comments/youtube.cr
+++ b/src/invidious/comments/youtube.cr
@@ -133,15 +133,16 @@ module Invidious::Comments
                   node_replies = node["replies"]["commentRepliesRenderer"]
                 end
 
-                if node["commentViewModel"]?
+                if cvm = node["commentViewModel"]?
                   # two commentViewModels for inital request
-                  cvm = node.dig?("commentViewModel", "commentViewModel")
                   # one commentViewModel when getting a replies to a comment
-                  cvm ||= node.dig("commentViewModel")
+                  cvm = cvm["commentViewModel"] if cvm["commentViewModel"]?
+
                   comment_key = cvm["commentKey"]
                   toolbar_key = cvm["toolbarStateKey"]
                   comment_mutation = mutations.find { |i| i.dig?("payload", "commentEntityPayload", "key") == comment_key }
                   toolbar_mutation = mutations.find { |i| i.dig?("entityKey") == toolbar_key }
+
                   if !comment_mutation.nil? && !toolbar_mutation.nil?
                     # todo parse styleRuns, commandRuns and attachmentRuns for comments
                     html_content = parse_description(comment_mutation.dig("payload", "commentEntityPayload", "properties", "content"), id)
@@ -160,17 +161,20 @@ module Invidious::Comments
                           end
                         end
                       end
-                      json.field "authorIsChannelOwner", comment_author["isCreator"].as_bool
-                      json.field "isSponsor", (comment_author["sponsorBadgeUrl"]? != nil)
-                      if sponsor_badge_url = comment_author["sponsorBadgeUrl"]?
-                        # Sponsor icon thumbnails always have one object and there's only ever the url property in it
-                        json.field "sponsorIconUrl", sponsor_badge_url
-                      end
+                    end
+
+                    json.field "authorIsChannelOwner", comment_author["isCreator"].as_bool
+                    json.field "isSponsor", (comment_author["sponsorBadgeUrl"]? != nil)
+
+                    if sponsor_badge_url = comment_author["sponsorBadgeUrl"]?
+                      # Sponsor icon thumbnails always have one object and there's only ever the url property in it
+                      json.field "sponsorIconUrl", sponsor_badge_url
                     end
 
                     comment_toolbar = comment_mutation.dig("payload", "commentEntityPayload", "toolbar")
                     json.field "likeCount", short_text_to_number(comment_toolbar["likeCountNotliked"].as_s)
                     reply_count = short_text_to_number(comment_toolbar["replyCount"]?.try &.as_s || "0")
+
                     if heart_state = toolbar_mutation.dig?("payload", "engagementToolbarStateEntityPayload", "heartState")
                       if heart_state.as_s == "TOOLBAR_HEART_STATE_HEARTED"
                         json.field "creatorHeart" do
@@ -181,8 +185,10 @@ module Invidious::Comments
                         end
                       end
                     end
+
                     published_text = comment_mutation.dig?("payload", "commentEntityPayload", "properties", "publishedTime").try &.as_s
                   end
+
                   json.field "isPinned", (cvm.dig?("pinnedText") != nil)
                   json.field "commentId", cvm["commentId"]
                 else
--- a/src/invidious/comments/youtube.cr
+++ b/src/invidious/comments/youtube.cr
@@ -104,6 +104,8 @@ module Invidious::Comments
       end
     end
 
+    mutations = response.dig?("frameworkUpdates", "entityBatchUpdate", "mutations").try &.as_a || [] of JSON::Any
+
     response = JSON.build do |json|
       json.object do
         if header
@@ -135,9 +137,9 @@ module Invidious::Comments
                   cvm = node.dig("commentViewModel", "commentViewModel")
                   comment_key = cvm["commentKey"]
                   toolbar_key = cvm["toolbarStateKey"]
-                  if mutations = response.dig?("frameworkUpdates", "entityBatchUpdate", "mutations")
-                    comment_mutation = mutations.as_a.find { |i| i.dig?("payload", "commentEntityPayload", "key") == comment_key}
-                    toolbar_mutation = mutations.as_a.find { |i| i.dig?("entityKey") == toolbar_key}
+                  if mutations.size != 0
+                    comment_mutation = mutations.find { |i| i.dig?("payload", "commentEntityPayload", "key") == comment_key }
+                    toolbar_mutation = mutations.find { |i| i.dig?("entityKey") == toolbar_key }
                     if !comment_mutation.nil? && !toolbar_mutation.nil?
                       html_content = comment_mutation.dig("payload", "commentEntityPayload", "properties", "content", "content").as_s
                       if comment_author = comment_mutation.dig?("payload", "commentEntityPayload", "author")
@@ -156,8 +158,8 @@ module Invidious::Comments
                             end
                           end
                         end
-                        json.field "authorIsChannelOwner",  comment_author["isCreator"].as_bool
-                        json.field "isSponsor", (comment_author["sponsorBadgeUrl"]?!= nil)
+                        json.field "authorIsChannelOwner", comment_author["isCreator"].as_bool
+                        json.field "isSponsor", (comment_author["sponsorBadgeUrl"]? != nil)
                         if comment_author["sponsorBadgeUrl"]?
                           # Sponsor icon thumbnails always have one object and there's only ever the url property in it
                           json.field "sponsorIconUrl", comment_author["sponsorBadgeUrl"].to_s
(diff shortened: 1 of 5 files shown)


ChunkyProgrammer:
The reason I didn't include channel emojis, links and text formatting in this PR is because the new format makes it difficult to escape the html as well as provide the necessary formatting. If we escape the html using HTML.escape then the indexes will be off which could cause formatting errors as well as potential for XSS vulnerabilities, if we don't use HTML.escape then we'd be more vulnerable to XSS but formatting would be fine unless the comment contains html code. 

Here's an example of how a comment with timestamp links would look (only included the first commandRun):
```json
"content": {
  "content": "00:48 eudcedit.exe - Private Char Editor\n02:25 charmap.exe - Character Map\n04:30 psr.exe - Problem Steps Recorder\n07:04 recdisc.exe - Repair Disk Tool\n08:04 sigverif.exe - File Signature Verification\n09:00 dialer.exe - Phone Dialer\n10:08 dxdiag.exe - DirectX Diagnostic\n11:30 MdSched.exe - Mem Diagnostic Scheduler\n12:38 mmc.exe - MS Management Console\n14:52 MRT.exe - Malicious Removal Tool\n16:00 msra.exe - Win Remote Assistance Tool\n 17:00 shrpubw.exe - Share Creation Wizard",
      "commandRuns": [
          {
              "startIndex": 0,
              "length": 5,
              "onTap": {
                  "innertubeCommand": {
                      "clickTrackingParams": "CBIQtnUiEwiXpu2xl7aFAxX8GdYAHZyeCIw=",
                      "commandMetadata": {
                          "webCommandMetadata": {
                              "url": "/watch?v=v3wm83zoSSY&t=48s",
                              "webPageType": "WEB_PAGE_TYPE_WATCH",
                              "rootVe": 3832
                          }
                      },
                      "watchEndpoint": {
                          "videoId": "v3wm83zoSSY",
                          "startTimeSeconds": 48
                      }
                  }
              }
          }
    ]...
```

Here is an example video: /watch?v=v3wm83zoSSY

src/invidious/comments/youtube.cr
```diff
@@ -131,89 +133,148 @@ module Invidious::Comments
...
+                if node["commentViewModel"]?
+                  cvm = node.dig("commentViewModel", "commentViewModel")
+                  comment_key = cvm["commentKey"]
+                  toolbar_key = cvm["toolbarStateKey"]
+                  if mutations.size != 0
+                    comment_mutation = mutations.find { |i| i.dig?("payload", "commentEntityPayload", "key") == comment_key }
+                    toolbar_mutation = mutations.find { |i| i.dig?("entityKey") == toolbar_key }
+                    if !comment_mutation.nil? && !toolbar_mutation.nil?
+                      # todo parse styleRuns, commandRuns and attachmentRuns for comments
+                      html_content = HTML.escape(comment_mutation.dig("payload", "commentEntityPayload", "properties", "content", "content").as_s)
```
SamantazFox, line 145: Parsing the `commandRuns` is trivial, actually.
  
We already have a function that does that for descriptions here:
https://github.com/iv-org/invidious/blob/b673695aa2704b880562399ac78659ad23b7940d/src/invidious/videos/description.cr

This function should be hardened against XSS, though (it's not a concern in descriptions, as `<` and `>` are not allowed in there).

```diff
     cp = iter.next
     break if cp.is_a?(Iterator::Stop)
 
-    str << cp.chr
+    if cp == 0x26  # Ampersand (&)
+      str << "&amp;"
+    elsif cp == 0x27  # Single quote (')
+      str << "&#39;"
+    elsif cp == 0x22  # Double quote (")
+      str << "&quot;"
+    elsif cp == 0x3C  # Less-than (<)
+      str << "&lt;"
+    elsif cp == 0x3E  # Greater than (>)
+      str << "&gt;"
+    else
+      str << cp.chr
+    end
 
     # A codepoint from the SMP counts twice
     copied += 1 if cp > 0xFFFF
     copied += 1
```

src/invidious/comments/youtube.cr
```diff
@@ -131,89 +133,148 @@ module Invidious::Comments
...
-
-                content_html = node_comment["contentText"]?.try { |t| parse_content(t, id) } || ""
-                author = node_comment["authorText"]?.try &.["simpleText"]? || ""
-
-                json.field "verified", (node_comment["authorCommentBadge"]? != nil)
+                if node["commentViewModel"]?
+                  cvm = node.dig("commentViewModel", "commentViewModel")
+                  comment_key = cvm["commentKey"]
+                  toolbar_key = cvm["toolbarStateKey"]
+                  if mutations.size != 0
```
SamantazFox, line 140: You should check if the mutations array is empty right after creating the array, rather than on each iteration. That way, you can return an empty response (and/or an error message) before doing any further processing.

I'm pretty sure that we'll never get the old format again.

src/invidious/comments/youtube.cr
```diff
@@ -131,89 +133,148 @@ module Invidious::Comments
...
+                                json.field "height", thumbnail["height"]
+                              end
+                            end
+                          end
+                        end
+                        json.field "authorIsChannelOwner", comment_author["isCreator"].as_bool
+                        json.field "isSponsor", (comment_author["sponsorBadgeUrl"]? != nil)
+                        if comment_author["sponsorBadgeUrl"]?
+                          # Sponsor icon thumbnails always have one object and there's only ever the url property in it
+                          json.field "sponsorIconUrl", comment_author["sponsorBadgeUrl"].to_s
```
SamantazFox, line 166: It improves performance a liiiitle bit:

```suggestion
                        if sponsor_badge_url = comment_author["sponsorBadgeUrl"]?
                          # Sponsor icon thumbnails always have one object and there's only ever the url property in it
                          json.field "sponsorIconUrl", sponsor_badge_url.to_s
```

src/invidious/comments/youtube.cr
ChunkyProgrammer: Would this not mess with the index included in the commandRuns?

src/invidious/comments/youtube.cr
ChunkyProgrammer: I'll remove the check for mutations.size as it should never be empty when we have a commentViewModel. I was able to get the old format a couple of days ago when I was testing this code + the old format is used for community posts.

src/invidious/comments/youtube.cr
SamantazFox: Not at all ^^

This function was specifically made to copy each code point individually while keeping track of the number of original characters this represents.

This is because the Crystal Stdlib doesn't allow UTF-16 Surrogates in an UTF-8 string (they are always decoded to their UTF-8 equivalent) but Youtube uses them everywhere.

src/invidious/comments/youtube.cr
ChunkyProgrammer: I got the old format for comments again today

src/invidious/comments/youtube.cr
ChunkyProgrammer: I was getting some weird formatting issues when trying to add parsing for styleRuns and attachmentRuns so I won't include it for now. This is the code I had, it would place links in the wrong spot in the description for some reason (it might be caused by overlapping runs?):
```crystal
def parse_runs(content : JSON::Any, video_id : String) : String
  text_content = content["content"].as_s
  return "" if text_content.empty?

  command_runs = content["commandRuns"]?.try &.as_a || [] of JSON::Any
  attachment_runs = content["attachmentRuns"]?.try &.as_a || [] of JSON::Any
  style_runs = content["styleRuns"]?.try &.as_a || [] of JSON::Any

  all_runs = (command_runs + attachment_runs + style_runs ).sort { |a, b| a["startIndex"].as_i <=> b["startIndex"].as_i }
  return text_content if all_runs.size == 0
  # Not everything is stored in UTF-8 on youtube's side. The SMP codepoints
  # (0x10000 and above) are encoded as UTF-16 surrogate pairs, which are
  # automatically decoded by the JSON parser. It means that we need to count
  # copied byte in a special manner, preventing the use of regular string copy.
  iter = text_content.each_codepoint
  index = 0
  return String.build do |str|
    all_runs.each do |command|
      cmd_start = command["startIndex"].as_i
      cmd_length = command["length"].as_i
      # Copy the text chunk between this command and the previous if needed.
      length = cmd_start - index
      index += copy_string(str, iter, length)
      # We need to copy the command's text using the iterator
      # and the special function defined above.
      cmd_content = String.build(cmd_length) do |str2|
        copy_string(str2, iter, cmd_length)
      end
      link = cmd_content
      if on_tap = command.dig?("onTap", "innertubeCommand")
        link = parse_link_endpoint(on_tap, cmd_content, video_id)
      elsif emoji_thumb = command.dig?("element", "type", "imageType", "image", "sources", 0)
        emoji_url = emoji_thumb["url"].as_s
        if emoji_url.starts_with?("https://lh3.googleusercontent.com/")
          link = String.build do |str|
            str << %(<img alt=") << cmd_content << "\" "
            str << %(src="/ggpht) << URI.parse(emoji_url).request_target << "\" "
            str << %(title=") << cmd_content << "\" "
            str << %(width=") << emoji_thumb["width"] << "\" "
            str << %(height=") << emoji_thumb["height"] << "\" "
            str << %(class="channel-emoji" />)
          end
        end
      end

      link = "<i>#{link}</i>" if command.dig?("italics")
      link = "<s>#{link}</s>" if command.dig?("strikethrough").try &.as_s == "LINE_STYLE_SINGLE"
      if weight_label = command.dig?("weightLabel").try &.as_s
        # Comments use MEDIUM for bold text and video descriptions use BOLD for bold text
        link = "<strong>#{link}</strong>" if weight_label == "FONT_WEIGHT_MEDIUM" || weight_label == "FONT_WEIGHT_BOLD"
      end
      str << link
      index += cmd_length
    end
    # Copy the end of the string (past the last command).
    remaining_length = content.size - index
    copy_string(str, iter, remaining_length) if remaining_length > 0
  end
end

```

wanniethecharacter:
I like invidious because I get to scroll down almost all the comments. But we really need to get the comments by the time this pull request gets merged with the repository later today or tomorrow.

Bubba8291:
Why isn't this merged yet?

Yetangitu:
See https://github.com/iv-org/invidious/issues/4566#issuecomment-2059070281, testing a number of video's with this PR applied shows comments loaded on every load. Given that this PR does not remove the original comment code but uses it as a fallback if the 'new' comment model is not present this seems to be a workable solution. Once YT finishes its A/ testing and settles on one of the current comment models the fallback (or the new code if they decide to ditch the new model) can be removed.

siddharth-narayan:
It's been a week since the request for review, and I know you guys must be busy but can we please get an update on the progress instead of minimizing comments? If anything needs to be done, I'm happy to help, and I can't wait for this to be merged.

Yetangitu:
For those who can not (or do not want to) build a release themselves and who feel confident (or silly) enough to trust a random binary put on the 'net  I put a release on my fork:

https://github.com/Yetangitu/invidious/releases/tag/0.20.1-PR4576

Here's the actual binary, built on Debian stable:

https://github.com/Yetangitu/invidious/releases/download/0.20.1-PR4576/invidious

This is what I'm running on my server, if it works for me it might work for you.

SamantazFox (approved):
Outside my comments below, everything works nicely, both on video comments as well as community post comments! Thanks for the PR :)

SamantazFox:
I'll definitely need to work on decoupling the upstream from the downstream JSON logic one day. This is becoming quite a problem. Also I realized that the community posts use a different chunk of logic, where all of that could be unified.

SamantazFox (approved):

TheFrenchGhosty:
I've been using it for more than a day now via the test instance, seems to work perfectly (seems also faster and more reliable (?) that the previous implementation.

unixfox:
have been running fine on yewtu.be for 2 days: https://github.com/yewtudotbe/invidious-custom/blob/master/patches/012-4576.patch

unixfox (approved):

PR #4587, opened 2024-04-20
API: Fix duplicated query parameters in proxied video URLs

absidue:
This pull request fixes that bug that was causing the query parameters to get doubled in the streaming URLs when `?local=true` is passed to the `/api/v1/videos/{id}` API endpoint.

Before: `host/path?parameters?parameters`
After: `host/path?parameters`


SamantazFox (approved):

SamantazFox:
Thanks for fixing that bug :)

PR #4589, opened 2024-04-21
Fix proxying live DASH streams

absidue:
The live DASH manifests returned by Invidious use `.c.youtube.com` subdomains, not the usual `.googlevideo.com` ones, YouTube's live DASH also uses numbered sequences/segments with the `sq` parameter instead of byte ranges.

This pull request updates the domain check in the proxy to also support the `.c.youtube.com` domains and prevents Invidious adding a range header when it is missing but the `sq` parameter is present.

### DIFF
--- a/src/invidious/routes/video_playback.cr
+++ b/src/invidious/routes/video_playback.cr
@@ -21,7 +21,7 @@ module Invidious::Routes::VideoPlayback
     end
 
     # Sanity check, to avoid being used as an open proxy
-    if !host.matches?(/[\w-]+.googlevideo.com/)
+    if !host.matches?(/[\w-]+\.(?:googlevideo|c\.youtube)\.com/)
       return error_template(400, "Invalid \"host\" parameter.")
     end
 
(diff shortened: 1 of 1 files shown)


src/invidious/routes/video_playback.cr
```diff
@@ -37,7 +37,8 @@ module Invidious::Routes::VideoPlayback
 
     # See: https://github.com/iv-org/invidious/issues/3302
     range_header = env.request.headers["Range"]?
-    if range_header.nil?
+    sq = query_params["sq"]?
+    if range_header.nil? && sq.nil?
```
SamantazFox, line 41: The `Range` header here is only added for the `HEAD` request below. After that, it's removed again. Does that change really make a difference?

absidue: Hmm odd, when I tested it without this change before it didn't work but I've just tested it again and now it worked without removing the range header too???

Personally I wouldn't send it, even if it doesn't make a difference, as YouTube's player would never send it for live stream segments either, but it's up to you to decide what the best approach is for Invidious.

SamantazFox: I should try to get rid of that HEAD request at some point.
I really want to split the logic between DASH and progressive streams/download, but I don't know where to start, this code has always broke on me every time I tried to make significant changes...

PR #4597, opened 2024-04-23
Rename native_notification.png to 07_native_notification.png to keep standards.

An0n-00:
# Pull Request

## Changes:
Rename native_notification.png to 07_native_notification.png to keep standards for all screenshots.


unixfox:
I'm for just keeping as it is. These screenshots are even too outdated. This one is particular is even not used on the homepage!

SamantazFox:
Yeah, there is no point in doing that change. We'll see when/if the UI changes enough to update the screenshots.

PR #4599, opened 2024-04-23
Fix to internal changes in HTTP::Client

beta-ziliani:
Crystal 1.12 changed the way timeouts are stored in `HTTP::Client`. Accessing the fields directly is a bad idea, but the interface doesn't provide getters.

Ref: https://github.com/crystal-lang/crystal/pull/14371

EDIT: This is PR to fix the problem in CI, but with no consideration to possible better alternatives.

### DIFF
--- a/src/invidious/yt_backend/proxy.cr
+++ b/src/invidious/yt_backend/proxy.cr
@@ -84,9 +84,9 @@ class HTTPClient < HTTP::Client
   def proxy_connection_options
     opts = {} of Symbol => Float64 | Nil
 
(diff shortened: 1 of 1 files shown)


syeopite:
Thanks for the pull request!

However the fix is more or less redundant since the whole proxy code there is set to get removed in the next batch of PRs merged.

See #4570

SamantazFox:
Closing, as I'm going to merge https://github.com/iv-org/invidious/pull/4570.

Issue #4596, opened 2024-04-22
[Bug] Music filter in trending section is returning empty page


keshavbhatt:
**Describe the bug**
The Trending music filter returns empty page on almost all instances

**Steps to Reproduce**
1. Go to https://vid.puffyan.us/feed/trending?type=Music&region=US
2. See the empty page, while in the past there were results

keshavbhatt:
Seems fixed.

PR #4600, opened 2024-04-24
Trending: Un-nest category if this is the only one

ChunkyProgrammer:
Don't get rid of category when there's only one. Trending music was returning less than 24 videos which was getting filtered out. Since multiple categories only appears on the default trending page, we don't need to do that filtering for Music, Gaming or movies.

closes https://github.com/iv-org/invidious/issues/4596


SamantazFox (approved):

SamantazFox:
No impact on search nor other trending feeds, that's perfect, thanks!

Issue #4584, opened 2024-04-18
[Bug] The video returned by YouTube isn't the requested one. (Android client) (VideoNotAvailableException)


markozajc:
Edit by @unixfox 

### The Invidious team is aware of the issue. Please only comment if you have new things to add to the subject.

This post will be updated with relevant infos as needed.

----------

**Describe the bug**
It seems like youtube is breaking something again, similar to #4498. ~~This only triggers rarely (probably A/B testing), but I assume it's going to get worse as time goes on.~~

**Steps to Reproduce**
1. Open a video
2. ~~Therr seems to be a rare (about 1/20?) chance of getting this error~~ The error now occurs on more or less every request 

**Logs**
Date: `2024-04-18T23:37:57Z`
Route: `/watch?v=ovju1aNan54&list=PLv20rxSKTWwmqrOOaSfgLIhnuMVjzIUKj&index=47&autoplay=1&listen=true`
Version: `2024.04.10-b673695a @ (HEAD detached from 5c0b6d8a)`

<details>
<summary>Backtrace</summary>
<p>
   
```
The video returned by YouTube isn't the requested one. (Android client) (VideoNotAvailableException)
  from /var/cache/invidious/updater/source/src/invidious/videos/parser.cr:155:5 in 'try_fetch_streaming_data'
  from /var/cache/invidious/updater/source/src/invidious/videos/parser.cr:112:10 in 'extract_video_info:video_id'
  from /usr/share/crystal/src/hash.cr:1299:13 in 'fetch_video'
  from /var/cache/invidious/updater/source/src/invidious/videos.cr:379:13 in 'get_video:region'
  from /var/cache/invidious/updater/source/src/invidious/routes/watch.cr:55:15 in 'handle'
  from /var/cache/invidious/updater/source/lib/kemal/src/kemal/route.cr:13:9 in '->'
  from /var/cache/invidious/updater/source/src/invidious/helpers/handlers.cr:30:37 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from /var/cache/invidious/updater/source/lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
[...]

PR #4652, opened 2024-04-27
update the url_search_issues variable to include current error in search bar

tracedgod:
## Changes
-  Changes the crash page's Github issue search option to include the ``issue_title`` in the search bar when opening the issue search page.
---
As seen with #4584 and other critical problems, duplicate issues end up unnecessarily flooding the issue reports. While this change won't entirely stop this behavior, this will help the user to easily identify if the error they have received has been reported yet and discourage them from creating a duplicate (hopefully...)

### DIFF
--- a/src/invidious/helpers/errors.cr
+++ b/src/invidious/helpers/errors.cr
@@ -43,6 +43,8 @@ def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exce
   # URLs for the error message below
   url_faq = "https://github.com/iv-org/documentation/blob/master/docs/faq.md"
   url_search_issues = "https://github.com/iv-org/invidious/issues"
+  url_search_issues += "?q=is:issue+is:open+"
+  url_search_issues += URI.encode_www_form("[Bug] " + issue_title)
 
   url_switch = "https://redirect.invidious.io" + env.request.resource
 


src/invidious/helpers/errors.cr
```diff
@@ -43,6 +43,8 @@ def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exce
   # URLs for the error message below
   url_faq = "https://github.com/iv-org/documentation/blob/master/docs/faq.md"
   url_search_issues = "https://github.com/iv-org/invidious/issues"
+  url_search_issues += "?q=is:issue+is:open+"
+  url_search_issues += URI.encode_www_form("[Bug] " + issue_title)
```
SamantazFox, line 47: Minor nitpick: [Using string interpolation prevents an extra memory allocation](https://crystal-lang.org/reference/1.12/guides/performance.html#use-string-interpolation-instead-of-concatenation).
```suggestion
  url_search_issues += URI.encode_www_form("[Bug] #{issue_title}")
```

tracedgod: that's a good nitpick to have! not entirely familiar with crystal yet so I appreciate the info.

SamantazFox (approved):

SamantazFox:
Thanks for your work!

PR #4654, opened 2024-04-27
CI: Bump Crystal version matrix

SamantazFox:


ChunkyProgrammer:
Should the dockerfiles be bumped too (they point to 1.8.2)?

SamantazFox:
@ChunkyProgrammer indeed, thanks :)

docker/Dockerfile.arm64
```diff
@@ -1,5 +1,5 @@
-FROM alpine:3.18 AS builder
-RUN apk add --no-cache 'crystal=1.8.2-r0' shards sqlite-static yaml-static yaml-dev libxml2-static zlib-static openssl-libs-static openssl-dev musl-dev xz-static
+FROM alpine:3.19 AS builder
+RUN apk add --no-cache 'crystal=1.10.1-r0' shards sqlite-static yaml-static yaml-dev libxml2-static zlib-static openssl-libs-static openssl-dev musl-dev xz-static
```
syeopite, line 2: Is crystal 1.12 not supported on Alpine ARM yet?

SamantazFox: Not on 3.19, no. It's on Edge, but the container would break almost as soon as a new Crystal release is available.

syeopite (approved):

SamantazFox:
As for testing, I've always been using the latest Crystal version on the test instance (except for 1.12.x) and it has been working great!

unixfox (approved):

PR #4661, opened 2024-04-28
remove the dislike feild

ashley0143:
this pr removes the dislike field from the api response as it just returns "0"


syeopite:
Even so this will break downstream apps that relies on Invidious' API. This will have to be something on the hypothetical V2 API if that ever gets developed but for now this is far too much of a breaking change.

PR #4670, opened 2024-04-30
update docker to crystal 1.12.1 and arm crystal 1.10.1

unixfox:
Now that invidious support crystal 1.12.1, upgrading to latest version.
And for ARM trying to keep a stable version of alpine so upgrading only to crystal 1.10.1: https://pkgs.alpinelinux.org/packages?name=crystal&branch=v3.19&repo=&arch=&maintainer=


syeopite:
The crystal versions in the docker builds is already set to be updated in PR #4654

unixfox:
oops yes, this went out of my head

PR #4675, opened 2024-05-02
HTML: Use full URL in the og:image property

Fijxu:
Some opengraph implementations don't support a URL without the domain therefore failing to fetch the video thumbnail and channel image. This pull request basically fixes that.


SamantazFox (approved):

SamantazFox:
Thanks for the PR :D

PR #4696, opened 2024-05-14
API: Fix out of bound error on empty playlists

### LINKED ISSUE #4679: [Bug] API returns 403 with error "index out of bound" when calling a playlist with 0 videos
**Describe the bug**
When I make a api call to an existing playlist `https://myinstance.tld/api/v1/auth/playlists/{playlist id}` that has no videos in it return a 403 error with `{ "error": "Index out of bounds" }`

**Steps to Reproduce**
I'm using the api call in a function for my userscript, but I tried it directly in the address bar of my browser and had the same result:
1. logged in to my instance
2. removed all videos from an existing playlist
3. made the api call in the browser using the playlist id
4. got the json with error mentioned.
5. added a video to the playlist
6. remade the api call
7. got the proper callback (all ok)
8. created a new playlist without adding a video to it
9. made a new api call with the new playlist's id
10.  got the error
11. added a video to the new playlist
12. remade the api call
13. and all is good again. 

**Additional context**
As I said, it seems to be happening to only empty playlists. As I said I'm using this api call to check if a playlist exists or not (it doesn't matter if it has no videos) and I think this error shouldn't be there for an existing playlist and I should be getting the playlist's object with the videoCount set to 0.

Fijxu:
Invidious assumes that every playlist will have at least one video because it needs to check for the `index` key. So if there is no videos on a playlist, there is no `index` key and Invidious throws `Index out of bounds`

This fixes this API endpoints when there is no videos in a playlist:
```
api/v1/playlists/:plid
api/v1/auth/playlists/:plid
```
*For some weird reason, the `api/v1/auth/playlists/:plid` endpoint was able the return the error `Index out of bounds` as said in #4679 but `api/v1/playlists/:plid` did not return anything. Weird*

Fixes #4679

### DIFF
--- a/src/invidious/routes/api/v1/misc.cr
+++ b/src/invidious/routes/api/v1/misc.cr
@@ -74,7 +74,9 @@ module Invidious::Routes::API::V1::Misc
       response = playlist.to_json(offset, video_id: video_id)
       json_response = JSON.parse(response)
 
-      if json_response["videos"].as_a[0]["index"] != offset
+      if json_response["videos"].as_a.empty?
+        json_response = JSON.parse(response)
+      elsif json_response["videos"].as_a[0]["index"] != offset
         offset = json_response["videos"].as_a[0]["index"].as_i
         lookback = offset < 50 ? offset : 50
         response = playlist.to_json(offset - lookback)


Fijxu:
Can be tested with a empty playlist that I created at https://debuginv.nadeko.net/api/v1/playlists/PL5ASK6Z87scO0GrGbMzxVz5qvMhRUitbG

SamantazFox (approved):

SamantazFox:
> _For some weird reason, the `api/v1/auth/playlists/:plid` endpoint was able the return the error `Index out of bounds` as said in #4679 but `api/v1/playlists/:plid` did not return anything. Weird_

This because the authentication middleware (`AuthHandler`) catches the error and displays it nicely:
https://github.com/iv-org/invidious/blob/bad92093bff66bfb0281e5276fa0e136a61ba330/src/invidious/helpers/handlers.cr#L127-L133

SamantazFox:
Thank you for fixing this :)

PR #4706, opened 2024-05-19
UI: Add search button to search bar

### LINKED ISSUE #529: [suggestion] search button

thansk:
Closes #529

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -318,12 +318,9 @@ input[type="search"]::-webkit-search-cancel-button {
 }
 
 .searchbar #searchbutton {
-  border: 0;
+  border: none;
   background: none;
-  text-transform: uppercase;
-  display: grid;
-  place-items: center;
-  width: 1.5em;
+  margin-top: 0;
 }
 
 .searchbar #searchbutton:hover {
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -278,7 +278,14 @@ div.thumbnail > .bottom-right-overlay {
   display: inline;
 }
 
-.searchbar .pure-form fieldset { padding: 0; }
+.searchbar .pure-form {
+  display: flex;
+}
+
+.searchbar .pure-form fieldset { 
+  padding: 0;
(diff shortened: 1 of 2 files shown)


unixfox:
Why?

ChunkyProgrammer:
Related: https://github.com/iv-org/invidious/pull/4098/

thansk:
Because I find that sometimes I enter something into the box and later want to submit it without my keyboard around. 

I think it's generally a good idea for accessibility to make invidious usable without a keyboard.

unixfox:
Ha ok you just added a button on the search bar.

The title is very confusing.

To be fair I quite like the way it is without any search button.

I think I good middle ground would be to just have the search icon. Not the full text "Search". Like on youtube.com

thansk:
I added an icon instead and used the aria-label for it with the translation like @ChunkyProgrammer suggested. Icon is using the currentColor so it works on dark and light theme. It's MIT licensed from heroicons.

unixfox:
Please use the built-in icons instead: https://github.com/iv-org/invidious/blob/master/assets/fonts/ionicons.svg

thansk:
Yeah, that makes sense, it should be good now.

SamantazFox (approved):

unixfox (approved):

assets/css/default.css
```diff
@@ -278,7 +278,14 @@ div.thumbnail > .bottom-right-overlay {
   display: inline;
 }
 
-.searchbar .pure-form fieldset { padding: 0; }
+.searchbar .pure-form {
+  display: flex;
+}
+
+.searchbar .pure-form fieldset {
```
SamantazFox, line 285:
```suggestion
.searchbar .pure-form fieldset {
```

SamantazFox:
Thanks for contributing to invidious :D

PR #4709, opened 2024-05-22
Use `make_client` instead of calling `HTTP::Client`

syeopite:
This PR is mostly so that we can customize the client Invidious uses for all HTTP requests. For example, to config a HTTP proxy to use. 

I left `images.cr` untouched as to not conflict with #4326

### DIFF
--- a/src/invidious/routes/api/v1/search.cr
+++ b/src/invidious/routes/api/v1/search.cr
@@ -31,7 +31,7 @@ module Invidious::Routes::API::V1::Search
     query = env.params.query["q"]? || ""
 
     begin
-      client = make_client(URI.parse("https://suggestqueries-clients6.youtube.com"), force_youtube_headers = true)
+      client = make_client(URI.parse("https://suggestqueries-clients6.youtube.com"), force_youtube_headers: true)
       url = "/complete/search?client=youtube&hl=en&gl=#{region}&q=#{URI.encode_www_form(query)}&gs_ri=youtube&ds=yt"
 
       response = client.get(url).body
--- a/src/invidious/routes/api/v1/search.cr
+++ b/src/invidious/routes/api/v1/search.cr
@@ -31,9 +31,7 @@ module Invidious::Routes::API::V1::Search
     query = env.params.query["q"]? || ""
 
     begin
-      client = HTTP::Client.new("suggestqueries-clients6.youtube.com")
-      client.before_request { |r| add_yt_headers(r) }
-
+      client = make_client(URI.parse("https://suggestqueries-clients6.youtube.com"), force_youtube_headers = true)
       url = "/complete/search?client=youtube&hl=en&gl=#{region}&q=#{URI.encode_www_form(query)}&gs_ri=youtube&ds=yt"
 
       response = client.get(url).body
--- a/src/invidious/yt_backend/connection_pool.cr
+++ b/src/invidious/yt_backend/connection_pool.cr
@@ -42,7 +42,7 @@ struct YoutubeConnectionPool
 
   private def build_pool
     DB::Pool(HTTP::Client).new(initial_pool_size: 0, max_pool_size: capacity, max_idle_pool_size: capacity, checkout_timeout: timeout) do
-      conn = make_client(url, force_resolve = true)
+      conn = make_client(url, force_resolve: true)
       conn.family = Socket::Family::INET if conn.family == Socket::Family::UNSPEC
       conn
     end
--- a/src/invidious/yt_backend/connection_pool.cr
+++ b/src/invidious/yt_backend/connection_pool.cr
@@ -55,7 +55,7 @@ def make_client(url : URI, region = nil, force_resolve : Bool = false, force_you
     client.family = Socket::Family::INET if client.family == Socket::Family::UNSPEC
   end
 
-  client.before_request { |r| add_yt_headers(r) } if url.host == "www.youtube.com" || force_youtube_headers
+  client.before_request { |r| add_yt_headers(r) } if url.host.try &.ends_with?("youtube.com") || force_youtube_headers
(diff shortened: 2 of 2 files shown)


src/invidious/yt_backend/connection_pool.cr
```diff
@@ -45,24 +42,22 @@ struct YoutubeConnectionPool
 
   private def build_pool
     DB::Pool(HTTP::Client).new(initial_pool_size: 0, max_pool_size: capacity, max_idle_pool_size: capacity, checkout_timeout: timeout) do
-      conn = HTTP::Client.new(url)
-      conn.family = CONFIG.force_resolve
+      conn = make_client(url, force_solve = true)
```
ChunkyProgrammer, line 45: Should this be?
```suggestion
      conn = make_client(url, force_resolve = true)
```

src/invidious/yt_backend/connection_pool.cr
```diff
@@ -45,24 +42,22 @@ struct YoutubeConnectionPool
 
   private def build_pool
     DB::Pool(HTTP::Client).new(initial_pool_size: 0, max_pool_size: capacity, max_idle_pool_size: capacity, checkout_timeout: timeout) do
-      conn = HTTP::Client.new(url)
-      conn.family = CONFIG.force_resolve
+      conn = make_client(url, force_resolve = true)
```
SamantazFox, line 45: It should be that, btw:
```suggestion
      conn = make_client(url, force_resolve: true)
```

I don't understand why the `=` isn't flagged by the compiler as invalid syntax.

SamantazFox: So, from a quick discussion in the Crystal matrix room, doing
```crystal
make_client(url, foo = true)
```
is the same as doing
```crystal
foo = true
make_client(url, foo)
```
i.e it assigns the value to a local variable and then passes that variable as a positional function parameter. In this case, this would be end up being assigned to `region`.

syeopite: Oops thanks!

I guess that also explains why this wasn't caught my the CI https://github.com/iv-org/invidious/pull/4709#discussion_r1612491743

SamantazFox: Yep! I've opened an issue at the Crystal bug tracker, because it seems like something that shouldn't be allowed by design: https://github.com/crystal-lang/crystal/issues/14722

src/invidious/routes/api/v1/search.cr
```diff
@@ -31,9 +31,7 @@ module Invidious::Routes::API::V1::Search
     query = env.params.query["q"]? || ""
 
     begin
-      client = HTTP::Client.new("suggestqueries-clients6.youtube.com")
-      client.before_request { |r| add_yt_headers(r) }
-
+      client = make_client(URI.parse("https://suggestqueries-clients6.youtube.com"), force_youtube_headers = true)
```
SamantazFox, line 34: This one has the same issue, btw

src/invidious/yt_backend/connection_pool.cr
```diff
@@ -45,24 +42,22 @@ struct YoutubeConnectionPool
 
   private def build_pool
     DB::Pool(HTTP::Client).new(initial_pool_size: 0, max_pool_size: capacity, max_idle_pool_size: capacity, checkout_timeout: timeout) do
-      conn = HTTP::Client.new(url)
-      conn.family = CONFIG.force_resolve
+      conn = make_client(url, force_resolve: true)
       conn.family = Socket::Family::INET if conn.family == Socket::Family::UNSPEC
```
SamantazFox, line 46: I think this was here to prevent `CONFIG.force_resolve` to pass `Socket::Family::UNSPEC` and hence should be moved to `make_client()` too!

src/invidious/yt_backend/connection_pool.cr
```diff
@@ -45,24 +42,22 @@ struct YoutubeConnectionPool
...
+def make_client(url : URI, region = nil, force_resolve : Bool = false, force_youtube_headers : Bool = false)
   client = HTTP::Client.new(url)
 
   # Force the usage of a specific configured IP Family
   if force_resolve
     client.family = CONFIG.force_resolve
   end
 
-  client.before_request { |r| add_yt_headers(r) } if url.host == "www.youtube.com"
+  client.before_request { |r| add_yt_headers(r) } if url.host == "www.youtube.com" || force_youtube_headers
```
SamantazFox, line 60: I think that should be `.ends_with("youtube.com")` so that headers are passed no matter which youtube subdomain we use (tv, mobile, web, search suggestions)

src/invidious/yt_backend/connection_pool.cr
```diff
@@ -45,24 +41,21 @@ struct YoutubeConnectionPool
 
   private def build_pool
     DB::Pool(HTTP::Client).new(initial_pool_size: 0, max_pool_size: capacity, max_idle_pool_size: capacity, checkout_timeout: timeout) do
-      conn = HTTP::Client.new(url)
-      conn.family = CONFIG.force_resolve
-      conn.family = Socket::Family::INET if conn.family == Socket::Family::UNSPEC
-      conn.before_request { |r| add_yt_headers(r) } if url.host == "www.youtube.com"
-      conn
+      next make_client(url, force_resolve: true)
```
syeopite, line 44: This line fails Ameba's `Style/RedundantNext` rule.

Should it also be disabled like `Style/RedundantBegin` and `Style/RedundantReturn`?

SamantazFox: I didn't know that next could be used like that! And yes, this will make the syntax consistent with the rest.

SamantazFox (approved):

PR #4717, opened 2024-05-30
Videos: Fix genre url being unusable

### LINKED ISSUE #4700: [Bug] Genre link has wrong (static) destination
**Describe the bug**
There doesn't seem to be a page for the genre, so `/channel` is taken instead, which does't make sense for a genre, let alone there being more than just a single static one

**Steps to Reproduce**
1. Go to a video page (`/watch`)
2. Scroll down below the player
3. Click on the link after "Genre:" in the left column
4. You'll be navigated to `/channel`, which probably redirects you to the default landing page, since `/channel` does not exist without a channel ID

**Logs**
_n/a_

**Screenshots**
grafik

**Additional context**
_n/a_

meatball133:
resolves #4700

### DIFF
--- a/spec/invidious/videos/regular_videos_extract_spec.cr
+++ b/spec/invidious/videos/regular_videos_extract_spec.cr
@@ -67,7 +67,7 @@ Spectator.describe "parse_video_info" do
     # Video metadata
 
     expect(info["genre"].as_s).to eq("Entertainment")
-    expect(info["genreUcid"].as_s).to be_nil
+    expect(info["genreUcid"].as_s?).to be_nil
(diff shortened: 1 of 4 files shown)


syeopite (approved):
I'm fine with this.

Though the `[]?` probably isn't needed as `genreUcid` is always present anyways.

And regarding spec compliance, if it makes more sense for `genreUcid` to default to nil and it doesn't mess with the API then its a better idea to just modify the specs rather than forcing compliance through an empty string.

SamantazFox:
Thanks for your contribution ;)

PR #4731, opened 2024-06-05
Player: Fix playback position of already watched videos

### LINKED ISSUE #3976: [Enhancement]  Disable or reset playback position for playlist and videos near end.
Hey,

is it already possible or would it be possible to disable the save_player_pos only for playlists?

Currently after hearing a playlist once it jumps to the last 10s of each video.

To keep the feature working with playlists i think a threshold would be useful.
Eg. Reset to beginning when saved playback position is 90% or more

thx

Fijxu:
Closes #3976

Trying to watch an already watched video will make the video start 15 seconds before the end of the video. This is not very comfortable when listening to music or watching/listening playlists over and over. This can be easily tested on any instance with the "Save playback position" enabled on Preferences.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -351,7 +351,12 @@ if (video_data.params.save_player_pos) {
     const rememberedTime = get_video_time();
     let lastUpdated = 0;
 
-    if(!hasTimeParam) set_seconds_after_start(rememberedTime);
+    if(!hasTimeParam) {
+      if (rememberedTime >= video_data.length_seconds - 20)
+        set_seconds_after_start(0);
+      else
+        set_seconds_after_start(rememberedTime);
+    }
 
     player.on('timeupdate', function () {
         const raw = player.currentTime();


unixfox (approved):

assets/js/player.js
```diff
@@ -352,6 +352,7 @@ if (video_data.params.save_player_pos) {
     let lastUpdated = 0;
 
     if(!hasTimeParam) set_seconds_after_start(rememberedTime);
+    if (rememberedTime >= video_data.length_seconds - 20 && !hasTimeParam) set_seconds_after_start(0);
```
SamantazFox, line 355: I think that setting the start time twice can add some delay, as the player would have to seek (= load more data) twice.

You can prevent that by extracting the common condition like so:
```js
if(!hasTimeParam) {
  if (rememberedTime >= video_data.length_seconds - 20)
    set_seconds_after_start(0);
  else
    set_seconds_after_start(rememberedTime);
}

Fijxu: Works well!

SamantazFox (approved):

SamantazFox:
Thanks for your PR!

Issue #4734, opened 2024-06-06
[Bug] "This helps protect our community." - "Please sign in to confirm you’re not a bot"


ggtylerr:
Update 21/09/2024: https://github.com/iv-org/invidious/issues/4734#issuecomment-2365205990

----

EDIT by @unixfox: The Invidious team is aware of this issue. It appears that it affects all the software using YouTube.
Please refrain from commenting if you have nothing new to bring up. Thank you.

------------

**Describe the bug**
This seems to be another block by YouTube - happens on any video, regardless of settings or instance.

**Steps to Reproduce**
1. [Go to any video on Invidious](https://iv.ggtyler.dev/watch?v=dQw4w9WgXcQ)

**Logs**
No logs on browser. In docker logs:
```
invidious-invidious-refresh-1  | 2024-06-06 22:35:23 UTC [error] get_video: PrMRuA3pd7g : This helps protect our community. Learn more
invidious-invidious-refresh-1  | 2024-06-06 22:35:23 UTC [warn] i18n: Missing translation key "This helps protect our community. Learn more"
```

**Screenshots**
![image](https://github.com/iv-org/invidious/assets/28498877/ae945bd5-4ef3-4ccd-8c70-e71ad6c8ee85)

**Additional context**
This seems to be a global update, done before 22:23 UTC. (10:23 PM.) From my brief testing, this is present throughout all instances, and regardless of IPv6 address.

EDIT: It looks like this is just rolling out throughout all servers, so some may not be affected yet.

RustedTerrier:
This doesn't happen to all instances; https://inv.tux.pizza/ still plays videos.

ggtylerr:
Interesting - I tried mine, yewtu.be, and puffyan, all of which seem to be blocked. Perhaps it's just targeting the more popular instances then?

Laurendus:
Cannot confirm this. My private instance with two users is affected as well.

ggtylerr:
Okay, it's clear this isn't global yet and it's a standard YouTube rollout. Please refrain from posting "i'm affected/not affected" comments :)

accessiblepixel:
On my own selfhosted (local) instance, I'm getting both "This helps protect our community." and also a "You may have found a bug" - some of the public interfaces are affected too... 

Here's the output (if it's any use) - 

Title: `Missing hash key: "videoDetails" (KeyError)`
Date: `2024-06-07T04:39:15Z`
Route: `/watch?v=EXxMCm941WA`
[...]

PR #4737, opened 2024-06-07
Video: try using the TvHtml5ScreenEmbed client type to get videos.

ashley0143:
this pull requests tries fixing the #4734 issue by using the tv client.


ashley0143:
needs testing

unixfox:
No that's not the intended fix. We want to still use the ANDROID client or the WEB client with a proper fix.

ggtylerr:
Should at the very least fix up the error handling on the "last hope" parser, switching to it if there's an actual error rather than just a null response.

PR #4746, opened 2024-06-12
Pending migrations check on startup. Includes index change migration

matthewmcgarvey:
When the app starts, it will check for any pending migrations and issue a warning (but allow the app to continue). Adds new functionality to migrations to mark them as required which, during the pending migration check, will cause the app to fail to start.

To test this, I took the migration from https://github.com/iv-org/invidious/pull/2469 and included it here. It will not break anything if the migration is not run and seems useful for performance issues.

I know some are pushing towards automatically running migrations, but I am not. Feel free to push back

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -135,6 +135,9 @@ end
 OUTPUT = CONFIG.output.upcase == "STDOUT" ? STDOUT : File.open(CONFIG.output, mode: "a")
 LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
 
+# Check pending migrations
+Invidious::Database::Migrator.new(PG_DB).check_pending_migrations
+
 # Check table integrity
 Invidious::Database.check_integrity(CONFIG)
 
--- a/src/invidious/database/migration.cr
+++ b/src/invidious/database/migration.cr
@@ -4,9 +4,14 @@ abstract class Invidious::Database::Migration
   end
 
   @@version : Int64?
+  @@required : Bool = false
 
   def self.version(version : Int32 | Int64)
     @@version = version.to_i64
+  end
+
+  def self.required(required : Bool)
+    @@required = required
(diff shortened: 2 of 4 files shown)


src/invidious.cr
```diff
@@ -135,6 +135,9 @@ end
 OUTPUT = CONFIG.output.upcase == "STDOUT" ? STDOUT : File.open(CONFIG.output, mode: "a")
 LOGGER = Invidious::LogHandler.new(OUTPUT, CONFIG.log_level)
 
+# Check pending migrations
+Invidious::Database::Migrator.new(PG_DB).check_pending_migrations
+
 # Check table integrity
 Invidious::Database.check_integrity(CONFIG)
```
matthewmcgarvey, line 142: I noticed this method while working... and this is a bit of a monster hidden behind a simple name. It performs database updates besides doing checks. This might have to be taken out completely once the migration files are for sure the way forward since it reads the sql files.

matthewmcgarvey:
One thing I just remembered is that, while currently no migrations are marked as required, it would probably be wise to mark most of them so that people get on the migration train.

SamantazFox (commented):
Looks good to me!

src/invidious.cr
SamantazFox: Yeah, this is a remnant I haven't dare to touch. I think that it should be safe to remove all of that code.

SamantazFox:
> One thing I just remembered is that, while currently no migrations are marked as required, it would probably be wise to mark most of them so that people get on the migration train.

We could only mark the latest as required, that'd be enough.
Though we need to agree on a process for docker users.

unixfox:
Thank you! I'll look into doing some guides for applying migrations manually.

matthewmcgarvey:
Need to update this PR to also drop the index that this new one replaces if it exists.

PR #4747, opened 2024-06-12
Rewrite transcript logic to be more generic

syeopite:
The transcript logic in Invidious was written specifically as a workaround for captions, and not transcripts as a feature.

This PR genericises the logic as so it can be used for actually implementing transcripts within Invidious' as well.

The most notable change is the added parsing of section headings when it was previously skipped over in favor of regular lines.

The changes in `src/invidious/videos/transcript.cr` should be reviewed as more of a rewrite than just changes.


SamantazFox (approved):

SamantazFox:
Works well, both captions and storyboards have no issues.

SamantazFox:
Thanks for your work ^^

PR #4753, opened 2024-06-16
CI: Run Ameba

syeopite:
Ameba has been technically apart of Invidious for a couple years now but its linting was never enforced. This PR changes that.

Following cases such as what was seen in #4709 it'd probably be a good idea to have Ameba check over everything, especially for basic mistakes.

And from the testing I've done its already been pretty helpful. For example a duplicated key in the i18next logic

https://github.com/iv-org/invidious/blob/1ae14cc22468ce6e0eb794752b113604b1d5582d/src/invidious/helpers/i18next.cr#L97-L98

This PR simply adds Ameba to the CI but doesn't actually fix any of the detected issues. I'd be opening more PRs in the near future to slowly address everything until Invidious is compliant.

### DIFF
--- a/.ameba.yml
+++ b/.ameba.yml
@@ -20,6 +20,9 @@ Lint/ShadowingOuterLocalVar:
   Excluded:
   - src/invidious/helpers/tokens.cr
 
+Lint/NotNil:
+  Enabled: false
+
 
 #
 # Style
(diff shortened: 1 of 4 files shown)


SamantazFox:
While you're here, can you bump the ameba version in `shard.lock` to `1.6.1`?
https://github.com/crystal-ameba/ameba/releases

syeopite:
Once Invidious is Ameba compliant it might be a good idea to switch to https://github.com/crystal-ameba/github-action rather than building and using Ameba ourselves.

This allows the CI to automatically annotate the offending code on Github's UI which is a better experience than manually checking the CI logs. 

However, this is a blocker: https://github.com/crystal-ameba/github-action/issues/19.

SamantazFox (approved):

PR #4760, opened 2024-06-21
Show available actions on embed error page

johansenja:
## Context

If you have privacyRedirect installed in your browser, any embedded youtube videos are automatically replaced by an invidious equivalent on pages you visit. The trouble is, often embeds don't always work - if 1/ the selected instance is down; 2/ the video is blocked or similar.

## Current flow
If you particularly want to watch the video, you might right-click the iframe, click open in new tab, then replace the `/embed/` part of the url with `/watch?v=` - or replace the domain with `youtube.com` to see the youtube version of it (but still as an embed)

## Desired flow
It would be great if you had the same actions available on regular video error pages (switch instance / watch on youtube / refresh)

## Caveats
It's worth noting that different browsers handle links inside iframes differently; eg. firefox often prompts you to open the link in a new tab - which IMO is ok, because if you are interested enough in the video to want to watch it, you would probably prefer to at least have this automatically, rather than doing it manually?

Chrome, on the other hand, _doesn't_ even give you the option to "open frame in new tab" when you right click it, but it _does_ seem to follow urls within the frame

## Summary of changes in PR
- Add `request_path.starts_with?("/embed")` check to `error_redirect_helper`
---
- Move HTML string from error_redirect_helper into separate template
- Add unit tests for `error_redirect_helper`
- Add `ContextWithPreferences` and `test_env` helpers to help with test setup/compilation

The first point is obviously the main functional change, and the latter 3 are IMO code quality improvements (so they can be modified/removed if they are not helpful here)

## Testing
I don't have this set up fully e2e locally so haven't checked this fully, but I found the unit tests went a decent way to verifying the behaviour I wanted

### DIFF
--- a/spec/helpers/errors_spec.cr
+++ b/spec/helpers/errors_spec.cr
@@ -8,7 +8,7 @@ Spectator.describe "error_redirect_helper" do
     test_env.set "current_page", current_url
 
     html = error_redirect_helper(test_env)
-    expect(html).to eq "<p style=\"margin-bottom: 4px;\">After which you should try to: </p>\n<ul>\n  \n    <li>\n      <a href=\"/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Refresh</a>\n    </li>\n  \n    <li>\n      <a href=\"/redirect?referer=/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Switch Invidious Instance</a>\n    </li>\n  \n    <li>\n      <a href=\"https://youtube.com/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Go to YouTube</a>\n    </li>\n  \n    <li>\n      <a href=\"/watch?v=IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Open in new page</a>\n    </li>\n  \n</ul>\n"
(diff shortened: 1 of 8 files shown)


spec/helpers/errors_spec.cr
```diff
@@ -0,0 +1,31 @@
...
+require "kilt"
+
+Spectator.describe "error_redirect_helper" do
+  it "shows next steps on embed page errors" do
+    current_url = "/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS"
+    test_env = test_env current_url
+    test_env.set "current_page", current_url
+
+    html = error_redirect_helper(test_env)
+    expect(html).to eq "<p style=\"margin-bottom: 4px;\">After which you should try to: </p>\n<ul>\n  \n    <li>\n      <a href=\"/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Refresh</a>\n    </li>\n  \n    <li>\n      <a href=\"/redirect?referer=/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Switch Invidious Instance</a>\n    </li>\n  \n    <li>\n      <a href=\"https://youtube.com/embed/IeQT18gaB-c?si=YxBQzH-GBSTS4vBS\">Go to YouTube</a>\n    </li>\n  \n</ul>\n"
```
SamantazFox, line 11: For a second I though that we had an URL handling issue, but it's just that you're not passing an encoded URL like the HTTP server handlers would do (see how in the "Switch Invidious Instance" URL nothing after `referer=` is URL encoded?).

It also made me realize that the `si=` parameter isn't removed from the "Watch on Youtube URL" (for the record, this is used to track who shared the video)

PR #4763, opened 2024-06-24
CI: Add release based containers

syeopite:
- Changes the current master based container to use "master" tag instead of "latest" tag
- Adds a new workflow to build a container on each new release which has  the "latest" tag, and a tag based on the current released version.

Once this process is done, it should allow us to quickly merge PRs which passed the code-review but not the testing stage. Instead, the testing-stage can be pushed to when each release occurs.


iBicha:
> Changes the current master based container to use "master" tag instead of "latest" tag

`stable` might be more commonly used/understood. Or even `production`.
Very good initiative!

syeopite:
The release containers are tagged as "latest" as to ensure a smooth transition for instances. 

As testing is being moved to per-release with this PR, instances should be using the release containers instead of the container built off master as to ensure instance stability.

iBicha:
> The release containers are tagged as "latest" as to ensure a smooth transition for instances.
> 
> As testing is being moved to per-release with this PR, instances should be using the release containers instead of the container built off master as to ensure instance stability.

Fair enough, I take my comment back

SamantazFox:
Could you rename the two files to something in the odds of `container-release` and `container-nightly` (or whatever you see fit)? The two names are too similar, that's confusing!

.github/workflows/build-stable-container.yml
```diff
@@ -0,0 +1,90 @@
...
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: quay.io/invidious/invidious
+          tags: |
+            type=semvar,pattern={{version}}
```
SamantazFox, line 50: Small typo
```suggestion
            type=semver,pattern={{version}}
```
Reference: https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver

.github/workflows/build-stable-container.yml
```diff
@@ -0,0 +1,90 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Crystal
+        uses: crystal-lang/install-crystal@v1.8.2
+        with:
+          crystal: 1.12.2
```
SamantazFox, line 19: Commenting here, but can you bump the other file to the same version?

.github/workflows/build-stable-container.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Docker meta
+        id: meta-arm64
+        uses: docker/metadata-action@v5
+        with:
+          images: quay.io/invidious/invidious
+          flavor: |
+            suffix=-arm64
+          tags: |
+            type=semvar,pattern={{version}}
```
SamantazFox, line 75:
```suggestion
            type=semver,pattern={{version}}
```

SamantazFox (changes requested):

SamantazFox:
Marking as ready as the only option to trigger that is to merge the code and publish a new tagged release.

unixfox (approved):

PR #4772, opened 2024-07-03
Add support for an external signature server

### LINKED ISSUE #4649: [Enhancement] Add decryption of the sig parameter for bypassing throttling
Currently we do not support fetching the streams using WEB client because by default the stream is too slow.  So we are only relying on the ANDROID client. But recently the ANDROID client has started to be blocked from youtube unless you pass "droidguard" challenge: https://github.com/iv-org/invidious/issues/4584

This issue is to discuss how to implement the decryption of the sig parameter in order to avoid throttling.

An atempt was done back in https://github.com/iv-org/invidious/pull/2222 but was closed due to potential issues around missing isolation when executing untrusted JavaScript code from YouTube.

So there are other solutions:
- Execute a separate [node](https://nodejs.org/en) process which is not part of the main invidious process. May be too permissive.
- Execute a separate [deno](https://deno.com/) process which is not part of the main invidious process. [Good potential because by default deno lock down a lot of permissions.](https://docs.deno.com/runtime/manual/basics/permissions)
- Create bindings for using a C library that allows to evaluate isolated javascript code. I do not know any library like this.
- Use the minimal JS interpreter from yt-dlp by executing it using python: https://github.com/dirkf/youtube-dl/blob/master/youtube_dl/jsinterp.py
- Other minimal JS interpreter from YouTube.js in Typescript: https://github.com/LuanRT/Jinter

More ressources:
- [Calling python functions from crystal](https://fridgerator.github.io/2019/12/21/calling-python-functions-from-crystal_part_1.html)
- [crython](https://github.com/RomainFranceschini/crython)
- [crystal-nodejs](https://github.com/fukaoi/crystal-nodejs)

SamantazFox:
Closes #4649

This PR adds support for [inv_sig_helper](https://github.com/iv-org/inv_sig_helper), which offloads the player fetching, function extraction and signature parsing, which in turn allows to use the web client to watch videos.

Side note: I had to update the crystal overrides because the stdlib changed quite a while ago (See https://github.com/crystal-lang/crystal/pull/11049) and it was required to properly use TCP/unix sockets.

-----

TODO:
- [x] Only activate if something like `inv_sig_helper_url: XX` is filled. This way, this functionality is only enabled when the user needs it. And the user can specify a TCP endpoint in case of Docker, for example.
- [ ] ~~Additionally, if this is not activated, ANDROID_TEST_SUITE will still be used. (As long as it still works fine for "unblocked IP addresses")~~ This will be addressed when this and #4789 will be merged


src/invidious/helpers/sig_helper.cr
```diff
@@ -0,0 +1,321 @@
...
+      end
+
+      case host_or_path
+      when .starts_with?('/')
+        @socket = UNIXSocket.new(host_or_path)
+      when .starts_with?("tcp://")
+        uri = URI.new(host_or_path)
+        @socket = TCPSocket.new(uri.host.not_nil!, uri.port.not_nil!)
+      else
+        uri = URI.new("tcp://#{host_or_path}")
```
iBicha, line 276:
```suggestion
        uri = URI.parse(host_or_path)
        @socket = TCPSocket.new(uri.host.not_nil!, uri.port.not_nil!)
      else
        uri = URI.parse("tcp://#{host_or_path}")
```

syeopite (approved):
Overall I don't really see anything wrong with the logic here. I just have a couple questions to make sure I understand everything correctly.

Also, what do you think about splitting the signature communication logic into its own shard?

src/invidious/videos.cr
```diff
@@ -98,20 +100,47 @@ struct Video
 
   # Methods for parsing streaming data
 
+  def convert_url(fmt)
+    if cfr = fmt["signatureCipher"]?.try { |h| HTTP::Params.parse(h.as_s) }
```
syeopite, line 104: ```[C] Naming/BlockParameterName: Disallowed block parameter name found```

Can you pick a more descriptive name than `h` here?

src/invidious/helpers/sig_helper.cr
```diff
@@ -0,0 +1,325 @@
...
+    def get_sts : UInt64?
+      request = Request.new(Opcode::GET_SIGNATURE_TIMESTAMP, nil)
+
+      return send_request(request) do |bytes|
+        IO::ByteFormat::NetworkEndian.decode(UInt64, bytes)
+      end
+    end
+
+    # Return the current player's version
+    def get_player : UInt32?
```
syeopite, line 130: I think `get_player_version` would be clearer here

```suggestion
    def get_player_version : UInt32?
```

src/invidious/helpers/sig_helper.cr
```diff
@@ -0,0 +1,325 @@
...
+      request = Request.new(Opcode::GET_PLAYER_STATUS, nil)
+
+      send_request(request) do |bytes|
+        has_player = (bytes[0] == 0xFF)
+        player_version = IO::ByteFormat::NetworkEndian.decode(UInt32, bytes[1..4])
```
syeopite, line 135:
```suggestion
        player_version = IO::ByteFormat::NetworkEndian.decode(UInt32, bytes[1..4]) # ameba:disable Lint/UselessAssign
```

src/invidious/helpers/sig_helper.cr
```diff
@@ -0,0 +1,325 @@
...
+        IO::ByteFormat::NetworkEndian.decode(UInt64, bytes)
+      end
+    end
+
+    # Return the current player's version
+    def get_player : UInt32?
+      request = Request.new(Opcode::GET_PLAYER_STATUS, nil)
+
+      send_request(request) do |bytes|
+        has_player = (bytes[0] == 0xFF)
```
syeopite, line 134:
```suggestion
        has_player = (bytes[0] == 0xFF) # ameba:disable Lint/UselessAssign
```

src/invidious/yt_backend/youtube_api.cr
```diff
@@ -474,19 +476,29 @@ module YoutubeAPI
     params : String,
     client_config : ClientConfig | Nil = nil
   )
+    # Playback context, separate because it can be different between clients
+    playback_ctx = {
+      "html5Preference" => "HTML5_PREF_WANTS",
+      "referer"         => "https://www.youtube.com/watch?v=#{video_id}",
+    } of String => String | Int64
+
+    if {"WEB", "TVHTML5"}.any? { |s| client_config.name.starts_with? s }
```
syeopite, line 485: ```[C] Naming/BlockParameterName: Disallowed block parameter name found```

Can you pick a more descriptive name than `s` here?

src/invidious/helpers/sig_helper.cr
```diff
@@ -0,0 +1,325 @@
...
+
+      sig_dec = send_request(request) do |bytes|
+        StringPayload.from_bytes(bytes).string
+      end
+
+      return sig_dec
+    end
+
+    # Return the signature timestamp from the server's current player
+    def get_sts : UInt64?
```
syeopite, line 121: Can you add an `ameba:disable Naming/AccessorMethodName` to the definitions of `get_*`? Ameba seems to be mistakenly classifying these as getter methods.

src/invidious/helpers/signatures.cr
```diff
@@ -1,73 +1,46 @@
...
+    self.check_update
+    return SigHelper::Client.decrypt_sig(str)
+  rescue ex
+    LOGGER.debug(ex.message || "Signature: Unknown error")
+    LOGGER.trace(ex.inspect_with_backtrace)
+    return nil
+  end
 
-    return "&#{sp}=#{sig.join("")}"
+  def get_sts : UInt64?
```
syeopite, line 38: I think using `get_signature_timestamp` directly instead of abbreviating it would be clearer here.

syeopite (commented):
Just some Ameba nitpicks. Feel free to ignore.

src/invidious/helpers/signatures.cr
```diff
@@ -1,73 +1,46 @@
-alias SigProc = Proc(Array(String), Int32, Array(String))
+require "http/params"
+require "./sig_helper"
 
-struct DecryptFunction
-  @decrypt_function = [] of {SigProc, Int32}
-  @decrypt_time = Time.monotonic
+struct Invidious::DecryptFunction
+  @last_update = Time.monotonic - 42.days
```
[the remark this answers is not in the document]
SamantazFox: I wanted invidious to always check for a network update on each start, so I just took a "random" number that causes the condition in `update` to be `true` the first time it is called. The Number can be anything, really.

src/invidious/helpers/sig_helper.cr
```diff
@@ -0,0 +1,325 @@
...
+
+        @queue[transaction_id] = transaction
+      end
+
+      write_packet(transaction_id, request)
+
+      return transaction.channel
+    end
+
+    def receive_data
```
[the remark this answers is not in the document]
SamantazFox: I didn't think about it. Would you want me to make it private?

src/invidious/helpers/sig_helper.cr
SamantazFox: Is that a bug in ameba? I don't see how this assign is useless...

src/invidious/helpers/sig_helper.cr
SamantazFox: I'd rather disable `Naming/AccessorMethodName` altogether. Maybe `Naming/BlockParameterName` too!

SamantazFox:
> Also, what do you think about splitting the signature communication logic into its own shard?

Why though? Sound harder to maintain and I don't see anyone using this code except us.
The connection multiplexing code might actually be a good candidate for a shard, but I don't think the code is versatile enough in its current state.

src/invidious/helpers/sig_helper.cr
syeopite: Its because of the block.

In Ameba's eyes a block could potentially yield multiple time and each time it yields the assigns there would get overwritten without being used within the block.

syeopite:
Ah right. I did not think about the additional maintenance burden. 

I requested it to potentially open the opportunity for other YouTube projects within the Crystal ecosystem though the possibility of that is pretty low...

Let's just keep it within Invidious.

src/invidious/helpers/sig_helper.cr
syeopite: Disabling `Naming/BlockParameterName` is fine imo but the general conventions in Crystal is to suffix accessor  methods with `?`

Shouldn't Invidious be adhering to that convention? I'm not sure if the amount of false positives within Ameba is enough to warrant disabling that rule altogether.

SamantazFox:
Important note: with the current state of the PR, invidious can't start if the socket is not present. I will add some logic to fallback to another client in that case.

src/invidious/helpers/sig_helper.cr
syeopite: I take back what I said. Ameba complains about `Naming/AccessorMethodName` as a false positive more often than not

unixfox:
I don't see the way to configure to which hostname the inv_sig_helper is going to listen on. Is it hardcoded to use a localhost address with a fixed port?

I'm thinking about docker where each service has its separate private IP address.

techmetx11:
> I don't see the way to configure to which address the inv_sig_helper is listening on. Is it done on purpose or an oversight?
> 
> I'm thinking about docker, where each service has its separate private IP address.

Currently, it's listening on a UNIX socket (aka. on a path, by default: `/tmp/inv_sig_helper.sock`)

unixfox:
@techmetx11 

> > I don't see the way to configure to which address the inv_sig_helper is listening on. Is it done on purpose or an oversight?
> > I'm thinking about docker, where each service has its separate private IP address.
> 
> Currently, it's listening on a UNIX socket (aka. on a path, by default: `/tmp/inv_sig_helper.sock`) I could change it to let it listen on UNIX or TCP

Ah, yes that would be a great improvement because we have people in the community that run Invidious over multiple servers, and you can only communicate through TCP if you have multiple servers, a UNIX socket won't work.

I'm not sure if the sig parameter can be shared between multiple public IP addresses or not, though.

unixfox:
Continuing the discussion about signature helper server here: https://github.com/iv-org/inv_sig_helper/issues/1

unixfox:
There will be a need to add the support for TCP/IP, as such that it can be used from multiple invidious processes.

Issue opened on inv_sig_helper side: https://github.com/iv-org/inv_sig_helper/issues/2

999eagle:
i've been running this and #4789 on `iv.catgirl.cloud` for a while now to circumvent the login requirements, but occasionally invidious workers fail to start correctly after their hourly restart due to waiting indefinitely for the sig helper to force update the player. i've opened https://github.com/iv-org/inv_sig_helper/issues/4 for the underlying issue here

unixfox:
Missing feature:
- Only activate if something like `inv_sig_helper_url: XX` is filled. This way, this functionality is only enabled when the user needs it. And the user can specify a TCP endpoint in case of Docker, for example.
  Additionally, if this is not activated, ANDROID_TEST_SUITE will still be used. (As long as it still works fine for "unblocked IP addresses")

src/invidious/helpers/sig_helper.cr
```diff
@@ -135,7 +136,15 @@ class Invidious::SigHelper
         player_version = IO::ByteFormat::NetworkEndian.decode(UInt32, bytes[1..4])
         has_player ? player_version : nil
       end
+    end
+
+    # Return when the player was last updated
+    def get_player_timestamp : UInt64?
+      request = Request.new(Opcode::GET_SIGNATURE_TIMESTAMP, nil)
```
999eagle, line 143: this should probably be `Opcode::PLAYER_UPDATE_TIMESTAMP`, right?

src/invidious/helpers/signatures.cr
```diff
@@ -10,7 +10,16 @@ struct Invidious::DecryptFunction
...
+
+    # If we have updated in the last 5 minutes, do nothing
+    return if (now - @last_update) > 5.minutes
+
+    # Get the time when the player was updated, in the event where
+    # multiple invidious processes are run in parallel.
+    player_ts = Invidious::SigHelper::Client.get_player_timestamp
+    player_time = Time.unix(player_ts || 0)
+
+    if (now - player_time) > 5.minutes
```
999eagle, line 22: at least according to https://github.com/iv-org/inv_sig_helper#player_update_timestamp-0x05, the player update timestamp is already seconds since the last update, so this should be something like

```
player_time = (player_ts || 42).minutes

if player_time > 5.minutes
```

src/invidious/helpers/sig_helper.cr
SamantazFox: You're right, thanks!

src/invidious/helpers/signatures.cr
SamantazFox: Damn, I should learn to read :x I was expecting a unix timestamp here....

MMaster:
To also write it here: 
Video caching in invidious also caches and reuses data from player so if the player changes the video that is pulled from cache will cause 403s to be returned for videoplayback requests.

SamantazFox:
> To also write it here: Video caching in invidious also caches and reuses data from player so if the player changes the video that is pulled from cache will cause 403s to be returned for videoplayback requests.

Yeah, I should rewrite the URLs during parsing (i.e before caching) not during use.

PR #4784, opened 2024-07-09
Carry over audio-only mode in playlist links

krystof1119:
Currently, when a video is viewed in audio-only mode (/watch?listen=1...), the 'related video' links shown on the video include listen=true in the URL parameters, but if the video was opened as part of the playlist, the links to other videos in that playlist don't include it. This PR changes it such that the links have '&listen=1' appended to them when the user is watching a video in audio-only mode.

I made the same change for mixes, and from a cursory look, it seems to me as though it's working (but I personally don't use mixes, so I can't say for sure).

### DIFF
--- a/src/invidious/routes/api/v1/misc.cr
+++ b/src/invidious/routes/api/v1/misc.cr
@@ -42,8 +42,8 @@ module Invidious::Routes::API::V1::Misc
     format = env.params.query["format"]?
     format ||= "json"
 
-    listenParam = env.params.query["listen"]?
-    listen = (listenParam == "true" || listenParam == "1")
+    listen_param = env.params.query["listen"]?
+    listen = (listen_param == "true" || listen_param == "1")
 
     if plid.starts_with? "RD"
       return env.redirect "/api/v1/mixes/#{plid}"
@@ -114,8 +114,8 @@ module Invidious::Routes::API::V1::Misc
     format = env.params.query["format"]?
     format ||= "json"
 
-    listenParam = env.params.query["listen"]?
-    listen = (listenParam == "true" || listenParam == "1")
+    listen_param = env.params.query["listen"]?
+    listen = (listen_param == "true" || listen_param == "1")
 
     begin
       mix = fetch_mix(rdid, continuation, locale: locale)
--- a/src/invidious/routes/api/v1/misc.cr
+++ b/src/invidious/routes/api/v1/misc.cr
@@ -42,6 +42,9 @@ module Invidious::Routes::API::V1::Misc
     format = env.params.query["format"]?
     format ||= "json"
 
+    listenParam = env.params.query["listen"]?
+    listen = (listenParam == "true" || listenParam == "1")
+
     if plid.starts_with? "RD"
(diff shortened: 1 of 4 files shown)


syeopite (approved):

krystof1119:
Sorry, misclicked and rerequested the review, didn't mean to do that.

src/invidious/routes/api/v1/misc.cr
```diff
@@ -42,6 +42,15 @@ module Invidious::Routes::API::V1::Misc
...
 
+    listen = env.params.query["listen"]?
+    listen ||= "false"
+
+    if listen == "true" || listen == "1"
+      listen = true
+    else
+      listen = false
+    end
+
```
SamantazFox, line 53: You can directly save the result of the comparison (= a boolean) to a variable! This is much more compact ^^
```suggestion
    listen = env.params.query["listen"]? || "false"
    listen = (listen == "true" || listen == "1")
```

src/invidious/routes/api/v1/misc.cr
```diff
@@ -109,6 +118,15 @@ module Invidious::Routes::API::V1::Misc
...
```
SamantazFox, line 129:
```suggestion
    listen = env.params.query["listen"]? || "false"
    listen = (listen == "true" || listen == "1")
```

src/invidious/routes/api/v1/misc.cr
krystof1119: Good point, I changed this slightly to eliminate the `|| "false"` (since it wasn't needed). I also split the listen variable in two, since in retrospect, changing the type of a variable feels worse than adding one extra name to the code.

krystof1119:
Sorry for the late response, time got away from me. I've incorporated the suggestion and rebased the patch onto the current version of the code.

SamantazFox (approved):

src/invidious/routes/api/v1/misc.cr
```diff
@@ -42,6 +42,9 @@ module Invidious::Routes::API::V1::Misc
     format = env.params.query["format"]?
     format ||= "json"
 
+    listenParam = env.params.query["listen"]?
+    listen = (listenParam == "true" || listenParam == "1")
```
SamantazFox, line 46:
```suggestion
    listen_param = env.params.query["listen"]?
    listen = (listen_param == "true" || listen_param == "1")
```

src/invidious/routes/api/v1/misc.cr
```diff
@@ -111,6 +114,9 @@ module Invidious::Routes::API::V1::Misc
...
```
SamantazFox, line 118:
```suggestion
    listen_param = env.params.query["listen"]?
    listen = (listen_param == "true" || listen_param == "1")
```

SamantazFox (approved):
I made these changes for you :)

unixfox (approved):

syeopite:
Thank you for your work! Apologies for how long it has taken for this to be merged 😅

PR #4795, opened 2024-07-17
Ameba: Fix Lint/UselessAssign

syeopite:


src/invidious/channels/channels.cr
```diff
@@ -232,7 +232,7 @@ def fetch_channel(ucid, pull_all_videos : Bool)
       id:                 video_id,
       title:              title,
       published:          published,
-      updated:            Time.utc,
+      updated:            updated,
```
syeopite, line 235: I'm not too sure if this should be using the current time `Time.utc` or the parsed updated time

SamantazFox: I think it's supposed to report the time when the channel content was fetched by the background job. That's especially useful for channels that don't post often, when the last update is for a very old video.

syeopite: In that case should there be a new attribute that reports when the video was updated on YouTube's side? Or should we just remove the logic that parses the updated time?

SamantazFox: Okay, my bad. All that time I though we were talking about `InvidiousChannel.updated` while you want to change `ChannelVideo.updated`.

My previous message applies to the former, and your fix in indeed welcome in the latter case :)

SamantazFox (approved):

SamantazFox:
Thank you!

PR #4807, opened 2024-07-27
Ameba: Miscellaneous fixes

syeopite:


src/invidious/helpers/i18next.cr
```diff
@@ -189,7 +189,7 @@ module I18next::Plurals
 
     # Emulate the `rule.numbers.size == 2 && rule.numbers[0] == 1` check
     # from original i18next code
-    private def is_simple_plural(form : PluralForms) : Bool
+    private def simple_plural?(form : PluralForms) : Bool
```
SamantazFox, line 192: Can you revert that change, as it's already in #4806 ?

src/invidious/routes/video_playback.cr
```diff
@@ -131,7 +131,7 @@ module Invidious::Routes::VideoPlayback
       end
 
       # TODO: Record bytes written so we can restart after a chunk fails
-      while true
+      loop do
```
SamantazFox, line 134: Is that change requested by Ameba?
`loop` is just a macro for `while true` anyway:
https://github.com/crystal-lang/crystal/blob/4cea10199d5006000a129413f6f607697185c83a/src/kernel.cr#L114

syeopite: Yep, `Style/WhileTrue`. It does seem like a pretty useless change to me though. 

Should that rule be disabled?

src/invidious/helpers/i18next.cr
syeopite: Done. Sorry about that

src/invidious/routes/video_playback.cr
SamantazFox: Nah, in that case it's good to have a tool that unifies the syntax.

src/invidious/videos.cr
```diff
@@ -355,11 +370,12 @@ struct Video
   getset_i64 likes
   getset_i64 views
 
+  # TODO: Make predicate_bool the default as to adhere to Crystal conventions
   getset_bool allowRatings
   getset_bool authorVerified
   getset_bool isFamilyFriendly
   getset_bool isListed
-  getset_bool isUpcoming
+  predicate_bool upcoming, isUpcoming
```
SamantazFox, line 378: Why are you only changing that one? There are multiple other `is*` methods!

syeopite: This PR is mostly to just fix the Ameba issues which for some reason does not detect the other methods as violating `Naming/PredicateName`

Hence the TODO comment a few lines above

https://github.com/iv-org/invidious/blob/b75e5cfba6b4683307a5755506deefaebee5c3dd/src/invidious/videos.cr#L373

Do you want me to also resolve those in this PR?

SamantazFox: No, please keep that for a future PR, I'm worried it could cause too much conflicts with other open PRs!

SamantazFox (approved):

syeopite:
Rebased against master; and all CI has passed. Once merged it looks like Invidious will finally be Ameba compliant!

SamantazFox (approved):

SamantazFox:
Thanks for your Ameba compliance PR series!

PR #4816, opened 2024-08-04
Ignore .DS_Store

broquemonsieur:
A very common macOS cruft file. Help proprietary software devs contribute to free software!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /invidious
 /sentry
 /config/config.yml
+.DS_Store


syeopite (approved):
Well there's nothing wrong with the change itself.

But I'm personally of the opinion that the gitignore file should only exclude files relevant to Invidious and Crystal itself.

Anything else, especially files for a developers own specific environment should be excluded in `.git/info/exclude` imo.

Those environment specific files that are currently listed in the gitignore file are pretty much only there because that was what Omarroth added when he first created the repository.

I'll wait for some comments from the other Invidious team members before merging.

broquemonsieur:
I am fine using `.git/info/exclude`

PR #4667, opened 2024-04-30
HTML: Add rel="noreferrer noopener" to external links

ulmemxpoc:
Related to https://github.com/iv-org/invidious/issues/4297. CC: @SamantazFox.

Adds `rel="noreferrer noopener"` to constant YouTube links on Invidious frontend.
~~Also change to [`YT_URL`](https://github.com/iv-org/invidious/blob/eda7444ca46dbc3941205316baba8030fe0b2989/src/invidious.cr#L67).~~

Does not add  `rel="noreferrer noopener"` to:
- links in channel description
- links in video descriptions
- links in video comments

PR #4826, opened 2024-08-10
Confirmation before connecting to youtube.com/youtu.be

### LINKED ISSUE #4798: [Enhancement/Feature Request] Add confirmation box when leaving invidious to connect to youtube instead of immediately connecting to YouTube.

On at least 10 separate occasions, I have accidentally clicked/tapped **"Watch on YouTube (Embed)"** .

And also when I'm reading the comment section on links(text based web browser) using `&nojs=1` , I have accidentally clicked/tapped on **[YT]** which links to the comment on youtube .

I do not want to accidentally ping YouTube servers. Please implement a confirmation box before connecting to YouTube.

A confirmation box like...

** "You are leaving Invidious to visit <insert YouTube/comment link> . Are you Sure?" **

** "Yes" Or "No" **

Click "No" returns the page back to where the user was before.

I know modern web browsers(Firefox/Chromium) allow for editing the CSS, so I could just find the element and `{display: none;}` in Stylus, but this won't work in text based web browsers like links.

I realize this is entirely a user error(my error), and I've tried to be careful not to click YT links, but I've failed so many times now.
So I thought I'd propose this feature request. Please consider it.

Thanks!

unlxam:
Closes #4798 

 I also thought that since we are adding a confirmation before going to youtube.com/youtu.be in the "Watch on YouTube (Embed)" and "[YT]" in the comments. It would make sense adding also to external domains from links in the description and comments too. What you think about that?

### DIFF
--- a/src/invidious/routes/misc.cr
+++ b/src/invidious/routes/misc.cr
@@ -46,17 +46,14 @@ module Invidious::Routes::Misc
 
   def self.confirm_leave(env)
     locale = env.get("preferences").as(Preferences).locale
-    
-    link = env.params.query["link"]?
-    link = HTML.escape(link.to_s)
-
     referer = get_referer(env)
 
-    if link && !link.empty?
+    if env.params.query["link"]? && !env.params.query["link"].empty?
+      link = HTML.escape(env.params.query["link"].to_s)
+      
       templated "confirm_leave"
     else
       env.redirect "#{referer}"
     end
-
   end
 end


src/invidious/routes/misc.cr
```diff
@@ -43,4 +43,14 @@ module Invidious::Routes::Misc
     instance_url = fetch_random_instance
     env.redirect "https://#{instance_url}#{referer}"
   end
+
+  def self.confirm_leave(env)
+    locale = env.get("preferences").as(Preferences).locale
+    
+    link = env.params.query["link"]?
```
SamantazFox, line 50: Change 1: the link parameter is mandatory
Change 2: HTML escape the data, otherwise this introduces an XSS vulnerability

NB: you will need to add `require "html"` at the top of the file to make it work.

```suggestion
    link = HTML.escape(env.params.query["link"])
```

src/invidious/views/confirm_leave.ecr
```diff
@@ -0,0 +1,22 @@
+<% content_for "header" do %>
+<title><%= translate(locale, "Leave") %> - Invidious</title>
+<% end %>
+
+<div class="h-box">
+    <h3><%= translate(locale, "You are leaving Invidious. Continue to external link?") %></h3>
```
SamantazFox, line 6: You need to add this line to `locales/en-US.json`, otherwise Weblate won't be able to detect it, and in turn that won't be proposed to translators:

```json
"confirm_dialog_external_link": "You are leaving Invidious. Continue to external link?"
```

```suggestion
    <h3><%= translate(locale, "confirm_dialog_external_link") %></h3>
```

src/invidious/frontend/comments_youtube.cr
```diff
@@ -148,8 +150,10 @@ module Invidious::Frontend::Comments
         END_HTML
 
         if comments["videoId"]?
+          permalink = "https://www.youtube.com/watch?v=#{comments["videoId"]}&lc=#{child["commentId"]}"
+          permalink_confirm = "/confirm_leave?link=#{URI.encode_path(permalink)}"
           html << <<-END_HTML
-            <a href="https://www.youtube.com/watch?v=#{comments["videoId"]}&lc=#{child["commentId"]}" title="#{translate(locale, "YouTube comment permalink")}">[YT]</a>
+            <a href="#{permalink_confirm}" title="#{translate(locale, "YouTube comment permalink")}">[YT]</a>
```
SamantazFox, line 156: Can you do the same in the next if block (3 lines after this one)?

SamantazFox (changes requested):

src/invidious/comments/youtube.cr
```diff
@@ -303,6 +303,7 @@ module Invidious::Comments
     if format == "html"
       response = JSON.parse(response)
       content_html = Frontend::Comments.template_youtube(response, locale, thin_mode)
+      content_html = Comments.replace_external_links(content_html)
```
SamantazFox, line 306: Can you explain why you added this here? Is it because you missed one link to replace, [as mentioned here?](https://github.com/iv-org/invidious/pull/4826#discussion_r1712965531)

SamantazFox:
Also, I realized that there was https://github.com/iv-org/invidious/pull/4667 pending. I merged, so can you rebase your change on that please?

PR #4844, opened 2024-08-14
Update pl.json

temp-byte:


unixfox:
https://github.com/iv-org/invidious#translations

locales/pl.json
```diff
@@ -310,207 +410,90 @@
...
     "next_steps_error_message": "Po czym należy spróbować: ",
-    "next_steps_error_message_refresh": "Odśwież",
-    "next_steps_error_message_go_to_youtube": "Przejdź do YouTube",
-    "invidious": "Invidious",
-    "tokens_count_0": "{{count}} token",
-    "tokens_count_1": "{{count}} tokeny",
-    "tokens_count_2": "{{count}} tokenów",
-    "generic_videos_count_0": "{{count}} film",
-    "generic_videos_count_1": "{{count}} filmy",
-    "generic_videos_count_2": "{{count}} filmów",
```
SamantazFox, line 359: Do note that because polish has three plural forms (e.g: 1 miesiąc / 2 miesiące / 5 miesięcy, [as per the unicode standard](https://www.unicode.org/cldr/charts/42/supplemental/language_plural_rules.html#pl)) you need to keep the suffixes `_0`, `_1` and `_2`. If you use no suffix / `_plural`, then the strings won't be translated.

PR #4850, opened 2024-08-15
Videos: Convert URL before putting result into cache

### LINKED ISSUE #4837: [Bug] Don't cache videoplayback if player has been updated
Continuation of https://github.com/iv-org/invidious/pull/4772#issuecomment-2284243481

>> To also write it here: Video caching in invidious also caches and reuses data from player so if the player changes the video that is pulled from cache will cause 403s to be returned for videoplayback requests.

> Yeah, I should rewrite the URLs during parsing (i.e before caching) not during use.

@SamantazFox

SamantazFox:
Closes #4837


src/invidious/videos/parser.cr
```diff
@@ -132,10 +132,33 @@ def extract_video_info(video_id : String)
...
+
+        new_streaming_data[key] = JSON::Any.new(new_formats)
+      else
+        new_streaming_data[key] = value
+      end
+    end
+
+    params["streamingData"] = JSON::Any.new(new_streaming_data)
+  end
+
```
straight-shoota, line 161: **suggestion:** It's entirely possible to mutate data in a `JSON::Any`. So I think this can be simplified quite a lot to something like this:
```cr
  if streaming_data = player_response["streamingData"]?
    %w[formats adaptiveFormats].each do |key|
      streaming_data.as_h[key]?.try &.as_a.each do |format|
        format.as_h["url"] = JSON::Any.new(convert_url(format))
      end
    end
  end
```

SamantazFox: Thanks a lot! Now that seems obvious ^^

PR #4851, opened 2024-08-15
add more parameter to look a lot like official youtube client

unixfox:
Taken from a request done by youtube.js: https://github.com/LuanRT/YouTube.js and I tried to add as much new parameters as possible.

[example_yt_request_youtubejs.txt](https://github.com/user-attachments/files/16627563/example_yt_request_youtubejs.txt)

We are missing:
```
       "mainAppWebInfo"     => {
          "graftUrl"                  => "https://www.youtube.com",
          "isWebNativeShareAvailable" => true,
          "pwaInstallabilityStatus"   => "PWA_INSTALLABILITY_STATUS_UNKNOWN",
          "webDisplayMode"            => "WEB_DISPLAY_MODE_BROWSER",
        },
```
inside `client_context = {` but I can't get my head around the compiler issue:
```
Done checking player dependencies, now compiling Invidious...
Showing last frame. Use --error-trace for full trace.

In src/invidious/yt_backend/youtube_api.cr:320:33

 320 | "mainAppWebInfo"     => {
                               ^----------
Error: expected argument #2 to 'Hash(String, Int64 | String)#[]=' to be (Int64 | String), not Hash(String, Bool | String)

Overloads are:
 - Hash(K, V)#[]=(key : K, value : V)
```

Other parameters we are missing:
- timeZone: in youtube.js this is dynamically generated from the actual timezone but I don't know if we should or just set a default one
- utcOffsetMinutes: same as above
- visitorData: this one is tricky because it requires to generate a visitordata from scratch
- remoteHost: we will have to fetch the public IP of the server at the start; not sure how
- "configInfo"."appInstallData": like visitordata, it's an ID

Remarks:
- Some parameters do not make sense in the context of the ANDROID client but I don't think this will do any harm like the size for screenHeightPoints or memoryTotalKbytes
- I required useragent to be passed because youtube is expecting it now on all the requests whatever the client name

### DIFF
--- a/src/invidious/yt_backend/youtube_api.cr
+++ b/src/invidious/yt_backend/youtube_api.cr
@@ -665,8 +665,6 @@ module YoutubeAPI
       "x-goog-api-format-version" => "2",
       "x-youtube-client-name"     => client_config.name_proto,
       "x-youtube-client-version"  => client_config.version,
-      "accept"                    => "*",
-      "accept-language"           => "*",
     }
 
     if user_agent = client_config.user_agent
--- a/src/invidious/yt_backend/youtube_api.cr
+++ b/src/invidious/yt_backend/youtube_api.cr
@@ -21,6 +21,10 @@ module YoutubeAPI
   private IOS_USER_AGENT  = "com.google.ios.youtube/19.16.3 (iPhone14,5; U; CPU iOS 17_4 like Mac OS X;)"
   private IOS_VERSION     = "17.4.0.21E219" # Major.Minor.Patch.Build
 
+  # From reddit: https://web.archive.org/web/20210506123457/https://www.reddit.com/r/youtube/comments/n660l1/how_to_access_tv_mode_in_2020/
+  private YOUTUBE_TV_USER_AGENT = "Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/538.2 (KHTML, like Gecko) Large Screen Safari/538.2 LG Browser/7.00.00(LGE; 24LF4820-BU; 03.20.14; 1; DTV_W15L); webOS.TV-2015; LG NetCast.TV-2013 Compatible (LGE, 24LF4820-BU, wireless),gzip(gfe)"
+
+  private WEB_USER_AGENT  = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36,gzip(gfe)"
   private WINDOWS_VERSION = "10.0"
 
   # Enumerate used to select one of the clients supported by the API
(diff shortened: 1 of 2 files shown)


src/invidious/yt_backend/youtube_api.cr
```diff
@@ -459,8 +503,14 @@ module YoutubeAPI
...
-      "referer"         => "https://www.youtube.com/watch?v=#{video_id}",
+      "html5Preference"       => "HTML5_PREF_WANTS",
+      "referer"               => "https://www.youtube.com/watch?v=#{video_id}",
+      "lactMilliseconds"      => "-1",
+      "splay"                 => false,
+      "vis"                   => 0,
+      "autoCaptionsDefaultOn" => false,
+      "autonavState"          => "STATE_ON",
+      "currentUrl"            => "/watch?v=#{video_id}",
     } of String => String | Int64
```
SamantazFox, line 514: You need to add `| Bool` here to solve your compiler problem ^^

SamantazFox:
> Other parameters we are missing:
> 
>     * timeZone: in youtube.js this is dynamically generated from the actual timezone but I don't know if we should or just set a default one
>     * utcOffsetMinutes: same as above
>     * remoteHost: we will have to fetch the public IP of the server at the start; not sure >     * "configInfo"."appInstallData": like visitordata, it's an ID

As far as I know, these can be safely omitted. On some browsers with stronger privacy/adblocking features (like Tor), I've often seen these empty. 

> Some parameters do not make sense in the context of the ANDROID client but I don't think this will do any harm like the size for screenHeightPoints or memoryTotalKbytes

Maybe it's time to split the context logic by client, like NewPipe does?

src/invidious/yt_backend/youtube_api.cr
```diff
@@ -610,6 +665,8 @@ module YoutubeAPI
       "x-goog-api-format-version" => "2",
       "x-youtube-client-name"     => client_config.name_proto,
       "x-youtube-client-version"  => client_config.version,
+      "accept"                    => "*",
+      "accept-language"           => "*",
```
SamantazFox, line 669: U suspect these needs to coincide with what's already set in `youtube_pool.cr`

unixfox: good point, I have switched `accept` to `*` like on the official youtube.com website.

unixfox:
> > Other parameters we are missing:
> > ```
> > * timeZone: in youtube.js this is dynamically generated from the actual timezone but I don't know if we should or just set a default one
> > * utcOffsetMinutes: same as above
> > * remoteHost: we will have to fetch the public IP of the server at the start; not sure >     * "configInfo"."appInstallData": like visitordata, it's an ID
> > ```
> 
> As far as I know, these can be safely omitted. On some browsers with stronger privacy/adblocking features (like Tor), I've often seen these empty.

ok then we won't set those.

> > Some parameters do not make sense in the context of the ANDROID client but I don't think this will do any harm like the size for screenHeightPoints or memoryTotalKbytes
> 
> Maybe it's time to split the context logic by client, like NewPipe does?

Kinda don't have the time to do it today. I added all the parameters that I found on youtube.js because I have a separate server for invidious that exclusively use youtube.js and this one has never seen any issue about "watch on the latest version".

unixfox:
@SamantazFox can we merge into master branch?

src/invidious/yt_backend/youtube_api.cr
```diff
@@ -46,39 +50,51 @@ module YoutubeAPI
...
-      platform:   "MOBILE",
+      name:            "MWEB",
+      name_proto:      "2",
+      version:         "2.20240304.08.00",
+      os_name:         "Android",
+      os_version:      ANDROID_VERSION,
+      browser_name:    "Chrome",
+      browser_version: "128.0.0.0",
+      platform:        "MOBILE",
+      user_agent:      WEB_USER_AGENT,
```
SamantazFox, line 85: We need a different UA here. "Windows NT 10.0" won't cut it I guess!

src/invidious/yt_backend/youtube_api.cr
```diff
@@ -46,39 +50,51 @@ module YoutubeAPI
...
-      os_version: WINDOWS_VERSION,
-      platform:   "DESKTOP",
+      name:            "WEB",
+      name_proto:      "1",
+      version:         "2.20240304.00.00",
+      screen:          "WATCH_FULL_SCREEN",
+      os_name:         "Windows",
+      os_version:      WINDOWS_VERSION,
+      browser_name:    "Chrome",
+      browser_version: "128.0.0.0",
```
SamantazFox, line 60: can you put that in a constant at the top? It will be easier to maintain in the future!

unixfox:
Closing as we have moved to companion.

syeopite:
Wouldn't this still be applicable to the other Inner tube requests?

unixfox:
Maybe but youtube never really blocked anyone on the other endpoints, except when doing too many requests but I don't think this PR could have solved the issue.

We should from time to time update the version number though.

PR #4859, opened 2024-08-19
API: Proxy formatStreams URLs too

colinleroy:
Hello,

I have noticed that the /api/v1/videos endpoint does not proxify the formatStreams URLs when passed ?local=true, whereas the adaptiveFormats URLs are correctly proxified. 

The Web UI does proxify when clicking Download with fmt=18 for example, so I suppose this is an oversight. This PR aims to fix that.

Hope this helps.


SamantazFox (approved):

unixfox (approved):

SamantazFox:
Thank you for fixing this!

PR #4871, opened 2024-08-22
removed version line for docker-compose.yml, docker version is obsolete

volumedata21:
removed the line `version: "3"` in the docker-compose.yml file. Specifying version in a docker compose is now obsolete and will throw a warning message when launching a docker compose.


SamantazFox (commented):
In theory that should be harmless, but I'm worried about people on systems that still ship the older python-based `docker-compose`

unixfox:
I said previous times that I don't want to drop the backward compatibility yet.

PR #4923, opened 2024-09-16
use web screen embed for fixing potoken functionality

unixfox:
Helps for #4734

Since potoken is generated from a YouTube embed page, YouTube has started to enforce this token on exclusive usage of WEB_EMBEDDED_PLAYER client.

Thanks to `"embedUrl":"https://www.google.com/"`, this allows to have WEB_EMBEDDED_PLAYER working even on video that disallowed the embedding. As tested on https://www.youtube.com/watch?v=JfJYHfrOGgQ

I did found this parameter by entering the youtube video url on google.com and seeing that it worked there but not on the youtube embed page.

image

This also deactivates TVHTML5_SIMPLY_EMBEDDED_PLAYER when po_token is configured, since this doesn't work with that client.

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -53,6 +53,8 @@ end
 def extract_video_info(video_id : String)
   # Init client config for the API
   client_config = YoutubeAPI::ClientConfig.new
+  # use WEB screen embed for potoken functionality
+  client_config.client_type = YoutubeAPI::ClientType::WebEmbeddedPlayer
 
   # Fetch data from the player endpoint
   player_response = YoutubeAPI.player(video_id: video_id, params: "2AMB", client_config: client_config)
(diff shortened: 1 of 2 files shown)


unixfox:
@SamantazFox @syeopite I hope you don't mind that I merged this PR very quickly. I found a fix for the new issue https://github.com/iv-org/invidious/issues/4734 and I wanted the public to test it as soon as possible.

We can look into improving the code latter one.

unixfox:
I reverted the commit. It doesn't work as expected.

The issue being that WEB screen embed doesn't provide microformat info so this fails to work. Even WEB on a blocked IP doesn't provide it either so we will have to use next endpoint in order to fetch all the information that we previously extracted from microformat (description, publishedDate and more).

unixfox:
Someone on matrix said that embed www.google.com does not work for all videos. Example for https://www.youtube.com/watch?v=0RyInjfgNc4

rustynail1984:
@unixfox indeed, not all videos working on embed player. So i researched to find a working alternative.. please see it here https://github.com/LuanRT/YouTube.js/issues/748#issuecomment-2348418093

PR #4928, opened 2024-09-19
use WEB_CREATOR when po_token with WEB_EMBED as a fallback

unixfox:
This is a continuation of the work done in https://github.com/iv-org/invidious/pull/4923

- This uses WEB_CREATOR client for fetching the streams data when po_token is configured. Thanks @rustynail1984 for the notification
- WEB_EMBEDDED_PLAYER is configured as a fallback for po_token in case WEB_CREATOR doesn't work anymore
- Since neither WEB_CREATOR nor WEB_EMBEDDED_PLAYER returns microformat data, I have configured so Invidious does not crash if it doesn't find it anymore. This means that we will be missing some info until we switch the retrieval of this info from the next endpoint.  
  But this is better than nothing that works.
- TvHtml5ScreenEmbed won't be used when po_token is configured since it doesn't work with it.

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -108,7 +108,6 @@ def extract_video_info(video_id : String)
 
   # Second try in case WEB_EMBEDDED_PLAYER doesn't work with po_token.
   # Only trigger if reason found and po_token configured.
-  # TvHtml5ScreenEmbed now requires sig helper for it to work but doesn't work with po_token.
   if reason && CONFIG.po_token
     client_config.client_type = YoutubeAPI::ClientType::WebEmbeddedPlayer
     new_player_response = try_fetch_streaming_data(video_id, client_config)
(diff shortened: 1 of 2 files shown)


src/invidious/videos/parser.cr
```diff
@@ -102,6 +106,14 @@ def extract_video_info(video_id : String)
 
   new_player_response = nil
 
+  # Second try in case WEB_EMBEDDED_PLAYER doesn't work with po_token.
+  # Only trigger if reason found and po_token configured.
+  # TvHtml5ScreenEmbed now requires sig helper for it to work but doesn't work with po_token.
```
syeopite, line 111: This comment seems unrelated here

```suggestion
```

syeopite (approved):

rustynail1984:
Found the latest verion in YouTube Studio.

```
clientName:  62 // corresponds to WEB_CREATOR
clientVersion:  "1.20240918.03.00"
```

PR #4931, opened 2024-09-20
Logger: Add color support for different log levels

Fijxu:
Looks like this:

image

As I said in the code, this is useful when debugging things in Invidious. Having to find the type of error just by the name of the log level is not comfortable in my opinion.

Feel free to edit anything.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -226,11 +226,13 @@ https_only: false
 ## Enables colors in logs. Useful for debugging purposes
 ## This is overridden if "-k" or "--colorize" 
 ## are passed on the command line.
+## Colors are also disabled if the environment variable
+## NO_COLOR is present and has any value
 ##
 ## Accepted values: true, false
-## Default: false
+## Default: true
 ##
-#colorize_logs: false
+#colorize_logs: true
 
 # -----------------------------
 #  Features
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -224,7 +224,7 @@ https_only: false
 
 ##
 ## Enables colors in logs. Useful for debugging purposes
-## This is overridden if "-k" or "--colorize" 
+## This is overridden if "-k" or "--colorize"
 ## are passed on the command line.
 ## Colors are also disabled if the environment variable
 ## NO_COLOR is present and has any value
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -232,7 +232,7 @@ https_only: false
 ## Accepted values: true, false
 ## Default: true
 ##
-#colorize_logs: true
+#colorize_logs: false
 
 # -----------------------------
 #  Features
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -222,6 +222,15 @@ https_only: false
 ##
 #log_level: Info
 
+##
+## Enables colors in logs. Useful for debugging purposes
+## This is overridden if "-k" or "--colorize" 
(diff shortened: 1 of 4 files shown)


src/invidious/helpers/logger.cr
```diff
@@ -10,7 +12,7 @@ enum LogLevel
 end
 
 class Invidious::LogHandler < Kemal::BaseLogHandler
-  def initialize(@io : IO = STDOUT, @level = LogLevel::Debug)
+  def initialize(@io : IO = STDOUT, @level = LogLevel::Debug, @color : Bool = true)
```
syeopite, line 15: Pretty minor nitpick but I'll prefer if you can change `@color` to something that doesn't collide with `#color`. Such as `@use_color`.

It is valid Crystal code but it might cause some confusion to someone starting out with Crystal, or when switching from another language.

config/config.example.yml
```diff
@@ -222,6 +222,15 @@ https_only: false
...
 
+##
+## Enables colors in logs. Useful for debugging purposes
+## This is overridden if "-k" or "--colorize" 
+## are passed on the command line.
+##
+## Accepted values: true, false
+## Default: false
+##
+#colorize_logs: false
```
syeopite, line 233: I think this can just be enabled by default

syeopite (approved):
LGTM. I'll await some reviews from everyone else to see if the colors should be changed before merging.

Fijxu:
Done, I also added information about the `NO_COLOR` env variable.
Color output also works on docker logs which is pretty neat.

image

unixfox:
I don't think it's a good idea to have this as enabled by default.

I can see the color messing up with some existing log parsing system or something like that.

Fijxu:
> I can see the color messing up with some existing log parsing system or something like that.

Yeah you are right. This could break some automation system that does not support ANSI escape codes. It's better to leave it off by default and it would be great to point it on the wiki that this could be turned on (probably because most people would like to leave this setting on) if you want.

src/invidious/helpers/logger.cr
```diff
@@ -39,10 +41,22 @@ class Invidious::LogHandler < Kemal::BaseLogHandler
...
+    when LogLevel::Fatal then :magenta
+    else                      :default
+    end
+  end
+
   {% for level in %w(trace debug info warn error fatal) %}
     def {{level.id}}(message : String)
       if LogLevel::{{level.id.capitalize}} >= @level
-        puts("#{Time.utc} [{{level.id}}] #{message}")
+        puts("#{Time.utc} [{{level.id}}] #{message}".colorize(color(LogLevel::{{level.id.capitalize}})).toggle(@use_color))
```
SamantazFox, line 59: follow-up to my previous comment:
```suggestion
        puts("#{Time.utc} [{{level.id}}] #{message}".colorize(color(LogLevel::{{level.id.capitalize}}))
```

config/config.example.yml
```diff
@@ -222,6 +222,17 @@ https_only: false
 ##
 #log_level: Info
 
+##
+## Enables colors in logs. Useful for debugging purposes
+## This is overridden if "-k" or "--colorize"
```
SamantazFox, line 227: Also, git reports a trailing white space here!

SamantazFox (approved):

SamantazFox:
I forgot to say thanks for your contribution :)

PR #4984, opened 2024-10-06
Videos: use WEB client instead of WEB CREATOR

unixfox:
This switch to using WEB client when a potoken is configured. Otherwise try with Android test suite if there is no potoken configured.

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -53,10 +53,6 @@ end
 def extract_video_info(video_id : String)
   # Init client config for the API
   client_config = YoutubeAPI::ClientConfig.new
-  # Use the WEB when po_token is configured
-  if CONFIG.po_token
-    client_config.client_type = YoutubeAPI::ClientType::Web
-  end
 
   # Fetch data from the player endpoint
   player_response = YoutubeAPI.player(video_id: video_id, params: "2AMB", client_config: client_config)
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -53,9 +53,9 @@ end
 def extract_video_info(video_id : String)
   # Init client config for the API
   client_config = YoutubeAPI::ClientConfig.new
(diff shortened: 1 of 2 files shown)


SamantazFox (approved):

syeopite (approved):

unixfox:
Switching to a draft since after a while (5 minutes) all the requests get blocked. But I guess it's better than nothing since ANDROID_TEST_SUITE is straight blocked.

unixfox:
On matrix, iOS is the only candidate for keeping a working condition without po_token.

unixfox:
PR description updated

src/invidious/videos/parser.cr
```diff
@@ -53,9 +53,9 @@ end
 def extract_video_info(video_id : String)
   # Init client config for the API
   client_config = YoutubeAPI::ClientConfig.new
-  # Use the WEB_CREATOR when po_token is configured because it fully only works on this client
+  # Use the WEB when po_token is configured
   if CONFIG.po_token
-    client_config.client_type = YoutubeAPI::ClientType::WebCreator
+    client_config.client_type = YoutubeAPI::ClientType::Web
   end
```
syeopite, line 59: The default client type is already WEB so this section can be removed

unixfox: doesn't this allow to explicitly say that we use WEB? I was thinking the same, though. I don't know.

syeopite: I mean the logic will also default to the WEB client when the conditional fails. Its moot to include that.

Might be easier to just do 

```crystal
  if CONFIG.po_token
    client_config.client_type = YoutubeAPI::ClientType::Web
  else
    client_config.client_type = YoutubeAPI::ClientType::AndroidTestSuite
  end
```

As so we only have a single request. Though I guess the android client also doesn't return all of the data we need

syeopite (approved):

SamantazFox (approved):

SamantazFox:
Thank you for reverting this, it was much needed!

PR #4992, opened 2024-10-09
Videos: Fix missing host parameter on playback URLs when `local=true`

SamantazFox:
In #4850, the `host=` parameter ceased to be added to videoplayback URLs by default, as it should only be added when the video is proxied through invidious.

This change revealed that some parts of the code didn't properly "proxify" the playback URL (= adding back that `host=` parameter) when `local=true` was passed to those endpoints.

Thanks to @Fijxu for raising that problem!

### DIFF
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -27,28 +27,21 @@ module Invidious::Routes::API::Manifest
         haltf env, status_code: response.status_code
       end
 
-      manifest = response.body
-
-      manifest = manifest.gsub(/<BaseURL>[^<]+<\/BaseURL>/) do |baseurl|
-        url = baseurl.lchop("<BaseURL>")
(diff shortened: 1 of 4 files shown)


syeopite (approved):

fpletz (approved):
Fixes playback on my invidious instance.

unixfox (approved):

Fijxu:
Oh it got merged. Did you guys tried livestreams? I had to revert https://github.com/iv-org/invidious/pull/4992/files#diff-8e4c5ae2068b27382f39df8cf4ebcb0a30c6252c91d4fe94cc3f71add850fac3L180-L181 because it didn't work for livestreams. (Don't test on my instance, I recently changed something and livestreams are broken)

I tested it locally with the current standard setup of invidious (tokens and inv_sig_helper), no custom source code.

image

Issue #4651, opened 2024-04-26
[Enhancement] Unify "recommended steps" section for error pages


syeopite:
**Is your enhancement request related to a problem? Please describe.**

The recommended steps section on the generic error page returned via InfoException is different from the recommended steps section returned from the standard error page.  

**Describe the solution you'd like**

For these error pages to share the same section instead of having two different ones.

**Describe alternatives you've considered**

Keep it as-is.

**Additional context**

Standard exceptions
![recommended steps on most error pages](https://github.com/iv-org/invidious/assets/70992037/e8299aee-8d7e-4fd7-80ca-2ccd85201c1c)

Info exceptions:

[...]

PR #5003, opened 2024-10-13
trying to only use next endpoint for the major data

### LINKED ISSUE #4073: [Bug] Unavailable videos break playlists
**Describe the bug**

Unavailable videos show an error page with no option to continue playing a playlist. There should be a way to skip to the next video.

Related: #3862.

**Steps to Reproduce**

Open a playlist with an unavailable video, for example: https://yewtu.be/watch_videos?video_ids=JNy_hCQTQVg,jNQXAC9IVRw

screenshot

### LINKED ISSUE #4744: [Feature request / Enhancement] Allow viewing video info when Youtube blocks video access for signed-out users.
Creating a new issue thread for this suggestion as per @unixfox's order.

As I mentioned in #4734, it seems that when Youtube blocks access for signed-out users with the message mentioned in the main comment of the linked issue, Youtube only seems to block access to the actual video player and not the remainder of the page that contains most of the video info.

With that in mind, it might be possible to change Invidious's code to only disable the video player but show the remaining page with the video info when Invidious encounters the signed-out user block on the Youtube webpages. Currently, Invidious only shows the "This helps protect our community." error and nothing else. This would solve all problems for the Invidious users that are not interested in watching the video directly on the Invidious instance, but only in viewing the video info/comments and/or obtaining the actual video data through external means such as yt-dlp.

Of course, this won't be necessary if the actual solution to bypassing the signed-out access block will be found sooner or if Youtube stops engaging in blocking access in this way altogether.

**Additional context**
I am not an expert on web scraping or Youtube/Invidious's inner workings, so this may not be possible at all. However, if it is possible and if it will be implemented sooner than a true fix to Youtube's access blocking, I think this would solve a big chunk of problems for a big chunk of Invidious users.

Thank you.

### LINKED ISSUE #5435: Be more informative when youtube says that a video is private
When navigating to a URL like

https://www.youtube.com/watch?v=4w8ua3cB_nA

YouTube says 

```
Private video
Sign in if you've been granted access to this video
```

Navigating to the same video on an invidious instance results in the feedback:

```
Sign in if you've been granted access to this video

After which you should try to:

    [Refresh]
    [Switch Invidious Instance]
    [Go to YouTube]
```

The most informative bit of the YouTube text is, in my view, "private video". Knowing that this is what YouTube is saying should make it more or less clear that switching invidious instance probably isn't a way to access the video.

It'd be great if invidious prominently displayed the words "private video" in cases like this, or even something like "YouTube is saying that this video is private and that you should sign in if you've been granted access to this video".

unixfox:
## Description
This PR aims to avoid throwing a raw error and instead display all the data that we can extract from the /next endpoint but without displaying the "player" (video.js) element.

This is one big step towards https://github.com/iv-org/invidious/pull/4985. Invidious will have to be able to display the watch page even when YouTube is blocking the IP.

## What changed?

- The watch page is displayed properly but without the player if an error is found.
- All the important data like title, viewcount is fetched from /next endpoint when unavailable from the /player endpoint.
- The error message now throws the reason AND the subreason. Makes it clearer to the user what's really happening.
- Works without Invidious Companion (falls back to /next-only data).

## TODO
- [x] Integrate the embed page with the changes
- [x] Display the subreason on the watch page when there is an error
- [x] Fix description (nil safety in JSON::Any)
- [x] Add mocks for a video without videoDetails (requires iv-org/mocks#4)

## Try these videos ID for testing - on both a blocked IP and a not blocked one

- 98TyXNPdMRo (private video)
- 7KSfDyainYU (georestricted video)
- KIuDBxqz__k (familySafe no)
- yqaS07znJWY (unlisted video)
- aaaaaaaaaaaa (notfound video)

## Fixes 

- Closes #4744
- closes #4073
- Closes #5435

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -15,7 +15,7 @@ struct Video
   # NOTE: don't forget to bump this number if any change is made to
   # the `params` structure in videos/parser.cr!!!
   #
-  SCHEMA_VERSION = 3
+  SCHEMA_VERSION = 4
 
   property id : String
 
@@ -182,6 +182,10 @@ struct Video
     info["reason"]?.try &.as_s
   end
 
+  def subreason : String?
+    info["subreason"]?.try &.as_s
+  end
+
   def music : Array(VideoMusic)
     info["music"].as_a.map { |music_json|
       VideoMusic.new(
@@ -314,7 +318,7 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
     end
   else
     video = fetch_video(id, region)
-    Invidious::Database::Videos.insert(video) if !region
+    Invidious::Database::Videos.insert(video) if !region && video.reason.nil?
   end
 
   return video
@@ -336,12 +340,8 @@ def fetch_video(id, region)
   end
 
   if reason = info["reason"]?
-    if reason == "Video unavailable"
+    if reason == "Video unavailable" && !info["title"]?
       raise NotFoundException.new(reason.as_s || "")
-    elsif !reason.as_s.starts_with? "Premieres"
-      # dont error when it's a premiere.
-      # we already parsed most of the data and display the premiere date
-      raise InfoException.new(reason.as_s || "")
     end
   end
 
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -318,7 +318,7 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
     end
   else
     video = fetch_video(id, region)
-    Invidious::Database::Videos.insert(video) if !region && video.reason.nil?
+    Invidious::Database::Videos.insert(video) if !region
   end
 
   return video
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -310,7 +310,8 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
        video.schema_version != Video::SCHEMA_VERSION # cache control
       begin
         video = fetch_video(id, region)
-        Invidious::Database::Videos.update(video)
+        Invidious::Database::Videos.update(video) if CONFIG.invidious_companion.present?
+>>>>>>> 465a2cc6 (Skip caching entirely when Invidious Companion is not configured)
       rescue ex
         Invidious::Database::Videos.delete(id)
         raise ex
@@ -318,7 +319,8 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
     end
   else
     video = fetch_video(id, region)
-    Invidious::Database::Videos.insert(video) if !region
+    Invidious::Database::Videos.insert(video) if !region && CONFIG.invidious_companion.present?
+>>>>>>> 465a2cc6 (Skip caching entirely when Invidious Companion is not configured)
   end
 
   return video
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -311,7 +311,6 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
       begin
         video = fetch_video(id, region)
         Invidious::Database::Videos.update(video) if CONFIG.invidious_companion.present?
->>>>>>> 465a2cc6 (Skip caching entirely when Invidious Companion is not configured)
       rescue ex
         Invidious::Database::Videos.delete(id)
         raise ex
@@ -320,7 +319,6 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
   else
     video = fetch_video(id, region)
     Invidious::Database::Videos.insert(video) if !region && CONFIG.invidious_companion.present?
->>>>>>> 465a2cc6 (Skip caching entirely when Invidious Companion is not configured)
   end
 
   return video
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -179,11 +179,13 @@ struct Video
   end
 
   def reason : String?
-    info["reason"]?.try &.as_s
+    # Use `as_s?` so a JSON `null` (stored by some error paths) yields `nil`
+    # instead of raising a TypeCastError.
+    info["reason"]?.try &.as_s?
   end
 
   def subreason : String?
-    info["subreason"]?.try &.as_s
+    info["subreason"]?.try &.as_s?
   end
 
   def music : Array(VideoMusic)
@@ -310,7 +312,7 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
        video.schema_version != Video::SCHEMA_VERSION # cache control
       begin
         video = fetch_video(id, region)
-        Invidious::Database::Videos.update(video) if CONFIG.invidious_companion.present?
+        Invidious::Database::Videos.update(video)
       rescue ex
         Invidious::Database::Videos.delete(id)
         raise ex
@@ -318,7 +320,7 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
     end
   else
     video = fetch_video(id, region)
-    Invidious::Database::Videos.insert(video) if !region && CONFIG.invidious_companion.present?
+    Invidious::Database::Videos.insert(video) if !region
   end
 
   return video
@@ -329,21 +331,9 @@ rescue DB::Error
 end
 
 def fetch_video(id, region)
+  # `extract_video_info` always returns a hash (it raises `NotFoundException`
+  # for genuinely unavailable videos), so there is no nil case to handle here.
   info = Invidious::Videos::Parser.extract_video_info(video_id: id)
-
-  if info.nil?
-    raise InfoException.new("Invidious companion is not available. \
-    Video playback cannot continue. \
-    If you are the administrator of this instance, install Invidious companion \
-    following the installation instructions \
-    <a href=\"https://docs.invidious.io/installation/\">https://docs.invidious.io/installation/</a>")
-  end
-
-  if reason = info["reason"]?
-    if reason == "Video unavailable" && !info["title"]?
-      raise NotFoundException.new(reason.as_s || "")
-    end
-  end
 
   video = Video.new({
     id:      id,
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -102,23 +102,23 @@ module Invidious::Videos::Parser
 
     # Fetch the /next endpoint for additional data.
     begin
-        next_response = YoutubeAPI.next({"videoId": video_id, "params": ""})
-        # Remove the microformat returned by the /next endpoint on some videos
-        # to prevent player_response microformat from being overwritten.
-        next_response.delete("microformat")
-        player_response = player_response.merge(next_response)
-      rescue ex
-        # If we're in an error state and /next fails, return what we have
-        if reason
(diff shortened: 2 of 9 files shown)


ic-scm:
Thank you very much for this. Looking at instance yewtu.be which uses [invidious-custom](https://github.com/yewtudotbe/invidious-custom), it seems to be working just as expected, exactly as described in my linked issue, and with a 100% success rate.

Thank you for this great work - this is a huge step towards making Invidious convenient to use again. For me personally, it solves the entire problem.

The only thing I can potentially suggest is for the video thumbnail to be loaded and displayed on the page, separately from the broken video player. But that would be just a nice-to-have feature, and I'm guessing it may not be trivial to implement, when the official Youtube website just shows a black error screen, with the thumbnail nowhere to be seen.

src/invidious/videos.cr
```diff
@@ -326,13 +326,18 @@ end
 def fetch_video(id, region)
   info = extract_video_info(video_id: id)
 
-  if reason = info["reason"]?
+  if info["reason"]? && info["subreason"]?
+    reason = info["reason"].as_s
+    puts info
```
syeopite, line 331: Please remove the debug puts here

src/invidious/helpers/errors.cr
```diff
@@ -86,8 +82,13 @@ def error_template_helper(env : HTTP::Server::Context, status_code : Int32, mess
...
 
-  error_message = translate(locale, message)
-  next_steps = error_redirect_helper(env)
+  error_message = <<-END_HTML
+    <div class="error_message">
+      <h2>#{translate(locale, "error_processing_data_youtube")}</h2>
+      <p>#{translate(locale, message)}</p>
+      #{error_redirect_helper(env)}
+    </div>
+  END_HTML
```
syeopite, line 91: Your changes here will show the `Error while processing the data sent by YouTube` message on any `InfoException` while also removing the next steps message on all `InfoException`s

And `InfoException`s are raised in places unrelated to YouTube as well such as when a user inserts the wrong password. 

Related: #4651

src/invidious/videos/parser.cr
```diff
@@ -226,17 +230,34 @@ def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any
...
+    published_txt = video_primary_renderer
+      .try &.dig?("dateText", "simpleText")
+
+    if published_txt.try &.as_s.includes?("ago") && !published_txt.nil?
+      published = decode_date(published_txt.as_s.lchop("Started streaming "))
+    elsif published_txt && published_txt.try &.as_s.matches?(/(\w{3} \d{1,2}, \d{4})$/)
+      published = Time.parse(published_txt.as_s.match!(/(\w{3} \d{1,2}, \d{4})$/)[0], "%b %-d, %Y", Time::Location::UTC)
+    else
+      published = Time.utc
+    end
```
SamantazFox, line 249: Here I'd recommend using `"scheduledStartTime"` (see below) in the player response, as it's more accurate and doesn't need complex parsing logic.

```json
{
	"playabilityStatus": {
		"status": "LIVE_STREAM_OFFLINE",
		"reason": "This live event will begin in 32 hours.",
		"playableInEmbed": true,
		"liveStreamability": {
			"liveStreamabilityRenderer": {
				"videoId": "hUoukmX_BRQ",
				"offlineSlate": {
					"liveStreamOfflineSlateRenderer": {
						"scheduledStartTime": "1729468800",
						"mainText": {
							"runs": [
								{"text": "Live in "},
                                {"text": "32 hours"}
							]
						},
						"subtitleText": {
							"simpleText": "October 21 at 12:00 AM"
						}
					}
				}
			}
		}
	}
}
```

SamantazFox: Note: if you plan to fallback to the `"subtitleText"` value, there is a non-breaking space before AM/PM:
image

src/invidious/videos/parser.cr
```diff
@@ -247,8 +268,24 @@ def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any
...
+  if video_badges = video_primary_renderer.try &.dig?("badges")
+    if has_unlisted_badge?(video_badges)
+      is_listed ||= false
+    else
+      is_listed ||= true
+    end
+    # if no badges but videoDetails not available then assume isListed
+  else
+    is_listed ||= true
+  end
```
SamantazFox, line 287: Here you need to check for nil-ness, otherwise these two `is_listed ||= true` will almost always override a previous `false` value.

```suggestion
  if is_listed.nil?
    if video_badges = video_primary_renderer.try &.dig?("badges")
      is_listed = !has_unlisted_badge?(video_badges)
    else
      # If video has no badges and videoDetails is not
      # available, then assume isListed
      is_listed = true
    end
  end
```

src/invidious/views/embed.ecr
```diff
@@ -31,7 +31,15 @@
 %>
 </script>
 
-<%= rendered "components/player" %>
+<% if video.reason.nil? %>
+<h3>
+    <%= video.reason %>
+</h3>
+<% else %>
+<div id="player-container" class="h-box">
```
SamantazFox, line 39: I'd prefer to keep a placeholder for the player and put the error message in the middle of that black square (i.e a `p` inside that `#player-container`), like what Youtube does. What do you think?

src/invidious/views/watch.ecr
```diff
@@ -96,7 +98,10 @@ we're going to need to do it here in order to allow for translations.
 
     <% if video.reason %>
         <h3>
-            <%= video.reason %>
+            <%= translate(locale, "error_from_youtube_unplayable") %> <%= video.reason %>
```
SamantazFox, line 101: Ditto: keeping the error message inside a player placeholder.

src/invidious/videos.cr
```diff
@@ -326,13 +326,18 @@ end
...
+    else
+      subreason = "No additional reason"
+    end
     if reason == "Video unavailable"
-      raise NotFoundException.new(reason.as_s || "")
-    elsif !reason.as_s.starts_with? "Premieres"
-      # dont error when it's a premiere.
-      # we already parsed most of the data and display the premiere date
-      raise InfoException.new(reason.as_s || "")
+      raise NotFoundException.new(reason + ": Video not found" || "")
```
SamantazFox, line 338: That will expand to `"Video unavailable: Video not found"`, which is not very explicit, imo.

src/invidious/videos.cr
```diff
@@ -326,13 +326,18 @@ end
...
 
-  if reason = info["reason"]?
+  if info["reason"]? && info["subreason"]?
+    reason = info["reason"].as_s
+    puts info
+    if info.dig?("subreason").nil?
+      subreason = info["subreason"].as_s
+    else
+      subreason = "No additional reason"
+    end
```
SamantazFox, line 336: This bloc of code can be summarized to the following, as you're already checking `info["subreason"]?` falsiness in the parent `if`
```suggestion
    subreason = info["subreason"].as_s
```

src/invidious/videos.cr
```diff
@@ -313,7 +313,7 @@ def get_video(id, refresh = true, region = nil, force_refresh = false)
     end
   else
     video = fetch_video(id, region)
-    Invidious::Database::Videos.insert(video) if !region
+    Invidious::Database::Videos.insert(video) if !region && !video.info.dig?("reason")
```
SamantazFox, line 316: You should make a getter dunction inside the `Video` class for reason/subreason.

src/invidious/videos/parser.cr
```diff
@@ -64,18 +64,20 @@ def extract_video_info(video_id : String)
...
+    # Stop if private video or video not found.
+    # But for video unavailable, only stop if playability_status is ERROR because playability_status UNPLAYABLE
+    # still gives all the necessary info for displaying the video page (title, description and more)
+    if {"Private video", "Video unavailable"}.any?(reason) && !{"UNPLAYABLE"}.any?(playability_status)
       return {
-        "version" => JSON::Any.new(Video::SCHEMA_VERSION.to_i64),
-        "reason"  => JSON::Any.new(reason),
+        "version"   => JSON::Any.new(Video::SCHEMA_VERSION.to_i64),
+        "reason"    => JSON::Any.new(reason),
+        "subreason" => JSON::Any.new(subreason),
```
SamantazFox, line 80: Don't forget to bump `SCHEMA_VERSION` in the `Video` class!

SamantazFox (changes requested):

NintendoManiac64:
>  Integrate the embed page with the changes

To clarify, does this mean using the YouTube embedded player within the usual invidious video page?  I know it wouldn't be "privacy respecting" in the same way as traditional invidious, so maybe as a sort of optional "compatibility mode" in the settings?


It sounds like some people have suggested something similar but instead mention just linking to the embed URL (youtube.com/embed/fo0bArBAZ) - I couldn't find anyone suggesting to directly embed the actual embedded YouTube player right on invidious.

(heck, I found out that you can even use various custom adblock filters to hide the "fluff" of the embed player such as suggestions & recommendations, and even end-screen video links/thumbnails, which makes me wonder if even invidious could similarly do that on the server side, or if that's only possible on the client side.)

...and now as I'm typing this, I'm remembering that YouTube supposedly has a newer type of embedded player that's "brandless" and lacks the fluff? Maybe that'd be better to use for embedding into the invidious video page...except, [now that I review where I read about it](https://www.theverge.com/2024/12/12/24318124/youtube-player-cant-click-title-sigh), sounds like it's not something the user can choose and it's determined by the video channel itself?

syeopite:
> To clarify, does this mean using the YouTube embedded player within the usual invidious video page?

No, it would just mean displaying video info even if the player is unable to be rendered for whatever reason. Everything will still be in Invidious.

ChunkyProgrammer:
This should also close https://github.com/iv-org/invidious/issues/4073

mk-pmb:
Found this via #4744 . Is there any feedback still needed? Can I do something to help this move forward?

unixfox:
@mk-pmb There is still all the reviews that needs to be taken care off.

If you know about crystal, I would be happy to receive some contribution. If that's what you can help us, feel free to do a PR on my branch: https://github.com/unixfox/invidious/tree/use-only-next-for-extra-info or create a new PR in the invidious project :).

mk-pmb:
I don't know crystal but I can try and replicate your changes based on current master, with the suggestions from this thread applied. :+1:

src/invidious/views/embed.ecr
mk-pmb: Do I understand correctly that we only display the reason if we don't have any? Should I negate this check?

mk-pmb:
@unixfox I'd like to discuss my draft branch with  you, but your repo doesn't have issues. Would you mind creating a `git branch use-only-next-for-extra-info-base 70e4eb7f` in your repo so I can open a merge request for that? Then we can discuss there.

unixfox:
hello @mk-pmb, yes sure sorry for the delay. I have created the branch: https://github.com/unixfox/invidious/tree/use-only-next-for-extra-info-base

I have opened the issues on my fork too. 

I have a better idea, I invited you as a collaborator on my forked invidious repo, this way it will be easier to work on the branch use-only-next-for-extra-info directly. Don't screw up my others branches please :smile:

Thanks

PR #5025, opened 2024-10-26
[Debug] print requests as curl commands

iBicha:
This makes invidious prints requests it makes as curl commands. For example, when visting the trending page, the following curl command would be printed:
> curl -X POST 'https://www.youtube.com/youtubei/v1/browse?prettyPrint=false' -H 'Content-Type: application/json; charset=UTF-8' -H 'x-goog-api-format-version: 2' -H 'x-youtube-client-name: 1' -H 'x-youtube-client-version: 2.20240814.00.00' -H 'Host: www.youtube.com' -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36' -H 'Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7' -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' -H 'Accept-Language: en-us,en;q=0.5' -H 'Cookie: ; CONSENT=PENDING+644' -d '{"browseId":"FEtrending","context":{"client":{"hl":"en","gl":"US","clientName":"WEB","clientVersion":"2.20240814.00.00","clientScreen":"WATCH_FULL_SCREEN","osName":"Windows","osVersion":"10.0","platform":"DESKTOP"}}}'

This is useful for a few things:

- Understanding what Invidious is doing with Innertube
- Can easily copy the command and run in a terminal, dump response to a file
- Easier to iterate on requests that have issues

### DIFF
--- a/src/invidious/yt_backend/connection_pool.cr
+++ b/src/invidious/yt_backend/connection_pool.cr
@@ -49,6 +49,9 @@ struct YoutubeConnectionPool
       conn.family = CONFIG.force_resolve
       conn.family = Socket::Family::INET if conn.family == Socket::Family::UNSPEC
       conn.before_request { |r| add_yt_headers(r) } if url.host == "www.youtube.com"
+      if CONFIG.log_level <= LogLevel::Debug
+        conn.before_request { |r| LOGGER.debug(to_curl(url, r)) }
+      end
       conn
     end
   end
@@ -63,6 +66,9 @@ def make_client(url : URI, region = nil, force_resolve : Bool = false)
   end
 
   client.before_request { |r| add_yt_headers(r) } if url.host == "www.youtube.com"
+  if CONFIG.log_level <= LogLevel::Debug
+    client.before_request { |r| LOGGER.debug(to_curl(url, r)) }
+  end
   client.read_timeout = 10.seconds
   client.connect_timeout = 10.seconds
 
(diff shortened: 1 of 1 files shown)


unixfox:
Not really fond in doing that since we already have a set of tools to make it easier to talk to YouTube API: https://github.com/iv-org/youtube-utils

iBicha:
> Not really fond in doing that since we already have a set of tools to make it easier to talk to YouTube API: https://github.com/iv-org/youtube-utils

This is what I prefer (and use) to debug issues - I prefer to see exactly what Invidious is doing and when etc
I don't feel strongly about merging this PR, just sharing what I find productive for myself

Fijxu:
@iBicha I recommend you to use `proxychains` with `mitmproxy`. That is what I use to debug requests on Invidious like I did here: https://github.com/iv-org/invidious/issues/5033

unixfox:
> @iBicha I recommend you to use `proxychains` with `mitmproxy`. That is what I use to debug requests on Invidious like I did here: #5033

Now you can use the http_proxy feature directly with mitmproxy: https://github.com/iv-org/invidious/blob/master/config/config.example.yml#L181

iBicha:
Closing the PR since the purpose was to share my approach.
I like copying a curl command, run and dump response into a json file, then iterate on the command while diffing the response. Tweak header and payload to get innertube to return different things. This is a simple approach, doesn't require other tools, nor hooking up a proxy. Different tools for different things I guess.
Feel free to repurpose this if it's ever useful

PR #5027, opened 2024-10-26
Fix for channel live videos

### LINKED ISSUE #5021: [Bug] "Request contains an invalid argument." when going to channels
**Describe the bug**
Navigating to eg. `/channel/UCSJ4gkVC6NrvII8umztf0Ow/streams` returns `Could not extract JSON. Youtube API returned error 400 with message:
"Request contains an invalid argument." ` 

**Steps to Reproduce**
1. Go to https://inv.nadeko.net/channel/UCSJ4gkVC6NrvII8umztf0Ow/
2. Click on the 'Livestreams' tab

**Additional context**
Does not work on local instance nor when using the API

iBicha:
Closes #5021

This is an alternative to #5026 to keep the sort options working

After digging into the proto, live stream sorting has different values than channel videos

### DIFF
--- a/src/invidious/channels/videos.cr
+++ b/src/invidious/channels/videos.cr
@@ -23,6 +23,13 @@ def produce_channel_content_continuation(ucid, content_type, page = 1, auto_gene
     else                    15 # Fallback to "videos"
     end
 
+  sort_type_numerical =
+    case content_type
+    when "videos"      then 3
+    when "livestreams" then 5
+    else                    3 # Fallback to "videos"
+    end
+
   if content_type == "livestreams"
     sort_by_numerical =
       case sort_by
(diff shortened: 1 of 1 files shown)


absidue:
As this one fixes the problem properly, without removing functionality, I would suggest that this pull request should be considered the proper fix and the other one the alternative (so the opposite of what the titles currently say).

iBicha:
> As this one fixes the problem properly, without removing functionality, I would suggest that this pull request should be considered the proper fix and that the other one be considered as the alternative or even closed now that there is a proper fix (so the opposite of what the titles currently say).

fair enough

SamantazFox:
> This is an alternative to 5026 to keep the sort options working

I'll second absidue here: please close the other one as this one is more complete

src/invidious/channels/videos.cr
```diff
@@ -23,29 +23,57 @@ def produce_channel_content_continuation(ucid, content_type, page = 1, auto_gene
...
+      "110:embedded" => {
+        "3:embedded" => {
+          "#{content_type_numerical}:embedded" => {
+            "1:embedded" => {
+              "1:string" => object_inner_2_encoded,
+            },
+            "2:embedded" => {
+              "1:string" => "00000000-0000-0000-0000-000000000000",
+            },
+            "5:varint" => sort_by_numerical,
```
SamantazFox, line 55: For this field, I'd recommend doing the same as for the `embedded` above, to not have to duplicate the whole structure

SamantazFox (approved):

Fijxu:
Applied to my instance.
Test URL: https://inv.nadeko.net/channel/UC7_YxT-KID8kRbqZo7MyscQ/streams

SamantazFox:
> Applied to my instance. Test URL: https://inv.nadeko.net/channel/UC7_YxT-KID8kRbqZo7MyscQ/streams

Thanks for testing it out :)

src/invidious/channels/videos.cr
```diff
@@ -23,14 +23,31 @@ def produce_channel_content_continuation(ucid, content_type, page = 1, auto_gene
...
-    case sort_by
-    when "newest"  then 1_i64
-    when "popular" then 2_i64
-    when "oldest"  then 4_i64
-    else                1_i64 # Fallback to "newest"
+  sort_type_numerical =
+    case content_type
+    when "videos"      then 3
+    when "livestreams" then 5
+    else                    3 # Fallback to "videos"
```
lifo9, line 30:
```suggestion
    when "videos"      then 4
    when "livestreams" then 5
    else                    4 # Fallback to "videos"
```

src/invidious/channels/videos.cr
```diff
@@ -23,14 +23,31 @@ def produce_channel_content_continuation(ucid, content_type, page = 1, auto_gene
...
+      when "oldest"  then 13_i64
+      else                12_i64 # Fallback to "newest"
+      end
+  else
+    sort_by_numerical =
+      case sort_by
+      when "newest"  then 1_i64
+      when "popular" then 2_i64
+      when "oldest"  then 4_i64
+      else                1_i64 # Fallback to "newest"
```
lifo9, line 47:
```suggestion
      when "newest"  then 4_i64
      when "popular" then 2_i64
      when "oldest"  then 5_i64
      else                4_i64 # Fallback to "newest"
```

lifo9 (changes requested):
Hey,
thanks for finding out!
Protobuf values also changed for regular videos.
Tested on my local instance and it works now.

SamantazFox:
> Hey, thanks for finding out! Protobuf values also changed for regular videos. Tested on my local instance and it works now.

Thanks! It is being address in https://github.com/iv-org/invidious/pull/5059

iBicha:
Superseded by #5059

SamantazFox:
Thanks for your contribution btw!

PR #5036, opened 2024-10-29
Routes: Clean ajax actions handlers

SamantazFox:
Nothing much, but something that I wanted to do for quite some time, as this part of the code is confusing.


unixfox:
ping ones that know a bit about js on the client side :)
@ChunkyProgrammer @AHOHNMYC

unixfox (approved):

PR #5040, opened 2024-10-30
allow api endpoint '/api/v1/auth/tokens/register' to answer json requests with json response

lekma:
currently '/api/v1/auth/tokens/register' returns html even if the request is a json one when you register your first token (meaning there is no authorization header yet)


lekma:
sorry for the squash spam, but i realized after the fact that my assumptions were wrong, i think that should be enough

lekma:
and now i'm wondering if that is, in fact, worse...

lekma:
Hi,
no i think i got it right now, i'm just unsure of the 403 response code, is it enough?

SamantazFox:
> currently '/api/v1/auth/tokens/register' returns html even if the request is a json one when you register your first token (meaning there is no authorization header yet)

If you look at the code, it returns HTML because you have a session ID present in the current environment (`env.get? "sid"` returns a non-`nil` value), meaning that your API client is passing the `SID` cookie.

Are you trying to access that API endpoint from a browser where you're already logged in?

lekma:
no, that's the point, i'm trying to acces the api from a python script but i don't have a token yet, so i log in by posting a username/password (i couldn't find any documentation on how to login differently) then try to generate the token i would like to use for the rest of the session (mainly retrieving user feed).

the whole thing look like (pseudo-code):

```
# this is where i get the SID cookie it's a form-urlencoded request
login = session.post("instance.uri/login", data={"email": email, "password": password, "action": "signin"})
if login:
  # that request is a json one
  token =  session.post("instance.uri/api/v1/auth/tokens/register", json={"scopes": [":*"]})
```

and given that my second request is a json one i would expect the json api to return the same.
i thought it kinda odd to be given an html result...

lekma:
correct me if i'm wrong, but, I do need to be logged in to use all the 'api/v1/auth' endoints, right?

lekma:
also i forgot to mention but the idea is to store only the tokens (per instance) not the user/pass...

PR #5014, opened 2024-10-18
Apply upcoming formatting rules from Crystal 1.15

syeopite:

PR #5042, opened 2024-10-30
CI: Check Crystal lint only on latest version

### LINKED ISSUE #5017: CI: Test formatter only with latest Crystal version
Following a discussion in the Crystal matrix room

Formatter behavior shouldn't be expected to be compatible across various Crystal versions (but rather with one version of overlap instead)

We should switch to only testing the linter on the latest (or latest supported) Crystal version

SamantazFox:
Required for #5014

Run the linter against a single version of Crystal, as recommended by @straight-shoota in the Crystal Matrix room:

> I'd also suggest to run the formatter only against one version of the compiler. Formatter behaviour should not be expected to be forward compatible (we try to have at least one version overlap, but not necessarily more). So there might be a patch release for 1.14 to make it accept the trailing comma. But that won't be ported back to earlier versions.
>
> https://matrix.to/#/%23crystal-lang_crystal%3Agitter.im/%24_UuRZsBnmr5Q0OTQRiUaYDq1duOjXHZVfmoV3g0QJck?via=gitter.im&via=matrix.org&via=kde.org

I have also unified the way shards are installed and cached between the `build` and `lint` jobs to better use the cache.

Closes #5017

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
             shards install
           fi
 
-      - name: Run Crystal linter
+      - name: Check Crystal formatter compliance
         run: |
           if ! crystal tool format --check; then
             crystal tool format
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
 
       - name: Install required APT packages
         run: |
-          sudo apt install -y libsqlite3-dev 
+          sudo apt install -y libsqlite3-dev
         shell: bash
 
       - name: Install Crystal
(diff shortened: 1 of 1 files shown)


unixfox (approved):

.github/workflows/ci.yml
```diff
@@ -151,7 +149,18 @@ jobs:
           key: shards-${{ hashFiles('shard.lock') }}
 
       - name: Install Shards
-        run: shards install
+        run: |
+          if ! shards check; then
+            shards install
+          fi
+
+      - name: Run Crystal linter
```
syeopite, line 157:
```suggestion
      - name: Check Crystal formatter compliance
```

Since we have an actual linter now with Ameba I think we should probably be specific that the Crystal linter is just the formatter

PR #5045, opened 2024-10-31
Routing: Deprecate old channel API routes

SamantazFox:
Deprecate the following routes:
* `/api/v1/channels/videos/:ucid`
* `/api/v1/channels/latest/:ucid`
* `/api/v1/channels/playlists/:ucid`
* `/api/v1/channels/community/:ucid`
* `/api/v1/channels/search/:ucid`

in favor of:
* `/api/v1/channels/:ucid/videos`
* `/api/v1/channels/:ucid/latest`
* `/api/v1/channels/:ucid/playlists`
* `/api/v1/channels/:ucid/community`
* `/api/v1/channels/:ucid/search`

### DIFF
--- a/src/invidious/routing.cr
+++ b/src/invidious/routing.cr
@@ -270,11 +270,6 @@ module Invidious::Routing
 
       # Authenticated
 
-      # The notification APIs cannot be extracted yet! They require the *local* notifications constant defined in invidious.cr
-      #
-      # Invidious::Routing.get "/api/v1/auth/notifications", {{namespace}}::Authenticated, :notifications
-      # Invidious::Routing.post "/api/v1/auth/notifications", {{namespace}}::Authenticated, :notifications
-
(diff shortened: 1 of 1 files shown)


Fijxu:
Why remove them tho? The "deprecated" api routes won't be replaced by another routes so deleting them will just force everyone to change the route in their program.

This is to follow the same syntax that YouTube channels have right?

absidue:
@Fijxu This doesn't add any new routes, it just removes undocumented routes. The old routes were removed from the documentation in this commit on the 14th of January 2023 https://github.com/iv-org/documentation/commit/dca939835c8a5909b450440be29006360d682eae, that is almost 2 years ago, which should be more than enough time for people to switch to the correct endpoints, especially considering that the "new" endpoints have existed for even longer than that.

syeopite (approved):

Issue #5032, opened 2024-10-29
[Bug] Unexpected char '<' at line 1, column 1 (JSON::ParseException) | 429 rate limit errors


corincorvus:
UPDATE from @unixfox:

The issue is related to YouTube limiting the amount of requests done to their servers.

Invidious is not the only one affected: https://github.com/iv-org/invidious/issues/5032#issuecomment-2446578562

------

**Describe the bug**
just want to watch a video

**Steps to Reproduce**

**Logs**
```
Title: `Unexpected char '<' at line 1, column 1 (JSON::ParseException)`
Date: `2024-10-29T15:20:09Z`
Route: `/watch?v=Q314YvZvnDY`
Version: `2024.08.26-4782a67 @ (HEAD detached at v2.20240825.2)`

<details>
<summary>Backtrace</summary>
<p>
   
Unexpected char '<' at line 1, column 1 (JSON::ParseException)
  from /usr/share/crystal/src/json/lexer.cr:331:5 in 'raise'
  from /usr/share/crystal/src/json/lexer.cr:327:5 in 'unexpected_char'
  from /usr/share/crystal/src/json/lexer.cr:326:11 in 'unexpected_char'
  from /usr/share/crystal/src/json/lexer/io_based.cr:11:21 in 'consume_number'
  from /usr/share/crystal/src/json/lexer.cr:63:7 in 'next_token'
  from /usr/share/crystal/src/json/parser.cr:13:12 in '_post_json'
  from src/invidious/videos/parser.cr:58:21 in 'extract_video_info'
  from src/invidious/videos.cr:381:10 in 'fetch_video'
  from src/invidious/videos.cr:369:13 in 'get_video:region'
  from src/invidious/routes/watch.cr:55:15 in 'handle'
  from lib/kemal/src/kemal/route.cr:13:9 in '->'
  from src/invidious/helpers/handlers.cr:30:37 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:94:12 in 'call'
[...]

PR #5052, opened 2024-11-02
checking the status code returned by youtube

unixfox:
Should provide a better error message for the 429 rate limit: https://github.com/iv-org/invidious/issues/5032

### DIFF
--- a/src/invidious/yt_backend/youtube_api.cr
+++ b/src/invidious/yt_backend/youtube_api.cr
@@ -638,7 +638,7 @@ module YoutubeAPI
     # Send the POST request
     body = YT_POOL.client() do |client|
       client.post(url, headers: headers, body: data.to_json) do |response|
-        if response.status_code == 200
+        if response.status_code != 200
           raise InfoException.new("Error: non 200 status code. Youtube API returned \
             status code #{response.status_code}. See <a href=\"https://docs.invidious.io/youtube-errors-explained/\"> \
             https://docs.invidious.io/youtube-errors-explained/</a> for troubleshooting.")
--- a/src/invidious/yt_backend/youtube_api.cr
+++ b/src/invidious/yt_backend/youtube_api.cr
@@ -638,6 +638,10 @@ module YoutubeAPI
     # Send the POST request
     body = YT_POOL.client() do |client|
       client.post(url, headers: headers, body: data.to_json) do |response|
+        if response.status_code != 200
+          raise InfoException.new("Error: non 200 status code. Youtube API returned \
+            status code #{response.status_code}")
+        end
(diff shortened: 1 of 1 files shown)


src/invidious/yt_backend/youtube_api.cr
```diff
@@ -638,6 +638,11 @@ module YoutubeAPI
     # Send the POST request
     body = YT_POOL.client() do |client|
       client.post(url, headers: headers, body: data.to_json) do |response|
+        if response.status_code == 200
```
syeopite, line 641:
```suggestion
        if response.status_code != 200
```

syeopite (changes requested):

src/invidious/yt_backend/youtube_api.cr
unixfox: Sorry I left the wrong condition for testing the error message 😅

syeopite (approved):

SamantazFox (approved):

unixfox:
I think based on the URL alone it's going to be pretty obvious.

Usually even on errors, the yt api just returns a 200 status code.

iBicha:
> I think based on the URL alone it's going to be pretty obvious.
> 
> Usually even on errors, the yt api just returns a 200 status code.

I think it should at least contain the error from the body to make this more helpful than it was

unixfox:
We would need an error template which summarize the error AND print the full error stack, but we don't have one.

For the moment, that's the best that we can do for clearly explaining to the user what's going on.

I still insist on the case that when you receive another status code other than 200 then something really wrong is going on. Many people will discover it and create new issues, the URL causing the issue will come up too at some point.

iBicha:
I mean the YT response body can be just added to the body of InfoException, that is also fine

unixfox:
But then these messages will never be able to be translated.

Hence why I said, we do not have a page template with a summary of the error AND the full stack error.

unixfox:
And I forgot to say that we already parse the error given by youtube here: https://github.com/iv-org/invidious/blob/792d0d5f6df912039a58768e6ff503ae00abe7c0/src/invidious/yt_backend/youtube_api.cr#L654

Here https://github.com/iv-org/invidious/issues/5032 why Invidious is crashing. Is that it except youtube to give a JSON body but receive an HTML one.

Now that I think again. Maybe we could only throw a simple error status code message. When the status code is not 200 AND it's not a JSON body.

I think youtube probably return a JSON body with an error and a status code of 400

syeopite:
I have a PR opened upstream in Kemal to allow for routes pertaining to specific raised errors https://github.com/kemalcr/kemal/pull/688

I haven't had time to work on it lately but once it is finished and merged we should be able to create customized error pages for different exceptions

PR #5065, opened 2024-11-08
Remove stdlib override for proxy initialization

syeopite:
As we don't directly call `HTTP::Client` anymore the proxy initialization can just be done in `make_client` instead of when `HTTP::Client` is initialized. 

This is the only issue I've noticed with the Nov 8th merge batch.


SamantazFox (approved):
Thank you!

unixfox (approved):

PR #5066, opened 2024-11-08
Update shard.yml metadata

syeopite:
Changes are mostly based off of the Crystal compiler's own shard.yml

**Remember to bump the version attribute when creating a release!**

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,13 +1,12 @@
 name: invidious
-version: 0.20.1
+version: 2.20240825.2-dev
 
 authors:
   - Omar Roth <omarroth@protonmail.com>
-  - Invidious team
+  - Invidious team <contact@invidious.io>
 
-targets:
-  invidious:
-    main: src/invidious.cr
+description: |
+   Invidious is an alternative front-end to YouTube 
 
 dependencies:
   pg:
@@ -40,6 +39,10 @@ development_dependencies:
     github: crystal-ameba/ameba
     version: ~> 1.6.1
 
-crystal: ">= 1.0.0, < 2.0.0"
+crystal: ">= 1.10.0, < 2.0.0"
 
 license: AGPLv3
+
+repository: https://github.com/iv-org/invidious
+homepage: https://invidious.io
+documentation: https://docs.invidious.io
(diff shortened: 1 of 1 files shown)


unixfox (approved):

shard.yml
```diff
@@ -1,13 +1,12 @@
...
+version: 2.20240825.2-dev
 
 authors:
   - Omar Roth <omarroth@protonmail.com>
-  - Invidious team
+  - Invidious team <contact@invidious.io>
 
-targets:
-  invidious:
-    main: src/invidious.cr
```
SamantazFox, line 10: Why removing that, though?

shard.yml
```diff
@@ -1,13 +1,12 @@
 name: invidious
-version: 0.20.1
+version: 2.20240825.2-dev
```
SamantazFox, line 2: How should we handle the `-dev`, in term of the release process? Do we remove it for the release commit, then add a new commit right after that that adds back `-dev`, like so?
```
ac4596fb (HEAD) Prepare for next version
def09876 (v2.2024MMDD.0) Release v2.2024MMDD.0
b1f8ee56 Merge pull request #9876
82f1f891 Fix stuff
```

shard.yml
syeopite: I just modeled the file after the Crystal compiler one which doesn't have the target section.

Plus Invidious isn't intended to be built via shards

shard.yml
syeopite: Yeah that's pretty much the only option.

In the Crystal compiler for example, they also just remove and readd the -dev prefix before and after each release https://github.com/crystal-lang/crystal/commits/master/shard.yml

shard.yml
```diff
@@ -1,13 +1,12 @@
...
 
 authors:
   - Omar Roth <omarroth@protonmail.com>
-  - Invidious team
+  - Invidious team <contact@invidious.io>
```
syeopite, line 6: Regarding the authors section should we change it to only include the Invidious team as to reflect the current authors?

SamantazFox: Why not, though `Invidious team and contributors` is more representative of all the people that worked on invidious!

SamantazFox (approved):

PR #5059, opened 2024-11-07
Channels: Fix "Youtube API returned error 400"

SamantazFox:
Closes #4029
Closes #5021
Closes #5029

Requires #5027 to be merged first.
Does NOT address #4477

This PR also adds sort option for shorts.

PR #5071, opened 2024-11-09
API: Add "sort_by" parameter to channels/shorts endpoint

iBicha:
Small follow up to #5059 sort channel shorts on the api


SamantazFox (approved):

SamantazFox:
Thanks for your contribution!

PR #5073, opened 2024-11-10
Skip route if resp got closed by before handlers

syeopite:
This is the culprit behind https://github.com/iv-org/invidious/pull/4259#issuecomment-1828573067

This is why we should avoid overrides... [the problem was fixed upstream in Kemal 5 years ago in 2019!](https://github.com/kemalcr/kemal/pull/550)


SamantazFox (approved):

SamantazFox:
Holy moly...

unixfox:
:exploding_head:

unixfox (approved):

PR #5085, opened 2024-11-19
Docker compose.yml

asxpi:
This changes makes it work for us. Without inv_sig_helper it returned "The media could not be loaded, either because the server or network failed or because the format is not supported." instead of playing videos.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,6 +71,5 @@ services:
     security_opt:
       - no-new-privileges:true
 
-
 volumes:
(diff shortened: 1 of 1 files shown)


syeopite:
This compose file is only for developmental purposes. For production you should use the compose given here 

https://docs.invidious.io/installation/

---

Though maybe it's about time for the dev compose to also include inv sig helper...

unixfox:
Hello, you did not read the first lines of the file :).

PR #5087, opened 2024-11-20
send query string instead of range http header

unixfox:
Send the range using query string like the YouTube player do.

This avoids having to do some processing inside Invidious and just straight send the correct query string to YouTube servers.

Makes it easier to be compatible with other videoplayback proxies like [http3-ytproxy](https://github.com/TeamPiped/http3-ytproxy) or [piped-proxy](https://github.com/TeamPiped/piped-proxy). Especially the last one because piped only use query string now for the range.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -42,6 +42,15 @@ embed_url = location.origin + '/embed/' + video_data.id + embed_url.search;
 var save_player_pos_key = 'save_player_pos';
 
 videojs.Vhs.xhr.beforeRequest = function(options) {
+    if (options.uri.includes("videoplayback")) {
+        if (options.headers) {
+            if (options.headers.Range) {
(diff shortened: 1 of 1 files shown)


absidue:
Last time I tested this it didn't work with Invidious' own video proxy, which is why FreeTube still sends the range header instead of the query parameter. I haven't tested this specific pull request though, so maybe it is different with this one.

SamantazFox:
> This avoids having to do some processing inside Invidious and just straight send the correct query string to YouTube servers.

Invidious will always need some client-side processing for clients with JS disabled

unixfox:
> > This avoids having to do some processing inside Invidious and just straight send the correct query string to YouTube servers.
> 
> Invidious will always need some client-side processing for clients with JS disabled

True but at least for most people there is no processing involved. Same for the ones using DASH.

PR #5094, opened 2024-11-25
Set language info for dash audio streams and sort

GTechAlpha:
Set audio track info, if present, in dash manifest
  - language id
  - language display name
  - main/default track
 Sort audio formats so that main/default is first (for clients not using dash)

This resolves issue of incorrect/undesired audio when automatic translation is enabled by content creator (currently relatively rare, but likely to increase), and allows dash clients to correctly select default language and optionally offer multiple language streams.

* Note: this should be a non-breaking change; if audio track info is not available, the behavior does not change from current

### DIFF
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -70,17 +70,22 @@ module Invidious::Routes::API::Manifest
               # OTF streams aren't supported yet (See https://github.com/TeamNewPipe/NewPipe/issues/2415)
               next if !(fmt.has_key?("indexRange") && fmt.has_key?("initRange"))
 
+              audio_track = fmt["audioTrack"]?.try &.as_h? || {} of String => JSON::Any
+              lang = audio_track["id"]?.try &.as_s.split('.')[0] || "und"
+              is_default = audio_track.has_key?("audioIsDefault") ? audio_track["audioIsDefault"].as_bool : i == 0
+              displayname = audio_track["displayName"]?.try &.as_s || "Unknown"
+
               # Different representations of the same audio should be groupped into one AdaptationSet.
               # However, most players don't support auto quality switching, so we have to trick them
               # into providing a quality selector.
               # See https://github.com/iv-org/invidious/issues/3074 for more details.
-              xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true, label: fmt["bitrate"].to_s + "k") do
+              xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true, label: displayname, lang: lang) do
                 codecs = fmt["mimeType"].as_s.split("codecs=")[1].strip('"')
                 bandwidth = fmt["bitrate"].as_i
                 itag = fmt["itag"].as_i
                 url = fmt["url"].as_s
 
-                xml.element("Role", schemeIdUri: "urn:mpeg:dash:role:2011", value: i == 0 ? "main" : "alternate")
+                xml.element("Role", schemeIdUri: "urn:mpeg:dash:role:2011", value: is_default ? "main" : "alternate")
 
                 xml.element("Representation", id: fmt["itag"], codecs: codecs, bandwidth: bandwidth) do
                   xml.element("AudioChannelConfiguration", schemeIdUri: "urn:mpeg:dash:23003:3:audio_channel_configuration:2011",
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -106,7 +106,7 @@ struct Video
     if formats = info.dig?("streamingData", "adaptiveFormats")
       return formats
         .as_a.map(&.as_h)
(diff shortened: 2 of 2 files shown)


src/invidious/routes/api/manifest.cr
```diff
@@ -70,17 +70,22 @@ module Invidious::Routes::API::Manifest
...
+              lang = audio_track["id"]?.try &.as_s.split('.')[0] || "und"
+              is_default = audio_track.has_key?("audioIsDefault") ? audio_track["audioIsDefault"].as_bool : i == 0
+              displayname = audio_track["displayName"]?.try &.as_s || "Unknown"
+
               # Different representations of the same audio should be groupped into one AdaptationSet.
               # However, most players don't support auto quality switching, so we have to trick them
               # into providing a quality selector.
               # See https://github.com/iv-org/invidious/issues/3074 for more details.
-              xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true, label: fmt["bitrate"].to_s + "k") do
+              xml.element("AdaptationSet", id: i, mimeType: mime_type, startWithSAP: 1, subsegmentAlignment: true, label: displayname, lang: lang) do
```
syeopite, line 82: In the case of different representations of the same language you'll end up with two `AdaptationSet` with the same label and lang. I think that might confuse some downstream clients (and users) if two audio tracks end up having the same label and language.

For example see the dash manifest for this video with this PR https://youtube.comwatch?v=urevinis_PU

GTechAlpha: Hello.

If I am not misunderstanding your statement, yes that could happen if audio streams weren't being filtered for mime type "audio/mp4".

However, since they are being filtered this way, I have only seen one unique language representation per this mime type from YT.

With this PR and the current state of inv and YT, I have not experienced the issue you describe. And in the case of your example, no audio stream selection options are provided (by video.js and in my case) because there is no audio track information from YT, so the lang attribute is set to "und" and the client falls back to the default behavior of auto-selecting the first stream. When audio track info is available, video.js provides only one audio option per language in all of my testing.

Thank you for your time and attention.

syeopite: This isn't just about video.js. There are downstream clients that use other media players which also rely on the dash manifests generated by Invidious.

syeopite: I'm also concerned that this could reintroduce issue #3074.

Although the first `AdaptationSet` should always have the highest bitrate, YouTube's structure is very finicky and subject to change. Even if clients will always select the first stream it doesn't mean that it'll always have the highest bitrate.

PR #3162 also seemed to introduce a quality selector for audio but it seems to have broke somewhere down the line. If it was ever fixed, this issue will present a UX problem even in Invidious and video.js with the labels being simply "Unknown"

georgek: What's the way forward with this? What is the API that needs to be maintained for downstream clients? Does it just need some kind of unique ID?

There is also, of course, the possibility of doing a breaking change. We could argue that if clients can't support this any more then they need updating. Invidious is becoming more and more useless and more and more videos come with some AI-generated dub track.

unixfox:
> What's the way forward with this? What is the API that needs to be maintained for downstream clients? Does it just need some kind of unique ID?

> There is also, of course, the possibility of doing a breaking change. We could argue that if clients can't support this any more then they need updating. Invidious is becoming more and more useless and more and more videos come with some AI-generated dub track.

@georgek

Well first, you are not commenting in the right section. This is a comment section about a code change in the PR.

Second, This PR is about the DASH manifest, not the API that is used by the 3rd party clients.

If you are requesting the API using the `/api/v1/videos/`, then it's up to your client to handle the fact that there are multiple audio files with different languages. It's possible that we do not expose enough info to differentiate between each one, but if it's that the case, then please open a new GitHub issue.

WreckingBANG:
I build the Image myself and have been testing it for a bit and it is working great.

src/invidious/routes/api/manifest.cr
giuliano-macedo: @syeopite do you think something like this would solve theses issues?
```crystal
displayname = audio_track["displayName"]?.try &.as_s || "Unknown"
bitrate = fmt["bitrate"]
label = displayname == "Unknown" ? "[#{bitrate}k]" : "#{displayname} [#{bitrate}k]"
xml.element("AdaptationSet", ... label: label)
```

saukkico:
This seems to work well enough, at least for me. I'm using a third party app (Yattee) to watch videos on my Apple TV. With this PR Yattee can choose the right audio track. 
I've also been testing the companion setup, but then Yattee doesn't play the right audio track. So in my use case this PR works better than the companion setup.

src/invidious/routes/api/manifest.cr
syeopite: I believe so but the fallback bitrate label should be `#{lang} [#{bitrate}k]` instead of just simply a bitrate display.

giuliano-macedo: Great! I'm continuing this PR at #5149

PR #5112, opened 2024-12-14
Add the ability to listen on UNIX sockets

Caian:
This is a continuation of #2111 that was closed. 

As stated in the original PR, using unix sockets can have a beneficial impact on performance. I found that particularly relevant when running my invidious instance in a rootless container.

This PR adds the `socket_binding` option to config that specifies a socket to listen to. The `socket_binding` option will disable listening to TCP. I also added `socket_permissions` because often one needs to control who accesses the socket.


Caian:
Tested and running on my personal invidious behind a Caddy proxy with https enabled. Now videojs buffers the video much faster and the user-mode networking on the server consumes a less CPU and RAM.

unixfox (changes requested):
Maybe you can combine the two parameters into one.

And make sure to comment them otherwise socket binding will be on by default!

Caian:
> Maybe you can combine the two parameters into one.
> 
> And make sure to comment them otherwise socket binding will be on by default!

Ops, I fixed the configuration file to leave it disable. I prefer the idea of leaving separate parameters for clarity, but if the parameters were to be combined I was thinking of doing something like:

socket_binding: /tmp/invidious.socket,777

What do you think?

unixfox:
> > Maybe you can combine the two parameters into one.
> > 
> > And make sure to comment them otherwise socket binding will be on by default!
> 
> Ops, I fixed the configuration file to leave it disable. I prefer the idea of leaving separate parameters for clarity, but if the parameters were to be combined I was thinking of doing something like:
> 
> socket_binding: /tmp/invidious.socket,777
> 
> What do you think?

You can also do it this way:

socket_binding:
  path: /tmp/invidious.socket
  permission: 777

Caian:
> > > Maybe you can combine the two parameters into one.
> > > And make sure to comment them otherwise socket binding will be on by default!
> > 
> > 
> > Ops, I fixed the configuration file to leave it disable. I prefer the idea of leaving separate parameters for clarity, but if the parameters were to be combined I was thinking of doing something like:
> > socket_binding: /tmp/invidious.socket,777
> > What do you think?
> 
> You can also do it this way:
> 
> socket_binding: path: /tmp/invidious.socket permission: 777

I like this solution, implemented.

unixfox (changes requested):

config/config.example.yml
```diff
@@ -177,7 +191,7 @@ https_only: false
 ## Configuration for using a HTTP proxy
 ##
 ## If unset, then no HTTP proxy will be used.
-##
```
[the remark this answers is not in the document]
Caian: Sorry I'll revert that, my editor automatically removed a trailing space...

src/invidious/config.cr
```diff
@@ -138,6 +145,8 @@ class Config
   property port : Int32 = 3000
   # Host to bind (overridden by command line argument)
   property host_binding : String = "0.0.0.0"
+  # Path and permissions to make Invidious listen on a UNIX socket instead of a TCP port - Example: /tmp/invidious.sock,777
+  property socket_binding : SocketBindingConfig? = nil
```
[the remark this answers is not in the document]
Caian: Sure, I'll take a look!

src/invidious.cr
```diff
@@ -249,4 +247,19 @@ Kemal.config.app_name = "Invidious"
   Kemal.config.env = "production" if !ENV.has_key?("KEMAL_ENV")
 {% end %}
 
-Kemal.run
+Kemal.run do |config|
+  if CONFIG.socket_binding
+    socket_binding = CONFIG.socket_binding.not_nil!
```
syeopite, line 252:
```suggestion
  if socket_binding = CONFIG.socket_binding
```

src/invidious.cr
```diff
@@ -249,4 +247,19 @@ Kemal.config.app_name = "Invidious"
...
+    if File.exists?(socket_binding.path)
+      File.delete(socket_binding.path)
+    end
```
syeopite, line 255: This can be simplified to `File.delete?`

https://crystal-lang.org/api/1.14.0/File.html#delete%3F(path%3APath|String)%3ABool-class-method

```suggestion
File.delete?(socket_binding.path)
```

src/invidious/config.cr
```diff
@@ -249,6 +258,25 @@ class Config
       end
     end
 
+    # Check if the socket configuration is valid
+    if config.socket_binding
+      sb = config.socket_binding.not_nil!
```
syeopite, line 263:
```suggestion
    if sb = config.socket_binding
```

syeopite (changes requested):

syeopite (approved):

PR #5116, opened 2024-12-16
Fix video thumbnails in mixes

iBicha:


syeopite (approved):
Nice catch!

In the future Invidious should really start testing its API responses...

unixfox (approved):

PR #5120, opened 2024-12-19
Update Kemal to 1.6.0 and remove Kilt

syeopite:
Kilt is unmaintained and the ECR templating logic has been natively integrated into Kemal with the issues previously seen having been resolved.

This commit is mostly a precursor to support the next Kemal release which will add the ability to create error handlers for raised exceptions.

See https://github.com/kemalcr/kemal/pull/688


syeopite:
Looks like we'll need to drop support for Crystal versions prior to `1.12.X`

syeopite:
Yeah those templating issues seem to have been resolved. On my side I didn't notice any problems during testing

PR #5129, opened 2024-12-30
update Makefile

alindt:
- add uninstall target
- restructure and rename targets and vars

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,10 @@ MT := 0
 
 
 FLAGS ?=
+CRYSTAL_BIN ?= crystal
+SHARDS_BIN ?= shards
 
-
+# Build flags configuration
 ifeq ($(RELEASE), 1)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,60 +40,77 @@ ifeq ($(API_ONLY), 1)
...
 
 
 # -----------------------
 #  Development
 # -----------------------
 
-
 format:
-	crystal tool format
+	$(CRYSTAL_BIN) tool format --check
```
unixfox, line 74: Would prefer to have a separate arg for --check

unixfox (changes requested):

Makefile
alindt: Separate arg or a different target altogether maybe?

unixfox: not very familiar with makefiles but I guess separate arg

Issue #2007, opened 2021-04-10
[Enhancement] Add support for multiples audio tracks


SapeurPonpier:
I started watching a video ( https://yewtu.be/watch?v=fMfipiV_17o ) from MrBeast, which is a famous English-speaking channel. However, the audio was in Spanish. It seems like YouTube implemented the feature of separate audio tracks and that not only are we not able to pick which one we would like to hear (unlike on the youtube player) but that it defaulted to Spanish. Which is fine by me cause I speak a bit Spanish and it trains me but still ... it might become tedious when I come accross a video with Chinese dubs, I would most definetly not understand anything.

We probably need to add a feature to enable selection of audio track.

TheFrenchGhosty:
Adding supports for multiples audio track is needed in the player, or if videojs doesn't support multiples audio track, directly in the invidious UI.

Or alternatively just make it use the "default" one.

unixfox:
Notice: https://yewtu.be uses a custom backend, so the language selection is available.

![image](https://github.com/iv-org/invidious/assets/4016501/21a6f6d2-8b6b-4e02-b558-0607bb3f542d)

This does not mean that invidious support it natively. it's just a yewtu.be thing

d4g:
I would kindly ask to not see this as an enhancement but as a bug.
It's not an enhancement to play a video in the original language and it's not an enhancement to be able to select the original language if a different langauge is played back. This is a bug. As more and more issues regarding this are created, I would be interested if this is under investigation or postponed into oblivion?

mas1701:
> I would kindly ask to not see this as an enhancement but as a bug. It's not an enhancement to play a video in the original language and it's not an enhancement to be able to select the original language if a different langauge is played back. This is a bug. As more and more issues regarding this are created, I would be interested if this is under investigation or postponed into oblivion?

With more and more AI translations, the number of multilanguage videos will most likely increase rapidly. So IMO this is too significant to ignore. I just stumbled upon a german car review video playing in french. 👎

KwisatzJim:
I have one channel I subscribe to that I thought I hit some weird key combo to cause it to play audio in a foreign language, then I came across this thread. If I play the same video in a private tab in Firefox it plays in English. I'm using a self hosted invidious instance. The channel is Undecided with Matt Ferrell.

KwisatzJim:
[...]

PR #5149, opened 2025-01-21
[Continuing #5094] Set language info for dash audio streams and sort

giuliano-macedo:
Continuing #5094 made by **[GTechAlpha](https://github.com/GTechAlpha)** with its original commit since it hasn't had new changes for more than a month. I'm doing this by changing how the AdaptionSet label is set when there is multilang audio available but displays the birate so that it avoids #3074.

Screenshots:
screenshot1
screenshot2

### Original PR description
Set audio track info, if present, in dash manifest

    language id
    language display name
    main/default track

Sort audio formats so that main/default is first (for clients not using dash)

This resolves issue of incorrect/undesired audio when automatic translation is enabled by content creator (currently relatively rare, but likely to increase), and allows dash clients to correctly select default language and optionally offer multiple language streams.

closes https://github.com/iv-org/invidious/issues/2007

Note: this should be a non-breaking change; if audio track info is not available, the behavior does not change from current

### DIFF
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -74,12 +74,13 @@ module Invidious::Routes::API::Manifest
               lang = audio_track["id"]?.try &.as_s.split('.')[0] || "und"
               is_default = audio_track.has_key?("audioIsDefault") ? audio_track["audioIsDefault"].as_bool : i == 0
               displayname = audio_track["displayName"]?.try &.as_s || "Unknown"
+              bitrate = fmt["bitrate"]
 
               # Different representations of the same audio should be groupped into one AdaptationSet.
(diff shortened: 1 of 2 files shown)


giuliano-macedo:
@absidue Done, thanks!

syeopite (approved):

unixfox (approved):

syeopite:
Thanks for the contribution @GTechAlpha and @giuliano-macedo 😄. Great work on this  🎉

n3d1117:
Hi @giuliano-macedo thanks for your work! Now I can select the audio track on videos, however it still defaults to the AI translated instead of original language. Is there any way to change that or did I misunderstand this PR?

image

PR #5150, opened 2025-01-22
Warn when any top-level config is "CHANGE_ME!!"

syeopite:


unixfox (approved):

unixfox:
Good improvement. I won't need to check that in my companion PR anymore :)

PR #5153, opened 2025-01-22
Remove formatter check on container workflows

syeopite:


unixfox (approved):

unixfox:
Good idea since we already have it in a separate workflow :)

PR #5154, opened 2025-01-22
Pick a different instance upon redirect

epicsam123:
Currently, the redirect url may redirect the user back to the same instance they are currently on.

### DIFF
--- a/src/invidious/routes/misc.cr
+++ b/src/invidious/routes/misc.cr
@@ -46,13 +46,13 @@ module Invidious::Routes::Misc
     if instance_list.empty?
       instance_url = redirect_url
     else
-      # Sample returns an array
-      # Instances are packaged as {region, domain} in the instance list
       # Filter out the current instance
       other_available_instances = instance_list.reject! { |region, domain| domain == CONFIG.domain }
 
       # If there are any other instances, select a random one
       if other_available_instances.any?
+        # Sample returns an array
+        # Instances are packaged as {region, domain} in the instance list
         instance_url = other_available_instances.sample(1)[0][1]
       else
         # If the current instance is the only one, use the redirect URL as fallback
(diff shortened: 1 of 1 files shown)


syeopite (changes requested):
Crystal if blocks needs to be explicitly closed with an `end` statement

src/invidious/routes/misc.cr
```diff
@@ -40,14 +40,24 @@ module Invidious::Routes::Misc
...
+    redirect_url = "redirect.invidious.io"
 
     instance_list = Invidious::Jobs::InstanceListRefreshJob::INSTANCES["INSTANCES"]
     if instance_list.empty?
-      instance_url = "redirect.invidious.io"
+      instance_url = redirect_url
     else
       # Sample returns an array
       # Instances are packaged as {region, domain} in the instance list
-      instance_url = instance_list.sample(1)[0][1]
```
SamantazFox, line 50: Please move the two comments that were above this line to line to the new location of `.sample(1)[0][1]`!

syeopite (approved):
Please fix the lint issues.

src/invidious/routes/misc.cr
```diff
@@ -42,12 +42,17 @@ module Invidious::Routes::Misc
     referer = get_referer(env)
 
     instance_list = Invidious::Jobs::InstanceListRefreshJob::INSTANCES["INSTANCES"]
-    if instance_list.empty?
+    # Filter out the current instance
+    other_available_instances = instance_list.reject! { |_, domain| domain == CONFIG.domain }
```
syeopite, line 46:
```suggestion
    other_available_instances = instance_list.reject { |_, domain| domain == CONFIG.domain }
```

`reject!` modifies itself instead of creating a new Tuple. `other_available_instances` isn't necessary unless you're using `reject`

syeopite (approved):
LGTM

PR #5168, opened 2025-02-02
Feature: Reply Counts

epicsam123:
Productivity feature: include reply counts for both the "load more" and show/hide replies buttons.

Easily tested this through browser dev tools since core source code changes were purely JavaScript.


unixfox:
hey @AHOHNMYC @ChunkyProgrammer, could you please take a look at this PR for the JS side? thank you :D

assets/js/comments.js
```diff
@@ -1,5 +1,13 @@
 var video_data = JSON.parse(document.getElementById('video_data').textContent);
 
+var isRTL = (() => {
```
ChunkyProgrammer, line 3: This looks a bit messy tbh. Maybe just look at the html language and see if matches one of the rtl languages?
(Like here: https://github.com/iv-org/invidious/blob/164d764d553921e6ee98facda241f13c2103ec90/src/invidious/helpers/i18n.cr#L191-L198)

ghost (commented):
Haven't tested but I have this one comment

unixfox:
hello @ChunkyProgrammer, could do a final check in order to see if it is ok?

ghost (approved):
@unixfox  Code LGTM

It also LGTM on the video page and community page

# Here are some screenshots:
## Initial look:
image
## after opening replies (top)
image
## after opening replies (bottom)
image

## after loading all the comments:
image


## Comment count change:
### before loading all:
image

### after loading all:
image


## Community page:
image

Minecon724:
I don't see how the (x/x) is useful. I would keep "view x replies" loaded or not.

epicsam123:
> I don't see how the (x/x) is useful. I would keep "view x replies" loaded or not.

I personally like the prospect where with long comment chains you can know how far you have read into a comment's long thread of replies (characteristic of the first couple of comments due to popularity ranking) after initially collapsing the comment and then returning to the top of the comment section to find that same comment. I can modify the counter to only display the final count i.e. (40) vs (40/40) when all the comments are loaded.

Minecon724:
> > I don't see how the (x/x) is useful. I would keep "view x replies" loaded or not.
> 
> I personally like the prospect where with long comment chains you can know how far you have read into a comment's long thread of replies (characteristic of the first couple of comments due to popularity ranking) after initially collapsing the comment and then returning to the top of the comment section to find that same comment. I can modify the counter to only display the final count i.e. (40) vs (40/40) when all the comments are loaded.

Maybe do that in a "title" attribute? It's displayed on hover
Either way I suggest making it clearer like "(40/40 loaded)"

epicsam123:
> > > I don't see how the (x/x) is useful. I would keep "view x replies" loaded or not.
> > 
> > 
> > I personally like the prospect where with long comment chains you can know how far you have read into a comment's long thread of replies (characteristic of the first couple of comments due to popularity ranking) after initially collapsing the comment and then returning to the top of the comment section to find that same comment. I can modify the counter to only display the final count i.e. (40) vs (40/40) when all the comments are loaded.
> 
> Maybe do that in a "title" attribute? It's displayed on hover Either way I suggest making it clearer like "(40/40 loaded)"

Added to both title and sub text (for mobile users)

epicsam123:
Fractions are read as left to right in RTL languages:
https://en.wikipedia.org/wiki/Modern_Arabic_mathematical_notation

epicsam123:
@ChunkyProgrammer Would you be able to review this again? Thanks :)

ghost:
Code is looking even better now (haven't tested it this time around)

locales/en-US.json
```diff
@@ -220,6 +220,7 @@
     "View Reddit comments": "View Reddit comments",
     "Hide replies": "Hide replies",
     "Show replies": "Show replies",
+    "Replies loaded": "loaded",
```
ghost: Maybe this should be "replies loaded" instead of "loaded"? 

```suggestion
    "Replies loaded": "replies loaded",
```

epicsam123:
@unixfox Who else needs to review before this moves forward? Thanks!

src/invidious/views/watch.ecr
```diff
@@ -57,8 +57,10 @@ we're going to need to do it here in order to allow for translations.
     "reddit_comments_text" => HTML.escape(I18n.translate(locale, "View Reddit comments")),
     "reddit_permalink_text" => HTML.escape(I18n.translate(locale, "View more comments on Reddit")),
     "comments_text" => HTML.escape(I18n.translate(locale, "View `x` comments", "{commentCount}")),
+    "replies_loaded_text" => HTML.escape(I18n.translate(locale, "Replies loaded")),
     "hide_replies_text" => HTML.escape(I18n.translate(locale, "Hide replies")),
     "show_replies_text" => HTML.escape(I18n.translate(locale, "Show replies")),
+    replies")),
```
epicsam123:
```suggestion
```

PR #5169, opened 2025-02-06
include_youtube_links Setting

epicsam123:
I use Invidious mainly for preventing Google tracking. As a user, I thought it would be nice to include the option to hide YT links in the UI as I sometimes accidentally click on them (primarily the comment links when on mobile). I also included some missing "noreferrers".

Translations were provided by perplexity.ai

### DIFF
--- a/locales/ar.json
+++ b/locales/ar.json
@@ -73,7 +73,6 @@
     "preferences_thin_mode_label": "الوضع الخفيف: ",
     "preferences_category_misc": "تفضيلات متنوعة",
     "preferences_automatic_instance_redirect_label": "إعادة توجيه المثيل التلقائي (إعادة التوجيه إلى redirect.invidious.io): ",
-    "preferences_include_youtube_links_label": "عرض روابط إعادة التوجيه على يوتيوب: ",
     "preferences_category_subscription": "تفضيلات الاشتراك",
     "preferences_annotations_subscribed_label": "عرض الملاحظات في الفيديوهات تلقائيا في القنوات المشترك بها فقط: ",
     "Redirect homepage to feed: ": "إعادة التوجية من الصفحة الرئيسية لصفحة المشتركين (لرؤية اخر فيديوهات المشتركين): ",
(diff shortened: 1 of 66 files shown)


src/invidious/views/privacy.ecr
```diff
@@ -37,7 +37,7 @@
 2019-01-19 16:37:53 +00:00 200 GET /vi/r5P-f5arPXE/maxres.jpg 1085.41ms
 2019-01-19 16:37:54 +00:00 200 GET /watch 7.04ms</code></pre>
     <p>This website does not store the visitor's user-agent or IP address and does not use fingerprinting, advertisements, or tracking of any form.</p>
-    <p>This website provides links to googlevideo.com to provide audio and video playback. googlevideo.com is owned by Google and is subject to their <a href="https://policies.google.com/privacy">privacy policy</a>.</p>
+    <p>This website provides links to googlevideo.com to provide audio and video playback. googlevideo.com is owned by Google and is subject to their <a rel="noopener noreferrer" href="https://policies.google.com/privacy">privacy policy</a>.</p>
```
epicsam123, line 40: There are other external links that I didn't modify. Should we consider those as well?

locales/ar.json
```diff
@@ -73,6 +73,7 @@
     "preferences_thin_mode_label": "الوضع الخفيف: ",
     "preferences_category_misc": "تفضيلات متنوعة",
     "preferences_automatic_instance_redirect_label": "إعادة توجيه المثيل التلقائي (إعادة التوجيه إلى redirect.invidious.io): ",
+    "preferences_include_youtube_links_label": "عرض روابط إعادة التوجيه على يوتيوب: ",
```
ChunkyProgrammer, line 76: You only need to provide translations for en-US. 

Other translations are done through weblate (ex: https://github.com/iv-org/invidious/pull/4989/files ).

syeopite:
Although not mutually exclusive, this PR does overlap and change the same sections of code as https://github.com/iv-org/invidious/pull/4826, which adds a confirmation page before visiting YouTube links.

I'm not sure if I personally agree with adding more code to remove the links from the frontend—which can increase the maintenance burden—when a confirmation page will serve the same purpose for safeguarding against tracking.

@SamantazFox @unixfox what do you think?

SamantazFox:
@syeopite I'm in favor of the confirmation page proposed in that other pull request. I agress that we already have enough config options like that.

epicsam123:
Will close this in favor of https://github.com/iv-org/invidious/pull/4826, thanks!

Issue #5182, opened 2025-02-18
[Bug] Community tab disappeared from channels information


Fijxu:
**Describe the bug**

Invidious used to display a community tab on the channel information, but it randomly disappeared.

**Steps to Reproduce**

Open any channel with a "Posts" tab on it. Example: https://www.youtube.com/@SomeOrdinaryGamers/community

**Screenshots**

![Image](https://github.com/user-attachments/assets/0aececfa-19ed-4cf9-afa0-f7d8da74c88d)

(There should be a "Community" tab shown there)

**Additional context**

Easy to fix. Youtube changed the "Community" tab name by "Posts", so Invidious is not able to detect it because it's hardcoded to search for "community". 

![Image](https://github.com/user-attachments/assets/261d3cec-d2e0-43a2-a8c4-fae33585cd53)

How should we handle this name change? Should invidious continue to use the "Community" name on everything? (api, urls, channel info) or it should handle it the same way on the old urls? (`/api/v1/channels/:ucid/community` returns the same as `/api/v1/channels/:ucid/posts`. `/channel/:ucid/community` return the same as `/channel/:ucid/posts`)

PR #5183, opened 2025-02-18
Channels: Fix community tab

Fijxu:
May close https://github.com/iv-org/invidious/issues/5182

image

API works well too.

If I need to change anything, just tell me

### DIFF
--- a/src/invidious/routes/channels.cr
+++ b/src/invidious/routes/channels.cr
@@ -198,6 +198,8 @@ module Invidious::Routes::Channels
   end
 
   def self.community(env)
+    return env.redirect env.request.path.sub("posts", "community") if env.request.path.split("/").last == "posts"
+
     data = self.fetch_basic_information(env)
     if !data.is_a?(Tuple)
       return data
@@ -214,7 +216,7 @@ module Invidious::Routes::Channels
 
     continuation = env.params.query["continuation"]?
 
-    if !channel.tabs.includes? "community"
+    if !channel.tabs.includes? "community" && "posts"
       return env.redirect "/channel/#{channel.ucid}"
     end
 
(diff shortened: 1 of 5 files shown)


src/invidious/routes/channels.cr
```diff
@@ -308,6 +310,7 @@ module Invidious::Routes::Channels
   private KNOWN_TABS = {
     "home", "videos", "shorts", "streams", "podcasts",
     "releases", "playlists", "community", "channels", "about",
+    "posts",
```
Fijxu, line 313: Here I didn't replace "community" by "posts". I left it as a fallback.

src/invidious/routes/channels.cr
```diff
@@ -198,6 +198,8 @@ module Invidious::Routes::Channels
   end
 
   def self.community(env)
+    return env.redirect env.request.path.sub("posts", "community") if env.request.path.split("/").last == "posts"
+
```
syeopite, line 202: I think `/community` should be redirected to `/posts` instead. The canonical URL should more or less follow YouTube  imo. What do you think?

Fijxu: Youtube doesn't has a posts endpoint, that is why I made that redirect.

image

try it by yourself if you want https://www.youtube.com/@SomeOrdinaryGamers/posts

syeopite (approved):

ChunkyProgrammer:
Suggestion: Since the url didn't change and just the text, maybe this can be handled similarly to how channel live tab is handled? (1 ternary statement that changes `live` to `streams`).

https://github.com/iv-org/invidious/blob/164d764d553921e6ee98facda241f13c2103ec90/src/invidious/channels/about.cr#L105-L112

PR #5188, opened 2025-02-20
captions: provide "w", "o", "-", "+" keydowns for player from YT

epicsam123:
Added additional keydown events missing from YouTube's interface, including:
- toggle the size of the captions with "-", "+" ("+" is recognized as "=")
- toggle the opacity of the cations with "o"
- toggle the caption window with "w"

Big thanks to the Invidious team for all their work. Please keep this project alive!

### DIFF
--- a/assets/css/player.css
+++ b/assets/css/player.css
@@ -71,8 +71,10 @@
   padding-top: 2em
 }
 
-.video-js.player-style-youtube .vjs-progress-control .vjs-progress-holder, .video-js.player-style-youtube .vjs-progress-control {height: 5px;
-margin-bottom: 10px;}
+.video-js.player-style-youtube .vjs-progress-control .vjs-progress-holder, .video-js.player-style-youtube .vjs-progress-control {
+  height: 5px;
+  margin-bottom: 10px;
+}
 
 ul.vjs-menu-content::-webkit-scrollbar {
   display: none;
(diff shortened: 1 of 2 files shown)


assets/css/player.css
```diff
@@ -71,8 +71,10 @@
   padding-top: 2em
 }
 
-.video-js.player-style-youtube .vjs-progress-control .vjs-progress-holder, .video-js.player-style-youtube .vjs-progress-control {height: 5px;
-margin-bottom: 10px;}
+.video-js.player-style-youtube .vjs-progress-control .vjs-progress-holder, .video-js.player-style-youtube .vjs-progress-control {
```
syeopite, line 74: Are the changes in player.css necessary? It seems like to me that the changes in player.js will be enough to implement this feature.

assets/js/player.js
```diff
@@ -591,6 +598,31 @@ function increase_playback_rate(steps) {
     player.playbackRate(options.playbackRates[newIndex]);
 }
 
+function increase_caption_size(steps) {
```
syeopite, line 601: VideoJS has a native caption settings feature. If possible I think it'd be better to interface with those settings rather than to implement it on Invidious' side through manipulating the CSS.

assets/css/player.css
epicsam123: Reverted changes apart from one "line-height" modification that fit better for the caption window.

assets/js/player.js
epicsam123: Implemented changes with `player.textTrackSettings` interface.

assets/js/player.js
```diff
@@ -5,6 +5,9 @@ var video_data = JSON.parse(document.getElementById('video_data').textContent);
 var options = {
     liveui: true,
     playbackRates: [0.25, 0.5, 0.75, 1.0, 1.25, 1.5, 1.75, 2.0],
+    fontPercent: [0.5, 0.75, 1.25, 1.5, 1.75, 2, 3, 4],
```
epicsam123, line 8: I verified that these are the only values that are accepted by videojs for each of the three caption properties.

SamantazFox (approved):

syeopite (changes requested):
Hi sorry about the late response. Everything looks good and I'll be merging this soon!

But could you move the YouTube player style css to another PR? The line-height change for the captions are fine but the player changes are unrelated and best suited for another PR.

syeopite (approved):

syeopite:
This PR is still set to be merged. It's just that baring important fixes we generally try to merge PRs in batches to avoid extra maintenance burden from new image spam on the instance maintainers side.

Unfortunately this can lead to a PR waiting for ages if there aren't anything else to batch it with even if it has already been marked as ready... 

Perhaps we can start to merge singular PRs if there hasn't been any recent unstable image releases but the invidious team will have to decide on that.

epicsam123:
Ok I added another smaller PR that could help bundle this and other approved ones once ready: https://github.com/iv-org/invidious/pull/5368

PR #5189, opened 2025-02-20
Fix an issue with the HLS manifest check for livestream videos

alexmaras:
Originally, the HLS manifest check was essentially a boolean: if the HLS manifest field was present, it was assumed to be a livestream. Some videos include the HLS Manifest but aren't livestreams.
    
In the case where they are livestreams, the video contains a videoType field with the value "Livestream". In the case that they're normal videos, the videoType is "Video". This is exposed via the video.live_now method.
    
This commit just checks that video.live_now is true before treating it as a livestream

### DIFF
--- a/src/invidious/views/components/player.ecr
+++ b/src/invidious/views/components/player.ecr
@@ -4,7 +4,7 @@
     <% if params.autoplay %>autoplay<% end %>
     <% if params.video_loop %>loop<% end %>
     <% if params.controls %>controls<% end %>>
-    <% if (hlsvp = video.hls_manifest_url) && !CONFIG.disabled?("livestreams") %>
+    <% if (hlsvp = video.hls_manifest_url) && video.live_now && !CONFIG.disabled?("livestreams") %>
         <source src="<%= URI.parse(hlsvp).request_target %><% if params.local %>?local=true<% end %>" type="application/x-mpegURL" label="livestream">
     <% else %>
         <% if params.listen %>


alexmaras:
Fixes: https://github.com/iv-org/invidious-companion/issues/35

There's also another place where the simple check for an HLS manifest exists: https://github.com/iv-org/invidious/blob/164d764d553921e6ee98facda241f13c2103ec90/src/invidious/jsonify/api_v1/video_json.cr#L72

Does this also need to be changed?

The alternative approach to this would be to add a boolean property to the `video` - `video.is_livestream?` which can resolve to a boolean and centralise that check. Happy to adjust and take that approach. 

This change has been tested on my local instance and is working well with:
- videos that have an hls_manifest_url but aren't a livestream
- videos that have an hls_manifest_url and _are_ a livestream
- videos that don't have an hls_manifest_url

Let me know if anything makes sense to change.

src/invidious/views/components/player.ecr
```diff
@@ -4,7 +4,7 @@
     <% if params.autoplay %>autoplay<% end %>
     <% if params.video_loop %>loop<% end %>
     <% if params.controls %>controls<% end %>>
-    <% if (hlsvp = video.hls_manifest_url) && !CONFIG.disabled?("livestreams") %>
+    <% if (hlsvp = video.hls_manifest_url) && hlsvp.includes?("/source/yt_live_broadcast/") && !CONFIG.disabled?("livestreams") %>
```
syeopite, line 7: You should just be able to check whether `video.live_now` is true instead of the URL

syeopite:
If I'm understanding everything correctly this is something that should be fixed by https://github.com/iv-org/invidious/pull/4118, whenever that is merged.

This part of the PR prevents Invidious's frontend from using the HLS manifest unless the video is a livestream or the user has selected HLS quality
https://github.com/iv-org/invidious/blob/ffc44b0cff507206880365b17cf09e308f3f18ae/src/invidious/views/components/player.ecr#L7-L10

src/invidious/views/components/player.ecr
alexmaras: Thanks - great catch. I'm still digging through the codebase, this does a better job that checking the URL. 

I've updated to that in case it's worth merging this in in the meantime as some videos are broken at the moment.

alexmaras:
Thanks @syeopite - yep, looks like the fix in #4118 covers this completely. If that one is still a little while away, this fixes the issue for the moment until that PR gets merged in. I've amended the commit to handle the issue using the video.live_now method instead.

syeopite (approved):

PR #5196, opened 2025-03-01
Handle parse errors gracefully on timeline items

syeopite:
This PR makes it so that if one item in the timeline fails to parse, the rest of the page will still be usable and the failed item will instead be represented with an error card.

This is done creating a new `SearchItem` known as a `ProblematicTimelineItem` which represents an item that has failed to parse. The frontend and API will render this as a single error card while the rest of the page renders normally. 

graceful errors

### DIFF
--- a/src/invidious/helpers/errors.cr
+++ b/src/invidious/helpers/errors.cr
@@ -31,7 +31,7 @@ def get_issue_template(env : HTTP::Server::Context, exception : Exception) : Tup
 
   issue_template += github_details("Backtrace", exception.inspect_with_backtrace)
 
-  return {issue_title, issue_template}
+  return issue_title, issue_template
 end
 
 def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exception : Exception)
@@ -78,7 +78,7 @@ def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exce
       <p>#{translate(locale, "crash_page_report_issue", url_new_issue)}</p>
 
       <!-- TODO: Add a "copy to clipboard" button -->
-      <pre style="padding: 20px; background: rgba(0, 0, 0, 0.12345);">#{issue_template}</pre>
+      <pre class="error-issue-template">#{issue_template}</pre>
     </div>
   END_HTML
 
--- a/src/invidious/helpers/errors.cr
+++ b/src/invidious/helpers/errors.cr
@@ -18,16 +18,7 @@ def github_details(summary : String, content : String)
   return HTML.escape(details)
 end
 
-def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exception : Exception)
-  if exception.is_a?(InfoException)
-    return error_template_helper(env, status_code, exception.message || "")
-  end
-
-  locale = env.get("preferences").as(Preferences).locale
-
-  env.response.content_type = "text/html"
-  env.response.status_code = status_code
-
+def get_issue_template(env : HTTP::Server::Context, exception : Exception) : Tuple(String, String)
   issue_title = "#{exception.message} (#{exception.class})"
 
   issue_template = <<-TEXT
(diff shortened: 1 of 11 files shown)


src/invidious/helpers/errors.cr
```diff
@@ -40,6 +31,24 @@ def error_template_helper(env : HTTP::Server::Context, status_code : Int32, exce
 
   issue_template += github_details("Backtrace", exception.inspect_with_backtrace)
 
+  return {issue_title, issue_template}
```
SamantazFox, line 34: Regarding the comment on lines  47/48, maybe try without the `{}`:
```suggestion
  return issue_title, issue_template
```

src/invidious/playlists.cr
```diff
@@ -500,6 +500,10 @@ def extract_playlist_videos(initial_data : Hash(String, JSON::Any))
         index:          index,
       })
     end
+  rescue ex
+    videos << ProblematicTimelineItem.new(
+      parse_exception: ex
+    )
```
SamantazFox, line 506: No need to be so verbose ^^
```suggestion
    videos << ProblematicTimelineItem.new(ex)
```

src/invidious/routes/embed.cr
```diff
@@ -12,13 +12,15 @@ module Invidious::Routes::Embed
           url = "/playlist?list=#{plid}"
           raise NotFoundException.new(translate(locale, "error_video_not_in_playlist", url))
         end
+
+        get_first_video = videos[0].as(PlaylistVideo)
```
SamantazFox, line 16: Do you mind renaming these to `first_video`?
`get_first_video` sound too much like a method or function.

src/invidious/helpers/serialized_yt_data.cr
```diff
@@ -291,6 +291,55 @@ struct SearchHashtag
...
+  end
+
+  # Provides compatibility with PlaylistVideo
+  def to_json(json : JSON::Builder, *args, **kwargs)
+    return to_json("", json)
+  end
+
+  def to_xml(env, locale, xml : XML::Builder)
+    xml.element("entry") do
+      xml.element("id") { xml.text "iv-err-#{Random.new.base64(8)}" }
```
SamantazFox, line 323: Why not use `@id` here?

src/invidious/yt_backend/extractors.cr
```diff
@@ -45,13 +59,16 @@ private module Parsers
   # `videoRenderer`s can be found almost everywhere on YouTube. In categories, search results, channels, etc.
   #
   module VideoRendererParser
+    extend self
+    include BaseParser
+
     def self.process(item : JSON::Any, author_fallback : AuthorFallback)
```
SamantazFox, line 65: You can also remove `self` here
```suggestion
    def process(item : JSON::Any, author_fallback : AuthorFallback)
```

Note: please keep the `self` in `return self.parse(...)`, as it's easier to understand that we're using a module method and rathern than an external function.

src/invidious/views/components/item.ecr
```diff
@@ -97,6 +97,18 @@
...
+        <% when ProblematicTimelineItem %>
+            <div class="error-card">
+                <div class="explanation">
+                    <i class="icon ion-ios-alert"></i>
+                    <h4><%=translate(locale, "timeline_parse_error_placeholder_heading")%></h4>
+                    <p><%=translate(locale, "timeline_parse_error_placeholder_message")%></p>
+                </div>
+                <details>
+                    <summary class="pure-button pure-button-secondary"><%=translate(locale, "timeline_parse_error_show_technical_details")%></summary>
+                    <pre style="padding: 20px; background: rgba(0, 0, 0, 0.12345);"><%=get_issue_template(env, item.parse_exception)[1]%></pre>
```
SamantazFox, line 109: Can you move that inlined style to the CSS file?

src/invidious/helpers/errors.cr
syeopite: Same problem occurs once you change

```crystal
  issue_template_components = get_issue_template(env, exception)
  issue_title, issue_template = issue_template_components
```
to
```crystal
  issue_title, issue_template = get_issue_template(env, exception)
```

SamantazFox: And what happens if you replace:
`def get_issue_template(...) : Tuple(String, String)`
with:
`def get_issue_template(...) : {String, String}`

syeopite: It still occurs.

The compiler error I mentioned isn't due to a compile time syntax or type error but rather a bug in the compiler itself.

```
BUG: `__temp_6137 = get_issue_template(env, exception)` at /home/syeopite/Work/Software/iv-org/invidious/src/invidious/helpers/errors.cr:49:33 has no type (Exception)
  from crystal in '??'
  from crystal in '??'
  from crystal in '??'
  (...)
  from crystal in '__crystal_main'
  from crystal in '??'
  from crystal in 'main'
  from /usr/lib/libc.so.6 in '??'
  from /usr/lib/libc.so.6 in '__libc_start_main'
  from crystal in '_start'
  from ???
Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
````

syeopite: https://github.com/crystal-lang/crystal/issues/15638

SamantazFox (approved):

PR #5197, opened 2025-03-06
Release v2.20250314.0

syeopite:
I'm planning on making a new release next week on March 14th. 

Please take a look at the changelog to see if I missed anything, or if any revision is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Tamil is now available as an interface language
 
 Automatic instance redirects will no longer have the chance to annoyingly redirect to the same instance you're on.
 
+Due to their requirements for video playback, Invidious will log warning messages when either inv-sig-helper, `po_token` or `visitor_data` is not configured
+
 Invidious is now able to listen through a UNIX socket
 
 User notifications are now batched for each channel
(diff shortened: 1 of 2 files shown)


unixfox (approved):

YuchkinUA:
Hi, @syeopite I wanted to watch the video 18 +

This error came out, what to do?
image

Thanks :3

syeopite:
> Hi, @syeopite I wanted to watch the video 18 +
> 
> This error came out, what to do?
> image
> 
> Thanks :3 

There is nothing you can do. YouTube has made it basically impossible to watch age restricted videos without an account and as such Invidious is unable to play them.

Please do not hijack a pull request to ask your question next time.

unixfox:
I was also thinking that it would be a good idea to create a release where we enforce inv_sig_helper because now it is necessary for invidious to work.

syeopite:
That's still going to be a pretty significant breaking change. Some people use the Invidious frontend to browse and search only while watching the video else where. Enforcing sighelper would break those setups and force them to install an application they don't need.

Plus since sighelper is going to be replaced by companion in the near future imo it doesn't make sense to introduce two breaking changes of enforcing sighelper and then replacing it with companion in such a short timeframe

PR #5199, opened 2025-03-06
feat: add option to change username

Fijxu:
Tested and it works fine. I made this for people that want to change their username on public Invidious instance without having to create a new account or having to contact the instance maintainer to change their username. It also updates the materialized views sha256 ids, so if a user changed their account name from `user` to `user2` and a new user registers on the instance with the `username`, there shouldn't be a relation collision (when the collision happens, this is displayed: `relation "subscriptions_sha256" already exists`)

Tested:

- Changing username from `user1` to `user2`, subscriptions are maintained across username changes
- Playlist author (for Invidious playlists) updated on username change.
- Changing the username from `user2` to `user1` works well.
- If the user tries to change their username to one that already exists on the database, a "Username is already taken, use another one" will be returned. (Although someone can still register or change their username with an already taken username but with a slight unnoticeable SPACE character on it, it's possible to have a user called `user1` and `user1 `)

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -501,5 +501,11 @@
     "toggle_theme": "Toggle Theme",
     "carousel_slide": "Slide {{current}} of {{total}}",
     "carousel_skip": "Skip the Carousel",
-    "carousel_go_to": "Go to slide `x`"
+    "carousel_go_to": "Go to slide `x`",
+    "new_username": "New username",
+    "change_username": "Change username",
+    "username_required_field": "Username is a required field",
+    "username_empty": "Username cannot be empty",
+    "username_is_the_same": "This is your username, use another one",
+    "username_taken": "Username is already taken, use another one"
 }
--- a/src/invidious/database/users.cr
+++ b/src/invidious/database/users.cr
@@ -214,6 +214,13 @@ module Invidious::Database::Users
     PG_DB.exec(request, username, user.email)
   end
 
+  def update_user_materialized_view(user : User, username : String)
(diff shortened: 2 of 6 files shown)


Fijxu:
Note: This conflicts with https://github.com/iv-org/invidious/pull/2469 due to the use of materialized views.

syeopite (changes requested):

locales/en-US.json
```diff
@@ -501,5 +501,11 @@
...
     "carousel_slide": "Slide {{current}} of {{total}}",
     "carousel_skip": "Skip the Carousel",
-    "carousel_go_to": "Go to slide `x`"
+    "carousel_go_to": "Go to slide `x`",
+    "accounts_new_username": "New username",
+    "accounts_change_username": "Change username",
+    "accounts_username_required_field": "Username is a required field",
+    "accounts_username_empty": "Username cannot be empty",
+    "accounts_username_is_the_same": "This is your username, use another one",
+    "accounts_username_taken": "Username is already taken, use another one"
```
SamantazFox, line 510: We'd need to use UUIDs of some sort to differenciate accounts at some point. I'm worried that account name switching could be used to find if someone else is registered on this instance.

syeopite (changes requested):

locales/en-US.json
Fijxu: >I'm worried that account name switching could be used to find if someone else is registered on this instance.

That is already possible in the login page, if you insert a registered user with a wrong password it will return `Wrong username or password`.

PR #5201, opened 2025-03-10
fix: add missing check for c.youtube.com host

Fijxu:
Fixes that I said on https://github.com/iv-org/invidious/pull/4992#issuecomment-2608227126
Tested with `inv_sig_helper` and https://github.com/iv-org/invidious/pull/4985


absidue:
There is already an open PR for this: #4589

PR #5202, opened 2025-03-10
Warn when `po_token`, `visitor_data` and/or `inv-sig-helper` is not configured

syeopite:
https://github.com/iv-org/invidious/pull/5197#issuecomment-2708840771

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -167,13 +167,13 @@ DECRYPT_FUNCTION =
   if sig_helper_address = CONFIG.signature_server.presence
     IV::DecryptFunction.new(sig_helper_address)
   else
-    LOGGER.warn("WARNING: inv-sig-helper is required for video playback. See https://docs.invidious.io/installation/#run-inv_sig_helper-in-background for more information")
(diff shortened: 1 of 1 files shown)


unixfox:
Thanks :). Maybe point out a link to the doc? https://docs.invidious.io/installation/

src/invidious.cr
```diff
@@ -167,9 +167,16 @@ DECRYPT_FUNCTION =
   if sig_helper_address = CONFIG.signature_server.presence
     IV::DecryptFunction.new(sig_helper_address)
   else
+    LOGGER.warn("WARNING: inv-sig-helper is required for video playback. See https://docs.invidious.io/installation/#run-inv_sig_helper-in-background for more information")
     nil
   end
 
+{% for field in %w(po_token visitor_data) %}
+  if !CONFIG.{{field.id}}
+    LOGGER.warn("WARNING: {{field.id}} is required to view and playback videos. See https://docs.invidious.io/installation/#generate-po_token-and-visitor_data-identities for more information")
```
unixfox, line 176: I would prefer to have just https://docs.invidious.io/installation/. The URL you are linking to is for the Linux installation. If one is running Docker (which most people are) then they will get confused as to why they need to install crystal locally and so on.

PR #5205, opened 2025-03-12
Audio: Prioritize "original" audio tracks when selecting default

D-Hunt3r:
Check for "original" in audio track's display name when determining which track should be the default. If any track has "original" in its name, only such tracks will be marked as default. Otherwise, fall back to existing logic.

### DIFF
--- a/src/invidious/routes/api/manifest.cr
+++ b/src/invidious/routes/api/manifest.cr
@@ -59,13 +59,33 @@ module Invidious::Routes::API::Manifest
             mime_streams = audio_streams.select { |stream| stream["mimeType"].as_s.starts_with? mime_type }
             next if mime_streams.empty?
 
+            # First scan to check if any track has "original" in its name
+            original_track_found = false
+            mime_streams.each do |fmt|
+              audio_track = fmt["audioTrack"]?.try &.as_h? || {} of String => JSON::Any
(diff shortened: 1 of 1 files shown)


syeopite:
This approach only works if the YouTuber hasn't set any custom labels for the audio tracks. I'm also not sure if YouTube even adds the `original` tag if a video has multiple audio tracks of the same language.

absidue:
If you look at the xtags (either the property with the protobuf or the url query parameter) you can get the audio content type without having to rely on parsing display labels.

Here's how it is done in YouTube.js (removed code is the URL query parameter handling and the added code is the protobuf handling): https://github.com/LuanRT/YouTube.js/pull/909/files

D-Hunt3r:
It may not be a complete solution, but I think 80% is definitely more useful than nothing! Plus, I’ve included a fallback to the old logic. 

My code has been running smoothly for a few days, and I believe it enhances the current setup. 
I’m looking forward to your feedback

absidue:
As the maintainers have already expressed concerns about the implementation in this pull request and changing to the proper approach is not that difficult (you don't even have to think of another approach as I've already provided you with a link), it probably makes more sense to do it correctly from the start, than merging an incomplete solution and then having to do a second pull request afterwards to clean it up.

PR #5222, opened 2025-03-26
Add badge for channel-only videos and settings to exclude them

syeopite:


src/invidious/config.cr
```diff
@@ -52,6 +52,7 @@ struct ConfigPreferences
   property vr_mode : Bool = true
   property show_nick : Bool = true
   property save_player_pos : Bool = false
+  property exclude_members_only_videos : Bool = false
```
SamantazFox, line 55: Don't forget to add the corresponding entry to `config.example.yaml`!

efb4f5ff-1298-471a-8973-3d47447115dc:
FYI: https://github.com/FreeTubeApp/FreeTube/pull/7208#issuecomment-2816593838

src/invidious/yt_backend/extractors.cr
```diff
@@ -135,6 +135,10 @@ private module Parsers
         when "Premium"
           # TODO: Potentially available as item_contents["topStandaloneBadge"]["metadataBadgeRenderer"]
           badges |= VideoBadges::Premium
+        when "Members only"
```
PikachuEXE, line 138: There is also `Members first` (from https://github.com/FreeTubeApp/FreeTube/pull/7208#issuecomment-2816593838
image

image

syeopite:
Thanks!

src/invidious/config.cr
syeopite, line 55: Should we exclude the channel only videos by default?

SamantazFox: Honnestly, I don't see a reason to show member-only videos, as Invidious users can't watch them anyway. We always create a dedicated channel page/place for the various "special" playlists (`UUMO` and co).

src/invidious/views/feeds/trending.ecr
```diff
@@ -42,6 +42,7 @@
 
 <div class="pure-g">
 <% trending.each do |item| %>
+    <% next %>
```
SamantazFox, line 45: I suppose this is a leftover from something else?

src/invidious/views/components/item.ecr
```diff
@@ -200,6 +200,19 @@
...
+                <div class="video-card-row flexible video-badges">
+                    <%
+                        # TODO Other types of badges
+                    %>
+                    <% if item.badges.members_only? %>
+                        <span><i class="icon ion ion-md-lock"></i><%=translate(locale, "video_badges_members_only")%></span>
+                    <% end %>
+                    <% if item.badges.members_first? %>
+                        <span><i class="icon ion ion-md-lock"></i><%=translate(locale, "video_badges_members_first")%></span>
+                    <% end %>
```
SamantazFox, line 213: What do you think about putting the `lock` icon before the title?

PR #5228, opened 2025-04-01
fix: restore dmca_content functionality

Fijxu:
This restores (or adds) the functionality of the `dmca_content` config option that at this date, has been unused and makes no effect.

I prefer to leave it enabled only for the frontend, I'm not sure about the API, so tell me what do you guys (iv-org) prefer

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -501,5 +501,6 @@
     "toggle_theme": "Toggle Theme",
     "carousel_slide": "Slide {{current}} of {{total}}",
     "carousel_skip": "Skip the Carousel",
-    "carousel_go_to": "Go to slide `x`"
+    "carousel_go_to": "Go to slide `x`",
+    "dmca_content": "Sorry, this video cannot be played on this instance due to a DMCA/copyright infringement letter sent to the instance administrator."
 }
--- a/src/invidious/routes/watch.cr
+++ b/src/invidious/routes/watch.cr
@@ -30,6 +30,10 @@ module Invidious::Routes::Watch
       return env.redirect "/"
     end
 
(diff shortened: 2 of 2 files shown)


syeopite (commented):
The `dmca_content` config was originally just used to disable the download widget for specific videos

Seems like the configuration check was forgotten when the download widget and logic was extracted from `watch.ecr` in #2922

----

I don't really have any particular opinion on whether this should be restored or not, or whether the api should be affected.

But if this attribute is expanded to encompass the `/watch` page then the documentation in  `config.example.yml` and `config.cr` needs to be updated

Fijxu:
That should be fine. That restores the original functionality of what `dmca_content` setting used to do. I've also added a message showing the reason why (since the setting is called dmca_content)

image

I think is a great idea to add an option to block the whole video and not just the download, because instance administrators that are forced to comply with DMCA, will have not have an option to block the videos directly on Invidious and instead they will need to block it at the webserver level

SamantazFox:
> Seems like the configuration check was forgotten when the download widget and logic was extracted from `watch.ecr` in #2922

Yep, indeed.

> I don't really have any particular opinion on whether this should be restored or not, or whether the api should be affected.

I don't see why the API would be affected? This option only removes the easy download option, but it doesn't prenvent video playback.

In all cases, people making DMCA requests should be redirected to youtube, as we're not hosting any content, and thus unable to remove it from there.

Fijxu:
@SamantazFox We leave it like this or we also disable it from the API? Every public instance already has their API disabled and is impossible to keep it enabled, otherwise it will just be abused and rate limit the instance

unixfox (approved):

PR #5234, opened 2025-04-09
fix: content-disposition-header added

ekansh2:
Added content-disposition-header for downloading the file.

### DIFF
--- a/src/invidious/routes/watch.cr
+++ b/src/invidious/routes/watch.cr
@@ -321,6 +321,9 @@ module Invidious::Routes::Watch
       env.params.query["title"] = filename
       env.params.query["label"] = URI.decode_www_form(label.as_s)
 
+      # Set content disposition header to force download
+      env.response.headers["Content-Disposition"] = "attachment; filename=\"#{filename}\""
+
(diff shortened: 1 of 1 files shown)


src/invidious/routes/watch.cr
```diff
@@ -321,6 +321,9 @@ module Invidious::Routes::Watch
       env.params.query["title"] = filename
       env.params.query["label"] = URI.decode_www_form(label.as_s)
 
+      # Set content disposition header to force download
+      env.response.headers["Content-Disposition"] = "attachment; filename=\"#{filename}\""
```
syeopite, line 325: Please also add the `filename*` parameter needed for handling utf-8 strings on modern browsers. 

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_response_header_for_the_main_body for more information.

Implementation reference for the encoding: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#encoding_for_content-disposition_and_link_headers

PR #5081, opened 2024-11-12
Refactor connection pool

syeopite:
Brings improvements from https://github.com/jgaskins/http_client which is a shard that implements a HTTP connection pool through `DB::Pool` like Invidious.

The most noticeable change is that connection timeout errors will no longer say `DB::PoolTimeout` but rather `Invidious::ConnectionPool::Error`. Full backtrace here.

<details>
<summary>Backtrace</summary>

``````
Title: `Could not check out a connection in 5.0 seconds (Invidious::ConnectionPool::Error)`
Date: `2024-11-12T00:15:28Z`
Route: `/search?q=test`
Version: `2024.11.11-c026a7b1 @ connection-pool-improvements`

<details>
<summary>Backtrace</summary>
<p>
   
```
Could not check out a connection in 5.0 seconds (Invidious::ConnectionPool::Error)
  from src/invidious/connection/pool.cr:45:7 in '_post_json'
  from src/invidious/yt_backend/youtube_api.cr:569:12 in 'search'
  from src/invidious/search/processors.cr:10:7 in 'regular'
  from src/invidious/search/query.cr:127:9 in 'process'
  from src/invidious/routes/search.cr:61:17 in 'search'
  from src/invidious/routing.cr:184:20 in '->'
  from lib/kemal/src/kemal/route.cr:12:26 in '->'
  from src/invidious/helpers/handlers.cr:30:37 in 'process_request'
  from lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:21:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:157:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:94:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:146:5 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:70:5 in 'call'
[...]

PR #5235, opened 2025-04-10
Rework `HTTP::Client` logic

### LINKED ISSUE #4662: [Enhancement] Don't override stdlib for `force_resolve` implementation

The current `force_resolve` implementation relies on patching the `HTTP::Client` in stdlib. It really isn't good practice.

I've recently been made aware that it is possible to force the `HTTP::Client` to resolve a specific IP family without overriding the stdlib implementation.

See https://github.com/crystal-lang/crystal/issues/14196#issuecomment-1884003549

The downside is that we'll lose the automatic reconnect feature.

Keep it as-is.

**Additional context**

### LINKED ISSUE #5230: [Bug] COMPANION_POOL is fixed to one invidious-companion instance.
**Describe the bug**
Invidious COMPANION_POOL is fixed to only one invidious-companion instance at start. This makes request to the `/player` endpoint of invidious-companion go to only one instance of all the ones that could be listed inside the `CONFIG.invidious_companion` array. One would expect that the COMPANION_POOL randomizes between each invidious-companion instance as Invidious does with the `/watch` and `/embed` endpoints to prevent getting blocked from youtube after too many requests.

the COMPANION_POOL will only start switching to another invidious-companion instance if the connection to invidious-companion does not succeed (this doesn't include HTTP error codes)

**Steps to Reproduce**

- Add at least 2 invidious-companion instances to the `invidious-companion` array on `config.yml`
- Log requests made to invidious-companion or intercept all requests from Invidious using a mitm proxy like [mitmproxy](https://docs.mitmproxy.org/) for easy debugging
- Reduce `pool_size` config option to something like 3 for easy debugging
- If the logs/proxying are ready to be able to see the requests if Invidious, start clicking on random videos and see from which invidious-companion instance Invidious is requesting the video.

**Screenshots**

Image

As it can be seen here, the first set of requests were made to an invidious-companion instance running on 127.0.0.1 which is the one that the COMPANION_POOL randomly selected when starting Invidious, then I forced invidious-companion to return a 500 error code for testing but that doesn't make the COMPANION_POOL to switch to another invidious-companion instance, so it will show endless errors on the frontend at the time of requesting a video.

Then I killed the instance running on 127.0.0.1 and there is when the COMPANION_POOL jumped into this rescue block:

https://github.com/iv-org/invidious/blob/23ff6135bb4304c6b5e9de6fc06e67bfadc9651a/src/invidious/yt_backend/connection_pool.cr#L71-L78

finally changing the invidious-companion instance to another one from the `CONFIG.invidious_companion` array.

**Additional context**

Found this when I found something weird about the requests being made to the different invidious-companion instances with my Invidious fork, but it also happens here ;3 (upstream invidious) 

Reference that is probably going to be useless since I use another approach to balance between different invidious-companion instances: https://git.nadeko.net/Fijxu/invidious/commit/04e5e69d4800bc1fdc01169bca7d8807bb505eba

syeopite:
**Depends on #5081**

I'm making this pull request target a branch that has the same commit history as #5081 to make it easier to review 

----

This PR replaces the current force resolve implementation which patches stdlib to one that is based around a custom IO.

In addition to accommodate the lack of the automatic reconnect feature the connection pool now uses `DB::Pool#retry` 

`DB::Pool#retry` allows us to cycle until we find a client that that still has an open connection, or to create a new one if necessary. The poisoned clients are then removed once identified by `#retry`'s internal logic

The fact that clients with closed connections will now be slowly removed also means that Invidious will no longer use the same pattern of companion connections within the pool; distributing routes more evenly.

Closes #5230
Closes #4662


unixfox:
Looking forward to this PR because I wanted to use multiple invidious companions.

PR #5245, opened 2025-04-19
Update images in README

efb4f5ff-1298-471a-8973-3d47447115dc:
I noticed that the images are hard to look at on desktop because they are presented in multiple columns (on mobile its even worse). Someone that is skimming through the README probably doesnt want to open allot of tabs to look at the images in more detail. 

I split the big table up into three separate ones so that its easier to see the image contents on both desktop and mobile

See how it looks like here: https://github.com/efb4f5ff-1298-471a-8973-3d47447115dc/invidious/tree/patch-1?tab=readme-ov-file#screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,20 @@
 
 
 ## Screenshots
+| Player                              |
+|-------------------------------------|
+| ![](screenshots/01_player.png)      |
+| ![](screenshots/04_description.png) |
 
-| Player                              | Preferences                         | Subscriptions                         |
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
They're supposed to be small, they're basically thumbnails that you have to click to open the full version.

Your change makes it so the images take FAR too much space and is absolutely not desirable.

efb4f5ff-1298-471a-8973-3d47447115dc:
Got it. Sorry for the inconvenience.

PR #5247, opened 2025-04-23
fix #3933

### LINKED ISSUE #3933: [Bug] Redirect from short youtu.be URLs does not work

TeamMasse:
I changed the order of checking for youtu.be short links and channel links. When doing the channel checks for a short links it would previously link to / because response.body was empty. Now it just does these checks later, so that both channel and short links work.


syeopite (changes requested):
The channel redirects should have higher priority than the video link redirects. 

Please perform the checks here first and only to process a video id if it fails to obtain a channel UCID.

```crystal
      html = XML.parse_html(response.body)
      ucid = html.xpath_node(%q(//link[@rel="canonical"])).try &.["href"].split("/")[-1]

      if ucid
        env.response.headers["Location"] = "/channel/#{ucid}"
        haltf env, status_code: 302
      end
```

TeamMasse:
I fixed it now by adding a cookie, that prevents the cookie banner from appearing. That was the underlying problem, because  without deciding on a cookie policy it won't let you proceed.

src/invidious/routes/errors.cr
```diff
@@ -4,15 +4,17 @@ module Invidious::Routes::ErrorRoutes
...
+    
     if md = env.request.path.match(/^\/(?<id>([a-zA-Z0-9_-]{11})|(\w+))$/)
       item = md["id"]
 
       # Check if item is branding URL e.g. https://youtube.com/gaming
-      response = YT_POOL.client &.get("/#{item}")
-
+      headers = HTTP::Headers{
+        "Cookie" => "SOCS=CAE" # Cookies to prevent redirects to Cookie Consent Page CAE~Reject all, CAA~showing the cookie banner, CAI~Accept all
+      }
```
SamantazFox: A small change for legibility:
```suggestion
      # Cookie to prevent redirects to the cookie consent page.
      # Cookie values: CAE = Reject all, CAA = show the cookie banner, CAI = Accept all
      headers = HTTP::Headers{"Cookie" => "SOCS=CAE"}

```

src/invidious/routes/errors.cr
```diff
@@ -4,15 +4,17 @@ module Invidious::Routes::ErrorRoutes
     if HOST_URL.empty? && env.request.path.starts_with?("/v1/storyboards/sb")
       return env.redirect "#{env.request.path[15..]}?#{env.params.query}"
     end
-
+
```
SamantazFox: Please revert this change!

src/invidious/routes/errors.cr
```diff
@@ -49,4 +51,4 @@ module Invidious::Routes::ErrorRoutes
     env.response.headers["Location"] = "/"
     haltf env, status_code: 302
   end
-end
+end
```
SamantazFox: Please also revert this change. If your text editor doesn't want to, you might need to change a config option or two.

SamantazFox (changes requested):

src/invidious/routes/errors.cr
```diff
@@ -4,15 +4,18 @@ module Invidious::Routes::ErrorRoutes
     if HOST_URL.empty? && env.request.path.starts_with?("/v1/storyboards/sb")
       return env.redirect "#{env.request.path[15..]}?#{env.params.query}"
     end
-
+
```
TeamMasse: spaces

src/invidious/routes/errors.cr
```diff
@@ -49,4 +52,4 @@ module Invidious::Routes::ErrorRoutes
     env.response.headers["Location"] = "/"
     haltf env, status_code: 302
   end
-end
+end
```
TeamMasse: add back the end

SamantazFox (approved):

Issue #5142, opened 2025-01-12
[Bug] Proxy videos preference is ignored when `default_user_preferences.local` is set to `true` and user is logged in into Invidious.


Fijxu:
**Describe the bug**

When the user is logged in on Invidious and `default_user_preferences.local` is set to `true` on the config, the `Proxy videos` preference on Invidious is ignored, leaving users that want to use HD720 without proxying, unable to do it due to CSP. This only happens to logged in users. Anonymous users are able to change that preference regardless of the value of default_user_preferences.local

**Steps to Reproduce**

1. Set `default_user_preferences.local` to `true` on the config
2. Log in into Invidious (Make sure the cookie `PREFS` is not set on the browser!)
3. Disable `Proxy Videos` on Preferences
4. Set `Preferred video quality:` to `HD720`
5. Watch any video.

PR #5275, opened 2025-05-08
fix: set CSP header after setting preferences of registered users

Fijxu:
Fixes https://github.com/iv-org/invidious/issues/5142

### DIFF
--- a/src/invidious/routes/before_all.cr
+++ b/src/invidious/routes/before_all.cr
@@ -20,14 +20,6 @@ module Invidious::Routes::BeforeAll
     env.response.headers["X-XSS-Protection"] = "1; mode=block"
     env.response.headers["X-Content-Type-Options"] = "nosniff"
 
-    # Allow media resources to be loaded from google servers
-    # TODO: check if *.youtube.com can be removed
-    if CONFIG.disabled?("local") || !preferences.local
-      extra_media_csp = " https://*.googlevideo.com:443 https://*.youtube.com:443"
-    else
-      extra_media_csp = ""
-    end
-
     # Only allow the pages at /embed/* to be embedded
     if env.request.resource.starts_with?("/embed")
       frame_ancestors = "'self' file: http: https:"
@@ -45,7 +37,7 @@ module Invidious::Routes::BeforeAll
       "font-src 'self' data:",
       "connect-src 'self'",
       "manifest-src 'self'",
-      "media-src 'self' blob:" + extra_media_csp,
+      "media-src 'self' blob:",
       "child-src 'self' blob:",
       "frame-src 'self'",
       "frame-ancestors " + frame_ancestors,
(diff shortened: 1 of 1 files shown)


src/invidious/routes/before_all.cr
unixfox, line 1: Maybe add a comment to reference the fact that we want to load the CSP after the logic for user preferences?

unixfox (changes requested):

src/invidious/routes/before_all.cr
Fijxu: Code comment or commit comment?

unixfox: Code comment

Fijxu: If something is not clear, please tell me

syeopite (commented):
The problem with setting the CSP header later is that any error pages that gets rendered prior to it won't have any CSP headers. For example this code here a couple lines above:

https://github.com/iv-org/invidious/blob/6271c050df7abe768ad6f5847038bba79784a5df/src/invidious/routes/before_all.cr#L47-L49

Fijxu:
> The problem with setting the CSP header later is that any error pages that gets rendered prior to it won't have any CSP headers. For example the error that can be raised when loading user data a couple lines above:
> 
> https://github.com/iv-org/invidious/blob/6271c050df7abe768ad6f5847038bba79784a5df/src/invidious/routes/before_all.cr#L47-L49

Thanks, recent commit should be better, CSP header ends with something like this when watching a video:

```text
default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:; connect-src http://inv.nadeko.cl:8282/ 'self'; manifest-src 'self'; media-src http://inv.nadeko.cl:8282/ https://*.googlevideo.com:443 https://*.youtube.com:443 'self' blob:; child-src 'self' blob:; frame-src 'self'; frame-ancestors 'none'
```

src/invidious/routes/before_all.cr
```diff
@@ -110,6 +102,21 @@ module Invidious::Routes::BeforeAll
...
+    # `!preferences.local` has to be checked after setting and
+    # reading `preferences` from the "PREFS" cookie and
+    # saved user preferences from the database, otherwise
+    # `https://*.googlevideo.com:443 https://*.youtube.com:443`
+    # will not be set in the CSP header if
+    # `default_user_preferences.local` is set to true on the
+    # configuration file, causing preference “Proxy Videos”
+    # not to work while having it disabled and using medium quality.
+    if CONFIG.disabled?("local") || !preferences.local
+      env.response.headers["Content-Security-Policy"] = env.response.headers["Content-Security-Policy"].gsub("media-src", "media-src https://*.googlevideo.com:443 https://*.youtube.com:443")
```
syeopite, line 117:
```suggestion
      env.response.headers.update("Content-Security-Policy", &.gsub("media-src", "media-src https://*.googlevideo.com:443 https://*.youtube.com:443"))
```

syeopite (approved):

Issue #5248, opened 2025-04-25
[Bug] get_referer function strips `+` from referer URL.


Fijxu:
**Describe the bug**

The `get_referer` strips `+` characters from the referer URL

**Steps to Reproduce**

1. Search for anything with an space on Invidious, for example: `meow meow`, `some search` 
2. Go to the settings page (to call the `get_referer` function later) (Example: https://yewtu.be/preferences?referer=%2Fsearch%3Fq%3Dmeow%2Bmeow )
3. Save the settngs
4. You will get redirected to the URL but with `+` characters stripped from the query parameters, ending with a search that says `meowmeow` or `somesearch`, both with their SPACE character stripped

**Screenshots**

![Image](https://github.com/user-attachments/assets/e341a860-4d0e-4e90-8b5d-c972afec0fba)

**Additional Context**

Someone sent me a patch for this bug. I'll try it locally and if it works well, I'll do a PR to fix this.

Issue #2012, opened 2021-04-12
[Bug] Spaces in search queries gets removed when returning from preferences page


syeopite:
**Describe the bug**
The spaces in the search query are removed when returning from preferences page.

**Steps to Reproduce**

1. Search anything up on Invidious
2. Enter preferences on the search results page
3. Click "Save preferences" to return
4. You'll see that the spaces are removed in the search query

**Logs**

**Screenshots**

Before:
![before](https://user-images.githubusercontent.com/70992037/114344464-67bb9380-9b4f-11eb-831c-34ce382a958b.png)

After:
![after](https://user-images.githubusercontent.com/70992037/114344469-6a1ded80-9b4f-11eb-83e2-ebbda25c38bd.png)

**Additional context**

PR #5276, opened 2025-05-08
fix: do not strip '+' character from referer

Fijxu:
Fix that a user of my instance (https://inv.nadeko.net) sent me by email.

Fixes https://github.com/iv-org/invidious/issues/5248
Fixes https://github.com/iv-org/invidious/issues/2012


unixfox (approved):
Awesome thank you :)

syeopite (approved):

PR #5297, opened 2025-05-11
Add Javascript licence information automatically

syeopite:
This commit automates the process of documenting the licenses of Invidious Javascript files through a compile time macro in the licenses.ecr template file.

The script it calls is much smaller and does not depend on external IO so Invidious' compile-time is not impacted significantly. 

This should hopefully help keep the licence documentation up-to-date and allow extensions like LibreJS to always be able to load the latest Javascript files of Invidious.

Currently only Invidious's first-party Javascript files are supported. In the future it should be possible to leverage videojs-dependencies.yml to automatically document the Javascript licenses for VideoJS and co. as well.

This should be merged after #5292

Screenshot of Invidious' /licenses page with this PR</img>

### DIFF
--- a/src/invidious/views/licenses.ecr
+++ b/src/invidious/views/licenses.ecr
@@ -11,90 +11,6 @@
     <table id="jslicense-labels1">
         <tr>
             <td>
-                <a href="/js/_helpers.js?v=<%= ASSET_COMMIT %>">_helpers.js</a>
-            </td>
-
-            <td>
-                <a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL-3.0</a>
-            </td>
-
-            <td>
-                <a href="/js/_helpers.js?v=<%= ASSET_COMMIT %>"><%= translate(locale, "source") %></a>
-            </td>
-        </tr>
-
-        <tr>
-            <td>
-                <a href="/js/handlers.js?v=<%= ASSET_COMMIT %>">handlers.js</a>
-            </td>
-
-            <td>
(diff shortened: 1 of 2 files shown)


Fijxu (approved):
I was just thinking about this after submitting my PR.

src/invidious/views/licenses.ecr
```diff
@@ -289,19 +177,9 @@
...
-                <a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL-3.0</a>
-            </td>
-
-            <td>
-                <a href="/js/watch.js?v=<%= ASSET_COMMIT %>"><%= translate(locale, "source") %></a>
-            </td>
-        </tr>
+        <%- {% for row in run("../../../scripts/generate_js_licenses.cr").stringify.split('\n') %} %>
+            <%-= {{row.id}} -%>
+        <% {% end %} -%>
```
alexmaras, line 182:
```suggestion
        <%-= {{ run("../../../scripts/generate_js_licenses.cr").stringify }} %>
```

alexmaras (commented):
This works in its current form, but the quoting around each `<tr>` block isn't really necessary. The suggestions here just make it so that you don't have to do that quote-wrapping, and so don't need to escape the quotes inside the here doc.

syeopite:
Unfortunately this approach will break the string interpolation needed for displaying the translated "source" string at runtime

alexmaras:
Ah, well caught. That's a pity.

PR #5302, opened 2025-05-14
CI: Use GitHub actions cache in docker builds

Fijxu:
We already make use of the mount cache on https://github.com/iv-org/invidious/commit/e2df12b7d66618c21cffc1f75cefeea083f3dbc7 to speed up the compiling on Invidious, but we can also use Github actions for the Dockerfile, which caches packages installed using `apk` and any other layer of the container that can be cached.

For installed packages, it will always use the same cached version (if the cache has not been used on 7 days according to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy), so on each build, the packages are not going to be updated (https://docs.docker.com/build/cache/invalidation/#run-instructions). I personally think this is better to prevent a possible breakage on an updated package (which is probably not going to happen anyways because we use a specific alpine tag).

### DIFF
--- a/.github/workflows/build-nightly-container.yml
+++ b/.github/workflows/build-nightly-container.yml
@@ -60,6 +60,8 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           build-args: |
             "release=1"
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
       - name: Docker meta
         id: meta-arm64
@@ -85,3 +87,5 @@ jobs:
           tags: ${{ steps.meta-arm64.outputs.tags }}
           build-args: |
             "release=1"
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/build-stable-container.yml
+++ b/.github/workflows/build-stable-container.yml
@@ -53,6 +53,8 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           build-args: |
             "release=1"
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
       - name: Docker meta
         id: meta-arm64
@@ -79,3 +81,5 @@ jobs:
           tags: ${{ steps.meta-arm64.outputs.tags }}
           build-args: |
             "release=1"
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,8 @@ jobs:
           file: docker/Dockerfile.arm64
           platforms: linux/arm64/v8
           build-args: release=0
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
       - name: Test Docker
         run: while curl -Isf http://localhost:3000; do sleep 1; done


Fijxu:
Seems to work fine: https://github.com/iv-org/invidious/actions/runs/15012807593/job/42184935828?pr=5302#step:5:189.

 I may open another PR to improve the speed of https://github.com/iv-org/invidious/actions/runs/15012807593/workflow?pr=5302#L91 since it needs a different approach to be able to cache the layers, possible solution: https://stackoverflow.com/a/75544124

Adding cache to the `build-docker` job will likely save like just 5 seconds, but better than nothing I guess.

unixfox:
> so on each build, the packages are not going to be updated

That's an issue if you need to quickly patch a vulnerability in a package or dependency.

syeopite (commented):
> That's an issue if you need to quickly patch a vulnerability in a package or dependency.

Yeah this is a problem. Though Crystal and Invidious also doesn't have too many system dependencies so I'm not sure how large of a problem it'd really be.

The container for `x86-64` already builds in ~ 5 minutes and 30 seconds which is imo is more than good enough for an automated image build. Any improvements to that is pretty negligible. 

Instead I think it'll make more sense to optimize the arm64 build process (mainly by switching away from QEMU to the native arm runner on Github Actions) which takes almost 30 minutes to complete.

EDIT: Accidentally used the duration for building the test images during CI instead off the actual images for nightly/stable builds

Fijxu:
>Though Crystal and Invidious also doesn't have too many system dependencies so I'm not sure how large of a problem it'd really be.

>That's an issue if you need to quickly patch a vulnerability in a package or dependency.

Yeah, getting a vulnerability with the low amount of packages Invidious uses is pretty unlikely, but there is a still a possibility, some day, a vulnerability on some package may be released and abused, and if we wanted to release a fast fix with the package already patched on Alpine Linux repositories, we would need to delete the caches on https://github.com/iv-org/invidious/actions/caches?query=sort%3Acreated-desc manually before letting GH Actions build the Invidious container image, which doesn't seem practical at all.

I was searching about how to only ignore cache for the `RUN` command, but that is not possible. What it's possible is to prevent cache of a stage (https://github.com/moby/moby/issues/1996#issuecomment-1763467972) but that will just complicate the `Dockerfile`.

>Instead I think it'll make more sense to optimize the arm64 build process (mainly by switching away from QEMU to the native arm runner on Github Actions) which takes almost 10 minutes to complete.

I'll go with this in a new PR instead.

syeopite:
Actually caching the videojs dependencies could help reduce build times significantly and without any potential security issues either since there's a version file to easily burst the cache.

A cacheless (compiler-wise) cold build of `fetch-player-dependencies.cr` takes ages to complete and even longer with `--release`

If VideoJS dependencies could be cached and  the `-Dskip_videojs_download` flag passed to the compiler when cached dependencies are available then there could be a noticeable drop in build times.

Fijxu:
>Actually caching the videojs dependencies could help reduce build times significantly and without any potential security issues either since there's a version file to easily burst the cache.

I'll get my hands dirty on that on another PR ;)
Thanks for the suggestion

PR #5305, opened 2025-05-15
CI: Use public ARM64 Github actions runners for ARM64 builds.

Fijxu:
Currently, Invidious uses QEMU to build it's ARM64 Invidious image, which is slow (since we are basically using a virtual machine).

This helps with the speed of building ARM64 binaries for Invidious on each release/commit by a lot.

More information about the public ARM64 runners here: https://github.com/orgs/community/discussions/148648

---

Take a look at: https://github.com/Fijxu/invidious/actions/runs/15037532092 and https://github.com/Fijxu/invidious/actions/runs/15037532093

Testing OCI images at https://git.nadeko.net/Fijxu/-/packages/container/invidious_do_not_pull/master-arm64 and https://git.nadeko.net/Fijxu/-/packages/container/invidious_do_not_pull/master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,26 +100,16 @@ jobs:
 
   build-docker-arm64:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
 
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
-        with:
-          platforms: arm64
+      - name: Build Docker
+        run: docker compose -f docker-compose-arm64.yml build --build-arg release=0
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-
-      - name: Build Docker ARM64 image
-        uses: docker/build-push-action@v6
-        with:
-          context: .
-          file: docker/Dockerfile.arm64
-          platforms: linux/arm64/v8
-          build-args: release=0
+      - name: Run Docker
+        run: docker compose -f docker-compose-arm64.yml up -d
 
       - name: Test Docker
         run: while curl -Isf http://localhost:3000; do sleep 1; done
--- a/docker-compose-arm64.yml
+++ b/docker-compose-arm64.yml
@@ -1,0 +1,55 @@
+# Warning: This docker-compose file is made for development purposes.
+# Using it will build an image from the locally cloned repository.
+#
+# If you want to use Invidious in production, see the docker-compose.yml file provided
+# in the installation documentation: https://docs.invidious.io/installation/
+
+version: "3"
+services:
+
+  invidious:
+    build:
+      context: .
+      dockerfile: docker/Dockerfile.arm64
+    restart: unless-stopped
+    ports:
+      - "127.0.0.1:3000:3000"
+    environment:
+      # Please read the following file for a comprehensive list of all available
+      # configuration options and their associated syntax:
+      # https://github.com/iv-org/invidious/blob/master/config/config.example.yml
+      INVIDIOUS_CONFIG: |
+        db:
+          dbname: invidious
+          user: kemal
+          password: kemal
+          host: invidious-db
+          port: 5432
+        check_tables: true
+        # external_port:
+        # domain:
+        # https_only: false
+        # statistics_enabled: false
+        hmac_key: "CHANGE_ME!!"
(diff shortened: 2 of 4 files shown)


Fijxu:
If possible, test the ARM64 OCI Image on a real ARM64 machine. I don't have access to one (well, my phone may work) and I would like to know if it works fine just as the QEMU built one.

.github/workflows/ci.yml
```diff
@@ -100,26 +100,16 @@ jobs:
...
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
-        with:
-          platforms: arm64
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+      - name: Build Docker
+        run: docker compose -f docker-compose-arm64.yml build --build-arg release=0
```
syeopite, line 109: I think we can get rid of the `--build-arg` for both the arm64 and amd64 build since the Dockerfiles only compile Invidious in release mode when `release` is preset and is set as 1

```suggestion
        run: docker compose -f docker-compose-arm64.yml build
```

docker-compose-arm64.yml
```diff
@@ -0,0 +1,55 @@
...
+          user: kemal
+          password: kemal
+          host: invidious-db
+          port: 5432
+        check_tables: true
+        # external_port:
+        # domain:
+        # https_only: false
+        # statistics_enabled: false
+        hmac_key: "CHANGE_ME!!"
```
syeopite, line 33: Looks like our CI is actually bugged... in both the arm64 and amd64 test builds the unchanged `hmac_key` line should've caused the `Run Docker` step to fail and exit as soon as Invidious loads the config during start-up.

And the `Test Docker` step should also not be passing...

syeopite (approved):

docker-compose-arm64.yml
Fijxu: Yeah, it should fail. I noticed that yesterday when executing `while curl -Isf http://localhost:3000; do sleep 1; done`. It always returns status code 0.

What about using `docker compose up` in the `Run Docker` step (to get debugging output) and `curl -If http://localhost:3000` on the `Test Docker` step?

Fijxu: I'll do this on another PR to keep the commits organized. The build time of `build-docker` and `build-docker-arm64` will take a bit more, but it will actually check if Invidious works with the current `docker-compose.yml` file.

`build-docker`: https://github.com/Fijxu/invidious/actions/runs/15053716209/job/42314461096
`build-docker-arm64`: https://github.com/Fijxu/invidious/actions/runs/15053716209/job/42314461089

unixfox (approved):

syeopite (approved):
Perfect. Looks good to me!

Fijxu:
It still bothers me a little to have two `docker-compose` files, one for AMD64 and ARM64. For sure we can just use `docker buildx build -t invidious_docker_build_image` instead of `docker compose -f composefile.yml build` and replace 

https://github.com/iv-org/invidious/blob/03f89be9291b70ea7234e4291348e80b6c2c5d4e/docker-compose.yml#L11-L13

To:

```
    image: invidious_docker_build_image
```

But I'm not sure :/, what do you think @syeopite

syeopite:
Considering there is only a single line difference between the two compose files I think the simplest solution is to literally just sed `dockerfile: docker/Dockerfile` to `dockerfile: docker/Dockerfile.arm64` and then add this step:

```yml
- name: Use ARM64 Dockerfile if ARM64
  if: ${{name}} == "ARM64"
  run: sed -i 's/Dockerfile/Dockerfile.arm64/' docker-compose.yml
```

Fijxu:
Oh right... Dumb me.

unixfox:
Would be interesting to merge both amd64 and arm64 into a single docker image tag. That's what @inetol did here at SearXNG: https://github.com/searxng/searxng/pull/4764

Maybe he can explain to us how to do :)?

Fijxu:
> Would be interesting to merge both amd64 and arm64 into a single docker image tag. That's what @inetol did here at SearXNG with two separate Dockerfile: [searxng/searxng#4764](https://github.com/searxng/searxng/pull/4764)
> 
> Maybe he can explain to us how to do :)?

Oh that should be easy (at least from what I know). But I would prefer to do that in another PR

SamantazFox (approved):

PR #5306, opened 2025-05-15
CI: Fix build-docker job not checking if Invidious starts successfully or not

Fijxu:
Depends on https://github.com/iv-org/invidious/pull/5305

As discussed on https://github.com/iv-org/invidious/pull/5305#discussion_r2090590763, we noticed that the `build-docker` job doesn't actually checks if Invidious works or not, because curl always exited with code `0` (success) and the current docker-compose.yml files has defaults that will make Invidious fail at start (`hmac_key` and missing Invidious companion configuration).


docker-compose-arm64.yml
```diff
@@ -30,7 +34,12 @@ services:
         # domain:
         # https_only: false
         # statistics_enabled: false
-        hmac_key: "CHANGE_ME!!"
+        hmac_key: "DEVELOPMENT_KEY__DO_NOT_USE"
+        invidious_companion_key: "DEVELOPMENT__KEY"
```
syeopite, line 38:
```suggestion
        hmac_key: "DEVELOPMENT_KEY__DO_NOT_USE"
        invidious_companion_key: "DEVELOPMENT__KEY"
```

Considering the amount of people that just blindly runs the compose file from the repo I'll rather not make their Invidious instances less secure than we have to. Since at least for `CHANGE_ME!!` Invidious will exit and force a different value.

Can we just have the Github actions runner `sed` that particular line to something else?

docker-compose-arm64.yml
```diff
@@ -30,7 +34,12 @@ services:
...
+        # Bogus companion to let Invidious start
+        invidious_companion:
```
syeopite, line 40: `invidious_companion` and `invidious_companion_key` shouldn't be required for Invidious to start

.github/workflows/ci.yml
```diff
@@ -104,7 +104,12 @@ jobs:
         run: docker compose up -d
 
       - name: Test Docker
-        run: while curl -Isf http://localhost:3000; do sleep 1; done
+        id: test
```
syeopite, line 107:
```suggestion
```

syeopite (changes requested):

.github/workflows/ci.yml
Fijxu: Required for https://github.com/iv-org/invidious/pull/5306/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR111 (Line 111)

syeopite (approved):
Looks good to me! I just have a single documentation request left

unixfox (approved):

syeopite:
Should've been merged with https://github.com/iv-org/invidious/commit/80ec027c8ff22ca3b8add1eba9c32c656fcb9927

I don't know why Github's UI isn't marking this as merged

Issue #5295, opened 2025-05-10
[Bug] Invidious is unable to generate text captchas | textcaptcha.com is down


Degenticrypto:
**Describe the bug**

Wrong time on clock, reported bug when i tried to update it with correct time

**Steps to Reproduce**

**Logs**

**Screenshots**

**Additional context**

syeopite:
I can reproduce this. 

This issue is caused by an outage on the text captcha service that Invidious uses. You may have misclicked on the "Text CAPTCHA" toggle instead of the "register" button when inputting the time for the clock captcha.


Title: `Unexpected char '<' at line 2, column 1 (JSON::ParseException)`
Date: `2025-05-11T10:53:53Z`
Route: `/login?referer=%2Fsearch%3Fq%3Dhello&type=invidious`
Version: `2025.05.11-1705e2e7 @ crystal-perf-tools`

<details>
<summary>Backtrace</summary>
<p>
   
```
Unexpected char '<' at line 2, column 1 (JSON::ParseException)
  from /usr/lib/crystal/json/lexer.cr:331:5 in 'raise'
  from /usr/lib/crystal/json/lexer.cr:327:5 in 'unexpected_char'
  from /usr/lib/crystal/json/lexer.cr:326:11 in 'unexpected_char'
  from /usr/lib/crystal/json/lexer.cr:243:10 in 'consume_number'
  from /usr/lib/crystal/json/lexer.cr:63:7 in 'next_token'
  from /usr/lib/crystal/json/parser.cr:108:11 in 'next_token'
  from /usr/lib/crystal/json/parser.cr:9:5 in 'initialize'
  from /usr/lib/crystal/json/parser.cr:6:3 in 'new'
  from /usr/lib/crystal/json.cr:137:5 in 'parse'
  from src/invidious/user/captcha.cr:66:18 in 'generate_text'
  from src/invidious/routes/login.cr:106:25 in 'login'
  from src/invidious/routing.cr:59:20 in '->'
  from lib/kemal/src/kemal/route.cr:12:26 in '->'
  from src/invidious/helpers/handlers.cr:31:37 in 'process_request'
  from lib/kemal/src/kemal/route_handler.cr:17:7 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:30:7 in 'call_next'
  from lib/kemal/src/kemal/websocket_handler.cr:13:14 in 'call'
  from /usr/lib/crystal/http/server/handler.cr:30:7 in 'call_next'
[...]

PR #5308, opened 2025-05-17
Remove text captcha due to textcaptcha.com being down

Fijxu:
Fixes https://github.com/iv-org/invidious/issues/5295

textcaptcha.com seems to be down since April and it does not appear that service will be restored.

Text captchas can be easily automated using free LLMs, so keeping the text captcha is more like a gate to create accounts in mass on public Invidious instances.

It also gives headaches like bots automating account creation to modify the videos that appear on the popular page of each public instance (since the popular page is based on the subscriptions of the registered users).

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -64,8 +64,6 @@
     "User ID": "User ID",
     "Password": "Password",
     "Time (h:mm:ss):": "Time (h:mm:ss):",
-    "Text CAPTCHA": "Text CAPTCHA",
-    "Image CAPTCHA": "Image CAPTCHA",
     "Sign In": "Sign In",
     "Register": "Register",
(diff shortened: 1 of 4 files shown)


unixfox (approved):

syeopite (approved):
Please also remove the `Text CAPTCHA` and `Image CAPTCHA` keys from `locales/en-US.json`. Weblate should propagate the changes automatically.

Fijxu:
> Please also remove the `Text CAPTCHA` and `Image CAPTCHA` keys from `locales/en-US.json`. Weblate should propagate the changes automatically.

I was going to ask that but I forgot. I'll do, if we need to add text captcha again for whatever reason we can just revert this PR I guess.

PR #5311, opened 2025-05-17
Some minor optimizations for companionless stream retrieval

syeopite:


src/invidious/videos/parser.cr
```diff
@@ -111,12 +111,11 @@ def extract_video_info(video_id : String)
   if !CONFIG.invidious_companion.present?
     if player_response["streamingData"]? && player_response.dig?("streamingData", "adaptiveFormats", 0, "url").nil?
```
syeopite, line 112: This could probably be reduced as well

```suggestion
    if player_response.dig?("streamingData", "adaptiveFormats", 0, "url").nil?
```

Fijxu: Works fine, and it's faster ;) :

```
? &&   3.42M (292.29ns) (± 1.08%)  0.0B/op   1.28× slower
dig?     4.37M (229.07ns) (± 1.05%)  0.0B/op        fastest
```

Fijxu (approved):

syeopite:
Superceded by #5312

PR #5314, opened 2025-05-18
Remove `@iv-org/developers` from codeowners

syeopite:
I feel like the random team review requests are more just noise rather than anything actually useful. I personally don't see much of a point in keeping it around. Let me know what you guys think


Fijxu:
Yeah, I prefer to manually select the reviewers instead of letting github picking a random one from the list.

unixfox (approved):

Fijxu (approved):

PR #5315, opened 2025-05-18
Fix local storyboard requests on Crystal >= 1.16.0

syeopite:
Crystal 1.16.0 fixed the parsing of some weird-looking paths meaning that `//api/v1/storyboards/sb/i/...` will no longer get parsed into a request target of `/v1/storyboards/sb/i/...` within requests.

But the fix is a bit more annoying than just changing a path. The "new" existence  of the double slash means that `StaticFileHandler` will process and redirect the request to a normalized path with only a single slash. This redirect however does not preserve url query parameters which are needed to request storyboards from YouTube.

Then there's the fact that the status code error handlers are disabled for the API routes, meaning that there is no handler to actually process the request with.

This PR patches Kemal's `StaticFileHandler` to preserve query parameters when redirecting, and adds the `/api/v1/storyboard/sb/*` -> `/sb/i/*/` redirect directly into Kemal's `ExceptionHandler` for now.

In the future a dedicated handler should be created to properly handle http status code conditions on the API routes.

### DIFF
--- a/src/ext/kemal_static_file_handler.cr
+++ b/src/ext/kemal_static_file_handler.cr
@@ -191,5 +191,13 @@ module Kemal
         end
       end
     end
+
+    # See https://github.com/crystal-lang/crystal/issues/15788
+    #
+    # URL fragments also aren't passed along but there isn't an easy way
+    # to retrieve that at the moment.
+    private def redirect_to(context, path)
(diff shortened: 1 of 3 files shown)


syeopite:
**Note:** extra care needs to be taken when merging alongside #5338 to also override the `redirect_to` method on the new static files handler.

PR #5332, opened 2025-05-31
Make caption text selectable

CyanEngineer:
I think it would be really nice to be able to select and copy the caption text. The caption container has `pointer-events` disabled as it would otherwise eat the mouse input for the whole player area. But re-enabling `pointer-events` for just the text container makes the text selectable without affecting the player outside of the area with text.


syeopite:
Tested and works as expected.

----

Though this isn't a problem right now I'm concerned that this feature could conflict with implementing freely movable captions like those seen on YouTube's official player. You can't really differentiate whether the user wants to move the captions box or select the text since it involves the same mouse movements over a very tiny space that overlaps.

PR #5333, opened 2025-06-01
Expose database connection pool settings

syeopite:
Unless an instance maintainer was using the `database_url` attribute, and knew about the connection pool query parameters, the database connection pool settings was relegated to the default settings...

And the default settings are not great for any large instance... With it essentially only allowing a single connection within the pool, a maximum checkout time of a 5 seconds, and basically no additional retries whatsoever its no wonder that PgBouncer has became an necessity among Invidious instances.

This PR changes that by exposing the ability to configure the database connection pool for Postgres. I've also added some saner defaults that should hopefully help Invidious run better out of the box. 

For reference here are the default values: https://crystal-lang.org/reference/1.16/database/connection_pool.html#configuration

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -15,6 +15,90 @@ db:
   port: 5432
   dbname: invidious
 
+
+  ## -----------------------------------
+  ## Database connection pool settings
+  ## -----------------------------------
+
+  ##
+  ## The maximum allowed number of connections within the connection pool.
+  ## When this number is reached and no connections are free, Invidious will
+  ## wait for up to `checkout_timeout` before raising an exception.
+  ##
+  ## Accepted values: a positive integer
+  ## Default: 100
+  ##
+  #max_size: 100
+
+  ##
+  ## How many idle connections should be allowed to exist within the connection pool.
+  ##
+  ## There is no concept of whether a connection is closed or open, or how long it has been
+  ## idly sitting by. Instead Idle connections are defined as any connections that are not
+  ## currently being used. When this number is exceeded but the amount of connections is still
+  ## under max_pool_size, new connections will be created on a checkout but immediately closed
+  ## and destroyed during an release operation.
+  ##
+  ## For the most part, this should just be set to the same number as the `max_pool_size`.
+  ##
+  ## Accepted values: a positive integer
+  ## Default: 100
+  ##
+  #max_idle_pool_size: 100
+
+  ##
+  ## The amount of connections to establish on start-up.
+  ##
+  ## When Invidious starts up, the database connection pool is immediately populated
+  ## with this many connections. This could allow an instance to avoid potentially experiencing
+  ## any degraded service during start-up; since a good amount of connections will be available
+  ## from the start rather than needing to be created on the fly as requests come in.
+  ##
+  ##
+  ## Accepted values: a positive integer
+  ## Default: 1
+  ##
+  #initial_pool_size  : 1
+
+  ##
+  ## How long to wait (in seconds) before timing out during a checkout
+  ##
+  ## Accepted values: a positive float
+  ## Default: 5.0
+  ##
+  #checkout_timeout: 5.0
+
+  ##
+  ## How many attempts to retry a connection
+  ##
+  ## This allows Invidious to gracefully handle network problems that can
+  ## cause a connection to fail or be unable to get established in the first
+  ## place.
+  ##
+  ## Note: The mechanisms of the underlying library will first try to reuse all the
+  ## currently available idle connections within the pool, as well as one additional attempt,
+  ## at no delay. Afterwards, it will began counting down the `retry_attempts` and wait
+  ## `retry_delay` seconds between each attempt.
+  ##
+  ## This heuristic ensures that unnecessary new connections are not being made for no reason, and
+  ## in the event of widespread network failures, will help to throw out dead connections from the
+  ## pool.
+  ##
+  ## Accepted values: a positive integer
+  ## Default: 5.0
+  ##
+  #retry_attempts: 5.0
+
+  ##
+  ## How long to wait (in-seconds) between each retry attempt
+  ##
+  ## Note: See description for `retry_attempts` for more information
+  ##
+  ## Accepted values: a positive float
+  ## Default: 1.0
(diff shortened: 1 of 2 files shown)


config/config.example.yml
```diff
@@ -15,6 +15,90 @@ db:
...
+  ## currently being used. When this number is exceeded but the amount of connections is still
+  ## under max_pool_size, new connections will be created on a checkout but immediately closed
+  ## and destroyed during an release operation.
+  ##
+  ## For the most part, this should just be set to the same number as the `max_pool_size`.
+  ##
+  ## Accepted values: a positive integer
+  ## Default: 100
+  ##
+  #max_idle_pool_size: 100
```
syeopite, line 50: I originally also exposed the idle pool size config for the HTTP connection pool in #5081 but later reverted the change after I learnt what idle actually meant in the context of the connection pool.

I've added it here for completeness but I don't know if there's actually any benefits in exposing this setting over making it just always equal the max pool size.

config/config.example.yml
```diff
@@ -15,6 +15,93 @@ db:
...
+  ## When this number is reached and no connections are free, Invidious will
+  ## wait for up to `checkout_timeout` before raising an exception.
+  ##
+  ## Note: Setting to 0 will disable this limit and allow the the pool to
+  ## grow indefinitely.
+  ##
+  ## Accepted values: a positive integer
+  ## Default: 100
+  ##
+  #max_pool_size: 100
```
syeopite, line 34: This is a probably a non issue but

Although all of this is nested under the db field, the options here are still really similar to the ones for the HTTP connection pool which might get confusing. And this could be made worse if we ever decide to offer more granular control for the latter (eg configuring the pool options for each domain) which will lead to a lot of essentially duplicate options.

src/invidious/config.cr
```diff
@@ -6,6 +6,52 @@ struct DBConfig
...
+  # How many connections to construct on start-up, and keep it there.
+  property initial_pool_size : Int32 = 1
+  # The maximum size of the connection pool
+  property max_pool_size : Int32 = 100
+  # The maximum amount of idle connections within the pool
+  # idle connections are defined as created connections that are
+  # sitting around in the pool. Exceeding this number will cause new connections
+  # to be created on checkout and then simply dropped on release, till the maximum pool size
+  # from which there will be a checkout timeout.
+  property max_idle_pool_size : Int32 = 100
```
Fijxu, line 19: This should note two things.

The value of `max_idle_pool_size` has to be higher than `initial_pool_size`, if `initial_pool_size` is `500`, Invidious will start with `500` connections, and then it will reduce them to `100` if `max_idle_pool_size` is set to `100`.

`max_connections` on PostgreSQL has to be slightly higher than the value used here, the default value of `max_connections` is `100` (https://postgresqlco.nf/doc/en/param/max_connections/), so if someone reason Invidious needs to create new clients on the pool, it will hit `100` connections on the DB, depleting all the remaining connections and preventing other services using the same PostgreSQL server to connect to it.

syeopite: I'm not sure about forcing a user to change their default postgres configuration just to accommodate Invidious' default behavior.

Maybe we could just reduce the default `max_pool_size` instead? I'm not sure what'll be a reasonable value though.

Fijxu: >Maybe we could just reduce the default max_pool_size instead? I'm not sure what'll be a reasonable value though.

That will depend of the traffic of the instance. I think 32 or even 16 is fine. My instance used to run using a unix socket without pgbouncer (but with ~4 replicas, so at the end, the total connections to the database were 24 (4 replicas * 6 connections)) at all and everything worked fine.

I tested using `max_pool_size` 32 and this are the results:

When using `wrk` with 2 threads and 10 connections, I got 1772req/s with `~5ms` average, against an Invidious playlist with 1 video using the Invidious API `/api/v1/playlists/:plid`. Invidious used 10 db connections.

When using 2 threads and 50 connections, I got 1826req/s with `123ms` average, Invidious used 32 connections.

What matters is the latency of each operation. I'm not a database expert, but I still think this setting will depend of the current setup of Invidious people have. 

People that use a shared database used along other services like me may need to set a higher `max_connections` on PostgreSQL or set a lower `max_pool_size` so Invidious doesn't eat all the connections for any reason (mainly DoS).

People that use the Docker guide of https://docs.invidious.io/installation/#docker-compose-method-production should be fine since `max_pool_size` matches with the default `max_connections` value of PostgreSQL, and that instance of PostgreSQL runs exclusively for Invidious.

People that use Docker replicas or some sort of replicas, will have to adjust the `max_pool_size` value to limit how many connections each replica can use, which before this PR, it used to be fixed to 6

SamantazFox:
I am strongly in favor of creating a dedicated documentation page for the `database_url` parameters, rather than adding a ton of extra config options that will confuse (and not be useful to) most users.

syeopite:
The pool configurations at least are definitely useful since without it all connections are just temporary one-off connections and that could be slow on even one-person instances.

SamantazFox:
> The pool configurations at least are definitely useful

I don't disagree with that! All I'm saying is that given they're already available as `database_url` query parameters, we should document them on a dedicated page at docs.invidious.io rather than add discrete config options that need to be documented in the example config.

> since without it all connections are just temporary one-off connections and that could be slow on even one-person instances.

Don't we already provide good defaults for that use case? Or have I missed something lately?

syeopite:
Invidious only configures the connection pool for **web** requests so the **database** connection pool just use the base values defined in `crystal-db`

And the base values limits the maximum idle connections to only a single one. This means that connections will be frequently created and then disposed off to ensure that only a single idling connection can remain in the pool.

For example let's say a database operation is requested while the single idle connection is checked out somewhere else. To handle this operation the pool will create a new connection to the database.

If one of these two connections are then released back into the pool, the release of the other one will see that there is already an idling connection within the pool, and then close itself because of it.

This is why Invidious spams the database with connections and why instances have to deploy PGBouncer. It can only reuse a single connection.

---

This PR finally does add sane defaults (for single-user instances) for the **database** connection pool. So do you mean that these sane defaults should only be changeable when the database is configured through `database_url`?

PR #5337, opened 2025-06-04
Refactor `FilteredCompressHandler` to inherit from stdlib

syeopite:
This changes its behavior to align with the stdlib variant in that compression is now delayed till the moment that the server begins to send a response.

This allows the handler to avoid compressing empty responses,and safeguards against any double compression of content that may occur if another handler decides to compressi ts response.

This does however come at the drawback(?) of it now removing `content-length` headers on requests if it exists; since compression makes the value inaccurate anyway.

See: https://github.com/crystal-lang/crystal/pull/9625

### DIFF
--- a/src/invidious/helpers/handlers.cr
+++ b/src/invidious/helpers/handlers.cr
@@ -61,28 +61,13 @@ class Kemal::ExceptionHandler
   end
 end
 
-class FilteredCompressHandler < Kemal::Handler
+class FilteredCompressHandler < HTTP::CompressHandler
   exclude ["/videoplayback", "/videoplayback/*", "/vi/*", "/sb/*", "/ggpht/*", "/api/v1/auth/notifications"]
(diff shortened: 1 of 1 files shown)


src/invidious/helpers/handlers.cr
```diff
@@ -61,28 +61,13 @@ class Kemal::ExceptionHandler
   end
 end
 
-class FilteredCompressHandler < Kemal::Handler
+class FilteredCompressHandler < HTTP::CompressHandler
   exclude ["/videoplayback", "/videoplayback/*", "/vi/*", "/sb/*", "/ggpht/*", "/api/v1/auth/notifications"]
   exclude ["/api/v1/auth/notifications", "/data_control"], "POST"
```
syeopite, line 66: Its pretty weird that we're excluding the notification streams endpoints and /data_control from having their responses compressed

PR #5338, opened 2025-06-04
Replace `Kemal::StaticFileHandler` with direct subclass of stdlib `HTTP::StaticFileHandler` on Crystal < 1.17.0

syeopite:
Fixes the CI for Crystal nightly

----

Kemal's subclass of the stdlib `HTTP::StaticFileHandler` is not as
maintained as its parent, and so misses out on many enhancements and bug
fixes from upstream, which unfortunately also includes the patches for
security vulnerabilities...

Though this isn't necessarily Kemal's fault since the bulk of the stdlib
handler's logic was done in a single big method, making any changes hard
to maintain. This was fixed in Crystal 1.17.0 where the handler
was refactored into many private methods, making it easier for an
inheriting type to implement custom behaviors while still leveraging
much of the pre-existing code.

Since we don't actually use any of the Kemal specific features added by
`Kemal::StaticFileHandler`, there really isn't a reason to not just
create a new handler based upon the stdlib implementation instead which
will address the problems mentioned above.

This PR implements a new handler which inherits from the stdlib variant
and overrides the helper methods added in Crystal 1.17.0 to add the
caching behavior with minimal code changes. Since this new handler
depends on the code in Crystal 1.17.0, it will only be applied on
versions greater than or equal to 1.17.0. On older versions we'll
fallback to the current monkey patched `Kemal::StaticFileHandler`

### DIFF
--- a/src/invidious/http_server/static_assets_handler.cr
+++ b/src/invidious/http_server/static_assets_handler.cr
@@ -1,0 +1,138 @@
+{% skip_file if compare_versions(Crystal::VERSION, "1.17.0") < 0 %}
+
+module Invidious::HttpServer
+  class StaticAssetsHandler < HTTP::StaticFileHandler
+    # In addition to storing the actual data of a file, it also implements the required
+    # getters needed for the object to imitate a `File::Stat` within `StaticFileHandler`.
+    #
+    # Since the `File::Stat` is created once in `#call` and then passed around to the
+    # rest of the class's methods, imitating the object allows us to only lookup
+    # the cache hash once for every request.
+    #
+    private record CachedFile, data : Bytes, size : Int64, modification_time : Time
+
+    CACHE_LIMIT = 5_000_000 # 5MB
+    @@cached_files = {} of Path => CachedFile
+
+    # A simplified version of `#call` for Invidious to improve performance.
+    #
+    # This is basically the same as what we inherited but just with the directory listing
+    # features stripped out. This removes some conditional checks and calls which improves
+    # performance slightly but otherwise is entirely unneeded.
+    #
+    # Really, all the cache feature actually needs is to override the much simplifier `file_info`
+    # method to return a `CachedFile` or `File::Stat` depending on whether the file is cached.
+    def call(context) : Nil
+      check_request_method!(context) || return
+
+      request_path = request_path(context)
+
+      check_request_path!(context, request_path) || return
+
+      request_path = Path.posix(request_path)
+      expanded_path = request_path.expand("/")
+
+      # The path normalization can be simplified to just this since
+      # we don't need to care about normalizing directory urls.
+      if request_path != expanded_path
+        redirect_to context, expanded_path
+      end
+
+      file_path = @public_dir.join(expanded_path.to_kind(Path::Kind.native))
+
+      if cached_info = @@cached_files[file_path]?
+        return serve_file_with_cache(context, cached_info, file_path)
+      end
+
+      file_info = File.info?(file_path)
+
+      return call_next(context) unless file_info
+
+      if file_info.file?
+        # Actually means to serve file *with cache headers*
+        # The actual logic for serving the file is done in `#serve_file`
+        serve_file_with_cache(context, file_info, file_path)
+      else # Not a normal file (FIFO/device/socket)
+        call_next(context)
+      end
+    end
+
+    # Add "Cache-Control" header to the response
+    private def add_cache_headers(response_headers : HTTP::Headers, last_modified : Time) : Nil
+      super; response_headers["Cache-Control"] = "max-age=2629800"
+    end
+
+    # Serves and caches the file at the given path.
+    #
+    # This is an override of `serve_file` to allow serving a file from memory, and to cache it
+    # it as needed.
+    private def serve_file(context : HTTP::Server::Context, file_info, file_path : Path, original_file_path : Path, last_modified : Time)
+      context.response.content_type = MIME.from_filename(original_file_path.to_s, "application/octet-stream")
+
+      range_header = context.request.headers["Range"]?
+
+      if !file_info.is_a? CachedFile
+        retrieve_bytes_from = IO::Memory.new
+
+        File.open(file_path) do |file|
+          # We cannot cache partial data so we'll rewind and read from the start
+          if range_header
+            dispatch_serve(context, file, file_info, range_header)
+            IO.copy(file.rewind, retrieve_bytes_from)
+          else
+            context.response.output = IO::MultiWriter.new(context.response.output, retrieve_bytes_from, sync_close: true)
+            dispatch_serve(context, file, file_info, range_header)
+          end
+        end
+
+        return flush_io_to_cache(retrieve_bytes_from, file_path, file_info)
+      else
+        return dispatch_serve(context, file_info.data, file_info, range_header)
+      end
+    end
+
+    # Writes file data to the cache
+    private def flush_io_to_cache(io, file_path, file_info)
+      if @@cached_files.sum(&.[1].size) + (size = file_info.size) < CACHE_LIMIT
+        data_slice = io.to_slice
+        @@cached_files[file_path] = CachedFile.new(data_slice, file_info.size, file_info.modification_time)
+      end
+    end
+
+    # Either send the file in full, or just fragments of it depending on the request
+    private def dispatch_serve(context, file, file_info, range_header)
+      if range_header
+        # an IO is needed for `serve_file_range`
+        file = file.is_a?(Bytes) ? IO::Memory.new(file, writeable: false) : file
+        serve_file_range(context, file, range_header, file_info)
+      else
+        context.response.headers["Accept-Ranges"] = "bytes"
+        serve_file_full(context, file, file_info)
+      end
+    end
+
+    # Skips the stdlib logic for serving pre-gzipped files
+    private def serve_file_compressed(context : HTTP::Server::Context, file_info, file_path : Path, last_modified : Time)
+      serve_file(context, file_info, file_path, file_path, last_modified)
+    end
+
+    # If we're serving the full file right away then there's no need for an IO at all.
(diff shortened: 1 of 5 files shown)


src/invidious/http_server/static_assets_handler.cr
```diff
@@ -0,0 +1,119 @@
...
+      def directory?
+        false
+      end
+
+      def file?
+        true
+      end
+    end
+
+    CACHE_LIMIT = 5_000_000 # 5MB
```
syeopite, line 22: The `CACHE_LIMIT` is actually way too lenient imo. It actually allows you to fit everything from the assets folder since the entire non-compressed folder with the videojs dependencies is 4.9MB. And if you minified the videojs scripts, its only 2.7MB. Adding gzip compression to that can then get it down to around 840kB.

Adding some simple LRU cache and/or compressing the cached files might be something to consider.

src/invidious/http_server/static_assets_handler.cr
```diff
@@ -0,0 +1,119 @@
...
+    # Serves and caches the file at the given path.
+    #
+    # This is an override of `serve_file` to allow serving a file from memory, and to cache it
+    # it as needed.
+    private def serve_file(context : HTTP::Server::Context, file_info, file_path : Path, original_file_path : Path, last_modified : Time)
+      context.response.content_type = MIME.from_filename(original_file_path.to_s, "application/octet-stream")
+
+      range_header = context.request.headers["Range"]?
+
+      if !file_info.is_a? CachedFile
```
SamantazFox: Can you invert the condition to return early and reduce the indentation of the rest of the code?

src/invidious/http_server/static_assets_handler.cr
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    # Serves segments of a file based on the `Range header`
+    #
+    # An override of `serve_file_range` to allow using a generic IO rather than a `File`.
+    # Literally the same code as what we inherited but just with the `file` argument's type
+    # being set to `IO` rather than `File`
+    #
+    # Can be removed once https://github.com/crystal-lang/crystal/issues/15817 is fixed.
+    private def serve_file_range(context : HTTP::Server::Context, file : IO, range_header : String, file_info)
```
SamantazFox, line 107: Given the size of files we are serving, we probably could let the stdlib code load the file from disk if we ever receive a ranged requests.

src/invidious/http_server/static_assets_handler.cr
SamantazFox: Yeah, I would probably only save the gzipped + deflated versions to cache. Even storing the three versions would be less than 15M, which is perfectly reasonnable given the rest of the RAM usage of invidious.

src/invidious/http_server/static_assets_handler.cr
```diff
@@ -0,0 +1,119 @@
...
+    # If the requested file is cached, a `CachedFile` is returned instead of a `File::Stat`.
+    # This represents the metadata info of a cached file and implements all the methods of `File::Stat` that
+    # is used by the `StaticAssetsHandler`.
+    #
+    # The `CachedFile` also stores the raw bytes of the cached file, and this method serves as the place where
+    # the cached file is retrieved if it exists. Though the data will only be read in `#serve_file`
+    private def file_info(expanded_path : Path)
+      file_path = @public_dir.join(expanded_path.to_kind(Path::Kind.native))
+      {@@cached_files[file_path]? || File.info?(file_path), file_path}
+    end
```
SamantazFox, line 37: I have two questions here:

1. Do you think it could be interesting to keep track of the time a file was last read from disk, so that we could reload it if its modification time changed? I'm thinking about users that complained about not being able to make CSS changes without restarting.
2. Less important: is there a reason to use the full (system) path instead of the shorter `expanded_path` as the key of the `cached_files` Hash?

NovaAndrom3da:
Hi! I'm wondering if there's anything blocking these commits? I'm hoping to be able to use the new crystallang version hopefully soon.

src/invidious/http_server/static_assets_handler.cr
syeopite: Well keeping track of when the file was last read to update the cache after N time has passed and the last modified time has changed will still mean that things like CSS changes won't immediately show up till after that initial time to live.

Since we do already track the last modified time of the cached files already, we could just compare it with the actual last modified time on disk for each request. But that's going to cause a lot of disk reads on instances with high request loads. I don't know if that's going to be a nonissue on modern systems but Invidious already has a reputation of killing drives and I'm concerned that this could potentially strain disks further.

Alternatively we could also just forgo the file cache system entirely and let instance maintainers add caching during deployment instead.

---

As for why the full path is used, the later methods down the chain doesn't have access to the shorter `expanded_path` so they won't be able to use it as a key to write to the cache.

src/invidious/http_server/static_assets_handler.cr
syeopite: The compressed responses are created by the `FilteredCompressHandler` so I'm not sure how the compressed data can be stored without duplicating a bunch of code and skipping the `FilteredCompressHandler` entirely.

---

Design wise though I was thinking that maybe we could just store only the gzipped data and just decompress and recompress as needed to serve the uncompressed, deflate, and range requests. For most requests it'll just be sending the compressed gzip data and for the rest I don't think the extra (re)compression steps would cause too much of an issue.

syeopite:
> Hi! I'm wondering if there's anything blocking these commits? I'm hoping to be able to use the new crystallang version hopefully soon.

The PR is still going through the code review process but after that it can be merged

Fijxu:
This is now needed to build the latest invidious commits with Crystal 1.17.1 🫤 :

```
In src/ext/kemal_static_file_handler.cr:167:49

 167 | directory_listing(context.response, request_path, file_path)
                                           ^-----------
Error: expected argument #2 to 'Kemal::StaticFileHandler#directory_listing' to be Path, not String

Overloads are:
 - HTTP::StaticFileHandler#directory_listing(io : IO, request_path : Path, path : Path)
 ```

**Comment by Fijxu** [comment]
I just tested this PR and I can't get Invidious interface to load completely, css files are empty and js files do not load:

image

**Comment by syeopite** [comment]
Oops looks like #5337 is required for compression to work correctly... I just rebased against master which includes that commit and its fixed now.

A test case can't be added due to needing to import `handlers.cr` which depends on many other Invidious objects and constants... I guess we should separate the handlers there to make designing tests easier

**Comment by Fijxu** [comment]
I tested it and works fine now ;). Thanks

**Comment by dekeonus** [comment]
in contrast to @Fijxu I'm not seeing any of the static content load in the browser (neither Firefox-143.0 nor Google-Chrome-140.0.7339.207 on linux).  Even the favicon fails to load.
This seems to be related to content-type mismatch coupled with "x-content-type-options: nosniff"

here are some of the headers for the request for the favicon
**GET**
```
scheme: https
host: invidious.localdomain
filename:  /favicon-16x16.png
```

snippet of **Request Headers**
`Accept: image/avif,image/webp,image/png,image/svg+xml,image/*;q=0.8,*/*;q=0.5`

snippet of **Response Headers**
`content-type: text/html; charset=UTF-8`

Console log error
> The resource from “https://invidious.localdomain/feed/popular” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).

This same mismatch occurs with css files and js (i.e. they are sent with "content-type: text/html; charset=UTF-8")

EDIT:
This is Gentoo amd64 with
crystal-1.17.1
kemal-1.7.2¹
exception_page-0.5.0
invidious-v2.20250913.0
¹ kemal-1.6.0 is not in gentoo's guru repo (community maintained) due to the path transversal vulnerability that was fixed in kemal-1.7.0

**Comment by dekeonus** [comment]
After further examination, the issue I'm having is that the redirect to the static assets is not correct.
I restarted (after reconfiguring) invidious to serve directly (rather than by reverse proxy), to rule out some interaction there.
Note the location header is "/" in the response headers returned below.

GET
```
scheme: http
host: 10.144.1.5:3000
filename: /css/empty.css

v: b32b077
Address: 10.144.1.5:3000

Status: 302 Found
Version: HTTP/1.1
Transferred: 0 GB (0 GB size)
Referrer Policy: same-origin
DNS Resolution: System
```

Response Headers
```
HTTP/1.1 302 Found
Connection: keep-alive
Content-Type: text/html
Date: Sun, 28 Sep 2025 19:50:24 GMT
X-Frame-Options: sameorigin
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
Content-Security-Policy: default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:; connect-src 'self'; manifest-src 'self'; media-src 'self' blob:; child-src 'self' blob:; frame-src 'self'; frame-ancestors 'none'
Referrer-Policy: same-origin
Permissions-Policy: interest-cohort=()
Location: /
Content-Length: 0
```

Request Headers
```
GET /css/empty.css?v=b32b077 HTTP/1.1
Host: 10.144.1.5:3000
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:143.0) Gecko/20100101 Firefox/143.0
Accept: text/css,*/*;q=0.1
Accept-Language: en-AU,en-GB;q=0.7,en;q=0.3
Accept-Encoding: gzip, deflate
Referer: http://10.144.1.5:3000/
DNT: 1
Sec-GPC: 1
Connection: keep-alive
```

dekeonus:
Sorry for the noise, my issue was insufficient sed in the gentoo build script.
However, that said, it does raise the issue of less than ideal operation.

In gentoo's case the build script modifies several files to put *assets/*, *locales/*, and *config/{sql,migrate-scripts}/* under /usr/share/invidious.
When invidious fails to find one of the asset files, it continues running without it (that's fine) BUT without logging the fact that it failed to load the asset file, nor sending to the client a 404 response due to a non-existent resource.

PR #5345, opened 2025-06-12
Videos: Add fallback to TvSimply client

Fijxu:
Depends on: https://github.com/iv-org/invidious/pull/5344

Uses the same order Invidious companion uses: https://github.com/iv-org/invidious-companion/blob/d0c4bb79ae4688d019fb281257859e334adb7d8b/src/lib/helpers/youtubePlayerReq.ts#L75

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -111,7 +111,7 @@ def extract_video_info(video_id : String)
   if !CONFIG.invidious_companion.present?
     if player_response.dig?("streamingData", "adaptiveFormats", 0, "url").nil?
       LOGGER.warn("Missing URLs for adaptive formats, falling back to other YT clients.")
-      players_fallback = {YoutubeAPI::ClientType::TvHtml5, YoutubeAPI::ClientType::TvSimply, YoutubeAPI::ClientType::WebMobile}
+      players_fallback = {YoutubeAPI::ClientType::TvSimply, YoutubeAPI::ClientType::WebMobile}
 
       players_fallback.each do |player_fallback|
         client_config.client_type = player_fallback


unixfox:
IMO it's better to create just one PR instead of two.

Fijxu:
You're right. I'll fix it and test it later.

Fijxu:
I wasn't able to test it since Innertube didn't return `signatureCipher` to me. I added https://github.com/iv-org/invidious-companion/blob/d0c4bb79ae4688d019fb281257859e334adb7d8b/src/lib/helpers/youtubePlayerReq.ts#L92 to companion because when I was working on it, I got `signatureCipher` but not `url`. 

Maybe we should really add some mocks, test data and crystal tests :/

src/invidious/videos/parser.cr
```diff
@@ -146,6 +146,9 @@ def extract_video_info(video_id : String)
   if streaming_data = player_response["streamingData"]?
     %w[formats adaptiveFormats].each do |key|
       streaming_data.as_h[key]?.try &.as_a.each do |format|
+        if format.as_h["url"].nil?
```
syeopite:
```suggestion
        if format.as_h["url"]?.nil?
```

syeopite: The hash conversion should be called `as_h` only once:

```crystal
format = format.as_h
```

syeopite (changes requested):

syeopite:
Technically there already are mocks and tests, there's just not enough of it and imo the current system is confusing to understand.

I don't fully understand it myself if I'm being honest but I believe you'll need to update `iv-org/mocks` with the mock json for this, pull the latest version of the submodule in the invidious repo, and then finally add a test for it here.

src/invidious/videos/parser.cr
```diff
@@ -111,14 +111,14 @@ def extract_video_info(video_id : String)
...
-      players_fallback = {YoutubeAPI::ClientType::TvHtml5, YoutubeAPI::ClientType::WebMobile}
+      players_fallback = {YoutubeAPI::ClientType::TvHtml5, YoutubeAPI::ClientType::TvSimply, YoutubeAPI::ClientType::WebMobile}
 
       players_fallback.each do |player_fallback|
         client_config.client_type = player_fallback
 
         next if !(player_fallback_response = try_fetch_streaming_data(video_id, client_config))
 
-        if player_fallback_response.dig?("streamingData", "adaptiveFormats", 0, "url")
+        if player_fallback_response.dig?("streamingData", "adaptiveFormats", 0, "url") || player_fallback_response.dig?("streamingData", "adaptiveFormats", 0, "signatureCipher")
```
syeopite: Can you save the result of `player_fallback_response.dig?("streamingData", "adaptiveFormats", 0)` to a temporary variable first so that there isn't a double traversal?

Fijxu: is https://github.com/iv-org/invidious/pull/5345/commits/cf0a68bd77251528713404822a19c411a1c0aaca good enough?

Fijxu:
I have taken the opportunity to remove the TV fallback here because of https://github.com/iv-org/invidious-companion/pull/157
TvSimply works fine enough on my instance (https://inv.nadeko.net)

syeopite:
Merged with https://github.com/iv-org/invidious/commit/0308acb624657a30aff46417f848a2456c19025b

I'm not sure why Github's UI isn't marking this as merged.

PR #5346, opened 2025-06-12
Show message when connection to the database is not possible

Fijxu:
This gives a better message when something goes wrong on the database connection. May not be better for debugging purposes, but the error messages returned are from PostgreSQL

image

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -60,7 +60,13 @@ alias IV = Invidious
 CONFIG   = Config.load
 HMAC_KEY = CONFIG.hmac_key
 
-PG_DB       = DB.open CONFIG.database_url
+PG_DB = begin
+  DB.open CONFIG.database_url
+rescue ex
+  puts "Failed to connect to PostgreSQL database: #{ex.cause.try &.message}"
+  puts "Check your 'config.yml' database settings or PostgreSQL settings."
+  exit(1)
+end
 ARCHIVE_URL = URI.parse("https://archive.org")
 PUBSUB_URL  = URI.parse("https://pubsubhubbub.appspot.com")
 REDDIT_URL  = URI.parse("https://www.reddit.com")


unixfox (approved):

src/invidious.cr
```diff
@@ -60,7 +60,13 @@ alias IV = Invidious
 CONFIG   = Config.load
 HMAC_KEY = CONFIG.hmac_key
 
-PG_DB       = DB.open CONFIG.database_url
+PG_DB = begin
+  DB.open CONFIG.database_url
+rescue ex
+  puts "Failed to connect to PostgreSQL database: #{ex.cause.try &.message}"
```
SamantazFox, line 66: Just to make the linter happy:
```suggestion
rescue exc
  puts "Failed to connect to PostgreSQL database: #{exc.cause.try &.message}"
```

SamantazFox: vdfjvgbdkrfg
I should have checked the ameba docs, sorry :/
```
src/invidious.cr:65:1
[C] Naming/RescuedExceptionsVariableName: Disallowed variable name, use one of these instead: 'e', 'ex', 'exception', 'error'
> rescue exc
  ^
```

src/invidious.cr
```diff
@@ -60,7 +60,13 @@ alias IV = Invidious
 CONFIG   = Config.load
 HMAC_KEY = CONFIG.hmac_key
 
-PG_DB       = DB.open CONFIG.database_url
+PG_DB = begin
+  DB.open CONFIG.database_url
+rescue exc
+  puts "Failed to connect to PostgreSQL database: #{exc.cause.try &.message}"
```
SamantazFox:
```suggestion
rescue e
  puts "Failed to connect to PostgreSQL database: #{e.cause.try &.message}"
```

Fijxu: ~~We should change all `rescue ex` from the source code to make the linter happy or just this part :?~~

`ex` seems to be accepted 🤔

syeopite (changes requested):
Oops I didn't realize that lint also failed... It would be nice if the Github UI could indicate failures of unstable branches vs stable ones like the lint job.

Anyways the original error was because the ex variable ended up overshadowing the variable in this Kemal status code handler later on:

```crystal
error 500 do |env, ex|
```

The fix is to change that `ex` to something else rather than change the variable for the rescue block.

PR #5347, opened 2025-06-12
Document use of unix sockets for `db`

Fijxu:
Currently only `database_url` shows how to connect via unix sockets. This adds some docs for `db` too

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -8,6 +8,13 @@
 ## Database configuration with separate parameters.
 ## This setting is MANDATORY, unless 'database_url' is used.
 ##
+## Note: The 'db' setting allows the use of UNIX
+## sockets. To do so, set 'host' to ""
+## E.g:
+##    password: kemal
+##    host: ""
+##    port: 5432
+##
 db:
   user: kemal
   password: kemal


unixfox (approved):

Fijxu:
Note: When using `config.db`, Invidious will only search for this list of paths on the filesystem: https://github.com/will/crystal-pg/blob/cafe112f2847f366262460ee999e74f9c7e8b31c/src/pq/conninfo.cr#L7

It will not consider PostgreSQL installations running on another port like `5433` (this will make PostgreSQL create a unix socket called `.s.PGSQL.5433` and not `.s.PGSQL.5432`), so it will fail to connect. This can cause some annoyances on PostgreSQL cluster installations where you can run more than one PostgreSQL version and server on different ports (how Debian seems to handle it: https://wiki.debian.org/PostgreSql#Listing_existing_database_clusters). 

This is not a problem when using `database_url` since you can pass a custom unix socket path using the `host` query parameter like this: 

```
database_url: postgres://kemal:kemal@/invidious?host=/var/run/postgresql/.s.PGSQL.5433
```

Maybe we can add an option to point to a socket path on `db` 🤔?

SamantazFox:
> Note: When using `config.db`, Invidious will only search for this list of paths on the filesystem: https://github.com/will/crystal-pg/blob/cafe112f2847f366262460ee999e74f9c7e8b31c/src/pq/conninfo.cr#L7
>
> It will not consider PostgreSQL installations running on another port like `5433` (this will make PostgreSQL create a unix socket called `.s.PGSQL.5433` and not `.s.PGSQL.5432`), so it will fail to connect. 

This is the preconfigured fallback. If you pass a host, it will use that instead:
https://github.com/will/crystal-pg/blob/cafe112f2847f366262460ee999e74f9c7e8b31c/src/pq/conninfo.cr#L125

Then, if the host starts with a `/`, the PG driver will open a unix socket:
https://github.com/will/crystal-pg/blob/cafe112f2847f366262460ee999e74f9c7e8b31c/src/pq/connection.cr#L25-L26

I think that all we need to do is adapt our code to not build a `postgres://` URI if we see a host starting with `/`.

Issue #5351, opened 2025-06-15
[Enhancement] Detect and warn on common `domain` config misconfigurations


syeopite:
The `domain` property is a standard fully qualified domain name without any schemes or paths attached, including any trailing slashes. Misconfiguring this property will cause hard-to-diagnose problems in many of the core features of Invidious. 

To improve user experience it should at least warn (or maybe error) when the domain name appears to be misconfigured.

PR #5352, opened 2025-06-15
check domain configuration before starting Invidious

Fijxu:
Closes https://github.com/iv-org/invidious/issues/5351

image

### DIFF
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -319,6 +319,18 @@ class Config
       end
     end
 
+    # Check if the domain configuration is valid
+    if (domain = config.domain).nil? || domain.empty?
+      puts "Config: 'domain' is required/can't be empty"
+      exit(1)
+    elsif parsed_domain = URI.parse(domain)
+      if !parsed_domain.scheme.nil?
+        puts "Config: 'domain' cannot contain a scheme (#{parsed_domain.scheme})"
+        puts "Config: (Hint) Set 'domain' to '#{parsed_domain.host}'"
+        exit(1)
+      end
+    end
+
     return config
   end
 end


src/invidious/config.cr
```diff
@@ -319,6 +319,18 @@ class Config
       end
     end
 
+    # Check if the domain configuration is valid
+    if (domain = config.domain).nil? || domain.empty?
+      puts "Config: 'domain' is required/can't be empty"
```
SamantazFox: `domain` can be empty on local instances.

src/invidious/config.cr
```diff
@@ -319,6 +319,18 @@ class Config
...
+      exit(1)
+    elsif parsed_domain = URI.parse(domain)
+      if !parsed_domain.scheme.nil?
+        puts "Config: 'domain' cannot contain a scheme (#{parsed_domain.scheme})"
+        puts "Config: (Hint) Set 'domain' to '#{parsed_domain.host}'"
```
SamantazFox: Even though the scheme is an usual mistake, we should probably catch all other kinds of errors (trailing `/`, port number, etc)...
```suggestion
      if domain != parsed_domain.domain
        puts "Config: 'domain' is invalid.
        
        if parsed_domain.host.presence
          puts "Config: (Hint) Did you mean #{parsed_domain.host} ?"
        else
          puts "Config: (Hint) 'domain' should look like this: invidious.example.com"
        end
```

SamantazFox (changes requested):

src/invidious/config.cr
```diff
@@ -319,6 +319,23 @@ class Config
       end
     end
 
+    # Check if the domain configuration is valid
+    if (domain = config.domain).nil? || domain.empty?
+      puts "Config: 'domain' is required/can't be empty"
+      exit(1)
+    elsif parsed_domain = URI.parse(domain)
+      if domain != parsed_domain.domain
+        puts "Config: 'domain' is invalid.
```
SamantazFox:
```suggestion
        puts "Config: 'domain' is invalid."
```

src/invidious/config.cr
```diff
@@ -319,6 +319,23 @@ class Config
       end
     end
 
+    # Check if the domain configuration is valid
+    if (domain = config.domain).nil? || domain.empty?
```
SamantazFox: By the way you can use `String.presence` instead of a separate nil/empty check
```suggestion
    if !config.domain.presence
```

unixfox:
@Fijxu could you check @SamantazFox reviews :)?

Fijxu:
I did something more complicated while testing but this seems to be enough.

```
$ INVIDIOUS_DOMAIN=https://inv.nadeko.net:8080/meow/asd/meow ./invidious
Config (Hint): Remove the path from your domain: '/meow/asd/meow'
$ INVIDIOUS_DOMAIN=https://inv.nadeko.net:8080 ./invidious
Config (Hint): Remove the scheme from your domain: 'https://'
$ INVIDIOUS_DOMAIN=inv.nadeko.net:8080 ./invidious
Config (Hint): Remove the port from your domain: ':8080'
$ INVIDIOUS_DOMAIN=inv.nadeko.net/asd/asd/asd ./invidious
Config (Hint): Remove the path from your domain: '/asd/asd/asd'
$ INVIDIOUS_DOMAIN=https://inv.nadeko.net ./invidious
Config (Hint): Remove the scheme from your domain: 'https://'
$ INVIDIOUS_DOMAIN=tcp://inv.nadeko.net ./invidious
Config (Hint): Remove the scheme from your domain: 'tcp://'
$ INVIDIOUS_DOMAIN=meow ./invidious
[development] Invidious is ready to lead at http://0.0.0.0:30001

$ INVIDIOUS_DOMAIN=meow.localhost ./invidious
[development] Invidious is ready to lead at http://0.0.0.0:30001

$ INVIDIOUS_DOMAIN=meow.localhost.lol ./invidious
[development] Invidious is ready to lead at http://0.0.0.0:30001
```

unixfox:
Thanks @Fijxu and with empty domain?

Fijxu:
> Thanks @Fijxu and with empty domain?

```
INVIDIOUS_DOMAIN="" ./invidious
[development] Invidious is ready to lead at http://0.0.0.0:30001
```

The `Set-Cookie` header has no `domain` property when trying to log in since `INVIDIOUS_DOMAIN` is empty, as it should be: `SID=<some_sid>; expires=Wed, 16 Feb 2028 17:11:37 GMT; HttpOnly; SameSite=Lax`, so works fine ;)

PR #5367, opened 2025-06-26
pages/watch: URL encode 'action' in download widget

SamantazFox:
Caught in the review of PR 5224, but forgot to click on "send review" in time. I realized that too late, after the PR was already merged.

### DIFF
--- a/src/invidious/frontend/watch_page.cr
+++ b/src/invidious/frontend/watch_page.cr
@@ -32,7 +32,7 @@ module Invidious::Frontend::WatchPage
     return String.build(4000) do |str|
       str << "<form"
       str << " class=\"pure-form pure-form-stacked\""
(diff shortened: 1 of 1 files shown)


src/invidious/frontend/watch_page.cr
```diff
@@ -32,7 +32,7 @@ module Invidious::Frontend::WatchPage
     return String.build(4000) do |str|
       str << "<form"
       str << " class=\"pure-form pure-form-stacked\""
-      str << " action='#{url}'"
+      str << " action='" << URI.encode_www_form(url) << "'"
```
syeopite: I think this only needs to be HTML escaped rather than URL encoded since that'll also destroy the path slashes and scheme colon.

Issue #2142, opened 2021-06-10
[Bug] [Security issue] Server ip address leaked when behind a proxy.


moom0o:
**Describe the bug**
Bad actors can easily get the backend server ip by logging the web requests. This in turn can allow bad actors to easily ddos the backend server.

**Steps to Reproduce**
1. Click on a random video
2. Press F12
3. Go to network
4. Click play
5. Find a web request to googlevideo.com
6. The ip is located in the request url, making cloudflare/other ddos protection services useless.

**Screenshots**
![Image](https://i.imgur.com/0ZrYml7.png)
**Additional context**
 - Browser (if applicable): Brave 1.25.70
 - OS (if applicable): Windows 10 Pro 20H2

Is there any important reason why the server ip address needs to be sent in a request, or can this easily be removed? 
Thanks.

unixfox:
If I'm not mistaken, this is needed for the videoplayback endpoint, by removing it the YouTube server will drop the request.

I just tried a random URL from invidious.snopyta.org:

[...]

PR #5381, opened 2025-07-07
Add support for encrypted query parameters

Fijxu:
May close https://github.com/iv-org/invidious/issues/2142

Encrypts sensitive query parameters that should be hidden to users like `ip` and `pot` using the `hmac_key` config option. 

Taken from https://github.com/iv-org/invidious-companion/pull/65

There is some problems with this current implementation:
- The `hmac_key` config option needs to be at least 16 characters, if we force it, that will break some setups on a new update. I have thought of 3 ways to fix this:
  -  add characters to the `hmac_key` for the `ecb_without_salt` function until it has 16 (which is the minimum required for AES) if it's less than 16
  - use `invidious_companion_key`
  - introduce a new config option like `invidious_companion_key` but to encrypt the video query parameters.
- IP address and PoToken are still leaked if the video is not proxied and quality is set to medium, but that can be solved by forcing `Proxy videos` player preferences (Like I did here: https://git.nadeko.net/Fijxu/invidious/commit/7c9f79e1f146c59370bbe5e7abe82d37ec5fd5de). This will depend of the instance maintainer operator since proxying every video will use a lot of bandwidth, specially on public instances
- Any other videoplayback proxy without support for the encrypted parameters will not work, like https://github.com/TeamPiped/piped-proxy and https://github.com/TeamPiped/http3-ytproxy (But my fork supports it: https://git.nadeko.net/Fijxu/http3-ytproxy/src/commit/a6a33526a494a0719139602ec99c437dd64d81ef/internal/paths/videoplayback.go#L77)

Since Invidious companion will be become the official way to fetch youtube videos and videoplayback proxying, this may not be very useful on the future since companion is already able to encrypt the `ip` and `pot` query parameters, but will work for old setups that are still using `inv_sig_helper` for whatever reason (low memory capacity for example, since Invidious companion uses way more memory than `inv_sig_helper`)

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -568,6 +568,21 @@ hmac_key: "CHANGE_ME!!"
 ##
 #playlist_length_limit: 500
 
+##
+## Encrypts query params 'ip' and 'pot'
+## This is useful if you don't want to leak the IP address
+## and PoToken used when fetching a Youtube video.
+##
+## Note: This will only work if the 'Proxy videos' preference is enabled
+## on the Player preferences. Users that do not have 'Proxy videos' enabled
+## will be able to see the IP address and PoToken used on videoplayback
+## requests.
+##
+## Accepted values: true, false
+## Default: false
+##
+#encrypt_query_params: false
+
 #########################################
 #
 #  Default user preferences
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -183,6 +183,8 @@ class Config
   # Playlist length limit
   property playlist_length_limit : Int32 = 500
 
+  property encrypt_query_params : Bool = false
+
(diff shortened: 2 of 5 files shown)


Fijxu:
Oops, I didn't know that `require "uri/params/serializable"` was unsupported on Crystal <=1.13.3 (https://github.com/iv-org/invidious/actions/runs/16105127771/job/45439447127?pr=5381#step:7:11). I guess I'll do it another way then

syeopite:
Imo I think we can just drop support for `1.12.2`. Docker is technically the only official platform Invidious supports and our images are all on newer versions now.

Fijxu:
Closed since `inv_sig_helper` is now deprecated.

PR #5392, opened 2025-07-19
CI: Bump Crystal version matrix

Fijxu:
Add 1.17.0


Fijxu:
Requires https://github.com/iv-org/invidious/pull/5338

PR #5394, opened 2025-07-21
feat: Add configurable max_request_line_size to handle 414 errors

hellosunghyun:
## Summary
This PR adds a configurable `max_request_line_size` option to handle 414 (URI Too Long) errors that occur with very long continuation tokens in API requests.

## Problem
Users are experiencing 414 errors when accessing channel videos with long continuation tokens. The default HTTP request line size limit (8KB) is insufficient for some YouTube continuation tokens, which can be extremely long.

Example error:
```
414 Client Error: URI Too Long for url: http://localhost:8080/api/v1/channels/UChQ-VMvdGrYZxviQVMTJOHg/videos?sort_by=newest&continuation=4qmFsgLT...
```

## Solution
- Added `max_request_line_size` configuration option to the Config class
- Modified server initialization to apply the custom limit if configured
- Provided clear documentation and examples for both config.yml and environment variable usage

## Configuration
Users can now configure the limit in three ways:

1. In `config.yml`:
```yaml
max_request_line_size: 16384  # 16KB
```

2. Via environment variable:
```bash
INVIDIOUS_MAX_REQUEST_LINE_SIZE=16384
```

3. In `docker-compose.yml`:
```yaml
environment:
  INVIDIOUS_MAX_REQUEST_LINE_SIZE: 16384
```

## Default Behavior
- The default behavior remains unchanged (8KB limit)
- Only users experiencing 414 errors need to configure this option
- Conservative default ensures security for public instances

## Testing
- Tested with long continuation tokens that previously caused 414 errors
- Verified that the configuration is properly applied from all sources (config.yml, env var)
- Confirmed that default behavior is maintained when option is not set

## Security Considerations
- Added documentation warning about increasing this value on public instances
- The option is intentionally not enabled by default to maintain security posture

This change provides a flexible solution for users experiencing URL length issues without affecting existing installations.

Fixes issues related to long continuation tokens in channel video pagination.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -191,6 +191,20 @@ https_only: false
 #  path: /tmp/invidious.sock
 #  permissions: 777
 
+##
+## Maximum size of the HTTP request line (in bytes).
+## Increase this value if you encounter 414 errors when using URLs with long
+## query parameters (e.g., very long continuation tokens).
+##
+## Note: This directly sets the HTTP server's max_request_line_size.
+## Be cautious when increasing this value on public instances.
+##
+## Accepted values: integer (size in bytes)
+## Default: 8192 (8KB, Crystal's default)
+## Recommended: 16384 (16KB) if experiencing 414 errors
+##
+#max_request_line_size: 16384
+
 
 # -----------------------------
 #  Network (outbound)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,8 @@ services:
       # Please read the following file for a comprehensive list of all available
       # configuration options and their associated syntax:
       # https://github.com/iv-org/invidious/blob/master/config/config.example.yml
+      # Uncomment to increase max request line size (if you get 414 errors):
+      # INVIDIOUS_MAX_REQUEST_LINE_SIZE: 16384
       INVIDIOUS_CONFIG: |
         db:
           dbname: invidious
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -252,6 +252,11 @@ Kemal.config.app_name = "Invidious"
 {% end %}
 
 Kemal.run do |config|
+  # Set max request line size if configured
+  if max_size = CONFIG.max_request_line_size
+    config.server.not_nil!.max_request_line_size = max_size
+  end
+  
   if socket_binding = CONFIG.socket_binding
     File.delete?(socket_binding.path)
     # Create a socket and set its desired permissions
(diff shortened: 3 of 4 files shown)


config/config.example.yml
```diff
@@ -191,6 +191,20 @@ https_only: false
...
+##
+## Note: This directly sets the HTTP server's max_request_line_size.
+## Be cautious when increasing this value on public instances.
+##
+## Accepted values: integer (size in bytes)
+## Default: 8192 (8KB, Crystal's default)
+## Recommended: 16384 (16KB) if experiencing 414 errors
+##
+#max_request_line_size: 16384
+
```
syeopite, line 207: The default value should reflect the actual default value of `8192`

config/config.example.yml
```diff
@@ -191,6 +191,20 @@ https_only: false
...
+## Maximum size of the HTTP request line (in bytes).
+## Increase this value if you encounter 414 errors when using URLs with long
+## query parameters (e.g., very long continuation tokens).
+##
+## Note: This directly sets the HTTP server's max_request_line_size.
+## Be cautious when increasing this value on public instances.
+##
+## Accepted values: integer (size in bytes)
+## Default: 8192 (8KB, Crystal's default)
+## Recommended: 16384 (16KB) if experiencing 414 errors
```
syeopite, line 204: I don't think any default recommendations is needed here. Imo instances should decide this value according to their own requirements and needs if they decide to stray from the default 8 kilobytes.

Fijxu:
I just tested this, 16384 works fine for long URL with a large continuation string when trying to go to the next video page of a channel.

config/config.example.yml
Fijxu: What about now that Youtube started to return a even larger `continuation` string (https://github.com/iv-org/invidious/issues/5529)? `16384` seems like is a must now. Unless we adapt Invidious channel next page button to use a POST request to go to the next page with it's continuation as a body.

unixfox: I'm also in favor of a hardcoded value. This way we know everyone is on the same page.

Sommerwiesel:
Hi y'all,
FYI, I've been using 16384 on my Instance for ~ 3 months now due to an user bug report and it works fine.
https://git.nerdvpn.de/NerdVPN.de/invidious/issues/1

Imo 16384 is a good new default.

unixfox:
@hellosunghyun Could you please hardcode to `16384`? We will change that value if needed in the future. But at least everyone is on the same page.

PR #5406, opened 2025-08-04
Complete API-only mode implementation

hellosunghyun:
## Summary
This PR completes the API-only mode implementation for Invidious, allowing it to be built and run without GUI/frontend components. This significantly reduces the binary size and dependencies, making it ideal for API-focused deployments.

## Motivation
The API-only mode was partially implemented but had several missing pieces that prevented it from working correctly. This PR fills those gaps and ensures the API-only build works seamlessly.

## Changes
- Added conditional compilation flags (`{% if flag?(:api_only) %}`) throughout the codebase to exclude frontend-specific code
- Created stub implementations for database operations in API-only mode (`DummyDB`, `VideoNotification`, etc.)
- Updated Docker configurations to support API-only builds with proper build arguments
- Refactored require statements into a dedicated `requires.cr` file for better modularity
- Fixed route handlers to work correctly without frontend dependencies
- Added proper error handling for API-only mode

### Using Crystal directly:
```bash
crystal build src/invidious.cr -Dapi_only --release
```

### Using Make:
```bash
make API_ONLY=1
```

### Using Docker:
```bash
# For AMD64
docker build --build-arg release=1 --build-arg api_only=1 -t invidious-api-only .

# For ARM64
docker build --platform linux/arm64 --build-arg release=1 --build-arg api_only=1 -f docker/Dockerfile.arm64 -t invidious-api-only:arm64 .
```

## Benefits
- **Reduced binary size**: The API-only binary is significantly smaller without frontend assets and dependencies
- **Lower resource usage**: Less memory and CPU usage due to excluded frontend components
- **Faster startup**: Quicker initialization without frontend-related setup
- **Microservice friendly**: Perfect for API-only deployments in microservice architectures
- **Container optimization**: Smaller Docker images for cloud deployments

## Testing
- Built and tested API-only mode on both AMD64 and ARM64 architectures
- Verified all API endpoints work correctly without frontend dependencies
- Confirmed database stubs work as expected in API-only mode
- Tested Docker builds with the new build arguments

## Breaking changes
None. This change is fully backward compatible. The default build still includes all frontend components.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,6 +3,7 @@ FROM crystallang/crystal:1.16.3-alpine AS builder
 RUN apk add --no-cache sqlite-static yaml-static
 
 ARG release
+ARG api_only
 
 WORKDIR /invidious
 COPY ./shard.yml ./shard.yml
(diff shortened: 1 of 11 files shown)


SamantazFox:
What you propose here is much more than an "API only" mode. It's a whole "no database" and "no users" mode. Be aware that the API-only mode is meant to be a clipious/freetube backend (primarily a cloud account server) rather than a pure content proxy (for scrapping, [yt-dlp](https://github.com/yt-dlp/yt-dlp) might be more adapted).

PR #5417, opened 2025-08-13
Persist caption settings

### LINKED ISSUE #5140: [Enhancement] Caption size configuration option
Right now, if you set a caption size and then refresh the page, it isn't saved.

This should probably be stored in the cookie

n/a

**Additional context**
Perhaps this is a bug?

p-himik:
Fixes #5140


unixfox:
Are you sure this works in the specific video.js version we are using? So video.js 7.12.1

p-himik:
Huh. There's probably some caching issue on my side.
Initially I put the option at the level above, outside of `html5`. And it didn't work. Adding it also to `html5` made it work, removing from the top level didn't change things - still worked.

Now I've updated the local repo (got just a single `1ae0f45b` commit that doesn't seem to be related to anything), and made the PR.

Just rechecked after the update - the changes don't work.
Redid all the steps, but this time with disabled browser cache - seems like indeed the option should go at the top level and not inside `html5`. I'll update the PR in a minute.

PR #5423, opened 2025-08-18
Videos: Patch missing streamingData sections from fallback (formats + adaptive); fix blank formatStreams (#5420)

### LINKED ISSUE #5420: [Bug] `/api/v1/videos/<video_id>` endpoint has `formatStreams` blank.
Seems like YouTube changes a little while ago.
Any other way to send URLs to clients that aren't blocked and without proxying the request?
May not be fixable anymore.
Image

naoNao89:
Fixes #5420

Summary
- Some videos return a player response where streamingData lacks usable stream URLs on the default client, leaving `formatStreams` blank in `/api/v1/videos/<video_id>`.
- Historically, only `adaptiveFormats` were patched from a fallback client. When `formats` was missing/empty, it stayed blank.
- This PR patches whichever sections are actually missing (`adaptiveFormats` and/or `formats`) from alternate YouTube clients (TvHtml5, WebMobile), but only when the fallback provides usable entries (either a direct `url` or a `signatureCipher`/`cipher`).

Changes
- Add helper utilities:
  - `has_usable_stream?` detects usable entries by `url` or `signatureCipher`/`cipher`.
  - `patch_streaming_data_if_missing!` mutates only missing sections of `streamingData` and returns which parts were patched.
- In `extract_video_info`:
  - Detect when either `adaptiveFormats` or `formats` is missing a usable URL.
  - Try fallback clients `{TvHtml5, WebMobile}`; on success, patch only the missing sections.
  - Add debug log on successful patch: `fallback_patched: client=<...> video=<...> patched_adaptive=<...> patched_formats=<...>`.
  - Continue to run `convert_url` for all stream URLs after patching.
- In video playback route, add a triage log when playback fails due to missing URL: `playback_404: no URL for id=<...> itag=<...> fmt=<count> adaptive=<count>`.

Tests
- New spec `spec/streaming_fallback_spec.cr` validates:
  - Patches `formats`/`adaptiveFormats` when primary is missing and fallback has usable entries.
  - Does not overwrite valid primary data (guarding against downgrades).
  - Handles fallback without `formats` gracefully.

Risk/compatibility
- Minimal: We only patch when the primary section lacks any usable URL, and only if the fallback provides usable entries; we never overwrite valid primary data.
- Performance: At most two additional player calls (TvHtml5, WebMobile), short‑circuiting on first patch.
- No public API changes.

Context
- Aligns with maintainer note that only DASH (adaptive) used to be problematic; YouTube’s current behavior sometimes leaves progressive `formats` without URLs as well. This generalizes the fallback patching accordingly.

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -542,14 +542,25 @@ private def convert_url(fmt)
 
     LOGGER.debug("convert_url: Decoding '#{cfr}'")
 
-    unsig = DECRYPT_FUNCTION.try &.decrypt_signature(cfr["s"])
+    # When using Invidious Companion, streaming URLs should already be usable.
+    # Skip inv-sig-helper based signature decryption in that case.
+    unsig = if CONFIG.invidious_companion.present?
+              nil
+            else
+              DECRYPT_FUNCTION.try &.decrypt_signature(cfr["s"])
+            end
     params[sp] = unsig if unsig
   else
     url = URI.parse(fmt["url"].as_s)
     params = url.query_params
   end
 
-  n = DECRYPT_FUNCTION.try &.decrypt_nsig(params["n"])
+  # Skip nsig decryption when using Invidious Companion
+  n = if CONFIG.invidious_companion.present?
+        nil
+      else
+        DECRYPT_FUNCTION.try &.decrypt_nsig(params["n"])
+      end
   params["n"] = n if n
 
   if token = CONFIG.po_token
--- a/src/invidious/yt_backend/youtube_api.cr
+++ b/src/invidious/yt_backend/youtube_api.cr
@@ -473,8 +473,12 @@ module YoutubeAPI
     } of String => String | Int64
 
     if {"WEB", "TVHTML5"}.any? { |s| client_config.name.starts_with? s }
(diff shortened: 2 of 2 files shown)


unixfox:
Hello,

Thanks for the PR.

 I just would like to point out that inv-sig-helper will soon be deprecated.

It will be replaced by Invidious companion, which handle the streaming_data logic here: https://github.com/iv-org/invidious-companion/blob/master/src/lib/helpers/youtubePlayerReq.ts

naoNao89:
@unixfox 
- When Invidious Companion is enabled, we now skip inv-sig-helper signature paths (sig/nsig and STS).
Player requests already route through the Companion, and parser fallbacks behave accordingly.
- No behavior change when Companion is disabled.

I also amended the commit message to reflect this change clearly:
‘fix(companion): skip inv-sig-helper when Invidious Companion is enabled’ (refs #5423).

unixfox:
Thank you. But inv-sig-helper will be removed and replaced by Invidious companion. So we won't merge this PR.

If there is any related issue in Invidious companion itself. Please create a bug report at https://github.com/iv-org/invidious-companion/issues or create a PR in the companion code.

Issue #5413, opened 2025-08-11
[Enhancement] Switch to stdlib logger for logs


syeopite:
Kemal's (and in turn Invidious') logger originated from a time where the Crystal standard library did not possess a logging library.

But it does now and it's much better than what Invidious currently uses in both performance and flexibility. 

For example the stdlib version won't perform the string interpolation for the message when the severity level doesn't allow it unlike Invidious's current logger.

And it also provides the APIs necessary to emit logs to multiple different sources at once, and to toggle and change the severity level for different sections of the code as needed.

Though for a first implementation a simple search and replace on Invidious's log calls to the stdlib block syntax plus the necessary boiler plate should be enough imo.

PR #5426, opened 2025-08-23
Switch to stdlib log

Fijxu:
Closes https://github.com/iv-org/invidious/issues/5413

- Kemal was upgraded to 1.7.2 (1.7.0 is the one that introduced integration with the stdlib Log)
- `CONFIG.output` works again (it didn't seem to work when I tested it)
- ~Some functions were encapsulated into modules (`YoutubeJSONParser`, `Parser`, `I18n`, and others) to keep the `source` (from where the Log function is being called) consistent for some functions.~
- ~A macro `::Log.forf` was added to add the function name to the source so we can prevent using `Log.info { "fetch_channel: something" }`, `Log.info { "fetch_channel: something2" }`, etc.~

How it looks and behaves: https://asciinema.org/a/WQXsdnxcV1Hu7fOufJeMEaovZ

It's slower (~1.06x times slower) than using the old `Invidious::LogHandler`, but this is mostly because of heavy use of colorization. Without colorization, `Log` is 1.12x times faster.

With the `Invidious::Logger.formatter` added in this PR, it's possible to use `Log.info &.emit("msg", error: "someerror", debuginfo: response.data)` or something like that which may be useful for better context in the logs.

### DIFF
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -94,7 +94,7 @@ class Config
   property feed_threads : Int32 = 1
   # Log file path or STDOUT
   property output : String = "STDOUT"
-  # Default log level, valid YAML values are ints and strings, see https://crystal-lang.org/api/master/CLog.Severity.html#enum-members
+  # Default log level, valid YAML values are ints and strings, see https://crystal-lang.org/api/1.17.1/Log/Severity.html#enum-members
   property log_level : Log::Severity = Log::Severity::Info
   # Enables colors in logs. Useful for debugging purposes
   property colorize_logs : Bool = false


src/invidious/helpers/macros.cr
```diff
@@ -70,3 +70,9 @@ macro haltf(env, status_code = 200, response = "")
   {{env}}.response.close
   return
 end
+
+class Log
+  macro forf
+    Log.for({{@def.name.stringify}})
+  end
```
syeopite: Each `Log.for` call will eventually involve locking and unlocking a Mutex which adds considerable overhead 

https://github.com/crystal-lang/crystal/blob/19be240d1ac7f2887d2e69c45b8d4c311ae12e68/src/log/builder.cr#L69-L84

This is fine for constants since it is only done once but doing so for each message in a function like `fetch_channel` will hurt performance by a lot.

So I think adding prefixes like `fetch_channel: ...` is a necessity for now

src/invidious/helpers/logger.cr
```diff
@@ -1,72 +1,46 @@
...
-    elapsed_text = elapsed_text(elapsed_time)
-
-    # Default: full path with parameters
-    requested_url = context.request.resource
 
-    # Try not to log search queries passed as GET parameters during normal use
-    # (They will still be logged if log level is 'Debug' or 'Trace')
-    if @level > LogLevel::Debug && (
-         requested_url.downcase.includes?("search") || requested_url.downcase.includes?("q=")
-       )
```
syeopite: Although filtering out search queries still leaves a lot of other user usage info within the logs it still provides at least some basic level of privacy so imo I think it should be preserved (and maybe expanded in some future PR). Wdyt?

Since Kemal logs requests in the new `Kemal::RequestLogHandler` it should be pretty easy to bring back by bringing this logic into an override of that handler.

src/invidious/helpers/logger.cr
```diff
@@ -1,72 +1,46 @@
...
-      # Log only the path
-      requested_url = context.request.path
+    formatter = ::Log::Formatter.new do |entry, io|
```
syeopite, line 10: **Just a comment**

If we want to squeeze some extra performance calling a function is slightly faster (1.03x on my machine) than a proc so subclassing `Log::Formatter` or `Log::StaticFormatter` will result in slightly faster performance than using the proc constructor.

I don't know if such a small gain will be worth it though

src/invidious/helpers/signatures.cr
```diff
@@ -2,6 +2,8 @@ require "http/params"
 require "./sig_helper"
 
 class Invidious::DecryptFunction
+  Log = ::Log.for(self).for("Signature")
```
syeopite: Wdyt about `SignatureDecryption` instead? `Signature` by itself feels a bit vague imo

src/invidious/jobs/clear_expired_items_job.cr
```diff
@@ -1,11 +1,13 @@
 class Invidious::Jobs::ClearExpiredItemsJob < Invidious::Jobs::BaseJob
+  Log = ::Log.for(self)
```
syeopite: Since all the jobs inherit from `Invidious::Jobs::BaseJob` their specific loggers could actually just be defined under the `inherited` hook instead.

Fijxu: https://github.com/iv-org/invidious/pull/5426/commits/2a1a4f77984bcaa5a60a8dcbf02ac44057614cc5

src/invidious/helpers/signatures.cr
Fijxu: https://github.com/iv-org/invidious/pull/5426/commits/ad486ade39feedee629b08eaf4d0ce7ee841d3cd

src/invidious/helpers/logger.cr
Fijxu: Thanks, I didn't notice I removed this feature, sorry!

https://github.com/iv-org/invidious/pull/5426/commits/f8d56d77c1f2953c72c1781e431e22e969b9ca54

src/invidious/helpers/macros.cr
Fijxu: https://github.com/iv-org/invidious/pull/5426/commits/e354cf2502fd6e6808ee2ff5790afe1ff8059247

Fijxu:
>Can be split off into smaller PRs that can be merged after this main migration. Having it all in one adds a bunch of noise to the primary change of switching to the stdlib logger.

Done, that should be better ;)

src/invidious/config.cr
```diff
@@ -73,6 +73,7 @@ end
 
 class Config
   include YAML::Serializable
+  CLog = ::Log.for(self)
```
syeopite, line 76: For consistency can you also set this as `Log`? A namespace collision can be prevented by changing the `Log::Severity` below to `::Log::Severity`

Fijxu: I did that because I didn't know how to fix this error 😅:

```
...
 >  151 |                 # Sort types to avoid parsing nulls and numbers as strings
 >  152 |                 
 >  153 |                 [Log::Severity].each do |ivar_type|
 >  154 |                     if !success
 >  155 |                         begin
 >  156 |                             config.log_level = ivar_type.from_yaml(env_value)
 >  157 |                             success = true
 >  158 |                         rescue
 >  159 |                             # nop
 >  160 |                         end
 >  161 |                     end
 >  162 |                 end
 ...
 
 Error: undefined constant Log::Severity
 ```
 
 It should be `::Log::Severity` but is a passed as `Log::Severity` in the macro.

File: src/invidious/helpers/logger.cr [hunk 7]
```diff
@@ -1,68 +1,71 @@
...
+      io << (use_color ? source.colorize(:dark_gray) : source) << ": " if !source.empty?
+      io << message
+      if !data.empty?
+        io << " "
+        data.each do |dat|
+          io << (use_color ? dat[0].to_s.colorize(:light_cyan) : dat[0].to_s)
+          io << "="
+          io << dat[1].to_s
+        end
+      end
```
SamantazFox, line 28: I'd suggest to use the IO when using Colorize so that no intermediary string is created. The code is a bit convoluted due to the way `Colorize` is implemented in the standard library.

The `use_color` condition can also be removed, as `Colorize` keeps track of the enabled state (see lines 7/8 of this same file).

Finally, `.each` does nothing when the array is empty, thus making the `if` redundant. Note that in my suggestion the `key=value` pairs are separated by a single space, but a newline or another character might be more appropriate.

```suggestion
      entry.timestamp.colorize.dark_gray.to_s(io)
      io << " "
      colorize_severity(entry.severity).to_s(io)
      io << " "

      if !entry.source.empty?
        entry.source.colorize.dark_gray.to_s(io)
        io << ": "
      end

      io << entry.message

      data.each do |dat|
        io << " "
        dat[0].colorize.light_cyan.to_s(io)
        io << "=" << dat[1]
      end
```

PR #5442, opened 2025-09-02
CI: fix wrong if statement for build-docker job

Fijxu:
I saw the Dockerfile was always changed to `Dockerfile.arm64` on the AMD64 build-docker while watching at https://github.com/iv-org/invidious/pull/5441 because of a syntax error 😅

Now works fine and it's greyed out: https://github.com/Fijxu/invidious/actions/runs/17406402847/job/49411827393


unixfox:
True saw the same. Thanks

PR #5446, opened 2025-09-05
Fix 0 view count on related videos section

### LINKED ISSUE #5403: [Bug] 0 views on all related videos on any instances
**Describe the bug**

The view count of related videos are all 0 on any public instances and my self-hosted instance.

**Steps to Reproduce**
1. Enable related videos
2. Go to any videos
3. The issue is shown

**Logs**

**Screenshots**

Image

**Additional context**

Seems this issue occurred previously (#2830).

shiny-comic:
This PR fixes #5403.

### DIFF
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -354,9 +354,8 @@ we're going to need to do it here in order to allow for translations.
 
                                 <div class="pure-u-10-24" style="text-align:right">
                                     <b class="width:100%"><%=
-                                        views = rv["view_count"]?.try &.to_i?
-                                        views ||= rv["view_count_short"]?.try { |x| short_text_to_number(x) }
-                                        translate_count(locale, "generic_views_count", views || 0, NumberFormatting::Short)
+                                        views = short_text_to_number(rv["short_view_count"]? || "0")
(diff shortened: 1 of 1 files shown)


Fijxu (approved):
Tested, works fine!
Youtube doesn't seem to return `viewCountText` and it only returns `shortViewCountText` now.

Image

Edit: Since `viewCountText` is not returned anymore, we should add a comment about it above this variable (or just delete the variable) but I'm not sure what compact renderer is:

https://github.com/iv-org/invidious/blob/89c8b1b901062c729370370116aa9127a39cd214/src/invidious/videos/parser.cr#L28-L31

@SamantazFox There is any background about that line that was written 3 years ago?

syeopite:
Its either for an old Innertube structure which never got removed in Invidious or just how the view count is stored in the Innertube responses of some other InnerTube client different than the ones that Invidious currently uses

SamantazFox:
> @SamantazFox There is any background about that line that was written 3 years ago?

It was introduced in https://github.com/iv-org/invidious/pull/2856, when Youtube switched to the `compactVideoRenderer` for related videos (in the `next` endpoint response).

If that's not relevant anymore, that can be removed.

PR #5450, opened 2025-09-07
Store `preferences` in a variable when reused and rename `prefs` to `preferences`

Fijxu:
A little code cleanup on places where `preferences` is used more than one time.

### DIFF
--- a/src/invidious/routes/feeds.cr
+++ b/src/invidious/routes/feeds.cr
@@ -43,13 +43,14 @@ module Invidious::Routes::Feeds
   end
 
   def self.trending(env)
-    locale = env.get("preferences").as(Preferences).locale
+    preferences = env.get("preferences").as(Preferences)
+    locale = preferences.locale
 
     trending_type = env.params.query["type"]?
     trending_type ||= "Default"
 
     region = env.params.query["region"]?
-    region ||= env.get("preferences").as(Preferences).region
+    region ||= preferences.region
 
     begin
       trending, plid = fetch_trending(trending_type, region, locale)
--- a/src/invidious/routes/preferences.cr
+++ b/src/invidious/routes/preferences.cr
@@ -2,11 +2,10 @@
 
 module Invidious::Routes::PreferencesRoute
   def self.show(env)
-    locale = env.get("preferences").as(Preferences).locale
+    preferences = env.get("preferences").as(Preferences)
+    locale = preferences.locale
 
     referer = get_referer(env)
-
-    preferences = env.get("preferences").as(Preferences)
 
     templated "user/preferences"
   end
--- a/src/invidious/routes/watch.cr
+++ b/src/invidious/routes/watch.cr
@@ -2,7 +2,8 @@
 
 module Invidious::Routes::Watch
   def self.handle(env)
-    locale = env.get("preferences").as(Preferences).locale
+    preferences = env.get("preferences").as(Preferences)
+    locale = preferences.locale
     region = env.params.query["region"]?
 
     if env.params.query.to_s.includes?("%20") || env.params.query.to_s.includes?("+")
@@ -37,8 +38,6 @@ module Invidious::Routes::Watch
 
     nojs ||= "0"
     nojs = nojs == "1"
-
-    preferences = env.get("preferences").as(Preferences)
 
     user = env.get?("user").try &.as(User)
     if user
--- a/src/invidious/views/template.ecr
+++ b/src/invidious/views/template.ecr
@@ -1,6 +1,7 @@
 <%
-  locale = env.get("preferences").as(Preferences).locale
-  dark_mode = env.get("preferences").as(Preferences).dark_mode
+  preferences = env.get("preferences").as(Preferences)
+  locale = preferences.locale
+  dark_mode = preferences.dark_mode
 %>
 <!DOCTYPE html>
 <html lang="<%= locale %>">
--- a/src/invidious/frontend/misc.cr
+++ b/src/invidious/frontend/misc.cr
@@ -2,9 +2,9 @@ module Invidious::Frontend::Misc
   extend self
 
   def redirect_url(env : HTTP::Server::Context)
-    prefs = env.get("preferences").as(Preferences)
+    preferences = env.get("preferences").as(Preferences)
 
-    if prefs.automatic_instance_redirect
+    if preferences.automatic_instance_redirect
       current_page = env.get?("current_page").as(String)
       return "/redirect?referer=#{current_page}"
     else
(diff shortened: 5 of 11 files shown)


src/invidious/views/template.ecr
```diff
@@ -1,6 +1,7 @@
 <%
-  locale = env.get("preferences").as(Preferences).locale
-  dark_mode = env.get("preferences").as(Preferences).dark_mode
+  preferences = env.get("preferences").as(Preferences)
```
syeopite, line 2:
```suggestion
  preferences ||= env.get("preferences").as(Preferences)
```

`preferences` should already be available on most routes so I think we can just do something like this

Fijxu:
```
In src/invidious.cr:224:20               

 224 | Invidious::Routing.register_all
                          ^-----------
Error: instantiating 'Invidious::Routing.register_all()'


There was a problem expanding macro 'macro_140478978536432'

Called macro defined in src/invidious/routing.cr:19:5

 19 | {% unless flag?(:api_only) %}

Which expanded to:

 >  1 | 
 >  2 |       get "/", Routes::Misc, :home
 >  3 |       get "/privacy", Routes::Misc, :privacy
 >  4 |       get "/licenses", Routes::Misc, :licenses
 >  5 |       get "/redirect", Routes::Misc, :cross_instance_redirect
 >  6 | 
 >  7 |       self.register_channel_routes
 >  8 |       self.register_watch_routes
 >  9 | 
 > 10 |       self.register_iv_playlist_routes
 > 11 |       self.register_yt_playlist_routes
 > 12 | 
 > 13 |       self.register_search_routes
 > 14 | 
 > 15 |       self.register_user_routes
 > 16 |       self.register_feed_routes
 > 17 | 
 > 18 |       # Support push notifications via PubSubHubbub
 > 19 |       get "/feed/webhook/:token", Routes::Feeds, :push_notifications_get
 > 20 |       post "/feed/webhook/:token", Routes::Feeds, :push_notifications_post
 > 21 | 
 > 22 |       if CONFIG.enable_user_notifications
 > 23 |         get "/modify_notifications", Routes::Notifications, :modify
 > 24 |       end
 > 25 |     
Error: expanding macro


There was a problem expanding macro 'get'

Called macro defined in macro 'macro_140479049802448'

 3 | macro get(path, controller, method = :handle)

Which expanded to:

 > 1 |       unless Kemal::Utils.path_starts_with_slash?("/privacy")
 > 2 |         raise Kemal::Exceptions::InvalidPathStartException.new("get", "/privacy")
 > 3 |       end
 > 4 | 
 > 5 |       Kemal::RouteHandler::INSTANCE.add_route("GET", "/privacy") do |env|
 > 6 |         Routes::Misc.privacy(env)
 > 7 |       end
 > 8 |     
Error: instantiating 'Invidious::Routes::Misc.privacy(HTTP::Server::Context+)'


In src/invidious/views/template.ecr:2:18

 2 | preferences ||= env.get("preferences").as(Preferences)
                    ^
Error: '||=' before definition of 'preferences'
```

Doesn't seem to be available on most routes

src/invidious/routes/feeds.cr
```diff
@@ -43,13 +43,14 @@ module Invidious::Routes::Feeds
   end
 
   def self.trending(env)
-    locale = env.get("preferences").as(Preferences).locale
+    preferences = env.get("preferences").as(Preferences)
+    locale = preferences.locale
```
SamantazFox, line 47: Given that `locale = preferences.locale` is already present in the template, there is no need for a separate variable in the route handlers. Note that the `embed` handler is the only one that doesn't use the main template.

src/invidious/routes/feeds.cr
```diff
@@ -43,13 +43,14 @@ module Invidious::Routes::Feeds
...
 
     trending_type = env.params.query["type"]?
     trending_type ||= "Default"
 
     region = env.params.query["region"]?
-    region ||= env.get("preferences").as(Preferences).region
+    region ||= preferences.region
```
SamantazFox, line 53: You can keep that as-is if it's the only place in the handler that uses the preferences. Creating a separate variable only adds noise imo.

src/invidious/routes/preferences.cr
```diff
@@ -2,12 +2,11 @@
 
 module Invidious::Routes::PreferencesRoute
   def self.show(env)
-    locale = env.get("preferences").as(Preferences).locale
+    preferences = env.get("preferences").as(Preferences)
+    locale = preferences.locale
```
SamantazFox, line 6: Here you should be able to remove `preferences` and `locale`, as the required variables are defined in the template and are of no use in the route handler.

src/invidious/views/template.ecr
SamantazFox: Yeah, a lot of route handlers don't define a `preferences` variable, so that's why it doesn't work.

src/invidious/routes/watch.cr
```diff
@@ -2,7 +2,8 @@
 
 module Invidious::Routes::Watch
   def self.handle(env)
-    locale = env.get("preferences").as(Preferences).locale
+    preferences = env.get("preferences").as(Preferences)
+    locale = preferences.locale
```
SamantazFox, line 6: Similarly: no need to add a `locale` variable if it's not used in the handler.

SamantazFox (changes requested):

src/invidious/routes/watch.cr
Fijxu: It's used on this handler:

```
In src/invidious/routes/watch.cr:88:79

 88 | comment_html = JSON.parse(Comments.fetch_youtube(id, nil, "html", locale, preferences.thin_mode, region))["contentHtml"]
                                                                        ^-----
Error: undefined local variable or method 'locale' for Invidious::Routes::Watch:Module
```

src/invidious/routes/preferences.cr
Fijxu: `preferences` and `locale` don't seem to be able to be inherited from the template:

```
In src/invidious/views/user/preferences.ecr:2:22

 2 | <title><%= translate(locale, "Preferences") %> - Invidious</title>
                          ^-----
Error: undefined local variable or method 'locale' for Invidious::Routes::PreferencesRoute:Module
```

src/invidious/routes/feeds.cr
Fijxu:
```
In src/invidious/routes/feeds.cr:53:62

 53 | trending, plid = fetch_trending(trending_type, region, locale)
                                                             ^-----
Error: undefined local variable or method 'locale' for Invidious::Routes::Feeds:Module
```

Is needed

SamantazFox: Oh, I missed that line ^^> nevermind then

src/invidious/routes/preferences.cr
SamantazFox: Uh, weird. Maybe that's because we don't use kilt anymore?

Issue #5443, opened 2025-09-03
[Bug] Wrong date on video page


hans-aa:
**Description**
On the video page of some _old_ videos it shows todays date:
"Shared September 3, 2025"

**Steps to Reproduce**
1. Go to https://inv.nadeko.net/search?q=+Jeremy+Clarkson+on+the+Volkswagen+Jetta+%7C+Top+Gear
2. "Shared 13 years ago"
3. Go to https://inv.nadeko.net/watch?v=Zm7nlMHess8
4. " Shared September 3, 2025" - todays date, yesterday it was "September 2, 2025"

**Reproduced on**
 - Linux and Windows
 - FF, LibreWolf and Chrome
 - a couple invidious instances

Some affected Videos:
https://inv.nadeko.net/watch?v=JzClLWL-Eys
https://inv.nadeko.net/watch?v=Zm7nlMHess8
https://inv.nadeko.net/watch?v=LSYD4a3RzW4
https://inv.nadeko.net/watch?v=ma6syI9hecQ

brolykwotlakosta9:
I used this fix.
[4854a2cca1b1b524a23c0ffc.zip](https://github.com/user-attachments/files/22115033/4854a2cca1b1b524a23c0ffc.zip)

Fijxu:
It also happens on invidious.nerdvpn.de, so this is not a problem exclusive to my Instance (inv.nadeko.net). I'll take some time to fix it later. 

Thanks

PR #5453, opened 2025-09-07
Prevent player microformat from being overwritten by the next microformat

Fijxu:
Closes https://github.com/iv-org/invidious/issues/5443

The player microformat is what we need to get the published date, premiere timestamp, allowed regions and more information of the video.

Youtube introduced a new `microformat.microformatDataRenderer` in the next endpoint which overwrote the player microformat `microformat.playerMicroformatRenderer` when merged

Here is a small piece of code to see how the player microformat gets overwritten: https://gist.github.com/Fijxu/79ddab062a572bc1b1bf7cf48eefe72f

### DIFF
--- a/src/invidious/videos/parser.cr
+++ b/src/invidious/videos/parser.cr
@@ -102,6 +102,9 @@ def extract_video_info(video_id : String)
   # Don't fetch the next endpoint if the video is unavailable.
   if {"OK", "LIVE_STREAM_OFFLINE", "LOGIN_REQUIRED"}.any?(playability_status)
     next_response = YoutubeAPI.next({"videoId": video_id, "params": ""})
+    # Remove the microformat returned by the /next endpoint on some videos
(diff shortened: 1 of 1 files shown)


src/invidious/videos/parser.cr
```diff
@@ -102,6 +102,9 @@ def extract_video_info(video_id : String)
   # Don't fetch the next endpoint if the video is unavailable.
   if {"OK", "LIVE_STREAM_OFFLINE", "LOGIN_REQUIRED"}.any?(playability_status)
     next_response = YoutubeAPI.next({"videoId": video_id, "params": ""})
+    # Remove the microformat returned by the /next endpoint on some videos
+    # to prevent player_response microformat from being overwritten.
+    next_response.reject!("microformat")
```
syeopite:
```suggestion
    next_response.delete("microformat")
```

PR #5473, opened 2025-09-14
Dockerfile: Switch to 84codes crystal compiler container image

### LINKED ISSUE #5456: [Enhancement] Recompile Crystal's garbage collector (bdwgc) with `--enable-large-config`
~~Who knew that all of Invidious' problems can be resolved by recompiling various upstream dependencies~~

Apparently the default compilation flags for Bohem GC on most distros creates a GC that isn't set to handle "large" (> 100 mb) memory heaps well due to missing the `--enable-large-config` flag which results in significantly worse performance.

See this thread for more information

https://forum.crystal-lang.org/t/use-docker-cross-build-crystal-compiler-without-install-bdwgc-successful-but-why-shouldnt-bdwgc-be-a-required-dependency/6332/8

Fijxu:
Closes https://github.com/iv-org/invidious/issues/5456

The 84codes Crystal container image builds libgc (bdwgc) with `--enable-large-config`: https://github.com/84codes/crystal-packages/blob/b321bb4358b0140a63573d9d05ccf2f06c5ae040/alpine/Dockerfile#L13-L22

I tested it locally and I'm now going to test it on my fork of Invidious https://git.nadeko.net/Fijxu/invidious/commit/b38791e7ca6b7549e3a894f26f43cb77660c798e

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 ARG OPENSSL_VERSION='3.6.2'
 ARG OPENSSL_SHA256='aaf51a1fe064384f811daeaeb4ec4dce7340ec8bd893027eee676af31e83a04f'
 
-FROM crystallang/crystal:1.20.2-alpine AS dependabot-crystal
+FROM 84codes/crystal:1.20.2-alpine AS dependabot-crystal
 
 # We compile openssl ourselves due to a memory leak in how crystal interacts
 # with openssl


unixfox:
Ah so this image fixes the memory leak?

Fijxu:
> Ah so this image fixes the memory leak?

No, this is unrelated to the memory leak. It just improves GC performance and for now, my Invidious processes seem to use less memory. 

2fd67c34-000f-40ed-9d5b-04f1def58b93

At 00:00 is when I changed my instance to the 84codes container image to build Invidious.

Fijxu:
After approximately 1 day and 12 hours I can confirm that the 84codes container image which uses `--enable-large-config` for `bdwgc` has a more stable memory usage when combining it with https://github.com/iv-org/invidious/pull/5441. There is no memory limits (`mem_limit`) on my docker-compose file and I have around 8GB of memory free that Invidious could use, but the memory usage is stable ;):

image

unixfox:
https://github.com/iv-org/invidious/pull/5441 got merged instead of this one.

Fijxu:
#5441 is for memory leak

This one is for bdwgc compiled with `--enable-large-config` and not related to the memory leak. Is for better memory management in the GC

unixfox:
@Fijxu ok. I saw that they provide ARM64: https://hub.docker.com/r/84codes/crystal/tags. Could we unify our Dockerfile for both x86 and ARM64?

Fijxu:
@unixfox Since now we just use a single Docker file, we can just replace the current `crystallang` OCI by the `84codes` OCI and it should all work fine.

unixfox:
> @unixfox Since now we just use a single Docker file, we can just replace the current `crystallang` OCI by the `84codes` OCI and it should all work fine.

Sure go ahead :)

unixfox (approved):

PR #5480, opened 2025-09-18
Add Livestreams to trending page

Fijxu:
Taken from https://github.com/TeamNewPipe/NewPipeExtractor/blob/b469dcacf07978feb77ef68ff4d1753faf0476b4/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/kiosk/YoutubeLiveExtractor.java

Livestream trending page for US region:

image

Switching to another region shows the Livestream trending page for that region.

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -408,6 +408,7 @@
     "Default": "Default",
     "Music": "Music",
     "Gaming": "Gaming",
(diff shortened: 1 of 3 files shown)


SamantazFox (approved):
Looks good to me!

PR #5483, opened 2025-09-21
Create gl.json

ninjum:
Created the Galician language translation in case it can be added (gl)


unixfox:
Thank you, but you need to submit your translations on weblate: https://hosted.weblate.org/engage/invidious/

That's where we manage translations.

PR #5491, opened 2025-09-24
Fixed broken companion hyperlink

ndsvw:
I saw this link in my logs and it did not work. The PR fixes it.


Fijxu:
Good catch! Thank you

Fijxu (approved):

unixfox:
Please don't point to a certain section and instead to https://docs.invidious.io/installation/

People will see the warning anyway.

But good catch.

Issue #5454, opened 2025-09-08
[Bug] Mix playlists render as -1 videos


Fijxu:
Playlists cataloged as a Mix (which are playlists generated by Youtube) have a video count of **-1 video**. It should display `Mix` instead of the quantity of videos the playlist has.

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/1f95c7c4-ee78-4682-8046-3b75021ae341" />

Lines that need to be fixed:

https://github.com/iv-org/invidious/blob/89c8b1b901062c729370370116aa9127a39cd214/src/invidious/yt_backend/extractors.cr#L673-L679

JSON looks like this:

```
"thumbnailOverlayBadgeViewModel": {
	"thumbnailBadges": [
		{
			"thumbnailBadgeViewModel": {
				"icon": {
					"sources": [
						{
							"clientResource": {
								"imageName": "MIX"
							}
						}
					]
				},
				"text": "Mix", # THIS LINE NEEDS TO BE CONSIDERED
				"badgeStyle": "THUMBNAIL_OVERLAY_BADGE_STYLE_DEFAULT"
			}
		}
	],
	"position": "THUMBNAIL_OVERLAY_BADGE_POSITION_BOTTOM_END"
}
```

PR #5511, opened 2025-10-08
Fix playlists video count for playlists cataloged as "Mix" and "1 video"

Fijxu:
Fixes https://github.com/iv-org/invidious/issues/5454

It also fixes playlists with 1 video being shown as `-1 video`


Fijxu:
I think this can be improved, specially https://github.com/iv-org/invidious/pull/5511/files#diff-8660a2d72ed46b953aec736a988a47aeb1a7c598a6f52f17ab0e25beaeb3fc12R151-R154 

So I'll leave it as a draft for now.

PR #5513, opened 2025-10-14
Formatted prompt to input time at login more logically

Morty0Smith:
In this pull request, I changed the prompt for the task at signup to display "hh:mm:ss" instead of "h:mm:ss", because it more accurately tells the user, what to input. The actual formating is expected to be hh:mm:ss in the end anyways, because if you put in 8:22:03, you get an error message, in this case you need to input 08:22:03 to get through.


unixfox:
I think you need to do the language changes on weblate.

If i'm not mistaken @SamantazFox

Morty0Smith:
> I think you need to do the language changes on weblate.
> 
> If i'm not mistaken @SamantazFox 

That makes sense, but is there a way to bulk edit a change like that in weblate?

unixfox:
> > I think you need to do the language changes on weblate.
> > If i'm not mistaken @SamantazFox
> 
> That makes sense, but is there a way to bulk edit a change like that in weblate?

I'm sorry but I don't know. I do not use Weblate.

SamantazFox:
> That makes sense, but is there a way to bulk edit a change like that in weblate?

Well, yes, and no. The answer is not very straightforward, so let me explain why.

In this PR, you're doing two things: changing the **"key"** string, and changing the **translated** string.

## Difference between key and translated strings

The **keys** are the left-most strings of the JSON files, and is what the `translate()` function takes. It **uniquely identifies translation strings**.

The **translated** strings are the right-most strings of the same files and **contain the actual text** that will be **seen by the user**.

For legacy reasons, the key and translated string were the same in english, but we started to clean that up to reduce confusion. As an example, here are some lines that were updated to better differentiate the two:
https://github.com/iv-org/invidious/blob/5cfe294063c9317928d8da3387004e3eaddc991a/locales/en-US.json#L72-L80


## How weblate detects changes

The `en_US` language (and by extension the JSON file) is the **"source"**, meaning that whenever it is changed, Weblate will detect that and update the list of translations:
* if you remove a "key" from that file, Weblate will remove it in all languages
* if you update one of the translated strings (aka **"source" strings**), Weblate will mark the associated translated strings in other languages as **"unfinished"** (= requiring a new translation).
* if you add a new "key", Weblate will mark them as **"untranslated"**, unless the same key was also added to another language.

Note that changing a "key" is the same as deleting the line and adding a new one.

## Answer to the initial question

Now, if you only want to change the translated (source) string, that can now be done directly on Weblate, by editing the strings in the "english" language. As stated above, weblate will automatically mark strings in other languages as "unfinished" and primpt users for a translation.

However, if you want to change the identifier ("key"), it can only be done here, because both the code and the JSON files must be changed.

When you change both at the same time, make sure to only update the JSON files **where you can ALSO update the translated strings**. If you don't update them, Weblate won't be able to detect the change and thus mark them as **"translated"**. For instance, if you speak english and german, only update the `en_US.json` and `de.json`, and Weblate will take care of the rest.

PR #5528, opened 2025-11-07
feat: mark video as watched after a certain duration

### LINKED ISSUE #3794: [Enhancement] Mark as watched after certain
Currently, videos are marked as watched as soon as playback is started. If one decides to pause a video and continue watching it later, it has vanished from the subscription feed.

I would like to have the ability to configure a certain time when a video is marked as watched (preferably a configurable value like "mark as watched after __%").

I realise this probably requires JS to implement. It could therefore be opt-in (e.g. a checkbox "mark as watched immediately" that must be disabled).

Somewhat related to #1713. This has been brought up before in #607.

goestav:
This pull request marks videos as watched after watching them for a given amount of time if JavaScript is enabled and also indirectly fixes a bug that prevents YouTube/Invidious links from updating when seeking backwards for the video.

Closes #3794

### DIFF
--- a/assets/js/handlers.js
+++ b/assets/js/handlers.js
@@ -146,4 +146,15 @@
             event.preventDefault();
         }
     });
+
+    // Handle the video watched state with JavaScript (if enabled)
+    const anchorElements = document.querySelectorAll('a[href^="/watch"]');
+
+    for (const anchorElement of anchorElements) {
+        const url = new URL(anchorElement.href);
+
+        url.searchParams.set("mark_watched", "0");
+        
+        anchorElement.href = url.href;
+    }
 })();
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -127,13 +127,37 @@ function addCurrentTimeToURL(url, base, param = 't') {
  */
 var timeupdate_last_ts = 5;
 
+/** 
+ * Global variable to save the total video time watched (in seconds).
+ */
+let time_watched = 0;
+
 /**
- * Callback that updates the timestamp on all external links
+ * The duration of a short video (in seconds).
+ * This value is used to determine whether the video should be watched fully before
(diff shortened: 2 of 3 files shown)


unixfox:
```
It also removes the previous server-side logic that marked videos as watched upon navigation to a video page
```

I don't think that's a good thing because we still want to support non js users as long as possible.

goestav:
Apologies, that was supposed to be ignored only if this feature is enabled.
```diff
- if watched && preferences.watch_history
+ if watched && preferences.watch_history && !preferences.mark_watched_after_duration
  Invidious::Database::Users.mark_watched(user.as(User), id)
end
```
It should probably be mentioned under the checkbox in the preference page that the feature requires JavaScript though.
What do you think?

SamantazFox:
> ```
> It also removes the previous server-side logic that marked videos as watched upon navigation to a video page
> ```
> 
> I don't think that's a good thing because we still want to support non js users as long as possible.

We can do that, without the need for a config option.

The simplest way would be to add an extra URL query parameter (like `mark_watched=0`)  using JS to all links that go the the `/watch` page when clicked, so that the backend knows that is shouldn't mark a video as watched when loading the page.

Then the front-end JS will call the right endpoint after the video passes the 5s mark.

For non-JS users, it has no impact, as if the query parameter is absent, it continues to do as it did in the past.

goestav:
I think this PR is ready now, I implemented @SamantazFox's suggestion.
I can rebase if necessary.

goestav:
@SamantazFox or @unixfox could either of you review/merge this PR please?

Fijxu:
Hi, resolve the conflicts please. Thanks!

goestav:
The last commit takes the watch history preference into account.

Issue #5397, opened 2025-07-26
Missing hash key: "content" - trending removed


ghost:
Hi,
I started self-hosting Invidious, but when searching for videos, I get no results.
The only error I could find is when I go on the trending page and get this message:

Title: `Missing hash key: "content" (KeyError)`
Date: `2025-07-18T22:25:40Z`
Route: `/feed/trending`
Version: `2025.05.28-df8839d @ master`

<details>
<summary>Backtrace</summary>
<p>
   
```
Missing hash key: "content" (KeyError)
  from /usr/share/crystal/src/hash.cr:1198:9 in '??'
  from src/invidious/yt_backend/extractors.cr:839:17 in 'process'
  from src/invidious/yt_backend/extractors.cr:1061:20 in 'extract_items'
  from src/invidious/trending.cr:21:14 in 'trending'
  from src/invidious/helpers/handlers.cr:31:37 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:32:7 in 'call_next'
  from lib/kemal/src/kemal/filter_handler.cr:22:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:32:7 in 'call_next'
  from src/invidious/helpers/handlers.cr:95:12 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:32:7 in 'call_next'
  from src/ext/kemal_static_file_handler.cr:106:14 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call'
  from /usr/share/crystal/src/http/server/request_processor.cr:51:11 in 'handle_client'
  from /usr/share/crystal/src/fiber.cr:170:11 in 'run'
  from ???
```
</p>
</details>

syeopite:
This is caused by YouTube removing the aggregated trending page

ghost:
> This is caused by YouTube removing the aggregated trending page

But when I access other instances, everything works, including search and trending page?

[...]

PR #5537, opened 2025-11-18
Remove trending page from Invidious

Fijxu:
Closes https://github.com/iv-org/invidious/issues/5397

This completely strips the Trending page from Invidious since it's not available on Youtube anymore. Is not worth trying to find workarounds to keep the Trending page working as it will eventually break later.

### DIFF
--- a/src/invidious/views/components/feed_menu.ecr
+++ b/src/invidious/views/components/feed_menu.ecr
@@ -1,5 +1,11 @@
 <div class="feed-menu">
     <% feed_menu = env.get("preferences").as(Preferences).feed_menu.dup %>
+    <% 
+      # Remove the Trending page from the menu for users with the Trending
+      # item inside their PREFS cookie or in their user preferences saved
+      # in the database (if registered user)
+      # https://github.com/iv-org/invidious/issues/5397
+    %>
     <% feed_menu.reject! {|item| "Trending" == item} %>
     <% if !env.get?("user") %>
         <% feed_menu.reject! {|item| {"Subscriptions", "Playlists"}.includes? item} %>


src/invidious/views/components/feed_menu.ecr
```diff
@@ -1,5 +1,12 @@
 <div class="feed-menu">
     <% feed_menu = env.get("preferences").as(Preferences).feed_menu.dup %>
+    <% 
+      # Remove the Trending page from the menu for users with the Trending
+      # item inside their PREFS cookie or in their user preferences saved
+      # in the database (if registered user)
+      # https://github.com/iv-org/invidious/issues/5397
+    %>
+    <% feed_menu.reject! {|item| "Trending" == item} %>
```
Fijxu, line 9: I'm not sure if this is a good way to remove the Trending page from the feed menu. Registered users and unregistered users with the PREFS cookie set with `Trending` inside their `feed_menu` preferences will get the Trending option in their Feed menu, but since Trending is removed, is going to be a dead option that does nothing. So I did this to basically strip Trending from the feed menu if it's available on the user preferences.

Registered users and unregistered users will have to save their preferences again in order to remove `Trending` from their `feed_menu` preferences.

unixfox:
Newpipe can still show a trending page, it's by default the livestream trending page.

Maybe we could introduce the same thing? Allow the ability to select which trending tab the user wants.

Fijxu:
> Maybe we could introduce the same thing? Allow the ability to select which trending tab the user wants.

Yes is possible, Livestream trending page is the only one that works of all trending pages.

For music, we should use https://charts.youtube.com/ instead. That is what NewPipe uses.

And gaming can be fixed: https://www.youtube.com/gaming/trending

So you are fine with just leaving the Livestream page in trending for now?

bbsixzz:
Don't forget the Trending Trailers chart please.

It's by country now though…

https://charts.youtube.com/charts/TrendingTrailers/us

src/invidious/views/components/feed_menu.ecr
SamantazFox: Sounds good to me!

src/invidious/config.cr
```diff
@@ -38,7 +38,7 @@ struct ConfigPreferences
   property quality : String = "dash"
   property quality_dash : String = "auto"
   property default_home : String? = "Popular"
-  property feed_menu : Array(String) = ["Popular", "Trending", "Subscriptions", "Playlists"]
+  property feed_menu : Array(String) = ["Popular", "Subscriptions", "Playlists"]
```
SamantazFox, line 41: Do you think that we should log a warning if `feed_menu` contains `Trending` in the server config?

src/invidious/views/feeds/trending.ecr
```diff
@@ -1,49 +0,0 @@
-<% content_for "header" do %>
-<meta name="description" content="<%= translate(locale, "An alternative front-end to YouTube") %>">
-<title>
-    <% if env.get("preferences").as(Preferences).default_home != "Trending" %>
-        <%= translate(locale, "Trending") %> - Invidious
```
SamantazFox, line 5: Please remove `"Trending"` from the `locales/en_US.json` file, as well as other related translation string (there are more below). Weblate will take care of the other languages.

Fijxu:
May be superseded by https://github.com/iv-org/invidious/pull/5555

Fijxu:
Superseded by https://github.com/iv-org/invidious/pull/5555

PR #5550, opened 2025-11-30
Remove signature helper completely from Invidious

Fijxu:
The official way to reproduce video with Invidious now is by using Invidious Companion which uses Youtube.JS with a Javascript Interpreter that can successfully decrypt youtube video URLs.

Sig helper has not been used for a long time, is beyond broken and no one has plans to fix it and maintain it so is better to remove it.

### DIFF
--- a/src/invidious/videos.cr
+++ b/src/invidious/videos.cr
@@ -327,7 +327,11 @@ def fetch_video(id, region)
   info = extract_video_info(video_id: id)
 
   if info.nil?
-    raise InfoException.new("Invidious companion is not available. Video playback cannot continue.")
+    raise InfoException.new("Invidious companion is not available. \
+    Video playback cannot continue. \
+    If you are the administrator of this instance, install Invidious companion \
+    following the installation instructions \
+    <a href=\"https://docs.invidious.io/installation/\">https://docs.invidious.io/installation/</a>")
   end
 
   if reason = info["reason"]?


src/invidious.cr
```diff
@@ -172,12 +172,7 @@ Invidious::Database.check_integrity(CONFIG)
 
 # Misc
 
-DECRYPT_FUNCTION =
-  if sig_helper_address = CONFIG.signature_server.presence
-    IV::DecryptFunction.new(sig_helper_address)
-  else
-    nil
-  end
+DECRYPT_FUNCTION = nil
```
Fijxu: Should we get rid of `DECRYPT_FUNCTION` and places that make use of the `DECRYPT_FUNCTION`? They are not going to work anyways.

src/invidious/yt_backend/youtube_api.cr
```diff
@@ -496,9 +492,6 @@ module YoutubeAPI
       "playbackContext" => {
         "contentPlaybackContext" => playback_ctx,
       },
-      "serviceIntegrityDimensions" => {
-        "poToken" => CONFIG.po_token,
-      },
     }
```
Fijxu: Should we get rid of this function too?

unixfox: yes I think so

src/invidious.cr
unixfox: yes I think so

Fijxu:
image

If Invidious companion is not available, the video will not be available to watch.

https://github.com/iv-org/invidious/pull/5003 will need some work to display most of the video information without the video player

unixfox:
Maybe give instructions on how to fix that message?

src/invidious/videos.cr
```diff
@@ -327,7 +327,11 @@ def fetch_video(id, region)
   info = extract_video_info(video_id: id)
 
   if info.nil?
```
unixfox, line 329: Does `\` do a return line on the invidious side?

Fijxu: ~~No~~ or yes? Screenshot attached.

image

unixfox (approved):

PR #5551, opened 2025-11-30
feat: Add support for POST requests on searches for privacy

Fijxu:
Add support for POST requests on the search field for people that do not want their search history saved in their browser history.

image


unixfox:
Sounds like a neat feature. I know searx(ng) have that feature too.

Fijxu:
Now looks like this:

image

PR #5561, opened 2025-12-12
Allow downloading via companion

JeroenBoersma:
`post` request where not proxied to the companion 

This made it impossible to download with the companion enabled

### DIFF
--- a/src/invidious/routes/companion.cr
+++ b/src/invidious/routes/companion.cr
@@ -1,5 +1,5 @@
 module Invidious::Routes::Companion
-  # /companion
+  # GET /companion
   def self.get_companion(env)
     url = env.request.path
     if env.request.query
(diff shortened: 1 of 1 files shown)


unixfox:
You mean when using the download feature in the invidious frontend?

Because that should work as implemented in https://github.com/iv-org/invidious/pull/5224

JeroenBoersma:
> You mean when using the download feature in the invidious frontend?
Yes, the frontend route is indeed changed, but it is a post request to `/companion/download?...` and the post data isn't proxied to the companion so a redirect to the `/` occurs.

This fix implements a POST forward to the companion.

Screenshot_2025-12-13_10-28-12

src/invidious/routes/companion.cr
```diff
@@ -16,6 +16,24 @@ module Invidious::Routes::Companion
     end
   end
 
+# /companion
```
Fijxu:
```suggestion
 # /companion
```

Fijxu (commented):
Works fine, good catch!

@unixfox Download didn't work when Invidious companion was being proxied due to that missing route

PR #5565, opened 2025-12-14
feat: Add preferences cookie URL to import and export preferences

Fijxu:
Adds a `preferences` query parameter to the `preferences` page to import preferences and it also displays and URL to export the preferences to copy and paste it anywhere:

image

Any suggestions are appreciated because looks kinda off to me.

### DIFF
--- a/assets/css/default.css
+++ b/assets/css/default.css
@@ -555,6 +555,10 @@ span > select {
   border: 1px solid black;
 }
 
+.light-theme .cookie-url-box {
+  border: 2px solid #444
+}
+
 @media (prefers-color-scheme: light) {
   .no-theme a:hover,
   .no-theme a:active,
@@ -604,6 +608,10 @@ span > select {
 
   .no-theme .error-card {
     border: 1px solid black;
+  }
+
+  .no-theme .cookie-url-box {
+    border: 2px solid #444
   }
 }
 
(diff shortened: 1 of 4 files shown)


unixfox:
The UI is a bit odd. Sounds like a weird "hacker" interface at the end of the preferences page.

SamantazFox:
> Any suggestions are appreciated because looks kinda off to me.

Maybe something like what DDG does?

Screenshot of DuckDuckGo's settings page with the 'Show Bookmarklet and Settings Data' button highlighted
Screenshot of the interface that appears when the 'Show Bookmarklet and Settings Data' button is clicked. The interface contains a Bookmarklet URL, the settings in JSON format followed by a small paragraph that explains how this works, with a link to the documentation for the various settings

---

You could put the cookie value (JSON) directly, and as for the bookmarklet option, you could pass the various options to `/preferences` as URL parameters.

By the way, instead of adding a ton of dedicated logic for that in `show()`, you can simply put the URL query parameters (`env.params.query`) in the body (`env.params.body`) and chain a call to the `update()` handler. A similar thing is done for the download handler in [routes/watch.cr](https://github.com/iv-org/invidious/blob/21d0d1041a749c7b8a4dec306371653a9a94e082/src/invidious/routes/watch.cr#L313-L315).

Maybe offer the option of using `redirect.invidious.io` as a "portal" for the favorite, so this also works cross instance?

PR #5575, opened 2025-12-21
Generate companion check id one time and add missing companion check id on captions

Fijxu:
Invidious instances with `server.verify_requests` set to `true` (https://github.com/iv-org/invidious-companion/blob/520eaceedf1ee2371d912daf7c4730ac8eadd8d4/config/config.example.toml#L22) were unable to load captions due to a missing check id.

It also has a commit that generates the check id one time to save some CPU on encryption ;)

### DIFF
--- a/src/invidious/views/components/player.ecr
+++ b/src/invidious/views/components/player.ecr
@@ -71,15 +71,17 @@
         <% preferred_captions.each do |caption|
             api_captions_url = "/api/v1/captions/"
             api_captions_url = invidious_companion.public_url.to_s + api_captions_url if (invidious_companion)
+            api_captions_check_id = "&check=#{invidious_companion_check_id}"
         %>
-            <track kind="captions" src="<%= api_captions_url %><%= video.id %>?label=<%= caption.name %>" label="<%= caption.name %>">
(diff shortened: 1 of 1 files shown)


src/invidious/views/components/player.ecr
```diff
@@ -23,7 +26,7 @@
                 src_url  = "/latest_version?id=#{video.id}&itag=#{fmt["itag"]}"
                 src_url += "&local=true" if params.local
                 src_url = invidious_companion.public_url.to_s + src_url + 
-                            "&check=#{invidious_companion_encrypt(video.id)}" if (invidious_companion)
```
unixfox, line 26: I don't remember why we encrypt the checkid?

Fijxu: To prevent bots generating valid checks IDs to be used on Invidious companion player, dash, and caption requests and rate limit the companion instance

unixfox: Yes indeed, I got confused with encrypting query parameters: https://github.com/iv-org/invidious/pull/5381

unixfox:
@Fijxu could you please fix the CI?

Fijxu:
Already fixed at https://github.com/iv-org/invidious/commit/f7a31aa3dee1f37cb90a22303b6d45bec0033a3f

When merged, the linting should be ok, but I can rebase if you really want.

PR #5577, opened 2025-12-22
Downgrade Crystal to 1.16.3 in OCI

Fijxu:
After upgrading my instance to 1.18.2 it started to leak memory for some reason, I had to downgrade to 1.16.3 to make it stop. Currently I have no clue what is making Invidious use huge amounts of memory but I suppose is a problem with recent commits. 1.16.3 works fine

image

### DIFF
--- a/docker/Dockerfile.arm64
+++ b/docker/Dockerfile.arm64
@@ -21,7 +21,7 @@ RUN tar -xzvf openssl-${OPENSSL_VERSION}.tar.gz
 RUN cd openssl-${OPENSSL_VERSION} && ./Configure --openssldir=/etc/ssl && make -j$(nproc)
 
 FROM dependabot-alpine AS builder
-RUN apk add --no-cache 'crystal=1.18.2-r0' shards \
+RUN apk add --no-cache 'crystal=1.16.3-r0' shards \
         sqlite-static yaml-static yaml-dev \
         pcre2-static gc-static \
         libxml2-static zlib-static \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.arm64
unixfox, line 1: You need to downgrade alpine Linux version too

unixfox (changes requested):

unixfox:
This could be due to only this PR then https://github.com/iv-org/invidious/pull/5338.
But we don't know since we can't test crystal 1.17 without that PR.

syeopite:
Technically you can if you just strip out all code related to the static file handler from both Invidious and Kemal

syeopite:
Just read through everything again and I can't see what could possibly even lead to a memory leak in that PR... 

Maybe there's some weird interaction in Kemal due to that PR also stripping it out from the framework itself in versions after `1.17.X`?

absidue:
@syeopite Would updating kemal help?

syeopite:
Only one way to find out.

Upgrade Kemal and remove the corresponding overrides that strips out its static files handler.

I doubt it's actually the cause though.

Fijxu:
>The memory management model for libxml2 changed in Crystal 1.17 to become manual instead of integrating the GC. The change introduced another memory leak that is now fixed.
(https://crystal-lang.org/2026/04/16/1.20.0-released/#standard-library)

We encountered this memory leak when we upgraded Crystal to 1.17. I will try running my instance (https://inv.nadeko.net) with Crystal 1.20 and see if this memory leak is gone.

PR #5587, opened 2026-01-08
Fix last character disappearance if emoji are in comment

### LINKED ISSUE #5503: [Bug] The last character of comments disappear if emoji are included.
**Describe the bug**
The last character of comments dissappear when there's any emoji in it.
~This only happens if comments are written in Japanese or Korean. (I'm not sure about other multibyte character language).~
EDIT: I found out this also occurs in English. 

**Invidious English**
Image
**Youtube English**
Image
[link](https://www.youtube.com/watch?v=Fq09bCSlnyE&lc=UgzdxCs6ikiZlyGfZEV4AaABAg)

**Invidious Japanese**
Image
**Youtube Japanese**
Image
[link](https://www.youtube.com/watch?v=l01_q8c6UM8&lc=UgyOQgx6DMIEot6ddvd4AaABAg)

**Invidious Korean**
Image
**Youtube Korean**
Image
[link](https://www.youtube.com/watch?v=YAYmEa5-q4o&lc=UgxHmtiklUrZ264seuB4AaABAg)

**Steps to Reproduce**

1. Go to the Japanese, Korean (or maybe other multibyte language) Videos
2. Scroll down to comments section
3. Compare Invidious and Original comment by `[YT]` link.
4. 

**Additional context**

I'm using self-host instance with latest commit, but this also occurs in other public instances.
The API also returned comment without last character.

EDIT: It seems the number of character disappear is almost same as that of emoji in the comments.

Sorry for poor English.

shiny-comic:
This closes #5503

### DIFF
--- a/src/invidious/videos/description.cr
+++ b/src/invidious/videos/description.cr
@@ -21,8 +21,6 @@ private def copy_string(str : String::Builder, iter : Iterator, count : Int) : I
       str << cp.chr
     end
 
-    # A codepoint from the SMP counts twice
-    copied += 1 if cp > 0xFFFF
     copied += 1
   end
 
@@ -44,10 +42,6 @@ def parse_description(desc, video_id : String) : String?
     end
   end
 
-  # Not everything is stored in UTF-8 on youtube's side. The SMP codepoints
-  # (0x10000 and above) are encoded as UTF-16 surrogate pairs, which are
-  # automatically decoded by the JSON parser. It means that we need to count
-  # copied byte in a special manner, preventing the use of regular string copy.
   iter = content.each_codepoint
 
   index = 0


src/invidious/videos/description.cr
```diff
@@ -29,6 +29,10 @@ private def copy_string(str : String::Builder, iter : Iterator, count : Int) : I
   return copied
 end
 
+private def utf16_length(content : String) : Int32
+  content.each_char.sum { |ch| ch.ord >= 0x10000 ? 2 : 1 }
+end
+
```
Fijxu: This is fine, is basically a stripped down version of `String.to_utf16` **to count the number of characters that are two UInt16 wide**:

```
    each_char do |char|
      u16_size += char.ord < 0x1_0000 ? 1 : 2
    end
```
From: https://github.com/crystal-lang/crystal/blob/8fa7f90c091aa3757821c04ee243c7ab5f67ac20/src/string/utf16.cr#L18-L20

However, I wanted to test it's performance and the crystal stdlib implementation is faster to execute:

```
crystal stdlib   2.36M (424.52ns) (± 1.64%)   0.0B/op        fastest
       this pr   1.56M (640.54ns) (± 1.60%)  64.0B/op   1.51× slower
```

Benchmark code: 

```
require "benchmark"

aux = 0
aux2 = 0

char = " 전 나이도 있고 워낙 동해를 좋아해 강원도 해안도시에서 쾌적하게 살고 싶은데 지병 때문에 갈 수 없는 형편입니다. 집값 비싼 서울서 꾸역 꾸역 살아야하는 이유랍니다.😢😢 "

def toutf16(string : String) : Int32
  u16_size = 0
  string.each_char do |char|
    u16_size += char.ord < 0x1_0000 ? 1 : 2
  end
  return u16_size
end

def utf16_length(content : String) : Int32
  content.each_char.sum { |ch| ch.ord >= 0x10000 ? 2 : 1 }
end

Benchmark.ips do |x|
  x.report("crystal stdlib") { aux = toutf16(char) }
  x.report("this pr") { aux2 = utf16_length(char) }
end
```

So it's better to replace this to how the crystal stdlib does it


```suggestion
# size < bytesize, so we need to count the number of characters that are
# two UInt16 wide.
# Taken from: https://github.com/crystal-lang/crystal/blob/8fa7f90c091aa3757821c04ee243c7ab5f67ac20/src/string/utf16.cr#L18-L20
private def utf16_length(content : String) : Int32
  u16_size = 0
  content.each_char do |char|
    u16_size += char.ord < 0x1_0000 ? 1 : 2
  end
  u16_size
end
```

src/invidious/videos/description.cr
```diff
@@ -76,7 +80,7 @@ def parse_description(desc, video_id : String) : String?
     end
 
     # Copy the end of the string (past the last command).
-    remaining_length = content.size - index
+    remaining_length = utf16_length(content) - index
```
Fijxu: Japanese, Chinese and Korean characters are not ASCII, so we only use utf16_length if it's not ASCII ;)

```suggestion
    content_size = content.ascii_only? ? content.size : utf16_length(content)
    remaining_length = content_size - index
```

src/invidious/videos/description.cr
```diff
@@ -40,7 +44,7 @@ def parse_description(desc, video_id : String) : String?
     # Slightly faster than HTML.escape, as we're only doing one pass on
     # the string instead of five for the standard library
     return String.build do |str|
-      copy_string(str, content.each_codepoint, content.size)
+      copy_string(str, content.each_codepoint, utf16_length(content))
```
Fijxu: Japanese, Chinese and Korean characters are not ASCII, so we only use utf16_length if it's not ASCII ;)

```suggestion
      content_size = content.ascii_only? ? content.size : utf16_length(content)
      copy_string(str, content.each_codepoint, content_size)
```

Fijxu (changes requested):

src/invidious/videos/description.cr
```diff
@@ -29,6 +29,16 @@ private def copy_string(str : String::Builder, iter : Iterator, count : Int) : I
...
+# size < bytesize, so we need to count the number of characters that are
+# two UInt16 wide.
+# Taken from: https://github.com/crystal-lang/crystal/blob/8fa7f90c091aa3757821c04ee243c7ab5f67ac20/src/string/utf16.cr#L18-L20
+private def utf16_length(content : String) : Int32
+  u16_size = 0
+  content.each_char do |char|
+    u16_size += char.ord < 0x1_0000 ? 1 : 2
+  end
+  u16_size
+end
```
Fijxu:
```suggestion
end

```

make the linter happy

Fijxu (changes requested):
Opps

src/invidious/videos/description.cr
```diff
@@ -40,7 +51,8 @@ def parse_description(desc, video_id : String) : String?
     # Slightly faster than HTML.escape, as we're only doing one pass on
     # the string instead of five for the standard library
     return String.build do |str|
-      copy_string(str, content.each_codepoint, content.size)
+      content_size = content.ascii_only? ? content.size : utf16_length(content)
+      copy_string(str, content.each_codepoint, content_size)
```
SamantazFox: The reason I wrote this special string handling logic back then was because the Crystal standard library automatically decodes UTF-16 surrogates pairs to UTF-8, thus creating an offset between the positions reported in `commandRuns` and the actual character position in the crystal String.

If youtube stopped using UTF-16 surrogates, or if the Crystal standard library now leaves the surrogates alone then the line `copied += 1 if cp > 0xFFFF` can be removed from `copy_string()` and that will solve the problem.

There is no need for any additionnal counting shenanigans.

shiny-comic: Deleting the line did fix the issue. Thnaks for your help.

SamantazFox (approved):
Looks good to me!

Fijxu:
Thanks for your contribution!

PR #5589, opened 2026-01-11
Ameba 1.7

Sija:
Hello 👋

I've compiled an update to the newest Ameba version, 1.7.0. Thanks to the changes under the hood, it's not taking up time for compilation, cutting down CI time from **minutes** to **seconds** 🚀 

I've split the changes on a per-commit basis for an easier review.

Best,
Siyavash


unixfox:
Thanks!

We had some memory leak when upgrading to crystal 1.18 so we had to downgrade to crystal 1.16: https://github.com/iv-org/invidious/pull/5577

So for the moment we can't unfortunately merge your PR.

Sija:
@unixfox Current setup works for CI regardless of the Crystal version (uses docker) and local development environments - as long as they use Crystal 1.17+, so this should still be mergeable, as I can imagine you're using Docker for local development anyway.

unixfox:
Yes I know, but ideally we would like to find out if something needs to be changed regarding the support of crystal 1.17+: https://github.com/iv-org/invidious/pull/5577#issuecomment-3682154722

So it's possible that we would do some code changes before changing some massive amount of other code.

Sija:
Gotcha. You can do this after merging this PR, though.

Sija:
Wanna merge this PR, or should I close it?

Sija:
ping @unixfox @syeopite @Fijxu

Fijxu:
> ping @unixfox @syeopite @Fijxu

I prefer to keep this PR in hold until the memory leak with 1.17+ is fixed. Some conflicts may appear later.

.ameba.yml
```diff
@@ -27,41 +19,17 @@ Lint/SpecFilename:
...
-
-Style/RedundantReturn:
-  Enabled: false
-
-Style/RedundantNext:
-  Enabled: false
-
-Style/ParenthesesAroundCondition:
-  Enabled: false
-
```
SamantazFox, line 46: Please don't change invidious' style conventions. This creates a lot of useless changes across the entire codebase!

Sija: I don't agree that these are _useless changes_. They're the reason you have a linter for.

Sija:
I'm closing this PR since the feedback I've got after 1,5 month is... non-inspiring to say the least.

.ameba.yml
nobodywasishere: A lot of these are subjective style changes, and can understand why someone would want to have them disabled.

PR #5588, opened 2026-01-11
Add support for SABR

unixfox:
This pull request is for adding SABR support to invidious.

Close https://github.com/iv-org/invidious/issues/5263

### TODO
- [ ] Add support for https://github.com/LuanRT/ytc-bridge
- [ ] Remove the dependency on Invidious companion (reuse work from https://github.com/iv-org/invidious/pull/5003)
- [ ] Make loading the video player faster
- [ ] Fix glitch with loading icon in Shakajs
- [ ] Avoid using esm.sh for fetch the dependencies.
- [ ] Support extensions like SponsorBlock and DeArrow

PR #5600, opened 2026-01-19
Add Invidious check ID for storyboards (used for video storyboards in the video timeline)

Fijxu:
The storyboards API endpoint is unprotected and it allows anyone,
  including bots and abusers to spam that endpoint without having to
  enter the `/watch` endpoint (most of the time, is protected by some
  sort of bot protection by current instance owners to prevent abuse)

  I wonder if there is a better way to do this, but this works fine

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -90,9 +90,8 @@ db:
 #invidious_companion_key: "CHANGE_ME!!"
 
 ##
-## API key for Invidious companion, used for securing the communication
-## between Invidious and Invidious companion.
-## The key needs to be exactly 16 characters long.
+## Verify requests for endpoints that use Invidious companion
+## to fetch data
 ##
 ## Accepted values: true, false
 ## Default: true


unixfox:
I don't fully like the fact that we are reusing a token for companion in Invidious itself for encrypting/decrypting an endpoint that is not in companion. This is odd.

verify_requests was introduced as a way to reduce the possibility of getting blocked by youtube on the endpoints that request /player on the youtube side. Here it's the storyboard endpoint, I don't think doing too many storyboard requests will increase the chance of getting blocked by youtube.

And if we were to do this on the storyboard, we might as well do it on the other endpoints like comments.

IMO this problem can be instead solved by introducing some rate limits on the reverse proxy side on the storyboard endpoint.

unixfox:
Also in https://github.com/iv-org/invidious/pull/5588, I'm going to make companion optional. What would happen if one wants to use verify_requests for storyboard but without using companion? I feel like in such case we should instead introduce a new verify token.

PR #5601, opened 2026-01-19
Fix Youtube and Invidious links not rewinding their time when video playback position is rewound

Fijxu:
When rewinding a video, the time in the links **Watch on Youtube**, **(Embed)**, **Switch Invidious Instance**, **Embed Link** was not rewound, so the time in the links only increased and it didn't decrease depending of the video playback timestamp position.

If it's not clear, here is a video of the fix, trying to rewind without this fix, makes the `t` parameter only increase.

https://github.com/user-attachments/assets/f86cc2ba-a574-4ce8-bdbf-02b9adc1bd34

This fix does not affect at all with the saved playback position of each video (**Save playback position:** preference in Invidious settings that saves the video playback positions in Local Storage), the `timeupdate_last_ts` was only used here.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -132,7 +132,7 @@ var timeupdate_last_ts = 5;
 player.on('timeupdate', function () {
     // Only update once every second
     let current_ts = Math.floor(player.currentTime());
-    if (current_ts > timeupdate_last_ts) timeupdate_last_ts = current_ts;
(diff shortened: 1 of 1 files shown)


unixfox:
I feel like removing code is probably not the best idea, that code was there for a reason.

I can indeed reproduce the issue though.

Fijxu:
That should be better!, it will only update every second

SamantazFox (approved):

SamantazFox:
Please squash your commits though!

PR #5608, opened 2026-01-23
Update shard.yml to include target that was removed in commit 9d54cf9

Harm133:
[shard.yml]
- Include a target for LSPs to use as an entrypoint: (https://github.com/elbywan/crystalline?tab=readme-ov-file#entry-point)

This helps Crystalline to determine requires and project structure

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -5,6 +5,10 @@ authors:
   - Invidious team <contact@invidious.io>
   - Contributors!
 
+targets:
+  invidious:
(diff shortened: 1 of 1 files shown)


Fijxu:
That does not only fixes the LSP but also fixes Invidious building with `shards build`, I wonder why it got removed here: 9d54cf9, @syeopite any context why it got removed in that commit?

Fijxu (approved):

PR #5625, opened 2026-02-10
Title listen button time updates

JeroenBoersma:
When switching between Listen and Watching the timestamp in the url of the listen of watch button is now updated automatically.

This means if you switch between listening and viewing you keep in sync with time.

image


Fijxu (approved):
Good catch, thank you!

PR #5628, opened 2026-02-16
Add `enable_api` config option to disable unauthenticated API endpoints

### LINKED ISSUE #5599: [Feature request] Add option to disable easily abusable API easily.
Most public instances have to yes or yes, disable the Invidious API to prevent bots and abusers from spamming API endpoints that trigger a video fetch in [Invidious companion](https://github.com/iv-org/invidious-companion), most of them (including me) have to disable access to those endpoints using reverse proxy and manual discovery since we do not have a list of the endpoints that an instance owner is recommended to disable for the public.

So because of that it would be great to have a built-in option that let's instance owners and self-hosters to disable the endpoints that can be abused, therefore rate-limiting the instance for *legit* users using the instance.

Related to: https://github.com/iv-org/invidious/issues/3032

pierreeurope:
## Summary

Adds a new `enable_api` configuration option (default: `true`) that allows instance administrators to easily disable unauthenticated API endpoints to prevent abuse by bots and scrapers.

## Changes

- **`src/invidious/config.cr`**: Added `enable_api : Bool = true` property
- **`src/invidious/helpers/handlers.cr`**: Added `DisableAPIHandler` Kemal handler that intercepts `/api/v1/*` requests and returns 403 when the API is disabled
- **`src/invidious.cr`**: Registered the new handler

## Behavior

When `enable_api: false` is set in the config:
- All unauthenticated API endpoints (`/api/v1/*`) return `403` with `{"error": "The API has been disabled by the administrator."}`
- Authenticated endpoints (`/api/v1/auth/*`) remain available
- Stats endpoint (`/api/v1/stats`) remains available
- Default is `true` (no change for existing instances)

## Usage

```yaml
# config.yml
enable_api: false
```

Fixes #5599

### DIFF
--- a/src/invidious/config.cr
+++ b/src/invidious/config.cr
@@ -127,6 +127,10 @@ class Config
   property login_enabled : Bool = true
   property registration_enabled : Bool = true
   property statistics_enabled : Bool = false
+  # When set to false, disables the unauthenticated API endpoints
+  # (videos, channels, search, etc.) that can be abused by bots.
+  # Authenticated API endpoints (/api/v1/auth/*) are unaffected.
+  property enable_api : Bool = true
   property admins : Array(String) = [] of String
   property external_port : Int32? = nil
   property default_user_preferences : ConfigPreferences = ConfigPreferences.from_yaml("")
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -217,6 +217,7 @@ end
 Kemal.config.powered_by_header = false
 add_handler FilteredCompressHandler.new
 add_handler APIHandler.new
+add_handler DisableAPIHandler.new
 add_handler AuthHandler.new
 add_handler DenyFrame.new
 
--- a/src/invidious/helpers/handlers.cr
+++ b/src/invidious/helpers/handlers.cr
@@ -133,6 +133,29 @@ class APIHandler < Kemal::Handler
   end
 end
 
+class DisableAPIHandler < Kemal::Handler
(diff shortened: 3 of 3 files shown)


unixfox:
Hi, thanks for the PR.

There is an issue though is that /api/v1/comments will be blocked and, so one will not be able to load any comments.

src/invidious/config.cr
unixfox, line 1: Please move the comment to https://github.com/iv-org/invidious/blob/master/config/config.example.yml

Fijxu:
Not all API endpoints should be disabled, only easy to abuse ones. Other API endpoints not related to gathering video information are perfectly fine and they should not be disabled. First we should focus on the easy to abuse ones. I'm going to modify this PR to make it only restrict the easy to abuse API endpoints

Fijxu:
It also does not return a response to the client saying that the API is disabled:

image

pierreeurope:
Thanks for the feedback @unixfox and @Fijxu! You're right that blocking all endpoints is too broad — comments would break. @Fijxu if you'd like to refine this to only restrict the easily-abusable endpoints, feel free to push changes to this branch or I can close this in favor of your approach. Happy either way!

pierreeurope:
Closing this in favor of a more targeted approach — @Fijxu's idea to only restrict the easily-abusable endpoints makes much more sense than blocking the entire API. Happy to help review when that's ready!

Thanks for the feedback @unixfox 👍

unixfox:
And thank you for your contribution!

PR #5630, opened 2026-02-17
Add option to disable easy to abuse API endpoints

### LINKED ISSUE #5599: [Feature request] Add option to disable easily abusable API easily.

Fijxu:
Supersedes https://github.com/iv-org/invidious/pull/5628

The API endpoints that will be disabled when this option are:
 - /api/v1/videos
 - /api/v1/clips
 - /api/v1/transcripts

Add a `disable_abusable_api` config option, although it could be changed to a more flexible option to let the instance owner disable the API endpoints they want to disable, but I'm not too sure, what do you think @unixfox?

There is still API endponts that need some sort of validation or connection/proxying to Invidious companion like
`/api/v1/captions` and `/api/v1/storyboards` since they also do a video request to Invidious companion. I'm not sure if the `/next` API endpoint could be used to gather that type of information, if so, that would be better.

Closes #5599

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -436,6 +436,19 @@ full_refresh: false
 ##
 feed_threads: 1
 
+##
+## Setting to disable easy to abuse API endpoints that can
+## be spammed and therefore blocking your Invidious instance.
+##
+## Notes: The following API endpoints will be disabled:
+##  - /api/v1/videos
+##  - /api/v1/clips
+##  - /api/v1/transcripts
+##
+## Accepted values: true, false
+## Default: false
+##
+disable_api: false
 
 jobs:
 
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -185,7 +185,6 @@ https_only: false
 #  path: /tmp/invidious.sock
 #  permissions: 777
 
-
 # -----------------------------
 #  Network (outbound)
 # -----------------------------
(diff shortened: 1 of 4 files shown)


config/config.example.yml
```diff
@@ -436,6 +436,19 @@ full_refresh: false
...
+##
+## Notes: The following API endpoints will be disabled:
+##  - /api/v1/videos
+##  - /api/v1/clips
+##  - /api/v1/transcripts
+##
+## Accepted values: true, false
+## Default: false
+##
+disable_api: false
```
ghost: Imo the key name should also make it clear that it's only easily abused API endpoints that are disabled (like the comment does)

Maybe `disable_abusable_api` (like how the handler is named)

unixfox (approved):

PR #5637, opened 2026-02-19
Encapsulate translation constants and functions inside it's own `I18n` module

Fijxu:
As the title says, it encapsulates all related code from translation into it's own module.

Required for the migration to the crystal stdlib logger: https://github.com/iv-org/invidious/pull/5426


Fijxu:
Also, I wonder, should be adhere to the naming the modules have across most of the codebase? Which is appending `Invidious::<module_name>` or we just leave it like `<module_name>`?

SamantazFox (approved):

PR #5617, opened 2026-02-03
Channels: parse pronouns and display them on channel page

radmorecameron:
This PR parses and displays channel pronouns.

# Testing:

## Channel with pronouns:
1- Go to `{INVIDIOUS INSTANCE}/api/v1/channels/UCw-aR42z5gUcarpPGN5OKfA`
2 - see pronoun field below channel name
image

1 - Go to `{INVIDIOUS INSTANCE}/channel/UCw-aR42z5gUcarpPGN5OKfA`
2- See pronoun field
image

## Channel with no prouns

1 - Go to `{INVIDIOUS INSTANCE}/channel/UCX6OQ3DkcsbYNE6H8uQQuVA`
2 - Don't see pronouns
image

1 - Go to `{INVIDIOUS INSTANCE}/api/v1/channels/UCX6OQ3DkcsbYNE6H8uQQuVA`
2 - see empty pronouns
image

PR #5356, opened 2025-06-21
Add typos to Invidious

syeopite:
Serious title: Add job to check spelling via `typos` tool as apart of CI

----

I recently learned that ameba has integrations with the `typos` tool to detect spelling mistakes within the source code, though it does need to be installed first but I'm not exactly sure how to do that so it'll be done in a separate job for now.

The next release of the Ameba Github Actions program (which should include the `typos` tool) should add support to create annotations without needing a token so we can just merge both the Ameba and spell check jobs into one once we switch to the action program.

PR #5640, opened 2026-02-20
Parse channel handle and display it on channel page and API

### LINKED ISSUE #5638: [Feature request] Add Channel Handles to channel page
Hello, it would be nice if we could see channel handles on invidious channel pages.

Image

Image

it would be nice if Author Handle was added https://github.com/iv-org/invidious/blob/864893f4c75d79b725aba2ddfbf0d55a2b71111e/src/invidious/routes/api/v1/channels.cr#L50 after the author name.

not needed, i think

**Additional context**
not needed, i think

Fijxu:
It also fixes some inconsistencies with the CSS and HTML that was introduced in https://github.com/iv-org/invidious/pull/5617 but I didn't really noticed about because it looked fine in the Invidious frontend.

I also added a comment about the `break` statement in the `metadata_rows` row iteration because I didn't get why that `break` was there in the first place.

Closes https://github.com/iv-org/invidious/issues/5638

---

Looks like this in the Invidious frontend (with borders enabled just to see each element sizes and limits):

image

The API looks like this:

```
  ...
  "author": "Fijxu - Topic",
  "authorHandle": "@Fijxuu",
  "authorId": "UCw-aR42z5gUcarpPGN5OKfA",
  ...
```

### DIFF
--- a/src/invidious/channels/about.cr
+++ b/src/invidious/channels/about.cr
@@ -174,18 +174,15 @@ def get_about_info(ucid, locale) : AboutChannel
     metadata_rows.each do |row|
       metadata_parts = row.dig?("metadataParts")
 
-      subscribe_metadata_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("subscribers") }
-      if !subscribe_metadata_part.nil?
+      if (subscribe_metadata_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("subscribers") })
         sub_count = short_text_to_number(subscribe_metadata_part.dig("text", "content").as_s.split(" ")[0]).to_i32
       end
 
-      channel_handle_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("@") }
-      if !channel_handle_part.nil?
+      if (channel_handle_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("@") })
         channel_handle = channel_handle_part.dig("text", "content").as_s
       end
 
-      pronoun_metadata_part = metadata_parts.try &.as_a.find { |i| i.dig?("tooltip").try &.as_s.includes?("Pronouns") }
-      if !pronoun_metadata_part.nil?
+      if (pronoun_metadata_part = metadata_parts.try &.as_a.find { |i| i.dig?("tooltip").try &.as_s.includes?("Pronouns") })
         pronouns = pronoun_metadata_part.dig("text", "content").as_s
       end
 
--- a/src/invidious/channels/about.cr
+++ b/src/invidious/channels/about.cr
@@ -190,7 +190,7 @@ def get_about_info(ucid, locale) : AboutChannel
       end
 
       # This is to prevent processing more metadata parts if we already have
-      # all the parts we care about, which are the ones bellow
+      # all the parts we care about, which are the ones below
       break if sub_count != 0 && !pronouns.nil? && !author_handle.nil?
     end
   end
--- a/src/invidious/channels/about.cr
+++ b/src/invidious/channels/about.cr
@@ -2,7 +2,7 @@
 record AboutChannel,
   ucid : String,
   author : String,
-  author_handle : String?,
+  channel_handle : String?,
   auto_generated : Bool,
   author_url : String,
   author_thumbnail : String,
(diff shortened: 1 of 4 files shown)


src/invidious/channels/about.cr
```diff
@@ -162,26 +163,37 @@ def get_about_info(ucid, locale) : AboutChannel
...
+      end
+
+      pronoun_metadata_part = metadata_parts.try &.as_a.find { |i| i.dig?("tooltip").try &.as_s.includes?("Pronouns") }
       if !pronoun_metadata_part.nil?
         pronouns = pronoun_metadata_part.dig("text", "content").as_s
       end
 
-      break if sub_count != 0 && !pronouns.nil?
+      # This is to prevent processing more metadata parts if we already have
+      # all the parts we care about, which are the ones bellow
```
ghost:
```suggestion
      # all the parts we care about, which are the ones below
```

Fijxu: we really need https://github.com/iv-org/invidious/pull/5356 lol

src/invidious/views/components/channel_info.ecr
```diff
@@ -12,10 +12,17 @@
...
                 <span class="channel-name"><%= author %></span><% if !channel.verified.nil? && channel.verified %>&nbsp;<i class="icon ion ion-md-checkmark-circle"></i><% end %>
-                <% if !channel.pronouns.nil? %><br /><span class="channel-pronouns"><%= channel.pronouns %></span><% end %>
-            </div>
+                <span class="channel-metadata">
+                    <br/>
+                    <% if !channel.author_handle.nil? %>
+                    <span class="channel-metadata-item"><b><%= channel.author_handle %></b></span>
+                    <span>•</span>
+                    <% end %>
+                    <% if !channel.pronouns.nil? %><span class="channel-metadata-item"><%= channel.pronouns %></span><% end %>
```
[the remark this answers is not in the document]
Fijxu: image


May need to be modified with the addition of https://github.com/iv-org/invidious/issues/5641

ashleyirispuppy143:
why rename `author_handle ` to `channel_handle`? the channel id and like the channel name and stuff is called author so why the handle wouldnt be?

Fijxu:
> `channel_handle`

Just to be consistent by how is set in `src/invidious/helpers/serialized_yt_data.cr`! If any other maintainer wants to change it to `author_handle`, I'll change it. I really have no preference, once the new key is set, people that use the API will use that and is not going to be changed as it will be a breaking change ;)

src/invidious/channels/about.cr
```diff
@@ -162,26 +163,37 @@ def get_about_info(ucid, locale) : AboutChannel
...
+      metadata_parts = row.dig?("metadataParts")
+
+      subscribe_metadata_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("subscribers") }
       if !subscribe_metadata_part.nil?
         sub_count = short_text_to_number(subscribe_metadata_part.dig("text", "content").as_s.split(" ")[0]).to_i32
       end
 
-      pronoun_metadata_part = row.dig?("metadataParts").try &.as_a.find { |i| i.dig?("tooltip").try &.as_s.includes?("Pronouns") }
+      channel_handle_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("@") }
+      if !channel_handle_part.nil?
```
SamantazFox: You can combine that in a single `if` statement:
```suggestion
      if channel_handle_part = metadata_parts.try &.as_a.find { |i| i.dig?("text", "content").try &.as_s.includes?("@") }
```

SamantazFox: (And similarly, you can do this to the other `if` statements in this `.each` loop)

ashleyirispuppy143:
weh, is this never gonna get merged... :c 

~~i have to patch my invidious every update lmao~~

Fijxu:
Should be ready to merge! I'll deploy it in the test instance right now.

Fijxu:
PR added to the test instance

Test URLs: 

https://test.invidious.io/channel/UCw-aR42z5gUcarpPGN5OKfA
https://test.invidious.io/channel/UCdsE4M-iXp49V01vTlJNAow
https://test.invidious.io/channel/UCHbqoL7uUbcq1480BPz0Apw
https://test.invidious.io/channel/UCX6OQ3DkcsbYNE6H8uQQuVA

Fijxu:
The test instance doesn't seem to show the pronouns for some reason, but locally, it shows them. Maybe the pronouns show based on the region? I will do some testing with the IP address of the test instance.

Fijxu:
Effectively Youtube does not return the pronouns of the channel on certain IP addresses or regions, for example, when using the test instance IP address (which is from Sweden) to get the channel pronouns, you get no pronouns, but when using another IP address from the same country, you get pronouns, so it seems to depend on Youtube heuristics to show pronouns or not.

PR #5642, opened 2026-02-20
Remove noreferer since youtube now requires referers on embeds

ashleyirispuppy143:
Fixes an issue where clicking the embed button resulted in Error 153. YouTube now strictly requires a referrer for embeds to load correctly. This PR updates the link/embed behavior to pass the required referrer, so the embeds now work.

image

 Close https://github.com/iv-org/invidious/issues/5645

### DIFF
--- a/src/invidious/views/watch.ecr
+++ b/src/invidious/views/watch.ecr
@@ -125,7 +125,7 @@ we're going to need to do it here in order to allow for translations.
                     end
                 -%>
                 <a id="link-yt-watch" rel="noreferrer noopener" data-base-url="<%= link_yt_watch %>" href="<%= link_yt_watch %>"><%= translate(locale, "videoinfo_watch_on_youTube") %></a>
-                (<a id="link-yt-embed" rel="noopener" data-base-url="<%= link_yt_embed %>" href="<%= link_yt_embed %>"><%= translate(locale, "videoinfo_youTube_embed_link") %></a>)
+                (<a id="link-yt-embed" rel="noopener" referrerpolicy="origin-when-cross-origin" data-base-url="<%= link_yt_embed %>" href="<%= link_yt_embed %>"><%= translate(locale, "videoinfo_youTube_embed_link") %></a>)
             </span>
 
             <p id="watch-on-another-invidious-instance">


Fijxu:
This will not work as Invidious uses `same-origin` as value for `Referrer-Policy` policy, so only Invidious links of the same domain will contain a referrer header, external links like `Watch on Youtube` and `(Embed)` will not contain a referrer:

https://github.com/iv-org/invidious/blob/fda8d1b528f1999b5eced404e4818f592a79702f/src/invidious/routes/before_all.cr#L58m

Is also not possible to change their values using Javascript: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header

unixfox:
@Fijxu I guess we could change the global policy to `origin-when-cross-origin`.

Fijxu:
> @Fijxu I guess we could change the global policy to `origin-when-cross-origin`.

Yeah, that seems like the only thing we can do, other solutions that I have thought about are kinda over engineered (like using an external redirector so it doesn't leak the Invidious domain to youtube servers!) so I guess it's fine. What do you think about adding a webpage that warns the user before redirecting (or something like that)?

unixfox:
> What do you think about adding a webpage that warns the user before redirecting (or something like that)?

I guess that was this PR was about: https://github.com/iv-org/invidious/pull/4826

But to be fair, IMO it's useless, people should be responsible for where they click, not us.

src/invidious/views/watch.ecr
```diff
@@ -125,7 +125,7 @@ we're going to need to do it here in order to allow for translations.
                     end
                 -%>
                 <a id="link-yt-watch" rel="noreferrer noopener" data-base-url="<%= link_yt_watch %>" href="<%= link_yt_watch %>"><%= translate(locale, "videoinfo_watch_on_youTube") %></a>
-                (<a id="link-yt-embed" rel="noreferrer noopener" data-base-url="<%= link_yt_embed %>" href="<%= link_yt_embed %>"><%= translate(locale, "videoinfo_youTube_embed_link") %></a>)
+                (<a id="link-yt-embed" rel="noopener" data-base-url="<%= link_yt_embed %>" href="<%= link_yt_embed %>"><%= translate(locale, "videoinfo_youTube_embed_link") %></a>)
```
Fijxu:
```suggestion
                (<a id="link-yt-embed" rel="noopener" referrerpolicy="origin-when-cross-origin" data-base-url="<%= link_yt_embed %>" href="<%= link_yt_embed %>"><%= translate(locale, "videoinfo_youTube_embed_link") %></a>)
```

This fixes it, of course, it will leak the domain to youtube embeds.

Fijxu (changes requested):

PR #5647, opened 2026-02-23
Add support for alternative domains for Invidious cookies

### LINKED ISSUE #1421: Instances are working well, but their onion version aren't
Hello,

Fijxu:
Closes https://github.com/iv-org/invidious/issues/1421

Adds a new configuration option called `alternative_domains`, a list of additional domains from where the Invidious instance is hosted, this will make hosting Invidious in other domains like Tor hidden services, I2P tunnels, Yggdrasil or any other type of domain that is inserted.

It needs the usage of a reverse proxy to pass the `X-Forwarded-Host` header to Invidious, since that is the header that will be used to detect the alternative domains. It could be more flexible and just use `Host`, in that way, there is no need to use a reverse proxy, but I decided to go with `X-Forwarded-Host` since it's **de-facto standard header for identifying the original host requested by the client in the Host HTTP request header.** (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Host)

Instance administrators will also need to add their additional domains in their reverse proxy configuration, but that will depend on their reverse proxy configuration.

For nginx, that would mean adding the additional domain to their `server_name` server directive, example:

```
  server {
    server_name
           invidious-github.localhost
           someotherdomain.lol
           meowdious.i2p
           superhiddenservicetpm6zs3j4f33elo6wkpbjpxela6uv7hzomeyd.onion
	   ;
    listen 80;
    location / {
      proxy_pass http://127.0.0.1:30001;
      proxy_set_header X-Forwarded-Host $host;
    }
  }
```

Here is a demonstration on how the `Set-Cookie` header looks like when using an alternative domain:

image

### DIFF
--- a/src/invidious/user/cookies.cr
+++ b/src/invidious/user/cookies.cr
@@ -6,17 +6,24 @@ struct Invidious::User
 
     # Note: we use ternary operator because the two variables
     # used in here are not booleans.
-    SECURE = (Kemal.config.ssl || CONFIG.https_only) ? true : false
+    @@secure = (Kemal.config.ssl || CONFIG.https_only) ? true : false
 
     # Session ID (SID) cookie
     # Parameter "domain" comes from the global config
     def sid(domain : String?, sid) : HTTP::Cookie
+      # Not secure if it's being accessed from I2P
+      # Browsers expect the domain to include https. On I2P there is no HTTPS
+      # Tor browser works fine with secure being true
+      if domain.try &.split(".").last == "i2p"
+        @@secure = false
+      end
+
       return HTTP::Cookie.new(
         name: "SID",
         domain: domain,
         value: sid,
         expires: Time.utc + 2.years,
-        secure: SECURE,
+        secure: @@secure,
         http_only: true,
         samesite: HTTP::Cookie::SameSite::Lax
       )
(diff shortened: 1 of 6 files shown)


src/invidious/user/cookies.cr
```diff
@@ -6,17 +6,24 @@ struct Invidious::User
...
     # Session ID (SID) cookie
     # Parameter "domain" comes from the global config
     def sid(domain : String?, sid) : HTTP::Cookie
+      # Not secure if it's being accessed from I2P
+      # Browsers expect the domain to include https. On I2P there is no HTTPS
+      # Tor browser works fine with secure being true
+      if (domain.try &.split(".").last == "i2p") && @@secure
+        @@secure = false
+      end
+
```
Fijxu, line 20: Just as a note, when using `.i2p` domains, `secure` has to be false, otherwise, the browser using an I2P proxy will reject the cookie. But this seems to be too hardcoded, so I was thinking about modifying `alternative_domains` to let the instance administrator to define if they want the Cookie to be secure or not secure, because it really depends on where they are hosting Invidious and how the browser of the client is configured. Something like this basically, but I'm not too sure so suggestions are appreciated!:

```
alternative_domains:
  - something.i2p
    secure: false
  - othersomething.onion
    secure: true
  - alternative-clearnet.lol
    secure: true
```

Fijxu:
PR has been added to the test instance https://test.invidious.io

To test alternative domains, here are the URLs for the overlay networks Tor and I2P:
 - http://g6vfi53ftdlrqpwt7aqdtcj42jookxihparvwxo3x5adis3wqckq.b32.i2p (Not working, is going to be fixed later)
 - http://ftsyofigkp4fbx56pp3gpsoegsojy5zl5biq3l3ue5fkqsoriueobmqd.onion

unixfox (approved):

TheFrenchGhosty:
A 10$ bounty has been rewarded to @Fijxu for this PR.

Transaction (9a121ca29afabd5bc4bc108916418fa54d0fb845913a170d49ba632f5ea925cf):

Amount sent: 0.028332625018 XMR (~8.77€ or ~10$)

Mining fees: 0.000044400000 XMR (~0.01€)

PR #5650, opened 2026-02-27
Add embed youtube link for error_template_helper

### LINKED ISSUE #5649: [Feature request] Add YouTube 'embed' link to error pages
When accessing a video page normally, it includes both a "full-fat" YouTube video page link as well as the embedded player YouTube video page link.

But if invidious instead returns an error, the YouTube link(s) listed is only the "full-fat" video page with no link to the embedded player page. This is rather troublesome because, due to former issue #5645 the end-user cannot just manually type the embed page URL into their browser and have it function - it needs a proper referrer.

Simply add the same sort of (Embed) YouTube link like exists on the standard video page, see my mock-up:
Image

Uhhh, I really can't think of any unless you count just never having an error ever again? :)

**Additional context**
I still think linking to the youtube-nocookie.com embedded player makes the most sense for invidious's privacy-focused policy - see issue #5317 (but at this rate I'm probably starting to sound like a broken record so I probably should just shut up about it already...)

Screenshot of what the error page currently looks like (though the actual error in the screenshot is only one of maybe 5 different errors that show up from time-to-time regardless of invidious instance):
Image

Fijxu:
Closes https://github.com/iv-org/invidious/issues/5649

It also checks if the endpoint is `/watch` and that the `v` query parameter isn't empty before generating the embed URL ^^

### DIFF
--- a/src/invidious/helpers/errors.cr
+++ b/src/invidious/helpers/errors.cr
@@ -197,11 +197,17 @@ def error_redirect_helper(env : HTTP::Server::Context)
       video_id = params.fetch("v", nil)
 
       if video_id.presence
-        params.delete("v")
         if params.present?
-          embed_link = "https://youtube.com/embed/#{video_id}?#{params}"
+          new_params = URI::Params.new
+          list = params["list"]?.presence
+          index = params["index"]?.presence
+          if list && index
+            new_params.add("list", list)
+            new_params.add("index", index)
+            embed_link = HTML.escape("https://youtube.com/embed/#{video_id}?#{new_params}")
+          end
         else
-          embed_link = "https://youtube.com/embed/#{video_id}"
+          embed_link = HTML.escape("https://youtube.com/embed/#{video_id}")
         end
         embed_html_element = "(<a rel=\"noopener\" referrerpolicy=\"origin-when-cross-origin\" href=\"#{embed_link}\">#{go_to_youtube_embed}</a>)"
       end
--- a/src/invidious/helpers/errors.cr
+++ b/src/invidious/helpers/errors.cr
@@ -199,7 +199,7 @@ def error_redirect_helper(env : HTTP::Server::Context)
       if video_id.presence
         params.delete("v")
         if params.present?
-          embed_link = "https://youtube.com/embed/#{video_id}?#{params.to_s}"
+          embed_link = "https://youtube.com/embed/#{video_id}?#{params}"
         else
           embed_link = "https://youtube.com/embed/#{video_id}"
         end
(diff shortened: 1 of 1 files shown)


src/invidious/helpers/errors.cr
```diff
@@ -184,13 +184,29 @@ def error_redirect_helper(env : HTTP::Server::Context)
...
     switch_instance = translate(locale, "Switch Invidious Instance")
 
+    if is_watch
+      params = URI.parse(env.request.resource).query_params
+      video_id = params.fetch("v", nil)
+
+      if video_id.presence
+        params.delete("v")
+        if params.present?
+          embed_link = "https://youtube.com/embed/#{video_id}?#{params}"
```
SamantazFox: Watch out when passing URL parameters directly to youtube. They might contain tracking identifiers! That's why the other youtube link does not pass any.

src/invidious/helpers/errors.cr
```diff
@@ -184,13 +184,29 @@ def error_redirect_helper(env : HTTP::Server::Context)
...
+        else
+          embed_link = "https://youtube.com/embed/#{video_id}"
+        end
+        embed_html_element = "(<a rel=\"noopener\" referrerpolicy=\"origin-when-cross-origin\" href=\"#{embed_link}\">#{go_to_youtube_embed}</a>)"
```
SamantazFox, line 212: If you ever pass URL parameters, make sure to HTML escape them, otherwise that creates an XSS vulnerability.

SamantazFox (changes requested):

src/invidious/helpers/errors.cr
Fijxu: >They might contain tracking identifiers! That's why the other youtube link does not pass any.

Right, but when watching videos in a playlist, it actually contains two parameters, `list` and `index`

https://github.com/iv-org/invidious/blob/21d0d1041a749c7b8a4dec306371653a9a94e082/src/invidious/views/watch.ecr#L121-L122

I'll whitelist them so we only pass those to youtube links

SamantazFox: You can use (and expand) `src/invidious/yt_backend/url_sanitizer.cr`

radmorecameron: If there's a `t` parameter, it would need to become `start` #5658

Fijxu:
works fine, should be ready to merge

NintendoManiac64:
It might be worth noting that the "Embed" youtube link seems to be broken again: #5706

unixfox (approved):

PR #5660, opened 2026-03-09
player: Use correct time parameter for YouTube embed redirects

### LINKED ISSUE #5658: [Bug] ?t=<time> doesn't work for official YouTube embedded player
**Describe the bug**
The `?t=<time>` URL parameter doesn't work on the YouTube embedded player.

**Steps to Reproduce**
1. Open a video in an invidious instance that properly supports passing the referrer to the YouTube embedded player (the only one I know of that does this is https://invidious.tiekoetter.com)
2. Click on a video that supports embedding ([here's an example video that I know does](https://youtube.com/watch?v=VQxBW7JgkHw))
3. [Hope the description page actually loads and you don't get an error](https://github.com/iv-org/invidious/issues/5649)
4. Play the video and click somewhere in the middle
5. Click the `(Embed)` next to the `Watch on YouTube`
6. Notice how the embedded player still starts over from the beginning

**Additional context**
I have no idea if this is simply a case of if the URL parameter needs to be different or if this function is straight-up not even supported by YouTube's official embedded player.

I did find in my browser history that, in mid-2021, I accessed a video on youtube.com/embed/ with the URL parameter `?start=<time>` but, because one cannot just manually type a youtube.com/embed/ address anymore, I have no way of being able to test if perhaps `?start=<time>` is what is instead used by the official YouTube embedded player.

radmorecameron:
closes #5658

YouTube uses `start` instead of `t` for the embed's start time. As mentioned in the linked issue and can be confirmed here: https://developers.google.com/youtube/player_parameters

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -104,14 +104,15 @@ if (video_data.params.quality === 'dash') {
  *
  * @param {String} url
  * @param {String} [base]
+ * @param {'t' | 'start'} param
  * @returns {URL} urlWithTimeArg
  */
-function addCurrentTimeToURL(url, base) {
+function addCurrentTimeToURL(url, base, param = 't') {
     var urlUsed = new URL(url, base);
     urlUsed.searchParams.delete('start');
     var currentTime = Math.ceil(player.currentTime());
     if (currentTime > 0)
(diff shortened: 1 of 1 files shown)


Fijxu:
Thanks for your contribution!

PR #5692, opened 2026-04-25
chore: update Crystal to 1.20.0 in OCI

Fijxu:
Supersedes #5690

The memory leak mentioned in the https://github.com/iv-org/invidious/pull/5577 PR seems to be fixed, I have been running my instance using Crystal 1.20.0 for 1 day and memory usage is stable (I use Docker for Invidious and each Invidious container has a hard memory limit of 1536MB, but they barely hit that limit, so the Crystal GC is working fine at freeing memory):

image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 ARG OPENSSL_VERSION='3.5.2'
 ARG OPENSSL_SHA256='c53a47e5e441c930c3928cf7bf6fb00e5d129b630e0aa873b08258656e7345ec'
 
-FROM crystallang/crystal:1.16.3-alpine AS dependabot-crystal
+FROM crystallang/crystal:1.20.0-alpine AS dependabot-crystal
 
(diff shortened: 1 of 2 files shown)


erusc:
Just a head up that they started serving [official arm64 image](https://hub.docker.com/layers/crystallang/crystal/1.20.0-alpine/images/sha256-81b0a58f37f487eaedf25370ebbcf39945a00e36eba0013a60d176b9f1dbd556) a while ago.

Fijxu:
> Just a head up that they started serving [official arm64 image](https://hub.docker.com/layers/crystallang/crystal/1.20.0-alpine/images/sha256-81b0a58f37f487eaedf25370ebbcf39945a00e36eba0013a60d176b9f1dbd556) a while ago.

Cool, I just noticed that they started to serve arm64 OCI like 5 months ago. I'm going to make another PR to unify the Dockerfiles we use into a single one.

PR #5700, opened 2026-04-28
CI: Unify ARM64 and AMD64 Dockerfiles

Fijxu:
As I mentioned in https://github.com/iv-org/invidious/pull/5692#issuecomment-4331372731, Crystal now offers a `linux/arm64` OCI that we can use to build Invidious on ARM: https://hub.docker.com/layers/crystallang/crystal/1.20.0/images/sha256-82a15cc24f2774cefb4bd4f5ad869db3656f8db2c4e16d2955f6a6338d8a1a55

Also note that `linux/arm64` equals to `linux/arm64/v8` (https://stackoverflow.com/questions/70819028/relation-between-linux-arm64-and-linux-arm64-v8-are-these-aliases-for-each-othe, and using `docker pull --platform linux/arm64/v8 crystallang/crystal:1.20.0-alpine` or `docker pull --platform linux/arm64 crystallang/crystal:1.20.0-alpine` pulls the same exact image), so we don't need to change the platform.


Fijxu:
After merging this we can also update Invidious to build using 84codes Crystal image, since they build Crystal using `--enable-large-config` and as I demonstrated before in https://github.com/iv-org/invidious/pull/5473, the memory usage is lower.

unixfox:
Maybe we can publish the arm version on :latest tag too?

Fijxu:
> Maybe we can publish the arm version on :latest tag too?

Yes, should be possible, I didn't do it so we don't break things with deployments that use the `-arm64` images, but we can keep releasing images with `-arm64` until we deprecate it.

unixfox:
> > Maybe we can publish the arm version on :latest tag too?
> 
> Yes, should be possible, I didn't do it so we don't break things with deployments that use the `-arm64` images, but we can keep releasing images with `-arm64` until we deprecate it.

Yes that's what I was thinking.

Fijxu:
That should do the trick, although not fully tested since OCI releases are not built in PRs (of course). I will adapt  my Invidious repository fork to push to a OCI registry to do some sort of "production testing".

The simpler and easy way to push multiple OCI platforms at one is using QEMU with the [Docker Setup QEMU Action](https://github.com/marketplace/actions/docker-setup-qemu), which is what Invidious used, but we migrated to native ARM runners hosted by GitHub for faster ARM builds, so using something like this is no longer possible (using multiple platforms in this job):

```
      - name: Build and push Docker ${{ matrix.name }} image for Push Event
        uses: docker/build-push-action@v7
        with:
          context: .
          file:  ${{ matrix.dockerfile }}
          file: ${{ matrix.dockerfile }}
          platforms: linux/amd64,linux/arm64 # <--- here
          labels: ${{ steps.meta.outputs.labels }}
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          build-args: |
            "release=1"
 ```

So I discovered you can build different OCI for different platforms and merge them into a single manifest using `docker manifest create` or `docker buildx imagetools create`, here is a good read on how those operations work: https://andrewlock.net/combining-multiple-docker-images-into-a-multi-arch-image/

I opted to use https://github.com/marketplace/actions/docker-manifest-create-action to create the manifest with multiple platform support, it uses `docker buildx imagetools create` internally.

https://github.com/marketplace/actions/docker-manifest-create-action also has examples to better organize our workflow files, we could take and adapt some of those later

Fijxu:
Works fine, ready to be merged:

https://github.com/Fijxu/invidious/actions/runs/25261680351
`master` image has two architectures: https://quay.io/repository/fijxu/invidious?tab=tags

unixfox (approved):

PR #5702, opened 2026-04-30
feat: add option to filter Shorts / videos under 1 minute (#5688)

### LINKED ISSUE #5688: Add Disable shorts or videos less than 1minute?
I know shorts are not there, but if there was an option that could filter videos less than 1minutes or category of shorts then it would be great! Thanks!

Sushanth012:
## Summary
Adds a user preference to hide YouTube Shorts / videos under 1 minute from subscription feeds and search results.

## Changes
- Added `filter_short_videos` boolean preference (default: false)
- Filters out videos with `length_seconds < 60` when preference is enabled
- Added checkbox in user preferences UI: "Hide Shorts / videos under 1 minute"

## Related
Closes #5688

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -104,6 +104,7 @@
     "preferences_captions_label": "Default captions: ",
     "Fallback captions: ": "Fallback captions: ",
     "preferences_related_videos_label": "Show related videos: ",
+    "preferences_filter_short_videos_label": "Hide Shorts / videos under 1 minute: ",
     "preferences_annotations_label": "Show annotations by default: ",
     "preferences_extend_desc_label": "Automatically extend video description: ",
     "preferences_vr_mode_label": "Interactive 360 degree videos (requires WebGL): ",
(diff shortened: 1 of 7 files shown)


acheong08:
In subscriptions, shorts are length == 0. Hiding videos under 1 minute will hide legitimate videos too. I think this should be a filter for search, but subscriptions `length_seconds != 0` is sufficient. 

There is a side effect of not just shorts being hidden but also livestreams since they appear as 0 as well.

FMGordillo:
I pulled your branch, compiled everything under `docker-compose.yml` and I see the option in the Settings menu

image

I activate the option, and then click on "Save preferences"

image

I see the payload being submitted, but I still see shorts on my feed

When I came back to the Settings page I still see the option unchecked

Is this ready for merge or am I missing something?

FMGordillo:
> In subscriptions, shorts are length == 0. Hiding videos under 1 minute will hide legitimate videos too. I think this should be a filter for search, but subscriptions `length_seconds != 0` is sufficient.
> 
> There is a side effect of not just shorts being hidden but also livestreams since they appear as 0 as well.

Also see https://github.com/iv-org/invidious/pull/5609#issuecomment-3791846435

unixfox:
Closing as there is already a pending PR here: https://github.com/iv-org/invidious/pull/5609

PR #4051, opened 2023-08-14
Show message when comments are turned off

syeopite:
Fixes: #1092
I want to be awarded the bounty associated to the issue this PR is fixing.

comments

I've also added a message for when comments are disabled via Invidious config.

PR #5709, opened 2026-05-11
[codex] Show message when YouTube comments are disabled

### LINKED ISSUE #1092: Display "Comments are turned off" when comments are disabled (currently the comment section is just empty)
**Description**

On Youtube.com it is possible to disable the comments of a certain video. If this is the case, YouTube.com shows "Comments are turned off" in the place where the comments are supposed to be.

If a user of Invidious opens a video where the comments are disabled, the place where the comments are supposed to be just stays empty.

The user may think that there was a problem with Invidious when it tried to load the comments. To remove this uncertainty, Invidious should display a message like "Comments are turned off" or something similar so that the user knows about the disabled comments.

**Example**

At https://www.youtube.com/watch?v=3x7T6PyHRGo it says "Comments are turned off" in the comment section.

At https://invidio.us/watch?v=3x7T6PyHRGo the comment section is just empty.

nickita-khylkouski:
Fixes #1092.

## Changes
- Return a localized "Comments are turned off." message when YouTube provides no comment continuation items.
- Add `commentsDisabled: true` to the JSON response for the same case.
- Add regression coverage for both HTML and JSON parser output.

## Validation
- `crystal tool format --check src/invidious/comments/youtube.cr spec/invidious/comments/youtube_spec.cr`
- `make verify`
- `crystal spec spec/invidious/comments/youtube_spec.cr --warnings all`
- `crystal spec --warnings all`

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -236,6 +236,7 @@
     "This channel does not exist.": "This channel does not exist.",
     "Could not get channel info.": "Could not get channel info.",
     "Could not fetch comments": "Could not fetch comments",
+    "Comments are turned off.": "Comments are turned off.",
     "comments_view_x_replies": "View {{count}} reply",
     "comments_view_x_replies_plural": "View {{count}} replies",
     "`x` ago": "`x` ago",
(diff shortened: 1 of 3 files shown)


unixfox:
PR already exist in https://github.com/iv-org/invidious/pull/4051

PR #5710, opened 2026-05-11
Fallback from DASH on iPhone and iPod

### LINKED ISSUE #2236: [Bug] DASH quality on safari iOS
**Describe the bug**

This has already been reported [here](https://github.com/iv-org/invidious/issues/692) and [here](https://github.com/videojs/video.js/issues/6762) and fixed in [this PR](https://github.com/iv-org/invidious/pull/1679) but I still have the issue, it just loads at 144p and there's no gear icon to change the quality.

**Steps to Reproduce**

Play any video with dash quality enabled on and iOS device (not iPadOS though since it seems like dash in supported on there).

**Additional context**

According to [this answer](https://stackoverflow.com/questions/48342562/how-to-mpeg-dash-video-streaming-in-ios/48345811#48345811), dash isn't supported on iOS and must be replaced with fmp4 in HLS.

nickita-khylkouski:
Fixes #2236.

## Summary
- Detect iPhone and iPod user agents, which do not support DASH/MSE playback in Safari/WebKit.
- When those clients request `quality=dash`, use the existing muxed `hd720` source path for watch and embed pages.
- Leave iPad and desktop Safari on the current DASH path.

## Testing
- `crystal tool format --check src/invidious/helpers/utils.cr src/invidious/routes/watch.cr src/invidious/routes/embed.cr spec/invidious/helpers_spec.cr`
- `make verify`
- `crystal spec --warnings all`

### DIFF
--- a/spec/invidious/helpers_spec.cr
+++ b/spec/invidious/helpers_spec.cr
@@ -53,4 +53,30 @@ Spectator.describe "Helper" do
       expect(sign_token("SECRET_KEY", token)).to eq(token["signature"])
     end
   end
+
+  describe "#user_agent_lacks_dash_support?" do
+    it "detects iPhone Safari" do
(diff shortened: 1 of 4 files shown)


unixfox:
We should fix DASH instead of falling back to a bad video quality.

PR #5712, opened 2026-05-11
fix: handle cookies correctly on onion mirrors

### LINKED ISSUE #1421: Instances are working well, but their onion version aren't

khanhsnd:
Fixes: #1421

This change makes SID and PREFS cookies host-aware so onion mirrors fall back to host-only cookies instead of reusing the configured clearnet domain.

It also clears SID and PREFS cookies with explicit domain and path attributes to avoid leaving stale cookies behind when signing out or replacing anonymous preferences.

Validation:
- `crystal tool format --check`
- `CRYSTAL_CACHE_DIR=/private/tmp/crystal-cache crystal spec spec/invidious/helpers_spec.cr`
- `CRYSTAL_CACHE_DIR=/private/tmp/crystal-cache crystal build src/invidious.cr -Dskip_videojs_download --no-codegen --progress --stats --error-trace`
- `CRYSTAL_CACHE_DIR=/private/tmp/crystal-cache bin/ameba`

### DIFF
--- a/spec/invidious/helpers_spec.cr
+++ b/spec/invidious/helpers_spec.cr
@@ -3,6 +3,24 @@ require "../spec_helper"
 CONFIG = Config.from_yaml(File.open("config/config.example.yml"))
 
 Spectator.describe "Helper" do
+  describe "#cookie_domain" do
+    it "keeps the configured cookie domain for the matching host" do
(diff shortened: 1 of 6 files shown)


unixfox:
https://github.com/iv-org/invidious/pull/5647

PR #5713, opened 2026-05-11
Use request host for login cookie domain

### LINKED ISSUE #1421: Instances are working well, but their onion version aren't

diegog956:
## Summary
- choose SID/PREFS cookie domain from the current request host
- keep the configured cookie domain for matching configured hosts/subdomains
- use host-only cookies for unrelated hosts, such as a companion .onion hostname, so login is not tied to the clearnet domain
- add coverage for matching, forwarded, and unrelated hosts

Closes #1421.

## Validation
- `git diff --check`

Not run locally: `crystal`/`shards` are not installed in this Windows environment, and Docker daemon is unavailable.

### DIFF
--- a/spec/invidious/helpers_spec.cr
+++ b/spec/invidious/helpers_spec.cr
@@ -1,8 +1,36 @@
 require "../spec_helper"
 
 CONFIG = Config.from_yaml(File.open("config/config.example.yml"))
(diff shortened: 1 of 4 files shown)


Fijxu:
https://github.com/iv-org/invidious/pull/5647

PR #5646, opened 2026-02-23
Only include '&' if params are present in `watch` urls for playlist RSS

Fijxu:
Related to https://github.com/iv-org/invidious/issues/1232

When passing parameters like `?params=listen%3Dtrue%26quality%3Dmedium` at the end of the feed URL, all videos (`/watch` links) will inherit the passed query parameters, so video links will end up being `/watch?v=<video_id>&listen=true&quality=medium`

PR #5714, opened 2026-05-12
Fix stray ampersand in playlist RSS links

### LINKED ISSUE #1232: Some URLs in RSS feeds end with "&"
Some URLs in RSS feeds will end with "&", when checking the file, it shows that ```&amp;``` is added at the end of the URL.

Example: https://invidio.us/feed/playlist/PL0AQIMR-tpolVRCh8SEK6V5Xi4mRJdFTL

Since the link will update, here's a copy of the page:

[PL0AQIMR-tpolVRCh8SEK6V5Xi4mRJdFTL.txt](https://github.com/omarroth/invidious/files/4761658/PL0AQIMR-tpolVRCh8SEK6V5Xi4mRJdFTL.txt)

saint0x:
Fixes #1232

This keeps the change in the existing RSS link rewrite path and only appends serialized params when params are present. That removes the stray trailing `&` from playlist RSS watch links while preserving the current behavior when feed params actually exist.

AI assistance disclosure: this PR was prepared with AI assistance, then manually reviewed and verified in an isolated Quilt sandbox with a repo-native compile check.

### DIFF
--- a/src/invidious/routes/feeds.cr
+++ b/src/invidious/routes/feeds.cr
@@ -320,7 +320,7 @@ module Invidious::Routes::Feeds
         case attribute.name
         when "url", "href"
           request_target = URI.parse(node[attribute.name]).request_target
-          query_string_opt = request_target.starts_with?("/watch?v=") ? "&#{params}" : ""
+          query_string_opt = (!params.empty? && request_target.starts_with?("/watch?v=")) ? "&#{params}" : ""
           node[attribute.name] = "#{HOST_URL}#{request_target}#{query_string_opt}"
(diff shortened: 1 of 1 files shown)


saint0x:
Verified in an isolated Quilt sandbox. The change stays in the existing RSS rewrite path and only skips appending empty params, so URLs without feed params no longer end with a stray &.

Fijxu:
Same as https://github.com/iv-org/invidious/pull/5646 but this makes `query_string_opt` equal to `""` instead of `nil`.

Please do not open redundant PRs.

PR #5735, opened 2026-05-21
fix #5727: Channel videos are detected as playlists instead of videos

### LINKED ISSUE #5727: [Bug] Channel videos are detected as playlists instead of videos
Now videos in channel appear as a playlist having `/playlist?list=<video_id>` as their path and query parameter. They also show `-1 video` instead of time duration of the video.

Image

acheong08:
Fixes #5727

### DIFF
--- a/src/invidious/yt_backend/extractors.cr
+++ b/src/invidious/yt_backend/extractors.cr
@@ -630,12 +630,18 @@ private module Parsers
     end
   end
 
-  # Parses an InnerTube lockupViewModel into a SearchPlaylist.
+  # Parses an InnerTube lockupViewModel into a SearchVideo or SearchPlaylist.
   # Returns nil when the given object is not a lockupViewModel.
   #
   # This structure is present since November 2024 on the "podcasts" and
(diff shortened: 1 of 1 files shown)


acheong08:
Tested working on https://iv.duti.dev/channel/UC1CGaG114jfEiiWzZ2SkyhQ

(Visit https://iv.duti.dev/supersecretlink first to set whitelist cookie. Instance not public by default)

Fijxu:
Thanks for the PR, but there is still things broken with this PR:

- Playlists do not show on searches and channels.
- View count of channel videos is parsed wrongly, it doesn't show thousand, million and billions of views.

I already fixed those two issues and I'm now just checking things locally before making a PR to fix https://github.com/iv-org/invidious/issues/5727. I will take a look at some parts of your PR to combine it with my PR tho, like `LOCKUP_CONTENT_TYPE_PODCAST` that I wasn't aware of.

acheong08:
Thanks. I'll wait for yours then

acheong08:
It's worth looking at yt-dlp. They fix these issues really quick & I was just using them as reference

PR #5751, opened 2026-05-27
fix: fix author verification in channels

### LINKED ISSUE #5730: [Bug] authorVerified always returns as false on channel pages and recommendations on video pages
**Describe the bug**
On the channel pages and recommendations, authorVerified value always now shows up as false, even if the channel is verified.

Fijxu:
Fixes https://github.com/iv-org/invidious/issues/5730

Author verification badge is not longer located in `c4TabbedHeaderRenderer`. It may be deprecated.


ashleyirispuppy143:
doesnt work with music channels that are verified (example: UCBQZwaNPFfJ1gZ1fLZpAEGw)

Fijxu:
> doesnt work with music channels that are verified (example: UCBQZwaNPFfJ1gZ1fLZpAEGw)

Fixed, ty!

PR #5759, opened 2026-05-31
[RefreshChannelsJob] Reduce backoff if no errors occur

unrealtournament:
Currently there is no code in refresh_channels_job.cr to gradually reduce the backoff, causing it to increase forever even if errors occur sporadically:

image

This PR halves the backoff whenever there are no errors fetching the channel info, until its initial value of 2 minutes.

### DIFF
--- a/src/invidious/jobs/refresh_channels_job.cr
+++ b/src/invidious/jobs/refresh_channels_job.cr
@@ -36,6 +36,11 @@ class Invidious::Jobs::RefreshChannelsJob < Invidious::Jobs::BaseJob
 
               LOGGER.trace("RefreshChannelsJob: #{id} fiber : Updating DB")
               Invidious::Database::Channels.update_author(id, channel.author)
+
+              if backoff > 2.minutes
+                backoff /= 2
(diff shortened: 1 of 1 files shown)


unixfox:
Could you explain the reasoning behind this PR? Did you encounter any underlying issue regarding the refreshchannelsjob?

unrealtournament:
> Could you explain the reasoning behind this PR? Did you encounter any underlying issue regarding the refreshchannelsjob?

As I've shown in the screenshot attached, with the current behavior the backoff **never** decreases (only by restarting Invidious it resets to its default 2 minute timer), even after successful requests after an error. In my home connection where I'm hosting invidious, a request sometimes fails (either because of a transient internet issue or a temporary YT ban), and causes the backoff to double.

Since it never decreases and only keeps doubling, eventually after enough one time errors the backoff reaches multiple hours or even days, causing unnecessary delays between refreshes for just one transient error.

The PR makes it so that after a successful channel refresh, if the backoff had been increased because of errors, it is decreased (halved) again until it reaches its lowest value of 2 minutes. Hopefully I've explained myself clearly.

Fijxu:
Makes sense to me, what do you think @unixfox ?

PR #5768, opened 2026-06-09
Use “www.youtube.com” consistently

janmoesen:
When using the “Go to YouTube” link on error pages, I sometimes get an “Access denied” error from YouTube. When using the canonical domain with the “www.” prefix, this does not seem to happen. Also, this saves an unnecessary redirect.

### DIFF
--- a/src/invidious/helpers/errors.cr
+++ b/src/invidious/helpers/errors.cr
@@ -201,7 +201,7 @@ def error_redirect_helper(env : HTTP::Server::Context)
           <a href="/redirect?referer=#{env.get("current_page")}">#{switch_instance}</a>
         </li>
         <li>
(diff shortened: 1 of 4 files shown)


Fijxu:
Hi, can you share an screenshot of the Youtube error you are getting? Thanks!

janmoesen:
As soon as I get the YouTube error (IIRC, some anti-bot CAPTCHA or HTTP 500-ish error), I will post it.

But I do want to stress that the exact error message does not matter, since this pull request is about making things better/more correct regardless of the error. It should be “www dot youtube dot com” instead of just “youtube dot com” without the “www.” prefix.

Fijxu:
Yeah makes perfect sense for me regardless of the error you are getting from Youtube. I'll merge it after the workflow pass. Thanks for your contribution!

PR #5770, opened 2026-06-09
Show message when YouTube comments are disabled

### LINKED ISSUE #1092: Display "Comments are turned off" when comments are disabled (currently the comment section is just empty)

capturartelab:
## Summary

- Show a localized "Comments are turned off" message when YouTube returns no comment continuation items.
- Include `commentsDisabled: true` in both JSON and HTML API responses for that state.
- Keep the existing empty comments array and zero comment count for API compatibility.

Fixes #1092

## Testing

- `git diff --check origin/master..show-disabled-comments-message -- src/invidious/comments/youtube.cr src/invidious/frontend/comments_youtube.cr locales/en-US.json`
- `Get-Content -Raw locales/en-US.json | node -e "...JSON.parse(stdin)..."`

I could not run the Crystal test suite in this environment because `crystal`/`shards` are not installed.


Fijxu:
There is already a PR that solves this: https://github.com/iv-org/invidious/pull/4051

PR #5772, opened 2026-06-09
Use progressive quality selector on iOS DASH

### LINKED ISSUE #2236: [Bug] DASH quality on safari iOS

capturartelab:
## Summary

- Use the progressive quality selector on Apple mobile devices when the user preference is DASH
- Keep the DASH source hidden from the progressive quality selector
- Keep the existing DASH/http-source-selector path for non-Apple mobile and desktop browsers

## Details

iOS Safari does not reliably support the DASH flow used by the player. When users have DASH quality enabled, Invidious still renders progressive fallback sources, but the UI initializes the DASH-specific source selector instead of the normal quality selector. On iOS this can leave users stuck on a low fallback source with no gear menu to change quality.

This change detects Apple mobile devices and, only for that case, uses the progressive quality selector instead of the DASH source selector. The DASH source is marked with `hidequalityoption="true"` so the progressive selector shows actual video qualities rather than a `dash` entry.

Fixes #2236

## Testing

- `Get-Content -Raw assets\js\player.js | node --check`
- `git diff --check -- assets\js\player.js src\invidious\views\components\player.ecr src\invidious\views\components\player_sources.ecr`

I could not run the Crystal test suite or validate on a real iOS Safari device in this environment.

### DIFF
--- a/assets/js/player.js
+++ b/assets/js/player.js
@@ -38,6 +38,17 @@ if (player_data.aspect_ratio) {
     options.aspectRatio = player_data.aspect_ratio;
 }
 
+function isAppleMobileDevice() {
+    return /iP(ad|hone|od)/.test(navigator.platform) ||
+        (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);
+}
+
+var shouldUseProgressiveQualitySelector = (
+    !video_data.params.listen &&
+    video_data.params.quality === 'dash' &&
+    isAppleMobileDevice()
+);
+
 var embed_url = new URL(location);
 embed_url.searchParams.delete('v');
 var short_url = location.origin + '/' + video_data.id + embed_url.search;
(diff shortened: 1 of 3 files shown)


Fijxu:
Do not open PRs if you don't even test your changes in real devices, browsers and compilers.

PR #5791, opened 2026-06-28
Create AI_POLICY.md

TheFrenchGhosty:
As discussed internally

Rendered version: https://github.com/iv-org/invidious/blob/ai-policy/AI_POLICY.md

Already proofread

PS: I wanted to add a political point - something something capitalism bad, like I'm doing on my own projects ( https://github.com/TheFrenchGhosty/trakt-export-to-markdown#ai-acknowledgement / https://github.com/TheFrenchGhosty/obsidian-to-html#ai-acknowledgement - but I decided to keep it neutral since I assume other team members might not want to bring politics too close to the project.

### DIFF
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -11,6 +11,8 @@ Invidious is written in an obscure language: Crystal.
 
 Because it is obscure the number of people knowing it is really low.
 
+Because the amount of people knowing it is really low, the amount of contributors to Invidious is really low: the vast majority of contributions are done by the Invidious maintainers themselves.
+
 Invidious is the biggest Crystal project that exists, bigger than Crystal itself [(yes, seriously)](https://shards.info/).
 
 The problem of being the biggest software in an obscure language is that you're often effectively the first project to encounter a problem and because it's an obscure language, not a lot of libraries exist to make it easier for you, meaning, you usually have to make everything you need yourself.
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -17,7 +17,7 @@ The problem of being the biggest software in an obscure language is that you're 
 
 This makes it so working on Invidious far harder than working on most open source projects because you are effectively not benefiting and not using any external libraries for the vast majority of things. Almost any time you need anything, you have to make it yourself, which overcomplicates everything.
 
-We are aware that some people like that and we might even end up on one of the "bad people" list, but we try to be reasonable. We ask that you do not fork the project out of spite because of this new policy - let's not split the thin list of people able to contribute even thinner. Contributions are welcome and highly preferred to anything made by AI. This policy comes from a place of *need* not from a place of *choice*.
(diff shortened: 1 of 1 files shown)


AI_POLICY.md
```diff
@@ -0,0 +1,33 @@
...
+This document has been fully written, from scratch, by a Human.
+
+
+# Motivation
+
+Invidious is written in an obscure language: Crystal.
+
+Because it is obscure the number of people knowing it is really low.
+
+Invideos is the biggest Crystal project that exists, bigger than Crystal itself [(yes, seriously)](https://shards.info/).
```
unixfox:
```suggestion
Invidious is the biggest Crystal project that exists, bigger than Crystal itself [(yes, seriously)](https://shards.info/).
```

unixfox (changes requested):

AI_POLICY.md
TheFrenchGhosty: Voice to text messed up here :')

Fijxu (approved):
I agree with the policy!

AI_POLICY.md
```diff
@@ -0,0 +1,36 @@
+Invidious is and has always been made by human first and foremost. However, things have changed recently with the rise of AI.
+
+This document is going to explain everything that you need to know if you ever contribute in any way to Invidious using any kind of AI.
+
+This document has been fully written, from scratch, by a Human.
+
+
+# Motivation
```
unixfox, line 8: My opinion is that this part should be more professional. This reads like a blog post and does not inspire confidence.

> Invidious is written in a lesser known programming language: Crystal.
>
> Because of this, the amount of contributors is low. And those that are contributing are mainly the Invidious maintainers themselves.
>
> Since the maintainers do not have enough free time to work on all the essential maintenance tasks and the new features. It has been decided to use AI for reducing the time needed to work on all of these tasks. And thus keep the project relevant with all the new YouTube changes that come up quite frequently.
>
> We welcome contributors using AI if they follow the policy underneath.

Feel free to add more things. But IMO this should look similar to what I have written. It is concise and professional way to explain why we are using AI.

unixfox (changes requested):

AI_POLICY.md
TheFrenchGhosty: @unixfox  The problem with making it sound professional is that it ends up sounding fake, I do want this part of the policy to be honest and "pure", which mean it has to read like a normal human wrote it, not a corporate suit.

I'm open to ideas, but I really don't think we should drastically change the way it's written.

Fijxu: I'm with @TheFrenchGhosty, I prefer it making it sound more human than corporate, we are just a set a people that collaborate in the same project without funding from any company.

unixfox: Mmh OK fine.

unixfox (approved):

PR #5795, opened 2026-06-30
Add require_login and require_login_for_embeds config options

Geekshere:
Related to https://github.com/iv-org/invidious/issues/446

## Summary
Adds two new boolean config options:

- `require_login` - when enabled, unauthenticated users are redirected to 
  `/login` when accessing any page on the instance
- `require_login_for_embeds` - when enabled alongside `require_login`, 
  also restricts `/embed/` routes (default: false, so embeds remain 
  accessible on external sites by default)

## Use case
This is useful for private instances where the owner wants to restrict access to approved users only e.g. a family server

## What's exempted from the redirect

Even with `require_login: true`, the following are always accessible without login to avoid breaking functionality:

- `/login`, `/register`
- `/api/` (API routes)
- `/embed/` (unless `require_login_for_embeds: true`)
- `/vi/`, `/sb/`, `/ggpht/`, `/yts/`, `/s_p/` (thumbnails/assets)
- `/videoplayback`, `/latest_version`, `/download`, `/companion/` 

## Configuration

In `config.yml`:

​```require_login: true
​```
​```require_login_for_embeds: true  # optional, defaults to false
​```

## Note
Due to third-party cookie restrictions in modern browsers, enabling `require_login_for_embeds` will effectively disable embedding on external sites even for authenticated users. Only useful if embeds are served on the same domain as the instance.

### DIFF
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -206,12 +206,13 @@ before_all do |env|
     path = env.request.path
     allowed = {
       "/login", "/register", "/api/",
+      "/embed/",
       "/sb/", "/vi/", "/s_p/", "/yts/", "/ggpht/",
       "/api/manifest/", "/videoplayback", "/latest_version",
       "/download", "/companion/",
     }
     unless allowed.any? { |p| path.starts_with?(p) }
-      env.redirect "/login?return_path=#{URI.encode_www_form(path)}"
+      env.redirect "/login?referer=#{URI.encode_www_form(path)}"
       halt env, 302
     end
   end
--- a/src/invidious.cr
+++ b/src/invidious.cr
@@ -206,12 +206,15 @@ before_all do |env|
     path = env.request.path
     allowed = {
       "/login", "/register", "/api/",
-      "/embed/",
       "/sb/", "/vi/", "/s_p/", "/yts/", "/ggpht/",
       "/api/manifest/", "/videoplayback", "/latest_version",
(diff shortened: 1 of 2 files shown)


Fijxu:
Did you use AI to make this PR? If so, please note it. The AI_POLICY.md file is still not merged (https://github.com/iv-org/invidious/pull/5791/changes) but it says:

>**Any one using AI to report bugs or submit code MUST properly disclose it, this includes mentioning the name of the EXACT model used and the tools used to interact with it.**

Just asking because the summary looks very AI made-ish.

unixfox:
For the record there has already been a PR opened here: https://github.com/iv-org/invidious/pull/3728

Geekshere:
@Fijxu 

Ah yes sorry I apologize, I did use AI for some assistance as I am not fully familiar with crystal. I am not a big fan of it but regardless it is helpful. I used Claude Sonnet 4.6 and I did thoroughly test the changes. I did not use Claude code I used the website and just used it for some clarification and assistance (and the summary as you can tell).

Geekshere:
> For the record there has already been a PR opened here: #3728

I don't know how I missed that, sorry. It seems to be stale, though, I don't mind closing this in favor of it.

tibequadorian:
this should also be added to `config.example.yml` right?

Fijxu:
> this should also be added to `config.example.yml` right?

Yes

PR #5800, opened 2026-07-09
Fix videos being incorrectly labelled as isUpcoming = true

absidue:
The tenary operator currently just checks if the field is not nil but the parser initialises the SearchVideo struct with `Time.new(0)` if the premiere timestamp is unavailable (e.g. in the lockup view model parser), so the `isUpcoming` field in the API responses was incorrectly being set to `true`.

### DIFF
--- a/src/invidious/helpers/serialized_yt_data.cr
+++ b/src/invidious/helpers/serialized_yt_data.cr
@@ -148,7 +148,7 @@ struct SearchVideo
   end
 
   def upcoming?
-    premiere_timestamp ? true : false
+    premiere_timestamp.try { |t| t.to_unix > 0 } ? true : false
   end
 end
 


Fijxu:
Thanks you for your contribution! Works fine when parsing videos in a Youtube channel

Fijxu (approved):

TheFrenchGhosty:
So that's why Yattee was showing all videos as upcoming! I thought this was a Yattee issue :') 

You have my personal thanks then :)

absidue:
People kept reporting it as a FreeTube issue too, none of them ever opened an issue on the Invidious side and as none of the public instances have the API enabled I knew that it was unlikely that it would get noticed and fixed by one of the Invidious maintainers. I'm glad it helps Yattee users too.

PR #5860, opened 2026-08-02
CI: Exclude development dependencies from build job

Fijxu:
## Checklist

## AI Disclosure

- [ ] AI was used to fully create this pull request

**Tool(s) used:**

**How was AI used?**

---

## Pull request description

The build CI to test if Invidious builds without problems on certain Crystal versions do not really need to install development dependencies like Ameba to build Invidious, so we skip the installation of them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,8 +80,7 @@ jobs:
       - name: Install Shards
         run: |
           if ! shards check; then
-            shards install --production
-            shards install spectator --frozen
+            shards install --skip-postinstall --skip-executables
           fi
 
       - name: Run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,8 @@ jobs:
       - name: Install Shards
         run: |
           if ! shards check; then
-            shards install --without-development
+            shards install --production
+            shards install spectator --frozen
           fi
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty (commented):
Would it be a good idea to use `--production` like the dockerfile? https://github.com/iv-org/invidious/blob/master/docker/Dockerfile#L33

Might be closer to what has to actually build

Not sure what the production flag imply though since it's not in shard.yml

Fijxu:
> Would it be a good idea to use `--production` like the dockerfile? https://github.com/iv-org/invidious/blob/master/docker/Dockerfile#L33
> 
> Might be closer to what has to actually build
> 
> Not sure what the production flag imply though since it's not in shard.yml

Yes, but I just noticed that the actions fails because the spec runs after, I'll take a look at this later since I'm trying to fix another issue now ;)

Fijxu:
Done, now Ameba building is skipped for this action task:

```
I: Installing http_proxy (0.10.3)
I: Installing spectator (0.10.6)
I: Installing ameba (1.6.1)
I: Postinstall of ameba: shards build -Dpreview_mt (skipped)
```

Fijxu:
Postinstall is only ran by Ameba, which needs to build to provide the `ameba` binary in `./bin`

PR #5861, opened 2026-08-02
chore: Move nix files out of root directory

Fijxu:
## Checklist

## AI Disclosure

- [ ] AI was used to fully create this pull request

**Tool(s) used:**

**How was AI used?**

---


TheFrenchGhosty:
@Fijxu I was wondering actually, is the .lock file needed? Like I know that in some language they're generated on the fly and including them is usually not necessary, is it the case for nix? Otherwise looks good to me

TheFrenchGhosty (approved):

Fijxu:
> @Fijxu I was wondering actually, is the .lock file needed? Like I know that in some language they're generated on the fly and including them is usually not necessary, is it the case for nix? Otherwise looks good to me

Yes, flake.lock file is 100% necessary and keeps the state of the Nix inputs (git repositories)

PR #4264, opened 2023-11-15
Videos: Use start time and end time for clips

ChunkyProgrammer:
This PR parses the start and end time for clips. 
closes https://github.com/iv-org/invidious/issues/3921
https://github.com/FreeTubeApp/FreeTube/issues/4323

Ex url: [https://www.youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_](https://www.youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_)
API url: `/api/v1/clips/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_?pretty=1`

API Response:

```typescript
{
  "startTime": 8842.645,
  "endTime": 8855.856,
  "clipTitle": "✂️ Kirby is pink!",
  "video": [Video Object](https://docs.invidious.io/api/common_types/#videoobject)
}
```

PR #5871, opened 2026-08-02
fix: make clip api less expensive

radmorecameron:
## Checklist

## AI Disclosure

## Pull request description

I added the clip api endpoint to Invidious for FreeTube to use and am now finally getting around to adding support in FreeTube.

The endpoint is unnecessarily expensive for no good reason with the `get_video` call so this PR replaces it with just the video_id instead. I haven't been able to find a client that uses this endpoint but would like to use it in FreeTube as part of: https://github.com/FreeTubeApp/FreeTube/pull/9532

## Testing

`{INVIDIOUS API}/api/v1/clips/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_`

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -460,7 +460,6 @@ feed_threads: 1
 ##
 ## Notes: The following API endpoints will be disabled:
 ##  - /api/v1/videos
-##  - /api/v1/clips
 ##  - /api/v1/transcripts
 ##
 ## Accepted values: true, false
--- a/src/invidious/helpers/handlers.cr
+++ b/src/invidious/helpers/handlers.cr
@@ -136,7 +136,7 @@ end
 class DisableAbusableAPIHandler < Kemal::Handler
   {% for method in %w(GET HEAD) %}
     # This endpoints make a video request to Invidious companion.
-    {% for endpoint in %w(videos clips transcripts) %}
+    {% for endpoint in %w(videos transcripts) %}
       only ["/api/v1/{{ endpoint.id }}/:id"], {{ method }}
     {% end %}
   {% end %}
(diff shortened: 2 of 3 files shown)


Fijxu:
Sorry but I don't think we are going to accept this PR as the whole point of the `video` key in the clips API is to show the video information from where the clip was created. I sadly have no clip URL to test but that is what the code does. This is a breaking change that could break some third party applications that make use of that endpoint.

In the other hand seems like the clips feature has been removed from Youtube, but since I don't have any clip ID to test that, I can't really verify that.

radmorecameron:
Hi @Fijxu  I added an example url to the description.

I'm not aware of any third party apps using this endpoint.

 I am the person that added the endpoint (in this PR: https://github.com/iv-org/invidious/pull/4264 ). With the API endpoint as is, FreeTube would likely be making two `get_video` requests instead of one

Fijxu:
I see, thank you for providing a test clip link, I'm still against removing the video information from the clips API because that would involve a breaking change. If you want to make the API less expensive, a query parameter like `fetch_video= [false | true]` would be better, being `true` the default value of the query parameter, and then you can send `false` to make it less expensive for your use case.

What do you think @unixfox ?

unixfox:
I think removing it is probably a bad idea, people might depend on it.

From a quick search on https://github.com/search?q=%2Fapi%2Fv1%2Fclips+invidious&type=code. I found playlet using that API endpoint: https://github.com/iBicha/playlet/blob/48b10e8618924cadfe2e96beac95061b97586bb9/playlet-web/src/lib/Api/InvidiousApi.ts#L84

radmorecameron:
Ahh, I think I must have had a typo in my original GitHub search; I wasn't aware playlet was using this. I'll close this PR and opt for using the resolveurl api endpoint (and parsing the params in FreeTube instead). Sorry about that

PR #5882, opened 2026-08-03
Enable the new locales translated at more than 20%

TheFrenchGhosty:
## Checklist

- [X] I have read the [AI Policy](https://github.com/iv-org/invidious/blob/master/AI_POLICY.md) and understand the disclosure requirements

## AI Disclosure

- [ ] AI was not used to create this pull request
- [X] AI was used to fully create this pull request

GLM-5.2 - Max

**Tool(s) used:**

OpenChamber

**How was AI used?**

AI did the math for the percentage of 20%, and made the change.

Second commit is me asking the AI to align all the comments and it said it's now aligned on "display column 43" even if it still doesn't look like it's aligned... Not sure, not like it really matters anyway.

---

## Pull request description

Relevant to the locale PR that was merged yesterday after I remembered that this is a thing that should be done at reasonable interval.

This PR adds all the languages that are translated at more than 20% to the list of available locales.

Details (LLM written but should be correct): Added be, gl, gsw, hy, lv to LOCALES_LIST in src/invidious/helpers/i18n.cr; updated the candidate-locales comment block to remove be, add sgn, and fix the "tok typo.

### DIFF
--- a/src/invidious/helpers/i18n.cr
+++ b/src/invidious/helpers/i18n.cr
@@ -15,7 +15,7 @@ module I18n
   #
   LOCALES_LIST = {
     "ar"      => "العربية",               # Arabic
-    "be"      => "Беларуская",          # Belarusian
+    "be"      => "Беларуская",            # Belarusian
     "bg"      => "български",             # Bulgarian
     "bn"      => "বাংলা",                 # Bengali
     "ca"      => "Català",                # Catalan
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
@copilot verify that the comment alignment is correct

TheFrenchGhosty:
Well, copilot confirms it so okay I guess

unixfox:
Why not

